PR #15, opened 2016-11-18
New Provider: DNS Made Easy

mhenderson-so:
Added a new provider for DNS Made Easy. Although we don't use this provider yet, I wrote it as a way of learning DNS Control and how the providers work. You can sign up to a sandbox account for testing, which is what I did for development.

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,6 +23,7 @@ import (
 	_ "github.com/StackExchange/dnscontrol/providers/activedir"
 	_ "github.com/StackExchange/dnscontrol/providers/bind"
 	_ "github.com/StackExchange/dnscontrol/providers/cloudflare"
+	_ "github.com/StackExchange/dnscontrol/providers/dnsmadeeasy"
 	_ "github.com/StackExchange/dnscontrol/providers/gandi"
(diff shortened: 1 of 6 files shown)


koenrh:
Do you need any help finishing this provider implementation? From past experiences, I know the sandbox environment behaves identical to that of the production environment. Even though the organisation I work for doesn't use DNS Made Easy in production anymore, it still does have Business membership, which would allow me to verify proper working of `dnscontrol` with their production API.

mhenderson-so:
Hi @koenrh - thanks for the info. I've just been too busy to get to this in my own time.

The primary thing that's stopped me from catching up is that the format for the driver has changed since this PR was submitted, and I haven't reviewed what the new format is. If you want to finish it, be my guest! I think the changes should be reasonably simple - one or two new methods required for the interface...

captncraig:
This certainly seems to have been forgotten. Needs to be significantly reworked because of api changes and things. Maybe worth starting over entirely. If anybody is interested, it might be a good project. This is not a provider we actively use, so if anyone wants to own it, that would be great. Otherwise will close this.

PR #26, opened 2017-01-11
Supporting ipv6 targets in transforms.

captncraig:
If an IP is not v4, we create AAAA records instead of A.


tlimoncelli:
I rebased this. PTAL

tlimoncelli:
Very old. Please re-open if desired.

PR #43, opened 2017-03-15
DNSimple provider

aeden:
This PR implements a DNSimple provider. I have implemented support for both the DNS provider as well as the registrar provider, however I have only tested the DNS provider at this time.

Currently system records (i.e. SOA and NS records) are not editable in DNSimple, so when they are pushed our API will respond with a 400. This does not stop the provider from functioning, but it'd be better to remove those types from the corrections.

It is possible to specify a baseurl in the creds.json to use our sandbox (at https://sandbox.dnsimple.com) which is useful in testing.


mhenderson-so:
Looks like `github.com/dnsimple/dnsimple-go/dnsimple` needs to be vendored

providers/dnsimple/dnsimple.md
```diff
@@ -0,0 +1,31 @@
+## Dnsimple Provider
```
captncraig: You can probably omit this file. Some other providers have them from before we made a seperate docs section.

providers/dnsimple/dnsimpleProvider.go
```diff
@@ -0,0 +1,314 @@
...
+
+	recordsResponse, err := client.Zones.ListRecords(accountId, domainName, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return recordsResponse.Data, nil
+}
+
+// Returns the name server names that should be used. If the domain is registered
```
captncraig: Sorry, this api may be a bit misleading. For a DNS provider we always want GetNameservers to return the authoritative nameservers for the zone on this provider (in this case the default DNSimple nameservers every time it seems). The registrar actually making sure those are delegated properly is an orthogonal operation.

This lets us do things like get 4 possible nameservers each from 2 providers and choose 2 from each to delegate to.

providers/dnsimple/dnsimpleProvider.go
```diff
@@ -0,0 +1,314 @@
...
+	corrections := []*models.Correction{}
+
+	records, err := c.getRecords(dc.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	var actual []*models.RecordConfig
+	for _, r := range records {
+		fqdn := r.Name + "." + dc.Name
```
captncraig: We try to use this helper function for this wherever we do it: https://godoc.org/github.com/miekg/dns/dnsutil#AddOrigin

providers/dnsimple/dnsimpleProvider.go
```diff
@@ -0,0 +1,314 @@
...
+		fqdn = strings.TrimLeft(fqdn, ".")
```
captncraig: Why does it have a "." on the left? Should this be TrimRight?

captncraig:
Is there a way to make a sandbox account to test with without providing a credit card up front?

aeden:
> Looks like github.com/dnsimple/dnsimple-go/dnsimple needs to be vendored

It does. I couldn't find documentation explaining what vendoring mechanism is being used - is it just godep?

captncraig:
Govendor actually.

aeden:
> Is there a way to make a sandbox account to test with without providing a credit card up front?

https://developer.dnsimple.com/sandbox/ provides information on how to set up a sandbox account, including with acceptable mock credit card details.

aeden:
> Govendor actually.

👍  - I haven't used it yet, so I'll get it set up and vendor the dnsimple lib today.

providers/dnsimple/dnsimpleProvider.go
aeden: This trims the dot from the case where `r.Name` is blank (i.e. it's a record on the apex).

I took this approach rather than branching with an if/else on `r.Name == ""`

providers/dnsimple/dnsimple.md
aeden: Done.

providers/dnsimple/dnsimpleProvider.go
aeden: Are both registrar and DNS provider expected to use this same function? If so, I'll need to separate those implementations.

providers/dnsimple/dnsimpleProvider.go
aeden: Done.

captncraig:
Made a few changes to get my tests working against it. Looks pretty good.

It does look like the web ui allows me to set additional nameservers from other domains, but that isn't exposed via the api. That is unfortunate. We still couldn't delete any of the dnsimple NS records, or change the ttls, but it would still open up more scenarios.

aeden:
> It does look like the web ui allows me to set additional nameservers from other domains, but that isn't exposed via the api. 

That's correct, the ability to control secondary DNS is not yet available via the API. It's on the short list of things coming in the future.

> We still couldn't delete any of the dnsimple NS records, or change the ttls, but it would still open up more scenarios.

That's true, but it's not completely out of the realm of possibility in the future. We've mostly kept the management of NS and SOA records a tightly controlled part of our system, simply to avoid mistakes with delegation that can be obscure and hard to grok. Having said that, as we see more people using APIs to control DNS (rather than traditional AXFR) I can see that we may need to change our stance on this. I'll mull it over.

aeden:
Is there anything else you need from me at this point @captncraig ?

captncraig:
No, I'm actually comfortable merging this now. Great work. I've been really impressed with dnsimple so far.

tlimoncelli:
FYI: I just updated provider-list.md to clarify the responsibilities of someone that submits a new provider. Basically you'll be asked to fix future bugs, but the DNSControl maintainers will take responsibility for maintaining a test suite that lets you do that easily.  May we list you as the maintainer of this provider?

aeden:
Yes, you may list me as the maintainer.

PR #45, opened 2017-03-16
Redundant use of multiple fmt.Println commands

daubaris:
Changed multiple fmt.Println("") commands to one using backticks


captncraig:
Thanks!

PR #49, opened 2017-03-17
NEW: convertzone/ Read BIND zone files, output records as DSL, TSV, o…

tlimoncelli:
convertzone parses a BIND-style zonefile and outputs it either as tab-separate records, the DNSControl DSL, or pretty-printed zone.

The code needs some cleanups but it is good enough to start reviews.


captncraig:
1. Can we reuse the code from the bind provider? If we exported a few things like `ReadZone(fname string) []*Record` and `MakeZoneFile(records) string`, couldn't we do this without much more duplication?

2. I wouldn't be opposed to making this a seperate command availible from main. We're at the point where we should seriously refactor how those commands are structured in main, but that can be done in the future.

tlimoncelli:
This is a hack that should be used once when getting set up and then thrown away.  Therefore I don't want to put too much work into it, or refactoring dnscontrol.  I think it is fine as an external command (just like gofmt isn't built into go).  As far as reusing the bind provider... it uses the important parts.  Looking at rr_format, it repeats some things from the bind provider but nothing that I could unify without things getting ugly.

What would make both the BIND provider and this cleaner would be if the dns module permitted stringifying individual parts of a RR, instead of stringifying the entire thing and then requiring people to parse it.

misc/convertzone/main.go
```diff
@@ -0,0 +1,153 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/StackExchange/dnscontrol/providers/bind"
+	"github.com/miekg/dns"
+	"github.com/miekg/dns/dnsutil"
+)
+
+var output_mode = flag.String("mode", "tsv", "tsv|dsl|pretty")
+var flag_defaultTtl = flag.Uint("ttl", 300, "Default TTL")
```
captncraig: lint is throwing fits about underscore variable names. lets nip this in the bud and use camelCase for Go variables.

misc/convertzone/main.go
```diff
@@ -0,0 +1,153 @@
...
+var flag_registrar = flag.String("registrar", "REG_FILL_IN", "registrar text")
+var flag_provider = flag.String("provider", "DNS_FILL_IN", "provider text")
+
+// parseargs parses the non-flag arguments.
+func parseargs(args []string) (zonename string, filename string, r io.Reader, err error) {
+	// 1 args: first arg is the zonename. Read stdin.
+	// 2 args: first arg is the zonename. 2nd is the filename.
+
+	if len(args) < 1 {
+		return "", "", nil, errors.New("Not enough parameters")
```
captncraig: use fmt.Errorf instead of errors.New. It is usually more useful.

misc/convertzone/main.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	zonename = args[0]
+
+	if len(args) < 2 {
```
captncraig: lets do `==1` here instead of `<2` just for clarity.

PR #52, opened 2017-03-20
Validate

captncraig:
Mostly refactoring to validation. 

Allows distinguishing Warning vs Error. Most things are errors. Use `return Warning{fmt.Errorf("whatever")}` to use a warning vs an error.

If an error is returned, dnscontrol will not proceed.


tlimoncelli:
LGTM

PR #53, opened 2017-03-20
Update bind docs and Getting Started

captncraig:
Relaxes a few things to make getting started easier.

1. Don't strictly need a `creds.json`. I it is not there, or a provider not in it, it is up to the provider to complain it does no thave the config it needs.
2. "bind" is no longer a default skipped provider. We need to update our own scripts to work around this.
3. Getting started now runs a simple bind domain.

### DIFF
--- a/providers/bind/bindProvider.go
+++ b/providers/bind/bindProvider.go
@@ -11,10 +11,6 @@ bind -
 	if an update is actually needed. The old zonefile is also used
 	as the basis for generating the new SOA serial number.
 
-	If -bind_skeletin_src and -bind_skeletin_dst is defined, a
-	recursive file copy is performed from src to dst.  This is
-	useful for copying named.ca and other static files.
-
 */
 
 import (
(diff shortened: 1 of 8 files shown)


providers/bind/bindProvider.go
```diff
@@ -107,12 +103,12 @@ func rrToRecord(rr dns.RR, origin string, replace_serial uint32) (models.RecordC
...
 	// Make a default SOA record in case one isn't found:
-	soa_rec := models.RecordConfig{
+	soaRec := models.RecordConfig{
 		Type: "SOA",
 		Name: "@",
 	}
-	soa_rec.NameFQDN = dnsutil.AddOrigin(soa_rec.Name, origin)
-
+	soaRec.NameFQDN = dnsutil.AddOrigin(soaRec.Name, origin)
+	//TODO(cpeterson): are these sane defaults?
```
captncraig: @tlimoncelli I'm not sure if there is anything we can put for these defaults that makes more sense, but this still strikes me as odd.

captncraig:
In latest change, can make provider "non-default" in creds.json:
```
{
"bind":{
  "_exclude_from_defaults": "true"
}
}
```

PR #58, opened 2017-03-23
Adding simple require functionality to js

### LINKED ISSUE #51: Allow splitting of config into multiple files
From twitter: 

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr"><a href="https://twitter.com/Nick_Craver">@Nick_Craver</a> with dnscontrol is there a recommended approach for a global.js (providers/vars) + a single .js file for each domain?</p>&mdash; Brandon Potter (@BrandonPotter) <a href="https://twitter.com/BrandonPotter/status/843850532594864129">March 20, 2017</a></blockquote>

I agree this might be a useful feature. Currently the javascript is processed by the following process, using https://github.com/robertkrimen/otto as its javascript interpreter:

1. Load and run the built in code from `helpers.js`. 
2. Load and run your user code from `dnsconfig.js`.
3. Run a small script to jsonify and dump your config object to be read in go.

It may be possible to have some kind of built in mechanism for combining / importing other js files if your config is large and complicated. 

Possibly `require("otherfile.js")` or `import("otherfile.js")`. It would simply act as string replacement, replacing the import statement with the contents of the specified file before running.

This is not a commitment to implement this, only a place for discussion. Please state cases for or against such a scheme.

captncraig:
Should resolve #51 if sufficient. Simply do:

`require("otherfile.js")` and it will load and run that in place.


js/js_test.go
```diff
@@ -25,7 +26,7 @@ func TestParsedFiles(t *testing.T) {
 		t.Fatal(err)
 	}
 	for _, f := range files {
-		if filepath.Ext(f.Name()) != ".js" {
+		if filepath.Ext(f.Name()) != ".js" || !unicode.IsNumber(rune(f.Name()[0])) {
```
tlimoncelli: Add a comment explaining what is being checked here.

vfstaboy:
I wanted to thank you for this feature.  We manage >300 zone files and this will help make managing them with dnscontrol much better.  Also it would be a priceless section in the getting started doc.  Thank you again!

PR #84, opened 2017-04-14
WIP: PTR Support

captncraig:
Still Needs

- [ ] Make tests better.
- [ ] PTR record type in DSL
- [ ] Validate that all FQDNs are full ip addresses of proper format
- [ ] Docs
- [ ] Route 53
- [ ] Bind

Other providers may or may not support. We take it as they come.


transform/arpa_test.go
```diff
@@ -0,0 +1,45 @@
...
+func TestReverse(t *testing.T) {
+	var tests = []struct {
+		in      string
+		isError bool
+		out     string
+	}{
+		{"174.136.107.0/24", false, "107.136.174.in-addr.arpa"},
+
+		{"174.136.0.0/16", false, "136.174.in-addr.arpa"},
+		{"174.136.43.0/16", false, "136.174.in-addr.arpa"}, //do bits set inside the masked range matter? Should this be invalid? Is there a shorter way to specify this?
```
tlimoncelli: This should be an error. Technically it is valid (the inside bits should be ignored) but operationally I have only see someone do this when it was a typo.  So, let's make it an error and save someone a lot of grief. Add a comment to that effect in the code.  #opinionated

captncraig:
Still Needs

- Make tests better.
- PTR record type in DSL
- Validate that all FQDNs are full ip addresses of proper format
- Docs

Issue #50, opened 2017-03-19
Should we support ALIAS records?


captncraig:
Some providers support an ALIAS or ANAME record type that acts like a CNAME, but converts it dynamically to an A or AAAA record at dns request time. 

The primary use case I believe is allowing a CNAME type record at the zone apex where it is not usually allowed. 

R53, DME, and DNSimple, and CF support them natively, while others do not at the moment. 

### Proposal:

1. Add an `ALIAS( name, aliasDomain)` record type.

2. Add some mechanism for a provider to indicate it supports these records (perhaps by implementing `interface { SupportsAlias() bool }`).

3. At validation time, check if you use `ALIAS` records with a non-supporting provider, and error if so.

4. Potentially add some "override" metadata, to resolve the name at "compile time" and use that A and/or AAAA records instead for non-supporting providers. This would be a one-time lookup, but could still "work" in a way. Perhaps you run dnsconrtol on a cron job to keep it updated it or something. I would want to use something like `ALIAS("@", "otherdomain.com", { lookup_on_build: true})` to make sure you are really ok with that behaviour. (name negotiable). 

### Concerns:

1. Inconsistent experience if using some supporting providers and some non-supporting. Dual host setups could end up with divergent zones.

2. Is this feature even needed? I do think CNAME apex records could be valuable, but don't want to do it if nobody will use it.

3. Is "compile time" lookup a horrible idea in general?

tlimoncelli:
Compile-time lookups are generally a bad idea. We currently don't have any compile-time DNS queries.

We could restrict it to domains in the dnsconfig.js file and simply resolve it by looking at the data.  It would have to be done after any transforms, etc.  I don't think there is currently a lookup function.  We could do a simple lookup, ignore wildcards, etc.

    ANAME("@", "dest.foo.com.", {lookup_if_unsupported: true}),

A benefit of restricting this to destination domains that are in dnsconfig.js is that you know that if the A record changes, DNScontrol will be run.  They can't get out of sync.

captncraig:
I see where you are coming from there. In that case I think there are really 2 categories:

1. ALL providers you are using support ALIAS/ANAME records, and we can use them appropriately.
2. They don't. And you use A records instead. 

If you control the domain, I don't see a real reason to use ALIAS vs just defining the A records multiple times (or a macro).

The main use case I see for ALIAS is if I want to use a 3rd party service at the zone apex, and they don't have a long-term stable IP address. 

[...]

PR #86, opened 2017-04-15
Fix apex CNAME validation

bfirsh:
Cloudflare supports apex CNAMEs, but the validation was throwing errors like this because there were also MX records there:

    ERROR: Cannot have CNAME and MX record with same name: example.com

A better solution would only allow apex CNAMEs on providers that actually support it, but since that wasn't checked before, I figure this is a good interim solution.

### DIFF
--- a/normalize/validate.go
+++ b/normalize/validate.go
@@ -277,7 +277,7 @@ func checkCNAMEs(dc *models.DomainConfig) (errs []error) {
 		}
 	}
 	for _, r := range dc.Records {
-		if cnames[r.Name] && r.Type != "CNAME" {
+		if cnames[r.Name] && r.Name != "@" && r.Type != "CNAME" {
 			errs = append(errs, fmt.Errorf("Cannot have CNAME and %s record with same name: %s", r.Type, r.NameFQDN))
 		}
 	}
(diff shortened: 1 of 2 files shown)


captncraig:
Our intention in adding that validation was to encourage and enforce best-practices. Putting a CNAME at the apex should not be done, and is often tried by people who don't understand the ramifications. 

Yes, cloudflare allows it, only because they "flatten" it into an A record. 

Our long-term plan is to allow an ALIAS record type that resolves differently on a provider-by-provider basis. That is not available at present though. Removing this validation because one provider supports it in a non-standard way makes me a tad uncomfortable. 

Thoughts @tlimoncelli ?

bfirsh:
Agreed that an ALIAS record would be better.

I wish Cloudflare used ALIAS instead of silently flattening CNAMEs, really.

tlimoncelli:
Apex CNAMEs should require something special (ALIAS instead of CNAME) because people should have to do something extra to acknowledge that this is special.

bfirsh:
Sounds good.

I now see #50. I can have a go at implementing this, if you'd like.

captncraig:
I'm working on it now, as we kinda want providers to communicate their capabilities when they register themselves. Then we can do validation up front that all providers for a domain support the features used.

Will have PR soon.

PR #89, opened 2017-04-17
Support ALIAS records with Cloudflare

captncraig:
Should support cloudflare. Other providers can be added.


captncraig:
@bfirsh can you confirm this works for your use case?

bfirsh:
@captncraig almost! problem outlined in #95

PR #90, opened 2017-04-17
Fix broken doc-link from getting-started

philpennock:
This appears to be the only one.  Manual audit with (zsh):

    pcregrep -o1 '}}/([^\)]+)\)' **/*(.)

yielded all the site-based URLs and I eye-compared them to the names of
existing docs.


tlimoncelli:
Thanks for finding that!

PR #92, opened 2017-04-18
Name.com now creates NS records for all nameserves created

pmoroney:
The warning is no longer needed.

### DIFF
--- a/providers/namedotcom/records.go
+++ b/providers/namedotcom/records.go
@@ -81,7 +81,7 @@ func checkNSModifications(dc *models.DomainConfig) {
 	newList := make([]*models.RecordConfig, 0, len(dc.Records))
 	for _, rec := range dc.Records {
 		if rec.Type == "NS" && rec.NameFQDN == dc.Name {
-			continue
+			continue // Apex NS records are automatically created for the domain's nameservers and cannot be managed otherwise via the name.com API.
 		}
 		newList = append(newList, rec)
 	}
--- a/providers/namedotcom/records.go
+++ b/providers/namedotcom/records.go
@@ -2,14 +2,12 @@ package namedotcom
 
 import (
 	"fmt"
-	"log"
-	"strings"
+	"strconv"
 
 	"github.com/miekg/dns/dnsutil"
 
 	"github.com/StackExchange/dnscontrol/models"
 	"github.com/StackExchange/dnscontrol/providers/diff"
-	"strconv"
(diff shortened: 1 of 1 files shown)


captncraig:
So that warning is in case you try to do something like this:
```
D("example.com", DnsProvider(namedotcom), DnsProvider(r53),...)
```

DnsControl attempts to make NS records in each provider to match the list of registered nameservers. If you don't do that, things will sort of work, but break in odd ways according to ttls and caches and so forth.

Name.com will not allow you to add an NS record for `@` pointing at any other nameservers. In fact they don't let you edit or manage apex NS records at all. So if we detect any apex NS records for non-name.com nameservers, we print a warning because we can't do what you asked for.

captncraig:
Oh, I think I see what you mean. The registrar bit feeds the NS records they serve. Very clever. I get super weird api failures trying to set things, but as I query the nameservers I see what you mean. I'm gonna do some more testing.

pmoroney:
If you want to email me those weird api failures to pmoroney@name.com I can take a look at those as well.

captncraig:
Maybe "super weird" failures was a bit extreme. Here's what I noticed:

1. Calls to create NS records for '@' now succeed if the target is non-name.com, and fail if I set to `ns1.name.com` and such.
2. NS records for '@' are never returned via api. I would perhaps like to see documented that you should not manage them via api, but via the registered nameservers for the domain.
3. Trying to add via the web ui gives a slightly odd error:
capture
Although the record type description is somewhat helpful.

So I guess the crux of my complaint is that apex NS records should not be manageable via the api at all if the magic is going to happen from the nameservers. Or else they should be fully manageable and returned via api. 

I was able to add some NS via api, and they got served, but then setting the nameservers seems to destroy those changes.

But regardless of how the api works, I guess the way we are avoiding the problem in dnscontrol is going to work ok anyway.

Sorry for the long-winded ranting.

captncraig:
So to merge this, I would just like to add a comment in there, something to the effect of:

"name.com will automatically serve NS records for whatever nameservers are registered for your domain. We don't need to manage the apex NS records."

pmoroney:
Thanks for the list of issues.
Allowing apex NS record creation via the API was an oversight. We now return an error in that case.

captncraig:
Thank you, super glad we open sourced this. Nothing like having the source themselves finding and fixing things in our code.

PR #93, opened 2017-04-19
providers/route53/route53Provider.go

### LINKED ISSUE #82: Route53 should not require explicit key ID/access key, as there are other credential sources
The AWS SDK can find credentials from local profiles (AWS_PROFILE env var), instance profiles, and/or environment variables. Right now it seems the R53 provider code deliberately requires passing in the credentials, which makes it impossible to use any other credential source.

AFAIK the recommended way to deal with credentials is to pass them if they are explicitly defined, but not care at all otherwise and let the SDK handle things. If no valid credentials actually exist API calls will simply fail with lack of authorization.

mathieuherbert:
Hi there,

I propose this PR to fix #82. 
if KeyId and SecretKey are set, go sdk will use it otherwise it uses environment variable or IAM profile in an AWS Instance (default behavior of go sdk).

When go sdk performs a request to AWS it throw an error if credentials are not good.

It's my first PR here, I can improve it if you have any comment :)
I think I will propose the same in a next PR fo region when I have time.

### DIFF
--- a/providers/route53/route53Provider.go
+++ b/providers/route53/route53Provider.go
@@ -13,6 +13,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	r53 "github.com/aws/aws-sdk-go/service/route53"
+	"github.com/pkg/errors"
 )
 
 type route53Provider struct {
@@ -21,7 +22,6 @@ type route53Provider struct {
 }
 
 func newRoute53(m map[string]string, metadata json.RawMessage) (providers.DNSServiceProvider, error) {
-
 	keyId, secretKey := m["KeyId"], m["SecretKey"]
 
 	config := &aws.Config{
@@ -34,6 +34,10 @@ func newRoute53(m map[string]string, metadata json.RawMessage) (providers.DNSSer
 	sess := session.New(config)
 
 	api := &route53Provider{client: r53.New(sess)}
+	err := api.getZones()
+	if err != nil {
+		return nil, err
+	}
 	return api, nil
 }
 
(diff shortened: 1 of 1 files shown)


mathieuherbert:
I see last comments, I can add later the check on Route53 domain list.

danielkza:
It probably is a good idea to remove the region hardcoding as well.

> I see last comments, I can add later the check on Route53 domain list.

You can make the call with DryRun enabled, and it will just check if the provided credentials allow sufficient access.

captncraig:
I have been a bit fuzzy on what the region means for route 53. Is it just the api urls you talk to? Would changing it allow you to talk to a closer api server, or is there more to it? Are zones tied to particular regions only, or will any region do? We hard coded it because it worked at the time. Totally open to better options. It could be specified with the credentials, or as provider-level metadata if it would even make sense to use different regions for different domains. Any help there is much appreciated.

captncraig:
Yeah, my recommendation for using it would be to simply call `getZones` from the initializer. Ideally detect an auth failure and print some useful message. 

That would also have the advantage of having all the zones pre-fetched if a provider exists, and we can get rif of any other calls to getZones and checks that the zone list is populated.

danielkza:
@captncraig

> I have been a bit fuzzy on what the region means for route 53. Is it just the api urls you talk to? Would changing it allow you to talk to a closer api server, or is there more to it? 

That's a good question, I'm not sure if it the API region really matters. But it would, of course, be better if we could follow usual SDK configuration if possible.

mathieuherbert:
Hi, 
I fetch credentials in getZones, fetch zones in the init method.
Also I add a check on DelegationSet if nil or not : I had a Hosted Zone Route53 without Delegation Set, it resulted to a segmentation fault.

mathieuherbert:
For the region, Route53 is indeed a global service so it is not mandatory to specify another region that us-west-2. But often, when you set your AWS credentials in environment variables or aws credential file, you set also the region.

mathieuherbert:
Up on this PR :)

captncraig:
Looks pretty nice. Thank you!

captncraig:
Could you just quickly update the docs to reflect the various configuration strategies possible?

PR #94, opened 2017-04-19
implement namecheap dns management

rbelnap:
I was disappointed to see that the namecheap provider was only for registration, so I figured I'd take a stab at implementing it.  I've only known go for 2 days, and this is shamelessly cobbled together from other providers and bits of chewing gum.  It seems to do the basics, so I'm hopeful some more experienced contributors can work out the kinks before my mistakes get too ingrained.


captncraig:
Awesome! I'll take a look at this shortly. Have you tried wiring it up to the provider integration test suite? That is a good way to make sure it covers a bunch of the corner cases we have found. Hopefully the readme in there is sufficient, but if not I can help.

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
+}
+
+func newReg(conf map[string]string) (providers.Registrar, error) {
+	return newProvider(conf, nil)
+}
+
+func splitDomain(domain string) (string, string) {
+	ds := strings.Split(domain, ".")
+	dl := len(ds)
+	tld := ds[dl-1]
```
captncraig: This method fails for tlds like `.co.uk'. Should use https://godoc.org/golang.org/x/net/publicsuffix instead.

Its kinda crappy that their api expects SLD/TLD, and https://www.namecheap.com/support/api/methods/domains/get-info.aspx doesn't even split it like that for you.

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
-func newReg(m map[string]string) (providers.Registrar, error) {
+func newDsp(conf map[string]string, metadata json.RawMessage) (providers.DNSServiceProvider, error) {
+	return newProvider(conf, metadata)
+}
+
+func newReg(conf map[string]string) (providers.Registrar, error) {
+	return newProvider(conf, nil)
+}
+
+func splitDomain(domain string) (string, string) {
```
captncraig: I'd name the return values sld and tld for clarity. May also make sense to return an error in case we can't figure it out from the given domain.

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
+		return nil, err
+	}
+
+	var actual []*models.RecordConfig
+	for _, r := range records.Hosts {
+		if r.Type == "SOA" || r.Type == "NS" {
+			continue
+		}
+
+		if r.Name == "@" {
```
captncraig: Shouldn't need to do this. Set Name to Name, and for NameFQDN use `dnsutil.AddOrigin` should be an example of that somewhere.

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
+
+	sld, tld := splitDomain(dc.Name)
+	records, err := n.client.DomainsDNSGetHosts(sld, tld)
+	if err != nil {
+		return nil, err
+	}
+
+	var actual []*models.RecordConfig
+	for _, r := range records.Hosts {
+		if r.Type == "SOA" || r.Type == "NS" {
```
captncraig: Does namecheap not let you change NS records? Even for subdomains?

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
+
+	differ := diff.New(dc)
+	_, create, delete, modify := differ.IncrementalDiff(actual)
+
+	// because namecheap doesn't have selective create, delete, modify,
+	// we bundle them all up to send at once.  We *do* want to see the
+	// changes though
+
+	changes := false
+	desc := ""
```
captncraig: Probably make desc a `[]string`, append each description to it, and then use `strings.Join(desc, "\n") to build a final string.

Then you can avoid the `changes` cariable altogether and use `if len(desc) > 0`

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
+					return n.UpdateRecords(dc)
+				},
+			})
+	}
+
+	return corrections, nil
+}
+
+func (n *Namecheap) GetNameservers(domainName string) ([]*models.Nameserver, error) {
+	// not exactly sure what is wanted here, but currently not needed (TODO)
```
captncraig: This should return the nameservers to register if you are using namecheap's dns. Looks like maybe 

`"dns1.registrar-servers.com", "dns2.registrar-servers.com"` for the basic dns servers?

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
+// UpdateRecords bundles up the expected zone and sends it wholesale to
+// namecheap.
+func (n *Namecheap) UpdateRecords(dc *models.DomainConfig) error {
+
+	var recs []nc.DomainDNSHost
+
+	id := 1
+	for _, r := range dc.Records {
+		name := strings.Split(r.NameFQDN, "."+dc.Name)[0]
+		if name == dc.Name {
```
captncraig: If you leave the Name as I recommend above, this may not need to be done either.

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
+			MXPref:  int(r.Priority),
+			TTL:     int(r.TTL),
+		}
+		recs = append(recs, rec)
+		id++
+	}
+
+	sld, tld := splitDomain(dc.Name)
+	result, err := n.client.DomainDNSSetHosts(sld, tld, recs)
+	fmt.Println("Namecheap returns:", result)
```
captncraig: Remove debug before finalizing.

providers/namecheap/namecheap.go
```diff
@@ -18,9 +20,136 @@ type Namecheap struct {
...
+
+// UpdateRecords bundles up the expected zone and sends it wholesale to
+// namecheap.
+func (n *Namecheap) UpdateRecords(dc *models.DomainConfig) error {
+
+	var recs []nc.DomainDNSHost
+
+	id := 1
+	for _, r := range dc.Records {
+		name := strings.Split(r.NameFQDN, "."+dc.Name)[0]
```
tlimoncelli: Are you trying to extract the name from a FQDN? (i.e. remove the domain suffix?).  I recommend using dnsutil.TrimDomainName() instead (warning: it FQDN == domain this returns "@" instead of "").

providers/namecheap/namecheap.go
rbelnap: I had suspected that and tried it, but namecheap doesn't like it.  When set that way : 
```CREATE NS dnscontrol.com dns1.registrar-servers.com. 300
CREATE NS dnscontrol.com dns2.registrar-servers.com. 300
RECREATING ZONE: dnscontrol.com
Namecheap returns: <nil>
FAILURE! Error 5050900: INVALID_NAME: NS record can't be set up for a @ or * host names
```
The best I could figure was to have it be a noop, but I'm open to interpretation

captncraig: Yeah, this is pretty common. Too many providers control those way too tightly. The general policy in this case is to remove any NS records with Name `@` at the beginning of `GetDomainCorrections`. Something like:

```
dc.Filter(func(r *models.RecordConfig) bool {
    if r.Type == "NS" && r.Name == "@" {
        //check to make sure it looks like a namecheap server and print a warning if it does not.
        return false
    }
    return true
})
```

providers/namecheap/namecheap.go
rbelnap: Yeah, I can see that - I completely stole this section from the gandi provider, so the same can apply there ;)

providers/namecheap/namecheap.go
rbelnap: That would be a way to protect against that error in general, but I don't think it is applicable to the GetNameservers method, which it sounds like will need to remain a noop.

providers/namecheap/namecheap.go
rbelnap: Yes indeed, thanks for the tip! (fixed)

providers/namecheap/namecheap.go
rbelnap: fixed

providers/namecheap/namecheap.go
rbelnap: Agree on the crappy - implemented pubblicsuffix (hopefully correctly)

providers/namecheap/namecheap.go
rbelnap: It does! (for subdomains) fixed

providers/namecheap/namecheap.go
rbelnap: fixed

providers/namecheap/namecheap.go
captncraig: No, GetNameservers needs to be implemented. It does more than just create NS records. It also feeds the nameservers we give to the registrar part.

captncraig:
Build is failing because the publicsuffix package needs to be copied into the repo. I just added it to master, so next time you push it should pass.

providers/namecheap/namecheap.go
rbelnap: Hmm.. so are you saying implement the defaults for GetNameservers, then drop it (or filter it out) in GetDomainCorrections?

captncraig: Correct.

rbelnap:
I think everything has been fixed here - I welcome further feedback.  I did wire it up to the integration tests, but it looks like namecheap adds in default records when a zone is emptied, and this breaks all tests.  Also note that I uncovered what I expect is a bug in the go-namecheap module and submitted it as https://github.com/billputer/go-namecheap/pull/6

captncraig:
That sounds crappy. I'll try it when I get back to my desk and see what I can figure out. If that PR doesn't get merged quickly, it is possible to vendor in your changes regardless.

captncraig:
Ok, I think I got past that particular issue, just by filtering out those records and pretending like they don't exist. 

Now I'm getting rate limited running the test suite.

Can you give me permission to push to your pr branch? It may be easier to just make a few edits directly.

captncraig:
Oh, somehow I did push to that. Hmm.

captncraig:
What the hell did I do? I can fix.

captncraig:
I have no idea what I did. I was trying to add changes to your branch, but it did not do what I expected. 

If you re-push your changes to your repo, this may reopen. I'm sorry.

I also pushed your changes as the namecheap branch to this repo.

captncraig:
I reopened it as #96. Its quite nearly ready anyway.

PR #97, opened 2017-04-25
add BaseURL option to namecheap creds to allow setting sandbox endpoint

rbelnap:
I found this to be useful.


tlimoncelli:
Thanks for the PR!

PR #98, opened 2017-04-25
Adding URL support to namecheap dns provider

rbelnap:
Note that this builds on the namecheap branch created for #96 and includes a merge from master to update that branch.  The relevant bits are in the last commit.  This change is modeled after the ALIAS addition in #89 .   

I did not include further validation in validate.go, because I would be copying checkALIASes with essentially s/ALIAS/URL, and I'm not sure if a more general way to check capabilities would be more DRY-like.  I'm afraid that level of refactoring is beyond my current golang level.  

Namecheap also supports several URL type records (see https://www.namecheap.com/support/api/methods/domains-dns/set-hosts.aspx) so I'd hate to see nearly identical checkURLs, checkURL301s, and checkFRAMEs functions.  I don't use these additional records, so I'm not keen to implement them, but it would be nice if you're after a full featured namecheap implementation.


rbelnap:
spoke too soon about my need for URL301 and FRAME support - implemented these as well, with the same lack of validation mentioned above for URL

captncraig:
I'm not sure how I feel about these. I added ALIAS since it is a pretty provider-independent concept, and can be applied to multiple providers. No other provider has these record types.

I would like to be able to support these, but kinda wonder if there wouldn't be a better way to represent them.

Maybe something like `REDIRECT("@", "foo.com/whatever")`, and maybe some metadata for determining "modes": `REDIRECT("@", "foo.com/whatever", {redirect_mode: "frame"})`

I'm just not sure proliferating every custom record type every provider has is going to be a sustainable strategy.

rbelnap:
I agree completely - but I don't see a way to keep provider specific things separate.  Unless there were some dynamic js loading per provider or something, and I don't want to look too far down that path.  

It is especially onerous for namecheap, because of their "all or nothing" approach to setting records (except for SRV records...)  because these records can't just be ignored - they will be removed if they aren't managed.

I'm not sure what would be gained having it wrapped up in a REDIRECT, since they would just need to be pulled apart again to be sent through namecheap.  It would also make the DSL confusing, since the api itself uses URL/URL301/FRAME as the record type, with no mention of REDIRECT.

captncraig:
Regarding the namcheap issue, we could potentially filter the records to only known types when we read it, and then keep a list of "other" types, and re-add those records when we rebuild the zone. 

Kinda hacky, but is an option.

captncraig:
I've thought about this a bit. How about a solution like this:

- Add a function `providers.RegisterCustomRecordType(name, provider, realType)`, that will allow us to add namecheap specific records.
- Name must be prefixed with provider name: `providers.RegisterCustomRecordType("NAMECHEAP_URL", "NAMECHEAP", "URL")` so that it is obvious where it is used.
- We can validate that only providers of the given type can use it. Cannot be used in dual-hosting setups.
- After validating, the validation routine can replace the record type with the real type for the record. (in this case URL), so the provider does not have to.

So a few questions:

1. Does that seem like a sensible approach?
2. Have I missed something?
3. Do you want to have a go at implementing it, or should I?

rbelnap:
It sounds sensible, but I'm not sure I follow completely.  The main hurdle for me was the addition of the record types in helpers.js so they could be parsed at all.  Once it gets to the go pieces, there are probably a variety of ways for providers to register/validate custom record types.  Some type of callback function for providers to validate their own records, perhaps.  Maybe this is partly what you're suggesting.  

How would these get parsed in the js, and how would they look in the DSL?

I'm not a go programmer, so any implementation of mine should be avoided.  😄  besides, you are significantly more familiar with the codebase.

captncraig:
Yeah, thats a good point. You would still need to add a function to helpers.js (essentially a copy of cname or whatever). My primary goal though is to protect people using other providers from thinking this is useful to them. `URL` is just generic enough to be confusing. I'll work up a version of what I mean.

captncraig:
I added some things to support custom record types that are only valid for one provider. See how cloudflare does page rules:

https://github.com/StackExchange/dnscontrol/blob/master/providers/cloudflare/cloudflareProvider.go#L38
https://github.com/StackExchange/dnscontrol/blob/master/pkg/js/helpers.js#L320

Will at least validate that only namecheap providers can use this type. I will hopefully get back to testing the base PR at some point this week. Sorry its taken so long, I just have been real busy. Going to close this. Let's get the other merged soon, then we can add this functionality.

PR #100, opened 2017-04-26
fix go-namecheap vendor, already applied upstream

rbelnap:
upstream fix applied via https://github.com/billputer/go-namecheap/pull/6


captncraig:
This just needs to be run through `govendor update github.com/billputer/go-namecheap` so that the metadata file updates as well.

PR #101, opened 2017-04-26
add delay option to work around rate limits

rbelnap:
I found this was useful to avoid hitting overzealous rate limits.  I'm open to other approaches.

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,6 +36,8 @@ var domains = flag.String("domains", "", "Comma seperated list of domain names t
 
 var interactive = flag.Bool("i", false, "Confirm or Exclude each correction before they run")
 
+var delay = flag.Int64("d", 0, "delay between domains to avoid rate limits (in ms)")
+
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
 	flag.Parse()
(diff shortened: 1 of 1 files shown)


captncraig:
The optimal solution for rate limits IMO would be for each provider to know the limits, and self-throttle across all domains. 

In the case of nameheap though, I have not been able to find any reliable hard numbers. What value of -d has been working for you?

rbelnap:
I've found 5 seconds gets me by - which is a far cry from namecheap's vague "you shouldn't hit this limit" comments.  It seems to be roughly 10 requests within 20 seconds will trigger it.  I'll take it up with namecheap as well.  

I wouldn't mind seeing it as a property of the provider (or set in the DSL perhaps) - I don't want to set an option on every invocation.

tlimoncelli:
Either a property of the provider (kind of like how BIND has default_ns defined (https://stackexchange.github.io/dnscontrol/providers/bind) or something listed in creds.json.  That way it doesn't have to be included in the DSL over and over.

captncraig:
This is a great idea. I still think the best idea is for the namecheap driver to handle this. We will be adding the ability to run all domains in parallel sometime soon, and this being in main makes that a bit trickier.

I'd recommend something in the namecheap package to throttle itself. Maybe a simple [time.Ticker](https://golang.org/pkg/time/#Ticker) would do:

```
//package level variable:
var throttle = time.NewTicker(time.Second) // or whatever delay between requests seems reasonable

//wherever we make a request:
<- throttle
http.Get(...
```

tlimoncelli:
I'm merging this.  It is sufficient for the intended purpose.  (i.e. a full rate-limiting solution would be grand, but it isn't going to happen any time soon.  It is probably easier to work with vendors to remove/increase rate limits than to code the perfect solution)

PR #103, opened 2017-04-27
updating cloudflare documentation

tiernano:
to show how to use proxy and alias. Took me a while to figure this out, though i would update the doc...

### DIFF
--- a/docs/_providers/cloudflare.md
+++ b/docs/_providers/cloudflare.md
@@ -48,3 +48,5 @@ D("example.tld", REG_NAMECOM, DnsProvider(CFLARE),
 ## Activation
 
 DNSControl depends on a Cloudflare Global API Key that's available under "My Settings".
+
+**Note** the domain must be in Cloudflare already before it can be worked on. DNSControl will not create new domains (currently) before working on them. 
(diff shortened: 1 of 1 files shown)


docs/_providers/cloudflare.md
```diff
@@ -39,7 +39,9 @@ var REG_NAMECOM = NewRegistrar("name.com","NAMEDOTCOM");
 var CFLARE = NewDnsProvider("cloudflare.com","CLOUDFLAREAPI");
 
 D("example.tld", REG_NAMECOM, DnsProvider(CFLARE),
-    A("test","1.2.3.4")
+    A("test","1.2.3.4"),
+    A("www","1.2.3.4", {cloudlfare_proxy:"on"}),
+    ALIAS("@","test.example.tld",{cloudflare_proxy:"on"})
 );
 {%endhighlight%}
```
tlimoncelli: I think a better practice is to define macros:

"""
var CF_PROXY_OFF = {'cloudflare_proxy': 'off'};     // Default/off.
var CF_PROXY_ON = {'cloudflare_proxy': 'on'};       // Proxy on.
var CF_PROXY_FULL = {'cloudflare_proxy': 'full'};   // Proxy on + railgun.
"""

Then you can do things like:

    A("test", "1.2.3.4", PROXY_ON),

(and switching things on and off become a lot easier).

tiernano:
Yea... makes sense. But given that's an advanced feature and this is more a quick start guide, would just showing how to get it working not be the idea?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

tlimoncelli:
We could define those macros in helpers.js and then individual users don't have to define them.

PR #104, opened 2017-04-27
updating getting started doc for multiple providers

tiernano:
tweaking the Getting Started doc for multiple providers.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -25,7 +25,20 @@ D("example.com", registrar, DnsProvider(bind),
 );
 {%endhighlight%}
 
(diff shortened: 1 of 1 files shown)


captncraig:
I'm not sure this is perfectly clear. 'name' is not very descriptive, and it is not used in the dnsconfig anywhere, so it is not clear what this new provider is doing there. 

I would like to see maybe some comments in the js like `//define your providers here`, `//now make a domain`, and so forth. The creds.json can similarly have comments in it.

tiernano:
thanks @captncraig. updated the naming and added some comments... hopefully it works!

tlimoncelli:
I just did a big rewrite of that file. I didn't add this kind of info specifically as I'm not sure if it would best first there or in nameservers.md.

tiernano:
@tlimoncelli I have updated my pull request with the document. Given that it is for multiple providers, not just DNS providers, but also registrars, leaving it where it is might be best...

tlimoncelli:
Thanks for the PR!

PR #107, opened 2017-04-28
NS RR (apex) TTL example

eliheady:
I stumbled a bit trying to set NS TTLs. This changes the TTL example to cover NS (apex) records and also to demonstrate DefaultTTL


tlimoncelli:
thanks for the PR!

PR #110, opened 2017-05-03
adding no_purge to cloudflare

tiernano:
my try at fixing #7 but only for cloudflare... new to Go, so still learning...


captncraig:
I'll merge this, and then make the correct fix.

PR #119, opened 2017-05-10
Cloudflare Redirects

captncraig:
Well this has quite a bit of features:

1. Ability to register custom record types that are only valid for one provider.
2. Added `CF_REDIRECT` and `CF_TEMP_REDIRECT` "virtual" record types to DSL.
3. Cloudflare provider can manage page rules for url forwarding.

From config:

```
D("kittens.world", reg,DnsProvider(cf),
    A("@","1.2.3.5", {cloudflare_proxy: "on"}),
    A("www","1.2.3.5", {cloudflare_proxy: "on"}),
    A("chat", "1.2.3.4",{cloudflare_proxy: "on"}),
    A("meta", "1.2.3.4",{cloudflare_proxy: "on"}),

    CF_TEMP_REDIRECT("chat.*kittens.world/*","https://chat.stackexchange.com/?host=electronics.stackexchange.com"),
    CF_TEMP_REDIRECT("meta.*kittens.world/*","https://judaism.meta.stackexchange.com/$2"),
    CF_TEMP_REDIRECT("*kittens.world/*","https://electronics.stackexchange.com/$2")
)
```

It runs: 

```
******************** Domain: kittens.world
----- Getting nameservers from: cloudflare
----- DNS Provider: cloudflare... 3 corrections
#1: CREATE PAGE_RULE kittens.world chat.*kittens.world/*,https://chat.stackexchange.com/?host=electronics.stackexchange.com,3,302 1
SUCCESS!
#2: CREATE PAGE_RULE kittens.world meta.*kittens.world/*,https://judaism.meta.stackexchange.com/$2,2,302 1
SUCCESS!
#3: CREATE PAGE_RULE kittens.world *kittens.world/*,https://electronics.stackexchange.com/$2,1,302 1
SUCCESS!
----- Registrar: namecheap
Done. 3 corrections.
```

Internally cloudflare provider generates a special record with a specially formatted target, so that it can leverage the common diff algorithm. Output can probably be improved.


providers/cloudflare/cloudflareProvider.go
```diff
@@ -193,6 +235,20 @@ func (c *CloudflareApi) preprocessConfig(dc *models.DomainConfig) error {
 				rec.Metadata[metaProxy] = val
 			}
 		}
+		// CF_REDIRECT record types. Encode target as $FROM,$TO,$PRIO,$CODE
+		if rec.Type == "CF_REDIRECT" || rec.Type == "CF_TEMP_REDIRECT" {
+			parts := strings.Split(rec.Target, ",")
+			if len(parts) != 2 {
+				return fmt.Errorf("Invalid data specified for cloudflare redirect record")
```
tlimoncelli: The error message should mention something about comma being a forbidden char.

tlimoncelli: Alternatively, encode the string as follows:

    DELIMITER + src + DELIMITER + dst

Set DELIMITER to comma. If comma is in src or dst, try other chars.

captncraig:
I realized something about this. It may potentially break someone who has manually added redirect rules, but has not added the matching `CF_REDIRECT` records to their config.

I have three options:

1. Leave it as is. Manually added redirects will be deleted.
2. Only check page rules if you have at least one `CF_REDIRECT` record in your domain. Will make it impossible to remove all via dnscontrol.
3. Make it "opt-in" on a per-domain basis. Either by including redirect records, or by adding some kind of {"cloudflare_manage_pagerules": "true"} metadata to the domain.

tlimoncelli:
We should do whatever is the least surprising to users.

That removes option 1.   Option 2 might be a surprise... and certainly not being able to delete all page rules is a bad thing.

How about:

4.  Default to `{"cloudflare_manage_pagerules": "false"}`
If either CF_REDIRECT or CF_TEMP_REDIRECT exists, this implies `{"cloudflare_manage_pagerules": "true"}`.
People can set `{"cloudflare_manage_pagerules": "true"}` if they want to delete all pagerules.
We can make and alias called `CF_MANAGE_PAGERULES()` that means `{"cloudflare_manage_pagerules": "true"}`.

captncraig:
Realized this should probably be a provider-level metadata field. Should be something like:

```
var CFLARE = NewDnsProvider("cloudflare.com","CLOUDFLAREAPI", {manage_redirects: "true"});
```

Then all domains on the provider will have redirects managed. If you want to choose on a domain-by-domain basis, maybe need two providers.

I think that strikes a pretty good balance between least-surprise and avoiding repetative boilerplate.

tlimoncelli:
+1 for making it a provider-level metadata field.

PR #123, opened 2017-05-23
DNS Resolvers

tlimoncelli:
Hi @captncraig!  I'd appreciate some feedback on the code so far.  The cmd/spftest/main.go program is just temporary and will be deleted eventually.


spf/dnscache.go
```diff
@@ -0,0 +1,26 @@
+package spf
+
+type dnsCache map[string]map[string][]string
```
captncraig: We should document the dimensions here. Looks like it is [name][rType] -> answers. Some questions:

1. Do we intend to ever do A/MX resolving, or just TXT? If just TXT, we can probably eliminate the record type dimension.
2. Why store all the TXT responses? If we filter to those with prefix `v=spf1` there should be 0 or 1. This could be as simple as `map[string]string`.

spf/dnscache.go
```diff
@@ -0,0 +1,26 @@
...
+
+func (c *dnsCache) dnsGet(label, rtype string) ([]string, bool) {
```
captncraig: dnsCache is a map type, so these functions can be declared `func (c dnsCache)` and save some `*`'s in the body.

captncraig: These functions can probably be named simply `get` and `put`

spf/resolver.go
```diff
@@ -0,0 +1,81 @@
...
+func (c *dnsLive) Close() {
+	// Write out and close the file.
+	m, _ := json.MarshalIndent(c.cache, "", "  ")
+	m = append(m, "\n"...)
+	ioutil.WriteFile(c.filename, m, 0666)
+}
+
+// The "Pre-Cached DNS" Resolver:
+
+type dnsPreloaded struct {
```
captncraig: this type may not be neccessary. If we just implement `GetTxt` on the `dnsCache` itself, it can serve as a resolver.

spf/dnscache.go
pmoroney: Just some speculation, but if we had the ability to do A and AAAA lookups, we may be able to do some CNAME flattening. Which would of course require some automation or something to keep it updated regularly, but a simple cron job may suffice.

captncraig: Correct, we could do a similar thing. Our current plan is to support spf flattening in some form, but we are quite opposed to anything in the regular flow that actually depends on DNS lookups. Hence the cache here, that would presumably be updated daily on cron or something like that.

captncraig: I don't see this as a general purpose dns cache as much as a utility for caching spf txt records.

captncraig: But maybe it could be both in time.

tlimoncelli: At this time code only uses it to cache TXT records. However I future-proofed it to support many record types.  CNAME flattening is a good example of a future use.

spf/dnscache.go
tlimoncelli: Renamed to get/put

PR #132, opened 2017-06-06
CAA support

whs:
This PR add [CAA](https://en.wikipedia.org/wiki/CAA_record) support.

CAA is supported by Google Cloud DNS. (I'm not sure whether you would have to get in DNSSEC beta to gain access or not, but it was just a simple Google Forms before I could get in) Also, BIND obviously supported this.

I almost never wrote Go before (and this is my first Go PR), so if you could be more specific when suggesting changes I'd be appreciated.

Design decision that I think worth discussing:

- I implement CAA with a metadata field `caa_tag` plus a new struct field `CAAFlag`, as the flag needs to be int. Should I move the tag into the struct as well?
   - edit: After reviewing the SRV changes I moved the tag into the struct
- Or, should CAA be implemented without special parsing like TXT records? (i.e. `CAA("@", "0 issue \"letsencrypt.org\"")`)
- Should CAA tag be validated? I think that it would be more future-proof without validation.
- Flag is implemented by modifier because most of the time people would probably use no flag (0 in record). However, this cause mismatch in function definition to record ordering (`CAA("@", "iodef", "mailto:test@example.com", CAA_CRITICAL)` vs `@ CAA 1 iodef "mailto:test@example.com"`). Should I move flag into the function definition?
- The target field are quoted automatically. Is this too much magic?


whs:
I'm getting a crash with bind. I'll try to fix this later:

```
github.com/StackExchange/dnscontrol/providers/bind.rrToRecord(0xe9d3a0, 0xc42013c230, 0xc4203baff0, 0x9, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/whs/build/go/src/github.com/StackExchange/dnscontrol/providers/bind/bindProvider.go:78 +0x53e
github.com/StackExchange/dnscontrol/providers/bind.(*Bind).GetDomainCorrections(0xc420057ab0, 0xc4203545b0, 0xc420272174, 0x4, 0xc4202fb298, 0x1, 0x0)
	/home/whs/build/go/src/github.com/StackExchange/dnscontrol/providers/bind/bindProvider.go:179 +0x4a8
main.main()
	/home/whs/build/go/src/github.com/StackExchange/dnscontrol/main.go:187 +0xcdc
```

whs:
I've fixed the issue with parsing CAA records in BIND

tlimoncelli:
Adding new rtypes will become a lot easier once https://github.com/StackExchange/dnscontrol/tree/tlim_srv is merged. At that point, we should rebase.  That particular panic might go away :-)

tlimoncelli:
https://github.com/StackExchange/dnscontrol/tree/tlim_srv is merged. Please rebase.  It should be a lot easier to add CAA now.

whs:
I'll try to get this done tomorrow.

whs:
Rebased. It seems that master is currently broken though.

models/dns.go
```diff
@@ -71,6 +71,8 @@ type RecordConfig struct {
 	SrvPriority  uint16            `json:"srvpriority,omitempty"`
 	SrvWeight    uint16            `json:"srvweight,omitempty"`
 	SrvPort      uint16            `json:"srvport,omitempty"`
+	CaaTag       string            `json:"caa_tag,omitempty"`
```
tlimoncelli: The json tag should be caatag and caaflags

docs/_functions/domain/CAA.md
```diff
@@ -0,0 +1,34 @@
...
+parameters:
+  - name
+  - tag
+  - value
+  - modifiers...
+---
+
+CAA adds a CAA record to a domain. The name should be the relative label for the record. Use `@` for the domain apex.
+
+Tag can be one of "issue", "issuewild" or "iodef". Value depends on the tag, and is automatically quoted.
```
tlimoncelli: It is unclear what "automatically quoted" means.  The DSL requires quotes on strings. Do you mean automatically escaped?

docs/_functions/domain/CAA.md
```diff
@@ -0,0 +1,34 @@
...
+
```
tlimoncelli: What about "target"?   Is it like NS or CNAME where if it doesn't end in a "." the domain will be added?  I assume not, but it would be good to note.

pkg/js/helpers.js
```diff
@@ -156,6 +170,18 @@ function ALIAS(name, target) {
     }
 }
 
+// CAA(name,tag,value, recordModifiers...)
+function CAA(name, tag, value){
+    checkArgs([_.isString, _.isString, _.isString], arguments, "CAA expects (name, tag, value)")
```
tlimoncelli: Clarification: CAA expects (label, tag, target) plus optional flag.

docs/_functions/domain/CAA.md
tlimoncelli: If tag can only be "issue", "ussuewild" or "iodef", then NormalizeAndValidateConfig() should produce an error if it is anything else.

pkg/js/helpers.js
whs: The flag is implemented as a record modifier (eg. `CAA_CRITICAL`) since it is optional. Should I add it to the function definition instead?

docs/_functions/domain/CAA.md
whs: What I meant is that the DNS record requires a quote around the value (eg. `issue "letsencrypt.org"`) which is done automatically (i.e. `CAA("@", "issue", "letsencrypt.org")` rather than `CAA("@", "issue", "\"letsencrypt.org\"")`).

English is not my primary language though, so please feel free to suggest improvements. Or, I think this can be omitted since the auto quoting behavior is similar to TXT records. (i.e. leave the DNS technical details to the library)

whs:
Thanks for your suggestions, I'll try to get this done by tomorrow.

pkg/js/helpers.js
tlimoncelli: I think I was unclear. I was referring to the error message.

    checkArgs([_.isString, _.isString, _.isString], arguments, "CAA expects (label, tag, target) plus optional flag as a meta argument")

That makes me think....

Usually the order of a record's data is "label", priorities and stuff (if it exists), "target", { "meta": "value"}. This is the first record there is data after the target that isn't metadata.  

On the one hand, it seems to be setting a bad precedent to have actual RFC-defined elements of the RR after the target.  On the other hand, it seems like the "flag" is something that will be rarely used, and should stand out visually (like by adding CAA_CRITICAL at the end).

So it comes down to the decision between these two syntaxes:

    CAA("@",  flag, "tag", "target", {meta:value))

or

    CAA("@",  "tag", "target", flag, {meta:value})

I'm not sure which is better.

One that thing would help us decide:

  * How frequently do you think CAA_CRITICAL will be used?
  * What syntax do you think will be the most natural to novices?
  * What will make the most sense to someone trying to translate between ISP/Cloud/Cert provider documentation and the DNSControl DSL?  For example, if someone reads this page https://support.google.com/a/answer/174125?hl=en it is pretty obvious what to do with MX().  Eventually providers will make similar documents explaining what to do with CAA record.  If you were to write a hypothetical document, what would it look like?  Would it easily translate into CAA() format?

CC: @captncraig

docs/_functions/domain/CAA.md
tlimoncelli: Ah!  Now I understand the technical issue.

I would write:

    Tag can be one of "issue", "issuewild" or "iodef".

   Value is a string. The contents depend on the tag. If it includes spaces or other characters that must be escaped, DNSControl will do that for you (similar to TXT records). For example use `CAA("@", "issue", "letsencrypt.org")` rather than `CAA("@", "issue", "\"letsencrypt.org\"")`.

docs/_functions/domain/CAA.md
whs: The target field doesn't always have to be a domain; the meaning is depending on which tag is used.

For example, issue/issuewild also accept an optional parameter: `issue "example.net; acme-methods=dns-01,http-01"`. iodef accept either http/https URL or mailto URL.

I think it is ok to leave target unparsed, except for automatic quoting (speaking of which, I haven't add escaping for `"`. I'll have to add that)

pkg/js/helpers.js
whs: - The flag field is actually a bitfield of 8 bits. Right now only the first bit is used for critical, but in the future there might be multiple flags active at once. Since we already have variable meta options, making flags variable as well could be hard to implement.
  - One possible way is to define CAA_CRITICAL as an integer instead of a meta setter function, and add a special case to `addRecord` similar to how MX priority is implemented right now. The user then can use binary operator to combine flags (`CAA_CRITICAL | CAA_FUTURE_USE`)
- The above suggestion also help users in translating to dnscontrol format as they also can write the numbers directly, but they would need to add it at the end instead of the given position. The docs could indicate this by writing the method definition as `CAA(name, tag, value[, flags], recordModifiers...)`.

tlimoncelli: @captncraig  Thoughts?

captncraig: I like the integer idea, even though I kinda hate the MX special case. 

I like the user-friendliness of `CAA("@", "tag", "target", {meta: "whatever"},  CAA_CRITICAL | CAA_FUTURE_USE)`. Or alternatively, have flags be an arg to a function:

`CAA("@", "tag", "target", {meta: "whatever"},  CAA_FLAGS(CAA_CRITICAL | CAA_FUTURE_USE))`

tlimoncelli: I certainly don't want people to have to do binary math. :-)

Let's keep it as-is.

    CAA("@", "tag", "target", {meta: "whatever"}, CAA_CRITICAL, {more_meta:"foo")

The json should show an integer, not CAA_CRITICAL.  That way we can add more flags (CAA_FUTURE_USE) easily.

pkg/js/parse_tests/014-caa.json
```diff
@@ -0,0 +1,48 @@
...
+                    "name": "@",
+                    "caatag": "issue",
+                    "caaflag": 0,
+                    "target": "letsencrypt.org"
+                },
+                {
+                    "type": "CAA",
+                    "name": "@",
+                    "caatag": "issuewild",
+                    "caaflag": 0,
```
tlimoncelli: Since caaflags is "omitempty", I"m not sure this `"caaflag": 0,`  appears here.

Ah, I see the problem. The actual tag is "caaflags" (with an "s").  How about making it "caaflag" everywhere to match the dns.CAA struct?

whs: Ah, that's a typo. The field is actually `caatags`. (I didn't really test this file before I push as master already have it broken)

whs:
That should be everything. I've ran the integration test against BIND and GCLOUD.

tlimoncelli:
I modified the PR and merged it in https://github.com/StackExchange/dnscontrol/pull/161
(I think that if I knew git better I could have done this in one branch but I have a lot to learn about git).

PR #139, opened 2017-06-11
Gandi does not support TTLs > 30 days

tlimoncelli:
@captncraig  Recommendation of how to write a test for this?


captncraig:
Unless we want to make this a global limit for all records, I think this is probably sufficient.

PR #141, opened 2017-06-12
added full context path - fixes #140

### LINKED ISSUE #140: package context: unrecognized import path "context"
When I try to install the package I get 

`package context: unrecognized import path "context" (import path does not begin with hostname)`

tibu:
Added full context path to build it correctly on go1.6

### DIFF
--- a/providers/google/google.go
+++ b/providers/google/google.go
@@ -1,9 +1,9 @@
 package google
 
 import (
-	"golang.org/x/net/context"
 	"encoding/json"
 	"fmt"
+	"golang.org/x/net/context"
 	"strings"
 
 	gauth "golang.org/x/oauth2/google"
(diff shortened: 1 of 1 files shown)


captncraig:
Is there a particular reason you cannot upgrade past go 1.6? I am hesitant to add support for it for a variety of reasons, and I'm actually kinda surprised this is the only thing that doesn't work.

captncraig:
Added note to readme that dnscontrol requires go 1.7+. Supporting 1.6 at this point would require a little more workaround than just this change, and we probably won't put in the effort to support more than the two most recent version of Go or so.

tibu:
OK, no worries. I used go 1.6 because that was supported by default on my Ubuntu, but I'll find out how to upgrade it.

Issue #76, opened 2017-04-13
Support PTR records


dattas:
Hello,

It'd be nice to be able to use this tool to also create reverse zones. Reverse zones use PTR records which are basically CNAME records. Can we set this up so that PTR records are like CNAME?

Examples are like:
```
2               IN      PTR     blah1.domain.com.
3               IN      PTR     blah2.domain.com.
4               IN      PTR     blah3.domain.com.
5               IN      PTR     blah4.domain.com.
6               IN      PTR     blah5.domain.com.
```

Currently it generates this error: `ReferenceError: 'PTR' is not defined`

Thanks!

tlimoncelli:
We should probably add PTR() records asap.

captncraig:
Would it make sense to have a mode where we automatically create PTR records for A records?

Option A: everything explicit:
```
D("example.com", myReg, ...,
  A("foo","1.2.3.4"),
  PTR("1.2.3.4.in-addr.arpa", "foo") //not sure if the in-addr.arpa part is necessary.
```

Option B: magic!
```
D("example.com", myReg, ...,
  A("foo","1.2.3.4"),
  AutoPTR, //generates PTRs as above. Essentially reverse of all A/AAAA in the domain.
  PTR("5.5.5.5", "foo.otherdomain.com.")
}
```

captncraig:
Ok, talked to Tom, here's the plan:

1. PTR records will be supported in all domains: `PTR("23","foo.google.com.")`.
2. Add a `REV` function to transform a domain name:
   `D(REV("174.136.107.0/24")...)` is equivalent to `D("107.136.175.in-addr.arpa"...)`
3. Make sure all providers that support PTR can manage them. Once we get BIND passing the test, convertzone should work as intended too.

f0:
@captncraig   any progress on this?

tlimoncelli:
Most of this is done. We just haven't tested it with any providers.  Which provider are you using?  We can test that first.

dattas:
I'm personally using plain zonefiles (bind)

bhuntley-cu:
BIND, CloudDNS and Route53 here, thanks!

f0:
plain bind zonefiles

tlimoncelli:
@bhuntley-cu @f0 @dattas  I've got BIND working.  I'll push that after @captncraig does a code review. If you want to try what's in the branch and give feedback, I'd appreciate it.

[...]

PR #146, opened 2017-07-05
Enable PTR records for BIND driver

tlimoncelli:
Partial implementation of https://github.com/StackExchange/dnscontrol/issues/76


captncraig:
Looks good. 2 things missing:

1. Validation that all fqdns are full ip addresses of the proper length.
2. A flag so we avoid passing or runnign with providers that don't support it. 

I can merge this and work on those items.

PR #147, opened 2017-07-06
Support passing the JSON as a parameter

### LINKED ISSUE #73: Support passing the intermediate JSON as a command line parameter?
So I already have my own internal project that generates DNS zones using a DSL in Ruby. It outputs bind files currently, but I heard about DnsControl and thought I might be able to reuse it's providers support.

evaryont:
This allows dnscontrol to be able to be used with other tools that
support outputting to it's intermediate JSON format.

Closes #73

### DIFF
--- a/main.go
+++ b/main.go
@@ -52,7 +52,8 @@ func main() {
 		if err != nil {
 			log.Fatalf("Error reading %v: %v\n", *jsonFile, err)
 		}
-		if err = json.Unmarshal([]byte(text), dnsConfig); err != nil {
+		err = json.Unmarshal(text, &dnsConfig)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


evaryont:
This PR doesn't handle validation of the incoming JSON beyond what already happens.

So passing in unsupported record types will error out as expected: `ERROR: Unsupported record type (CAA) domain=example.com name=www`. Adding extra keys to the various dictionaries cause them to be silently dropped, which I think is wanted.

tlimoncelli:
This looks good.  Thanks for the contribution!

Note to self:

* We need better docs for the flags
* The flags should be renamed.  They were added organically with no plan.

PR #151, opened 2017-07-12
Route53 provider: Pass the right correction message based on action

### LINKED ISSUE #150: Route53 provider prints wrong correction message on DELETEs
The route53 provider prints the wrong correction message when deleting records. 

Specifically, when both deleting and creating/modifying records, the `DELETE` correction just prints the `CREATE`/`MODIFY` correction message again. And when only `DELETE`ing records, a blank correction message is printed.

It appears to be because of [`providers/route53/route53Provider.go:231`](https://github.com/StackExchange/dnscontrol/blob/master/providers/route53/route53Provider.go#L231) always printing `changeDesc` despite using `delDesc` on [`:193`](https://github.com/StackExchange/dnscontrol/blob/master/providers/route53/route53Provider.go#L193) to track the DELETE messages.

To be clear, it is taking the correct actions when running a push, but it prints misleading messages. I'm happy to make a PR for this issue, but it almost certainly wouldn't be the correct idiomatic golang solution.

Thanks,
Norm

maclennann:
Fixes #150.

Correction messages are tracked in different variables between UPSERTs and DELETEs. Formerly, this lead to the wrong correction message (or a blank correction message) being printed when performing DELETEs.

I don't know if this is the correct idiomatic solution, but it does print the right messages now.

### DIFF
--- a/providers/route53/route53Provider.go
+++ b/providers/route53/route53Provider.go
@@ -224,14 +224,15 @@ func (r *route53Provider) GetDomainCorrections(dc *models.DomainConfig) ([]*mode
 	changeReq := &r53.ChangeResourceRecordSetsInput{
 		ChangeBatch: &r53.ChangeBatch{Changes: changes},
 	}
+
 	delReq := &r53.ChangeResourceRecordSetsInput{
 		ChangeBatch: &r53.ChangeBatch{Changes: dels},
 	}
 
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Awesome!  Thanks for catching that ... and providing the fix!!

Issue #68, opened 2017-04-03
Support Route53 as a registrar


ddrght12345:
`Error creating registrars: Registrar type ROUTE53 not declared.`
Self explanatory?
We just recently moved to r53, so this would be nice.

Since I have been wanted to learn Go, this may be a good chance for me, but Im not sure when I would be able to start this, in case someone else is capable of doing this faster/before I can start learning

captncraig:
Hey, this should be pretty easy to implement at this point, we have just never had the need for it. It is really as simple as implementing a single function on the [route53Provider](https://github.com/StackExchange/dnscontrol/blob/master/providers/route53/route53Provider.go#L18) type to set the nameservers appropriately. 

Just implement `GetRegistrarCorrections(dc *models.DomainConfig) ([]*models.Correction, error)` and register it appropriately, and it should work. See the [namecheap](https://github.com/StackExchange/dnscontrol/blob/master/providers/namecheap/namecheap.go#L32) implementation for an easy reference.

I could do it, but if you would like the opportunity to learn and contribute it is all yours.

ddrght12345:
Oh. Fantastic!
Then I may try to tackle this, after all. 
Thanks for the reference

captncraig:
closed via #155

PR #155, opened 2017-07-18
#68 Registrar support for Route53 provider

maclennann:
I'd like to get some feedback on this before I go any further since I'm not particularly proficient in golang.

This is my attempt at addressing #68, adding registrar support for Route53.

The code as-is below works and I've tested it with my own Route53 Domain. However, I'm not sure how to proceed around unit testing, etc or if it's even feasible. I have a couple other questions about the idiomatic way of doing things (both golang-idiomatic and the desired UX for dnscontrol) that I'm going to pose below in the diff.

I'd appreciate any feedback you can provide.


providers/route53/route53Provider.go
```diff
@@ -13,27 +14,41 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	r53 "github.com/aws/aws-sdk-go/service/route53"
+	r53d "github.com/aws/aws-sdk-go/service/route53domains"
+	"github.com/aws/aws-sdk-go/private/waiter"
```
maclennann: Is it kosher to use "private" packages like this? My gut says no, but waiters are super-convenient (related to another question below).

providers/route53/route53Provider.go
```diff
@@ -305,3 +385,31 @@ func (r *route53Provider) EnsureDomainExists(domain string) error {
 	return err
 
 }
+
+func (r *route53Provider) waitUntilNameserversUpdate(operationId *string) error {
```
maclennann: This is related to the desired UX for dnscontrol.

Updating nameservers in route53domains returns an operationid immediately, but doesn't complete for ~30-45 seconds. Is your expectation that DNS control would just see that the request completed successfully, or is it okay to wait until we have confirmation that the change was successfully made?

vendor/vendor.json
```diff
@@ -182,6 +194,12 @@
 			"revisionTime": "2016-08-11T16:24:59Z"
 		},
 		{
+			"checksumSHA1": "qS3eZZfhdDM6haTZp7OeyhMDlV8=",
+			"path": "github.com/aws/aws-sdk-go/service/route53domains",
+			"revision": "f80e7d0182a463dff0c0da6bbed57f21369d4346",
```
maclennann: This revision is ~1 year old at this point (1.4.x vs 1.10.x). I needed to use this particular commit to retain compatibility with the rest of the library. I'm pretty sure problems happened with the shared parts of the SDK if I tried anything else. At what point does it make sense to update aws-sdk-go as a whole, or is it more common to update individual packages within it piecemeal?

captncraig:
Awesome! I will review this today. I'll also talk to tom about buying a test domain to try it out.

providers/route53/route53Provider.go
captncraig: This is a very good question. I definitely do not love the code required to wait, so I'd rather not. I'm curious about the exact semantics though. 

The update operation returns an "Operation ID"? Is that a commitment that they will make the change, or are there situations where it gets lost? If we are just waiting for them to update the tld nameservers, I think we could probably forego this. No other registrars really give real-time guarantees, and ttls on those records can be fairly long either way. But we do want to wait until we are highly confident it will go through in the future.

I'm torn. Thoughts @tlimoncelli ?

vendor/vendor.json
captncraig: I'd be ok updating everything that needs it. We can run the integration tests before merging to make sure everything else still works.

providers/route53/route53Provider.go
maclennann: I'm not exactly sure what the possibilities are for operations. [According to the API](https://github.com/aws/aws-sdk-go/blob/6a6d169fb0936d588240e44238579aaa32842863/models/apis/route53domains/2014-05-15/api-2.json#L1029), operations can still still fail. I'm not sure what might cause failures, though.

Validation failures (like having duplicate nameservers) will fail before even generating an operation id, so I don't know what scenarios there are for failure after the operation has been created.

captncraig: If you query the nameservers via api after getting an operation id but before the wait completes, what do you get? If you get the updated nameservers, I'd vote for not bothering to wait. 

In other words, as long as I can run dnscontrol twice in a row and get no corrections on the second run, I'm probably ok with that consistency model.

vendor/vendor.json
maclennann: Cool, thanks for the info.

I don't think anything currently needs it (for this PR anyway) - I was just curious. There have been some changes to the route53domains api in the intervening year, but nothing to do with fetching or updating nameservers so far as I can tell.

providers/route53/route53Provider.go
maclennann: I just tried this out with the CLI. I have no reason the believe the experience would be any different for the SDK.

I issued a `update-domain-nameserver` and then started watching the `get-domain-detail`. It still showed the old NS records until the operation actually completed successfully.

I also managed to get a failed operation when trying to set the NS records to unresolvable names. So it turns out it definitely is possible for operations to fail. It just gives you `FAILED` on the CLI - you need to check your domain contact's email address to figure out _why_ it failed. Gross. But more of a reason to ensure it succeeded before calling dnscontrol's job done.

tlimoncelli:
A lot of things in DNS don't propagate very quickly.  The API is updating a database that is used to generate actual DNS service and there is always a lag between database updates and the real world. The way we deal with this in DNSControl is to only deal with the database. If the API says that an update is complete, we assume it is true.  Nothing in DNSControl relies on doing actual DNS queries.

maclennann:
My only concern is that the API in this case is asynchronous and it won't reflect your update for a relatively-substantial.

While your operation is in progress, fetching the nameservers directly from the API still returns the old values. And even once an operation has been accepted by the API, there are still circumstances under which it can fail to update at all.

So even though dnscontrol is querying the API for the nameservers, there is still a minute or so where subsequent runs may return inaccurate information.

If you all are fine with that, I'm happy to concede the point (I don't plan on performing multiple back-to-back runs that quickly anyway). But I thought it was worth calling out.

captncraig:
I vote we leave the wait in for now. Changing nameservers should be relatively infrequent, and we want to make sure the change will in fact get committed.

maclennann:
Anything else you need from me to get this rolling again?

tlimoncelli:
@captncraig and I both approve the "wait".

@maclennann is it ready for merge?

providers/route53/route53Provider.go
```diff
@@ -12,28 +13,42 @@ import (
...
+		Region: aws.String("us-east-1"),
 	}
 
 	if keyId != "" || secretKey != "" {
 		config.Credentials = credentials.NewStaticCredentials(keyId, secretKey, "")
 	}
 	sess := session.New(config)
 
-	api := &route53Provider{client: r53.New(sess)}
+	api := &route53Provider{client: r53.New(sess), registrar: r53d.New(sess)}
```
captncraig: I think this may run into issues if you are only using domains or only using r53, depending on the access level of the token. Perhaps there should be a seperate type for registrar and dns providers?

maclennann: If you aren't using R53 as a registrar, it shouldn't attempt any route53domains actions. You can see on `:52` that it does a `GetZones` to ensure the key is valid, but I didn't add any similar route53domains-y check. So if you try to use it as a registrar without the proper IAM permissions, you'll get an `AccessDeniedException` when it tries to do registrar stuff. But if you don't have r53domains as your registrar, everything should proceed as it has in the past. See below for an example.

If there's a good way to tell ahead of time if the provider is being instantiated as a dsp, a registrar, or both, I'd be happy to add a conditional r53domains access check. Other I'd be willing to break the registrar out into a separate type. But the short answer is that if we do nothing, everything should continue to Just Work for everyone using the r53 dsp.

```bash
# This profile has no IAM permissions to route53domains.
# With route53domains selected as a registrar:
$ ./dnscontrol-Linux preview
Initialized 1 registrars and 1 dns service providers.
******************** Domain: [snip]
----- Getting nameservers from: r53
----- DNS Provider: r53... 1 correction
#1: [snip]

----- Registrar: r53
Error getting corrections: AccessDeniedException: User: arn:aws:iam::868399730840:user/dnscontrol is not authorized to perform: route53domains:GetDomainDetail
	status code: 400, request id: 48b534a1-7902-11e7-afa6-a3fffd2ce139
Done. 1 corrections.

# With none selected as a registrar
$ ./dnscontrol-Linux preview
Initialized 1 registrars and 1 dns service providers.
******************** Domain: [snip]
----- Getting nameservers from: r53
----- DNS Provider: r53... 1 correction
#1: [snip]

----- Registrar: none
Done. 1 corrections.
```

maclennann:
This is good to go, as far as I'm concerned pending a resolution to [the above discussion](https://github.com/StackExchange/dnscontrol/pull/155/files#r131229727).

PR #156, opened 2017-07-19
add ALIAS support for Name.com provider

pmoroney:
Name.com just implemented ALIAS records, which we are calling ANAME records to match with the current draft https://tools.ietf.org/html/draft-ietf-dnsop-aname-00.


tlimoncelli:
Thanks for the PR!  I'm glad you were tracking this.

PR #158, opened 2017-07-20
add the new ANAME record type to the records list

pmoroney:
Since this ANAME/ALIAS record is new,

### DIFF
--- a/providers/namedotcom/records.go
+++ b/providers/namedotcom/records.go
@@ -106,7 +106,7 @@ func (r *nameComRecord) toRecord() *models.RecordConfig {
 		Original: r,
 	}
 	switch r.Type {
-	case "A", "AAAA", "CNAME", "NS", "TXT":
+	case "A", "AAAA", "ANAME", "CNAME", "NS", "TXT":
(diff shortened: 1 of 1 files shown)


tlimoncelli:
This looks great so far!

Not all providers support ANAME. Please mark which providers do support it:

* providers/providers.go: Add the constant CanUseANAME
* pkg/normalize/validate.go checkProviderCapabilities(): add ANAME to the list.
* Any provider that supports ANAME, mark this in providers/*/*provider.go

Basically if you search for CanUsePTR and CanUseSRV, you'll see what to do.

pmoroney:
ANAME is what we are calling ALIAS records here at name.com, and we are already passing `providers.CanUseAlias` to the register function. I'm just adding the type to these functions since they list the valid types now because of the SRV change that got merged after I added support for this in https://github.com/StackExchange/dnscontrol/pull/156

pmoroney:
Also it seems I got distracted and never finished my initial description up above ;)

tlimoncelli:
Oh i see!  Yes, that makes sense now!

Issue #157, opened 2017-07-19
Debt: Consolidate mod processing in addRecordSRV/addRecord


tlimoncelli:
There is a lot of repeated code in addRecordSRV/addRecord.  It should be collapsed into a helper function.

whs:
I could take on this.

I'm thinking of something like:

```js
var CNAME = recordBuilder({
    type: 'CNAME',
});
```

which should take care of the (name, target) type.

For special cases like SRV a hook could be added

```js
var SRV = recordBuilder({
    type: 'SRV',
    args: [
		['name', _.isString], // second argument can be omitted/null if validation is not needed
		['priority', _.isNumber],
		['weight', _.isNumber],
		['port', _.isNumber],
		['target', _.isString],
	],
    transform: function(record, args, modifiers){
		// record is {type: 'SRV', ttl:d.defaultTTL, meta:{}}
		// with modifiers already applied
		// args will be an object for parameters defined

		record.name = args.name;
		record.srvpriority = args.priority;
		record.srvweight = args.srvweight;
		record.srvport = args.srvport;
		record.target = args.target;

		return record;
	},
	// example of number modifier parsing (eg. in CAA)
	modifierNumber: function(record, value){
		record.caaflags |= value;
		return record;
	}
});
```

Record could implement `modifierNumber` and `modifierString` `modifierBoolean` (basically all primitive types but null/undefined) to support custom record modifier like what MX/CAA are using now. The default handler for those options will log "Warning: Number/String/Boolean modifier (value) is not supported. Skipping"

Is there any use case that I missed?

captncraig:
I was thinking even simpler. Have addRecord return an object, and SRV just adds some fields to it after it finishes.

whs:
That still retain the special cases of number modifier parsing for CAA/MX. As addRecord will warn on unknown modifier, handling the number modifiers in the CAA function will also trigger the warning.

tlimoncelli:
Fixed in #157 and #163

PR #163, opened 2017-07-31
Refactor addRecord to recordBuilder (#157)

whs:
- recordBuilder is added. The design could be found in #157 comment.
- All existing record types are converted to use recordBuilder.
- I added `@deprecated` to addRecord and its helper function `getModifiers`, both are no longer used in helpers.js. In the future we could add a deprecation warning.
  - addRecord have its special cases handling (MX, CAA) removed
- addRecordSRV is removed. It is new enough that I don't think anyone would depends on it directly

I believe this PR doesn't affect backwards compatibility. (All existing tests are passing without modification, except one that is already failing in master which I fixed in another commit)


eliheady:
Minor fix in helpers.js:SRV, transform "args.value" should be "args.target"

`diff --git a/pkg/js/helpers.js b/pkg/js/helpers.js`
`index 4e46685..2f4d4c4 100644`
`--- a/pkg/js/helpers.js`
`+++ b/pkg/js/helpers.js`
`@@ -192,9 +192,64 @@ var SRV = recordBuilder({`
`         record.srvpriority = args.priority;`
`         record.srvweight = args.weight;`
`         record.srvport = args.port;`
`-        record.target = args.value;`
`+        record.target = args.target;`
`    },
`});

whs:
Fixed, thanks

tlimoncelli:
Wow!  I just came back from vacation and saw this huge PR.  Looks good so far. Let me do a little more testing and get back to you.

captncraig:
I guess I'm still just not clear on why this approach is better than simply having `addRecord` return the object it creates, and then something like MX can be done without the odd hard-coded modifier like so:

```
function MX(name, priority, target) {
    checkArgs([_.isString, _.isNumber, _.isString], arguments, "MX expects (name, priority, target)")
    var mods = getModifiers(arguments,3)
    return function(d) {
        var r = addRecord(d, "MX", name, target, mods);
        r.mxpreference = priority;
    }
}
```

I do appreciate that your way can consolidate the validation / modifier stuff a little more, but I don't think its really that much repetition to worry about. I am uncomfortable with the big options object that changes behavior pretty significantly by adding various fields. Even with documentation, it feels heavy.

captncraig:
Actually, my way is really not any different than your way now that I look at it. 

If anything I would probably say we should break a few more things out of the options object into top level arguments to recordBuilder.

I feel like at a minimum we should have top level parameters for Type, argument validation metadata, and the modifier function.

whs:
I actually realize what you were talking about while implementing this PR. In fact, IMPORT_TRANSFORM was already using that. (By that point I was almost finished I decided to submit the PR anyway and we could discuss the pros/cons of this solution and maybe develop a better solution)

Another design I considered was making record a class, and record types can override needed method. I'm open to refactoring this PR to classes. The problems with this design is:

- The JS runtime in use doesn't support ES6, so we have to use the prototype syntax. I don't feel it is much of a different to the current design except that it is standardized.
- Class requires `new`. It is possible to make class auto detect if it is called without new and retry automatically. That would be a boilerplate for class implementors though

As for the reason why everything is in one argument, it's a common pattern in JavaScript as it allow [the API to change easily without leaving in unused parameters for backwards compatibility](http://www.ericfeminella.com/blog/2013/01/22/simplifying-designs-with-parameter-objects/). I do agree though that type should be moved to the top level as all users needs to change it.

captncraig:
Honestly, I'd rather not have the `modifierBoolean`, `modifierNumber`, or `modifierString` at all.

Only the number one is used anywhere. Once in import_transform, where it should be removed likely (no need to do ttl special for just that type), and once in CAA, where it is very odd as well. I'd rather not accept a variable number of args for the flags, especially with how non-standard it makes everything here. Even though that decision was made elsewhere, I'm willing to kill it now. 

I started breaking it down into seperate args, and I kinda agree with your way now, since the name of each part is written where you use it, rather than just a list of random arrays and functions.

whs:
Updated the code:

- I just realized that current CAA flag handling is broken (the flag is named caaflag in the Go side, and caaflags in JavaScript side) and the test never caught this. I've fixed and updated the tests.
- CAA flags are now record modifier
- IMPORT_TRANSFORM no longer have TTL support. This will probably break something, but it seems that this record type is not even documented.
- With special cases gone, primitive modifiers support is now removed
- The type argument is also moved to top level argument.

pkg/js/helpers.js
```diff
@@ -274,15 +262,17 @@ function format_tt(transform_table) {
...
-function IMPORT_TRANSFORM(translation_table, domain,ttl) {
-    return function(d) {
-        var rec = addRecord(d, "IMPORT_TRANSFORM", "@", domain, [
-            {'transform_table': format_tt(translation_table)}])
-        if (ttl){
-            rec.ttl = ttl;
-        }
-    }
-}
+var IMPORT_TRANSFORM = recordBuilder('IMPORT_TRANSFORM', {
```
tlimoncelli: In this code the TTL parameter to IMPORT_TRANSFORM() is removed. It needs to be included. The syntax shouldn't change.

whs: Ok, I added TTL back.

pkg/js/helpers.js
```diff
@@ -294,6 +286,9 @@ function NO_PURGE(d) {
   d.KeepUnknown = true
 }
 
+/**
+ * @deprecated
```
captncraig: I'm ok deleting "deprecated code" in this file. Nobody should be depending on the internal working things like this function.

captncraig: But don't worry. I'll give that a look over after I merge.

PR #167, opened 2017-08-02
Cloudflare documentation update

virtualroot:
Hello there! 

* Fix syntax errors.
* Example on how to enable manage_redirects.

PR was tested with `runDocker.sh` and looks good.
Thanks!


tlimoncelli:
Good catch. Thanks for spotting those!

PR #170, opened 2017-08-03
Namecom punycode

pmoroney:


integrationTest/integration_test.go
```diff
@@ -294,7 +294,7 @@ var tests = []*TestCase{
 	tc("Internationalized name", a("ööö", "1.2.3.4")),
 	tc("Change IDN", a("ööö", "2.2.2.2")),
 	tc("Internationalized CNAME Target", cname("a", "ööö.com.")),
-	tc("IDN CNAME AND Target", cname("öoö", "ööö.ööö.")),
+	tc("IDN CNAME AND Target", cname("öoö", "ööö.企业.")),
```
captncraig: is there a reason this test needed to change? Does name.com validate the tld or something?

pmoroney: We do validate CNAME and ANAME targets to make sure they have a valid tld, so this test fails to create the CNAME.

tlimoncelli:
Thanks for helping us make DNSControl more international!

PR #171, opened 2017-08-04
Digitalocean provider

### LINKED ISSUE #125: Digital Ocean Support
Having DO support would be the cherry on the cake (sorry for the bad pun). If it's on the roadmap, I would like to pick it up. API docs - [https://developers.digitalocean.com/documentation/v2/#domains](https://developers.digitalocean.com/documentation/v2/#domains)

Deraen:
Fixes #125 

The tests pass. Digitalocean has support for SRV records, but DO requires that the names follow format `_service._protocol`, e.g. `_ts3._tcp.`, and the test cases use just `@`. So I disabled SRV support for now.

I haven't used Go before, but the diff looks sane so hopefully I managed to handle the dependencies correctly.


captncraig:
Wow! exciting to have this. I'll give it a thorough look over this weekend.

Deraen:
Note: Digitalocean only supports PTR records together with Droplets (virtual machines), not through DNS api.

providers/digitalocean/digitaloceanProvider.go
```diff
@@ -0,0 +1,176 @@
...
+
+	return corrections, nil
+}
+
+func toRc(dc *models.DomainConfig, r *godo.DomainRecord) *models.RecordConfig {
+	// This handles "@" etc.
+	name := dnsutil.AddOrigin(r.Name, dc.Name)
+
+	target := r.Data
+	// Make target FQDN
```
tlimoncelli: Please change this comment to:

    // Make target FQDN  (#rtype_variations)

`#rtype_variations` is how we mark anything that will need to be updated when adding a new DNS record type.  (This will be introduced in https://github.com/StackExchange/dnscontrol/pull/169)

providers/digitalocean/digitaloceanProvider.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+}
+
+func toReq(dc *models.DomainConfig, rc *models.RecordConfig) *godo.DomainRecordEditRequest {
+	// DO wants the short name, e.g. @
+	name := dnsutil.TrimDomainName(rc.NameFQDN, dc.Name)
+
+	// DO uses the same property for MX and SRV priority
+	priority := 0
+	switch rc.Type {
```
tlimoncelli: Please change:

    switch rc.Type {

to

    switch rc.Type {  // #rtype_variations

tlimoncelli:
Re: https://github.com/StackExchange/dnscontrol/pull/171#issue-248061672 Yes, please change the test to be conforming.  I've added https://github.com/StackExchange/dnscontrol/issues/173 to remind us to enforce this validation in the future.

Deraen:
I started testing this with real data and noticed that retrieving domain records from DO requires using paging to retrieve all the results. Fixed that now.

What would be correct way to implement `GetNameservers`? I saw that other providers also used the default list, but looks like this causes there always to be difference in TTL?

I have DefaultTTL modifier on the domain, no NAMESERVER entries.

```
----- DNS Provider: digitalocean... 3 corrections
#1: MODIFY NS metosin.fi: (ns3.digitalocean.com. ttl=1800) -> (ns3.digitalocean.com. ttl=300), DO ID: 1128468
#2: MODIFY NS metosin.fi: (ns2.digitalocean.com. ttl=1800) -> (ns2.digitalocean.com. ttl=300), DO ID: 1128467
#3: MODIFY NS metosin.fi: (ns1.digitalocean.com. ttl=1800) -> (ns1.digitalocean.com. ttl=300), DO ID: 1128466
```

Though limiting default nameservers to zero (`DnsProvider(digitalocean, 0)`) and providing NAMESERVER entries has the same problem, even if I add TTL modifier to NAMESERVER entries.

captncraig:
You can add `{ns_ttl: 1800}` to your domains to override the ttls on the auto-generated NS records. You can set that with `DEFAULTS({ns_ttl: 1800})` to apply it to all domains.

Deraen:
Thanks, I now have 6 domains configured and `preview` shows 0 corrections.

providers/digitalocean/digitaloceanProvider.go
```diff
@@ -0,0 +1,197 @@
...
+	return corrections, nil
+}
+
+func toRc(dc *models.DomainConfig, r *godo.DomainRecord) *models.RecordConfig {
+	// This handles "@" etc.
+	name := dnsutil.AddOrigin(r.Name, dc.Name)
+
+	target := r.Data
+	// Make target FQDN (#rtype_variations)
+	if r.Type == "CNAME" && target == "@" {
```
captncraig: I'm not quite sure on the logic here. Is the first case for something like `CNAME('foo', '@')`? I'm not sure I've considered that very often. Why are the other record types not included in the first if?

I feel like maybe that is something worth adding an integration test for.

Deraen: It is case for `CNAME('www', 'foo.com')` where `foo.com` is the domain. Digitalocean API will return the target this as `@`. Yes, Integration test for this makes sense. Might also be problem with the other types, but didn't have any such records and didn't think about this.

Deraen: Any tips of how to create tests for this? The test case would need to be `CNAME("foo", domainName)` but `tests` can't access `domainName`.

providers/digitalocean/digitaloceanProvider.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func init() {
+	providers.RegisterDomainServiceProviderType("DIGITALOCEAN", newDo, providers.CanUseSRV)
+}
+
+func (api *DoApi) EnsureDomain(domain string) error {
+	ctx := context.Background()
+	_, resp, err := api.client.Domains.Get(ctx, domain)
+	if resp.Status == "404" {
```
Deraen: Done. Just had the wrong name.

integrationTest/integration_test.go
```diff
@@ -290,11 +293,13 @@ var tests = []*TestCase{
 	tc("Change it", cname("foo", "google2.com.")),
 	tc("Change to A record", a("foo", "1.2.3.4")),
 	tc("Change back to CNAME", cname("foo", "google.com.")),
+	tc("Record pointing to @", cname("foo", "**current-domain**")),
```
Deraen: I managed to create tests for this by using special value for test cases and rewriting the value to `domainName` on `runTests`. Not very pretty but works.

tlimoncelli:
FYI: I just updated provider-list.md to clarify the responsibilities of someone that submits a new provider. Basically you'll be asked to fix future bugs, but the DNSControl maintainers will take responsibility for maintaining a test suite that lets you do that easily.  May we list you as the maintainer of this provider? (your name is commented out in provider-list.md)

integrationTest/integration_test.go
tlimoncelli: Great!  I think we'll be able to make more interesting tests that way.

Deraen:
@tlimoncelli This is OK. I plan on eventually migrating our domains somewhere else (one reason I wanted to use dnscontrol), so I might not use this myself a lot, but looks like maintaining this should be quite easy.

providers/digitalocean/digitaloceanProvider.go
```diff
@@ -38,13 +39,25 @@ func newDo(m map[string]string, metadata json.RawMessage) (providers.DNSServiceP
...
 		oauth2.StaticTokenSource(&oauth2.Token{AccessToken: m["token"]}),
 	)
 	client := godo.NewClient(oauthClient)
 
-	return &DoApi{client: client}, nil
+	api := &DoApi{client: client}
+
+	// Get a domain to validate the token
+	_, resp, err := api.client.Domains.List(ctx, &godo.ListOptions{PerPage: 1})
+	if resp.StatusCode != http.StatusOK {
```
captncraig: Nitpicking now, but you should do the error check before the status code check.

Deraen: Done

tlimoncelli:
Thanks for contributing a new provider!  You rock!

PR #174, opened 2017-08-06
Implement SRV support for CloudFlare provider

pjeby:
This PR implements basic support for SRV records in the CloudFlare API.  It works for me but lacks automated tests.  This is my first time writing _anything_ in Go, so..  well, honestly I haven't even tried to figure out how to *run* the unit tests yet.  :)  But I did manually test creation, modification, and deletion of SRV records, with various protocols, priorities, targets, etc.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -325,14 +325,14 @@ var tests = []*TestCase{
 
 	//SRV
 	tc("Empty").IfHasCapability(providers.CanUseSRV),
-	tc("SRV record", srv("@", 5, 6, 7, "foo.com.")).IfHasCapability(providers.CanUseSRV),
-	tc("Second SRV record, same prio", srv("@", 5, 6, 7, "foo.com."), srv("@", 5, 60, 70, "foo2.com.")).IfHasCapability(providers.CanUseSRV),
-	tc("3 SRV", srv("@", 5, 6, 7, "foo.com."), srv("@", 5, 60, 70, "foo2.com."), srv("@", 15, 65, 75, "foo3.com.")).IfHasCapability(providers.CanUseSRV),
-	tc("Delete one", srv("@", 5, 6, 7, "foo.com."), srv("@", 15, 65, 75, "foo3.com.")).IfHasCapability(providers.CanUseSRV),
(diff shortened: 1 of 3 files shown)


pjeby:
...and I'm so new I didn't know there was such a thing as `go fmt`, let alone that I needed to run it.  :)

tlimoncelli:
Looks good so far!  Have you been able to run the integration tests?  (See Step 5 and 6 of https://stackexchange.github.io/dnscontrol/adding-new-rtypes)

pjeby:
If I understand that correctly then travis should've run it, since I added the CanUseSRV flag and the SRV tests are keyed off that.  Unless travis isn't configured to run the integration tests?

captncraig:
Travis does not run the integration suite. They require credentials we have not set up there, and we probably wouldn't run it on prs automatically either.

pjeby:
Ok, the integration tests pass if I apply the changes from [PR 171](https://github.com/StackExchange/dnscontrol/pull/171/files#diff-bd97799a27737ab0d0c3e8f5ea3721dfL328) to use properly formatted `_service._protocol` names instead of `@`.   (Like Digital Ocean, CloudFlare's API requires this.)  I used the same names so as not to conflict with that PR.

tlimoncelli:
Thanks for the contribution!

"This is my first time writing anything in Go!"  <<< If you hadn't told me, I would have never guessed!  Great job!

pjeby:
Heh.  Well, I do have around 35 years of experience in *other* langauges, so that helps.  :) It also helps that there was already a "data" field in the CloudFlare JSON, and that SRV records were basically implemented already.

I was actually surprised at how easy it is to build and run: I see now why it's becoming such a popular language.  Anyway, glad I could help.

Oh, just thought of something...  this was already there in the CF code, but there's basically code that uses `dnsutil.AddOrigin()` to just add a '.' to the end of a domain.  (See lines [354](https://github.com/StackExchange/dnscontrol/blob/master/providers/cloudflare/cloudflareProvider.go#L354) and 369 of `cloudflareProvider.go`.)  I copied the pre-existing code at 354 to 369, without initially realizing that all `AddOrigin()` does when a string ends in `.` is return it.

So... there's not really any reason to be calling `AddOrigin()` in either of those places.  But it's also harmless, so I leave it to your discretion how to handle it going forward.

Issue #143, opened 2017-06-14
OVH support


Oprax:
That would be pretty nice if OVH is support ! Here documentation : https://api.ovh.com/console/#/domain

tlimoncelli:
Good suggestion!

I've added this to the [list of requests](https://stackexchange.github.io/dnscontrol/provider-list). If someone would like to volunteer to implement this, please re-open this issue. 

Advice on how to write a provider is documented [here](https://stackexchange.github.io/dnscontrol/writing-providers). Also, speak up in this issue and someone will get back to you asap.

masterzen:
Hi,

I started working on the dns provider yesterday (and also want to work on the domain provider) .
I'm not yet ready to submit a PR though, there are lots of corner cases to fix :)

What level of automated test do you require (beside the integration test)?
I was thinking about mocking the OVH API and test how I call it, but looking at the other providers it doesn't seem to be required?

Thanks !
Brice

masterzen:
Hi,

I've pushed the PR for the OVH DNS Provider. Please feel free to review.

I don't have the privileges to reopen the issue as I'm not the OP, can someone do it for me ?

For the moment the DNS Provider ignores any apex NS change, as this requires a specific implementation that uses the OVH registrar API, which I hope I can do soon :)

One question though is how one is supposed to manage nameservers in the `dnsconfig.js` as there exists both the `NS()` and `NAMESERVER()` commands. And how does that play with the dnscontrol registrar api?

Thanks,
Brice

captncraig:
We try to let the nameservers automatically resolve themselves. Each provider has the `GetNameservers` method that is intended to help dynamically discover what nameservers are needed to serve the zone. The driver tries to infer the nameservers and set up NS records appropriately, and register them.

If that fails, you can specify them manually with `NAMESERVER`. As far as I know, you cannot use `NS` directly. See https://stackexchange.github.io/dnscontrol/nameservers for details.

masterzen:
@captncraig thanks for the answer, that makes sense (and I wonder how I missed this documentation).

So, in essence, I need to add a registrar provider for OVH so that one can manage NS records.

captncraig:
Your pr looks ok for using ovh as a dns provider. It looks like GetNameservers does the right thing. Using OVH as a rgistrar is another thing altogether, and does not need to be included if you don't want to right now.

Although, now that I look, I don't see any way to use dns without purchasing a domain through them, so maybe the registrar bits would be good to have.

[...]

PR #175, opened 2017-08-07
OVH DNS Provider (#143)

masterzen:
This adds the OVH provider along with its documentation:

- [x] DNS Provider
- [x] Registrar Provider
- [x] documentation

A few comments:
* the provider unfortunately can't use `CanUsePTR` because OVH supports only setting PTR with a target in the arpa zone. This makes the integration tests 31 and 32 fail horribly (can we make those use the arpa zone?)
* OVH now supports setting apex NS records, so dual providers scenario can be accomplished 
* OVH doesn't support CAA (I'm going to ask them when they plan to support those).

### DIFF
--- a/integrationTest/providers.json
+++ b/integrationTest/providers.json
@@ -66,5 +66,11 @@
   "VULTR": {
     "token": "$VULTR_TOKEN",
     "domain": "$VULTR_DOMAIN"
+  },
+  "OVH": {
+    "app-key": "$OVH_APP_KEY",
+    "app-secret-key": "$OVH_APP_SECRET_KEY",
+    "consumer-key": "$OVH_CONSUMER_KEY",
+    "domain": "$OVH_DOMAIN"
   }
 }
--- a/providers/ovh/ovhProvider.go
+++ b/providers/ovh/ovhProvider.go
@@ -1,0 +1,177 @@
+package ovh
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/StackExchange/dnscontrol/models"
+	"github.com/StackExchange/dnscontrol/providers"
+	"github.com/StackExchange/dnscontrol/providers/diff"
+	"github.com/miekg/dns/dnsutil"
+	"github.com/xlucas/go-ovh/ovh"
+)
+
+type ovhProvider struct {
+	client *ovh.Client
+	zones  map[string]bool
+}
+
+var docNotes = providers.DocumentationNotes{
+	providers.DocDualHost:            providers.Cannot(),
+	providers.DocCreateDomains:       providers.Cannot("New domains require registration"),
+	providers.DocOfficiallySupported: providers.Cannot(),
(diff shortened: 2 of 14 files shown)


masterzen:
I've rebased to the latest master.

tlimoncelli:
Sorry for the delay.

Could you rebase  and re-run the integration tests?  I'd be glad to merge this.

masterzen:
@tlimoncelli I'll do that this week-end. Thanks!

masterzen:
@tlimoncelli,

I've rebased and made sure documentation is generated (I also implemented the `docNotes`). 

I've also enabled support for TLSA, but that required a modification of the integration test because OVH doesn't allow TLSA records for a sha256 matching type with a checksum that doesn't have exactly 64 characters (this makes sense). I couldn't test the other providers with this integration test change, so I hope I haven't introduced any test regressions.
If that proves that it is an issue, let me know and I'll revert that part of the change and exclude those tests for the OVH provider.

Thanks!

providers/ovh/ovhProvider.go
```diff
@@ -0,0 +1,199 @@
...
+			Name:           r.SubDomain,
+			Type:           r.FieldType,
+			Target:         r.Target,
+			TTL:            uint32(r.TTL),
+			CombinedTarget: true,
+			Original:       r,
+		}
+		actual = append(actual, rec)
+	}
+
```
tlimoncelli: Insert this (all providers now must do this; though if you were passing the integration tests it might be a no-op):

```
  // Normalize
  models.Downcase(actual)
```

integrationTest/providers.json
```diff
@@ -66,5 +66,13 @@
   "VULTR": {
     "token": "$VULTR_TOKEN",
     "domain": "$VULTR_DOMAIN"
+  },
+  "OVH": {
+    "COMMENT": "47-48 OVH doesn't support TLSA records with matching type 0 (full certificate)",
```
tlimoncelli: Is there a matchingtype other than 0 that the tests could use so that the tests can run? (I don't know much about TLSA)

masterzen: Yes, matching type 2 is for sha512, but since they check the length of the checksum we'd have to have a string of 128 hex characters.
Do you want me to modify the integration test to check changing matching type to 2 instead of 0 ?

providers/ovh/ovhProvider.go
masterzen: Oh, I missed this change, I'll do that.

integrationTest/providers.json
tlimoncelli: Sure!  The extra 128 bytes won't fill my hard disk :-) . (but seriously... I'd rather have working tests than not)

masterzen:
@tlimoncelli I've pushed the fixes you asked for. Let me know if there's anything I should do.

tlimoncelli:
Thank you for your contribution!

PR #190, opened 2017-08-24
Adding SRV functionality to Gandi

chewrocca:
Adding functionality to allow SRV rtypes for Gandi.

### DIFF
--- a/models/dns.go
+++ b/models/dns.go
@@ -359,7 +359,7 @@ func SplitCombinedSrvValue(s string) (priority, weight, port uint16, target stri
     if err != nil {
         return 0, 0, 0, "", fmt.Errorf("SRV preference %#v does not fit into a uint16", parts[0])
     }
-	return uint16(priority), uint16(weight), uint16(port), parts[5], nil
+	return priority, weight, port, parts[5], nil
 }
 
 func copyObj(input interface{}, output interface{}) error {
--- a/models/dns.go
+++ b/models/dns.go
@@ -327,6 +327,41 @@ func SplitCombinedMxValue(s string) (preference uint16, target string, err error
 	return uint16(n64), parts[1], nil
 }
 
+// CombineSRVs will merge the priority into the target field for all srv records.
+// Useful for providers that desire them as one field.
+func (dc *DomainConfig) CombineSRVs() {
+	for _, rec := range dc.Records {
+		if rec.Type == "SRV" {
+			if rec.CombinedTarget {
+				pm := strings.Join([]string{"CombineSRVs: Already collapsed: ", rec.Name, rec.Target}, " ")
+				panic(pm)
+			}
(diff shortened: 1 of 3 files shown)


models/dns.go
```diff
@@ -327,6 +327,41 @@ func SplitCombinedMxValue(s string) (preference uint16, target string, err error
...
+	}
+
+	priority = uint16(priority)
+	weight = uint16(weight)
+	port = uint16(port)
+
+    if err != nil {
+        return 0, 0, 0, "", fmt.Errorf("SRV preference %#v does not fit into a uint16", parts[0])
+    }
+	return uint16(priority), uint16(weight), uint16(port), parts[5], nil
```
tlimoncelli: No need to specify the types here.  It can be simply:

    return priority, weight, port, parts[5], nil

models/dns.go
```diff
@@ -327,6 +327,41 @@ func SplitCombinedMxValue(s string) (preference uint16, target string, err error
...
+			}
+			rec.Target = fmt.Sprintf("%d %d %d %s", rec.SrvPriority, rec.SrvWeight, rec.SrvPort, rec.Target)
+			rec.MxPreference = 0
+			rec.CombinedTarget = true
+		}
+	}
+}
+
+// SplitCombinedSrvValue splits a combined SRV preference and target into
+// separate entities, some DNS providers want "10" and "foo.com.", others want "10 foo.com.".
```
tlimoncelli: Update this comment.

models/dns.go
```diff
@@ -327,6 +327,41 @@ func SplitCombinedMxValue(s string) (preference uint16, target string, err error
...
+func SplitCombinedSrvValue(s string) (priority, weight, port uint16, target string, err error) {
+	parts := strings.Fields(s)
+
+	if len(parts) != 5 {
```
tlimoncelli: I think this should be 4, not 5.  https://play.golang.com/p/GScs9YWFs6

models/dns.go
```diff
@@ -327,6 +327,41 @@ func SplitCombinedMxValue(s string) (preference uint16, target string, err error
...
+		return 0, 0, 0, "", fmt.Errorf("SRV value %#v contains too many fields", s, s)
+	}
+
+	priority = uint16(priority)
+	weight = uint16(weight)
+	port = uint16(port)
```
tlimoncelli: `port = uint16(port)` is just `x = x`.  I think what you want is to convert a part[] from ascii to an integer.

For each of the 3 integers you'll need to do the conversion and handle any errors:

    priority, err := strconv.ParseUint(parts[0], 10, 16)
    if err != nil { .... }
   (repeat 2 for weight and port)

then later...

    return uint16(priority), uint16(weight), uint16(port), part[4]

(Now I see why those type conversions where there before!  ah ha!)

chewrocca:
i believe i fixed this issue and i will submit another pull request with better changes and comments.

PR #201, opened 2017-09-14
Create a dynamic provider features matrix

captncraig:
2017-09-14 15_43_09-localhost_4000_provider-matrix


build/generate/featureMatrix.go
```diff
@@ -0,0 +1,137 @@
...
+		providerTypes = append(providerTypes, n)
+	}
+	sort.Strings(providerTypes)
+	matrix := &FeatureMatrix{
+		Providers: map[string]FeatureMap{},
+		Features: []FeatureDef{
+			{"Official Support", "This means the provider is actively used at stack exchange, and we offer a stronger guarantee it will work"},
+			{"Registrar", "The provider has registrar capabilities to set nameservers for zones"},
+			{"DNS Provider", "Can manage and serve DNS zones"},
+			{"ALIAS", "Provider supports some kind of ALIAS,ANAME or flattened CNAME record type"},
```
tlimoncelli: add a space after the comma

build/generate/featureMatrix.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	providerTypes := []string{}
+	for n := range allNames {
+		providerTypes = append(providerTypes, n)
+	}
+	sort.Strings(providerTypes)
+	matrix := &FeatureMatrix{
+		Providers: map[string]FeatureMap{},
+		Features: []FeatureDef{
+			{"Official Support", "This means the provider is actively used at stack exchange, and we offer a stronger guarantee it will work"},
```
tlimoncelli: actively used at Stack Exchange, bugs are more likely to be fixed, and failing integration tests will block a release. See docs for details.

providers/namedotcom/namedotcomProvider.go
```diff
@@ -19,6 +19,13 @@ type nameDotCom struct {
 	APIKey  string `json:"apikey"`
 }
 
+var docNotes = providers.DocumentationNotes{
+	providers.DocDualHost:            providers.Cannot("Apex NS records not editable"),
```
pmoroney: Apex NS records are editable, but they have to match the list of nameservers at the registry.

captncraig: That's interesting @pmoroney. It creates a bit of a chicken and egg problem. In dnscontrol we update the dns provider first, and then the registrar. It sounds like that wouldn't work with NDC unless we invert that order? 

We made that choice specifically so we never point at an uninitialized dns server.

pmoroney: The dns provider ignores the Apex NS records and just lets the registrar provider handle it, so it doesn't really matter which order it works in. I do it for pat.irish using dnscontrol. I just removed it and re-added it and didn't have any problems.

captncraig: Oh, that makes sense. Thats a slightly different paradigm than other providers, but it should work fine. I'll run a quick test and update the flag.

PR #205, opened 2017-09-15
Format helpers.js

whs:
Last time I worked with helpers.js, that file was a completely mess with so many coding standards mixed together.

This pull request use [Prettier](https://github.com/prettier/prettier) to cleanup the file. It should work similar to gofmt, although there is no one true style in JavaScript so I take some liberty in defining the prettierrc.

Integration test files are not ran though prettier, as it doesn't understand our DSL and will put all arguments on its own line.

If we want this to last, I'd suggest we should add a check on pull requests similar to gofmt.

I have a plan for another pull request which add eslint to perform static analysis (like warn on unused variables) plus some style guide check (for the ones I skipped like integration tests). I'll see the response from this pull request first.


pkg/js/helpers.js
```diff
@@ -445,69 +466,66 @@ function addRecord(d,type,name,target,mods) {
 
 //ip conversion functions from http://stackoverflow.com/a/8105740/121660
 // via http://javascript.about.com/library/blipconvert.htm
-function IP(dot)
-{
+function IP(dot) {
     var d = dot.split('.');
-    return ((((((+d[0])*256)+(+d[1]))*256)+(+d[2]))*256)+(+d[3]);
+    // prettier-ignore
+    return ((((((+d[0]) * 256) + (+d[1])) * 256) + (+d[2])) * 256) + (+d[3]);
```
whs: Without prettier-ignore, this line turns to

```js
return ((+d[0] * 256 + +d[1]) * 256 + +d[2]) * 256 + +d[3];
```

I'm not sure which is better, so I put spaces around the operands in the original line instead.

captncraig:
The other thing I'd like to check is documentation. If possible we should probably wrap that all in a function and explicitly export things to the global context that we expect to use. Then we can inspect the exports and make sure they all have a file in the docs folder. 

That just seems like a bit of work I haven't buckled down to attack yet.

PR #211, opened 2017-09-26
Add ca-certificates to Docker image

bfirsh:
It was causing errors like this:

```
Error fetching domain list from cloudflare: Get https://api.cloudflare.com/client/v4/zones/?page=1&per_page=50: x509: failed to load system roots and no roots provided
```


captncraig:
Thanks! I switched to alpine, and forgot to add those.

PR #216, opened 2017-10-03
Link providers to domains up front

captncraig:
The domains just have a list of provider names in the json. The main driver was making a single collection of providers, and a lot of different places in code were cross-referencing and possibly erroring in all kinds of odd places. This fully constructs and links all providers to domains up front.


captncraig:
Part of this is moving some of the core interfaces into models from providers, and making an additional interface that is like the old provider, but with some more metadata (name, is it on by default, how many nameservers to fetch). 

I'm not in love with the interface names yet.

captncraig:
Tests not compiling. I'll fix tomorrow.

captncraig:
This has a few implementation details I need to fix. Probably gonna scratch this and start from scratch.

- Moved main interface into models unnecessarily. Can still keep an alias in providers to avoid touching every reference.
- Merge conflicts seem pretty hairy at this point.
- Naming is all wrong.

Will circle around and address this when I get some time, as it is a pretty major QOL improvement IMO

Issue #217, opened 2017-10-10
Vultr support


geek1011:
Vultr DNS support would be nice to have. See https://www.vultr.com/api/#dns

captncraig:
Sure would be. Looks like their API is pretty simple, if a bit under-documented. Should be fairly straightforward to implement. Are you planning to work on it?

geek1011:
@captncraig yes, sometime this week most likely.

geek1011:
@captncraig Finished. See #219.

PR #219, opened 2017-10-10
Added Vultr provider (#217)

geek1011:
I have added the Vultr provider. This PR includes full support for all supported features from Vultr (SRV, CAA, CreateDomains).

I have tested this with both the integration tests, and my own domain.

- [X] DNS Provider
- [X] Documentation
- [X] Tests

See issue #217

### DIFF
--- a/providers/vultr/vultr.go
+++ b/providers/vultr/vultr.go
@@ -27,6 +27,9 @@ Info required in `creds.json`:
 var docNotes = providers.DocumentationNotes{
 	providers.DocCreateDomains:       providers.Can(),
 	providers.DocOfficiallySupported: providers.Cannot(),
+	providers.CanUseAlias:            providers.Cannot(),
+	providers.CanUseTLSA:             providers.Cannot(),
+	providers.CanUsePTR:              providers.Cannot(),
 }
 
 func init() {
--- a/providers/vultr/vultr.go
+++ b/providers/vultr/vultr.go
@@ -1,0 +1,330 @@
+package vultr
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/StackExchange/dnscontrol/models"
+	"github.com/StackExchange/dnscontrol/providers"
+	"github.com/StackExchange/dnscontrol/providers/diff"
+	"github.com/miekg/dns/dnsutil"
+
+	vultr "github.com/JamesClonk/vultr/lib"
+)
+
+/*
+
+Vultr API DNS provider:
+
+Info required in `creds.json`:
+   - token
+
+*/
+
+var docNotes = providers.DocumentationNotes{
+	providers.DocCreateDomains:       providers.Can(),
+	providers.DocOfficiallySupported: providers.Cannot(),
+}
+
+func init() {
+	providers.RegisterDomainServiceProviderType("VULTR", NewVultr, providers.CanUseSRV, providers.CanUseCAA, docNotes)
+}
+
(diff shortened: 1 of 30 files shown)


geek1011:
cc @captncraig

geek1011:
I am finished this PR.

geek1011:
I am finished this PR.

captncraig:
Wow! this looks great. I'll take a good look over it tomorrow.

geek1011:
@captncraig Thanks! Btw, I have written code in go before, unlike some of the other contributors :smiley:. 

Also, I will submit some more PRs for improving the docs, and some of the other providers.

captncraig:
Sure, some of these bits have grown organically, so I'd love any feedback you have to offer. Some of the pain points I know about, but its always good to have a new set of eyes for honest feedback.

geek1011:
@captncraig ok.

geek1011:
Review pls

providers/vultr/vultr.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func (api *VultrApi) isDomainInAccount(domain string) (bool, error) {
+	domains, err := api.client.GetDNSDomains()
+	if err != nil {
+		return false, err
+	}
+
+	var vd *vultr.DNSDomain
+	for _, d := range domains {
```
captncraig: does their api do any kind of paging we need to worry about?

providers/vultr/vultr.go
```diff
@@ -0,0 +1,342 @@
...
+		rc.SrvPort = uint16(port)
+
+		target := splitData[2]
+		if !strings.HasSuffix(target, ".") {
+			target = target + "."
+		}
+		rc.Target = dnsutil.AddOrigin(target, dc.Name)
+	}
+
+	// #rtype_variations
```
captncraig: I don't think we need this #comment everywhere here. Just at the top so this function shows up in grep.

geek1011:
Review please

providers/vultr/vultr.go
```diff
@@ -0,0 +1,342 @@
...
+		Original: r,
+	}
+
+	// #rtype_variations
+	if r.Type == "MX" {
+		rc.MxPreference = uint16(r.Priority)
+	}
+
+	// #rtype_variations
+	if r.Type == "SRV" {
```
captncraig: I believe there are some helper functions in the models package for splitting out srv/caa records from a single field representation. Is vultr doing something odd so that those don't work?

providers/vultr/vultr.go
geek1011: Nope. Passes integration tests and I tried myself with 300 entries.

providers/vultr/vultr.go
geek1011: I'll have to look into that, but Vultr does quite a few funny things, so it may be better to manually handle it.

providers/vultr/vultr.go
geek1011: Ok, the docs on adding rtypes were not all that clear, I guess I will fix that next.

geek1011: Done

providers/vultr/vultr.go
geek1011: The helper functions do not use the right order for Vultr.

geek1011:
@captncraig Ready for another review.

geek1011:
@captncraig Ready for another review.

tlimoncelli:
Looks great! Thanks!

geek1011:
@tlimoncelli No problem!

Issue #221, opened 2017-10-11
Website issues


geek1011:
I will work on these sometime soon.

- [x] Code formatting/indentation
- [x] No page titles (#223)
- [x] Not responsive (#223)
- [ ] Some pages are hard to find (improve nav) (#236)
- [ ] Some pages could use some styling improvements (#236)
- [X] Documentation is inconsistent (#222) (#224) (#226) (#227) (#228) (#229) (#230) (#231) (#232) (#233) (#234) (#235)
- [X] Various documentation issues (#224) (#225) (#226) (#227) (#228) (#229) (#230) (#231) (#232) (#233) (#234) (#235)

I will add to this list as I encounter more issues.

tlimoncelli:
I think you've done enough to consider this closed.  We can split anything else into a separate bug, eh?

geek1011:
Yep. Basically only the nav is left (it's still slightly confusing).

PR #234, opened 2017-10-11
Improve Route53 documentation (#221)

geek1011:
- Grammar fixes
- Formatting improvements
- Don't encourage using environment vars instead of the credentials file

See #221

### DIFF
--- a/docs/_providers/route53.md
+++ b/docs/_providers/route53.md
@@ -17,12 +17,14 @@ You can specify the API credentials in the credentials json file:
 }
 {% endhighlight %}
 
-You can also use environment variables, but this is discouraged.
+You can also use environment variables, but this is discouraged, unless your environment provides them already.
 
 ```
 $ export AWS_ACCESS_KEY_ID=XXXXXXXXX
 $ export AWS_SECRET_ACCESS_KEY=YYYYYYYYY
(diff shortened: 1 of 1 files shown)


docs/_providers/route53.md
```diff
@@ -24,13 +17,18 @@ It is also possible to specify an aws access key in the providers config json fi
 }
 {% endhighlight %}
 
-## Metadata
+You can also use environment variables, but this is discouraged.
```
tlimoncelli: Please add back this link `[go sdk configuration](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html)`

geek1011: Ok, just curious, what is the point of it?

captncraig: Some users configure it in different ways. If you are running from certain instance types, the environment might be implicitly set up, or they want to manage aws creds in the same way they manage other tools.

geek1011: Ok, I see now. I don't use that method myself, so I did not see the point. I'll add it.

geek1011: @captncraig @tlimoncelli Done.

PR #237, opened 2017-10-12
add initial URL support to namecheap provider

rbelnap:
I'm picking up where https://github.com/StackExchange/dnscontrol/pull/98 left off... That was several months ago, so I'm not sure what might have changed in the meantime.  This is an initial attempt to get URL working (which works for my tests).  If this is the right approach, I can add the other namecheap specific types, URL301 and FRAME, along with tests, etc.


tlimoncelli:
Looks good.  I'll CC: @captncraig for further review.

By the way... since #98, we've added a similar feature to Cloudflare (CF_TEMP_REDIRECT and CF_REDIRECT).  Maybe we should have a pseudo rtype and let each provider translate that into the right thing.

tlimoncelli:
Thinking out loud...

How about:

        REDIRECT("type", "url"),

example:

        REDIRECT("301", "http://the-url.com")

Where "type" is "301", "302", "iframe".  The provider can do the right thing.  Meanwhile, we can implement REDIRECT for CloudFlare and other providers.

Tom

onlyhavecans:
What about providers like DNSimple hat have a URL record but you cannot set the type? It might be better to have it be URL(“url”, “type”=default_for_provider) so that you can declare it without a type

captncraig:
The reason we made CF_REDIRECT a special thing is because they have their own special syntax for defining patterns with wildcards and such, and transforming matched segments into the final url with other replacements and things. I'm not convinced there is enough similarity between provider implementations to make a truly generic `URL` record type.

Are we doomed to have `NAMECHEAP_URL` and other such records that only work with one provider, or can we find some unity? I have not dug in enough to know definitively, but cloudflare's way was pretty specialized.

tlimoncelli:
Just to clarify @captncraig 's issue a bit...

We want to be able to change DnsProvider() on a domain and know immediately if the new provider supports the same functionality. We want to "fail early"... i.e. in "dnscontrol check" we should be able to error out if (for example) someone is switching from one provider to another and using a feature that the new provider doesn't support.  We want to be able to make that determination early.  For example, with SRV records, we know which providers do/don't support SRV. Therefore we can give an error ("provider doesn't support SRV") at the syntax check, not at the very end when an API call returns with an error.

If the syntax is `URL("url", "type")` and providers advertise which types they provide, we could have that kind of "early knowledge".  Types "301", "302", "frame" would have no wildcards not the ability to do rewrites. Type "cf_wildcard" (or whatever) would permit Cloudflare-style wildcards.

Or, we could implement URL() that way and leave the cloudflare-specific calls alone.

captncraig:
Yeah, if we could define a spec for it that is supported by two or more providers I'm game. As long as you can support those features you use `REDIRECT`, and for anything else you need a more specialized record? I could get behind that.

rbelnap:
I added URL301/FRAME only to match functionality of my current fork - I will roll with whatever becomes of the REDIRECT discussion when it solidifies.

captncraig:
Cool, let's get it working for you, then have a larger discussion. 

If we can find a single way to represent a generic redirect, complete with specific rules for how a "standard" redirect should work with subpaths and so forth, we may re-address this.

As long as you are willing to admit that this may be an area of breaking changes in the future.

captncraig:
@rbelnap would you mind adding a section to the namecheap docs page that talks about how to use these, maybe with some examples) ? Thanks.

PR #238, opened 2017-10-16
add rate based throttle to namecheap

rbelnap:
The existing implementation doesn't look like it actually throttles.  If I understand the intent correctly, it is meant to only throttle after 20 requests?  If that is the case, I believe it should sleep after the for loop in init (~line 89).  When done this way, if you have > 20 requests, they fire off as fast as possible until 20, then it just sits for 30+ seconds, which isn't too user friendly.  As it is now, without a sleep, there is no time limitation.  

This implements a rate so that every request is throttled (if needed).  This will be slightly slower for people that have <20 requests, but it is more desirable for those that have > 20.   There may be some further options or optimizations that could be made here, I'm open to them.  This implementation does throttle things as expected for our use though.

### DIFF
--- a/providers/namecheap/namecheap.go
+++ b/providers/namecheap/namecheap.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"golang.org/x/net/publicsuffix"
 
@@ -75,20 +76,12 @@ func splitDomain(domain string) (sld string, tld string) {
 // from support in SEP-2017:
 //    "The limits for the API calls will be 20/Min, 700/Hour and 8000/Day for one user.
 //     If you can limit the requests within these it should be fine."
-var throttle = make(chan bool, 20)
+
+var throttle <-chan time.Time
 
 func init() {
-	go func() {
-		for {
-			// add (up to) 20 requests every minute
-			for i := 0; i < 20; i++ {
-				select {
-				case throttle <- true:
-				default:
-				}
-			}
-		}
-	}()
+	rate := time.Minute / 20
+	throttle = time.Tick(rate)
 }
 
 func (n *Namecheap) GetDomainCorrections(dc *models.DomainConfig) ([]*models.Correction, error) {


captncraig:
Hmmm. I think you are right about it needing a sleep. I got them to increase the limit on my account after making tickets, and I probably deleted that for testing and forgot to re-add it.

I implemented it this way to match the rate-limit rules their support told me in a chat. I still feel that allowing up to 20 requests immediately is the right choice. Waiting 5 seconds between requests and waiting 30 after 20 are really the same thing if you are doing a large number. You can't get around it. Might be worth adding a `Wait` function everywhere we pull from the channel. Could do something like

```
select {
 case <- throttle:
 default: 
     log.Println("Namecheap throttle reached. Waiting for more requests to be available")
     <- throttle
}
```

Or something like that.

captncraig:
I added the sleep back on master for now. I think you're right this user-experience is not optimal, but I don't really agree applying sleep before we need to solves any of that.

geek1011:
Is the rate limit a rolling time window, or a fixed time window?

captncraig:
That is a really good question. Running some tests.

captncraig:
My testing has been inconclusive. It *appears* that you get your full allotment back all at once. I'm not clear on if that time is based on the first request or the last one that causes you to be limited. I did not observe any penalty for repeated calls once rate limited.

There is also some amount of lag where some requests succeed and others fail, as if it takes time for the limit to propagate their system, and the whole thing seems rather unpredictable. I'm not sure either of their approaches is really optimal seeing what I am.

What if instead, we just make all requests through a wrapper that will sleep and retry only after it detects we are limited. 

```
// pseudocode
func doWithRateLimit(f func()error)error{
    for{
        err := f()
        if err != nil && err is rate limit {
            fmt.Println("Namecheap rate limited us")
            time.Sleep(3 * time.Second)
            continue
        }
        return err
    }
}
```

Then to use it we do something like:

Before: 

`records, err := n.client.DomainsDNSGetHosts(sld, tld)`

After:
```
var records *namecheap.DomainDNSGetHostsResult
var err error
err = doWithRateLimit(func()error{
   records, err = n.client.DomainsDNSGetHosts(sld, tld)
   return err
})
```

Not super pretty, but could let us handle rate limiting "just in time". Would need to play with it to reduce spamming console, but I think maybe that is a better paradigm than trying to guess when they get a new allowance.

geek1011:
@captncraig ok, but we should have exponential back-off with an upper limit to avoid upsetting namecheap. Have you asked their support about the details of the rate-limiting?

captncraig:
I asked them about the specific numbers, and they gave me the per minute/hour/day numbers that I noted in the code. I didn't really have a technical enough person to get into the specifics of windows and timing policies. They were able to increase my limit after making the appropriate requests.

At the end of the day, I'd probably recommend anyone with a large number of domains move dns hosting to cloudflare or something more scalable. Just running registrar bits through namecheap should be fewer requests (1 per domain or 2 if changes needed), and give you more buffer.

I agree the backoff can be dynamic, but probably not exponential. I feel like 1 req. every three seconds up to a max of 90 seconds total or something like that should be fairly reasonable. 

My tests were doing ~ 45 failed requests/sec and still getting success when the window resets.

rbelnap:
I can live with the sleep as implemented, as long as there is an informative message to explain the delay to the user.  I also feel like it shouldn't be overthought too much.

That said, I can add a few more of my cents...  It seems to me there are two ways of approaching this:

* rate limit avoidance - implement delays so that we don't hit the rate limit.  It seems optimal to me here to have a fixed rate within the limit, but there is the tradeoff I mention in my initial post.  
* rate limit reaction/backoff - This is optimal for not slowing down requests unnecessarily, but requires reacting to and retrying failed/rate limited requests.

The current implementation is kind of a hybrid approach, not slowing down a small amount of requests, but slowing down a large number of requests (or at least distributing them unevenly.)  I think the reaction/backoff method is a more acceptable compromise generally, but obviously not if you incur a penalty for hitting the limit.  It may also be problematic for how the code is structured, although I'm not familiar with it enough to say.

rbelnap:
@captncraig I added a message similar to how you described.  This will satisfy my needs for now.  I do think that retrying with backoff is probably the right thing long term, but is more invasive than I care to get into right now.  I'll take "imperfect but merged" over "pending perfection" :)

captncraig:
Awesome. If I get a sec I'll try to throw together an on-demand retry loop setup. But this seems better for now at least.

PR #244, opened 2017-10-23
Namecheap: switch throttling to only as needed.

captncraig:
cc @rbelnap


captncraig:
This passes integration tests locally, and a stress tester I made seems to pass pretty well.

PR #245, opened 2017-10-23
namecheap url doc mention

rbelnap:
@captncraig requested a mention in https://github.com/StackExchange/dnscontrol/pull/237 here it is


captncraig:
Sweet. Thanks!

Issue #195, opened 2017-09-05
Duplicate domain definitions should not be allowed


pmoroney:
It seems duplicates are allowed:
The following config is legal but should probably display an error:
```
var namecom = NewRegistrar("name.com", "NAMEDOTCOM");
var namecom_dns = NewDnsProvider("name.com", "NAMEDOTCOM")

D("example.org", namecom,
  A("@", "127.0.0.1")
)

D("example.org", namecom)
```

tlimoncelli:
Good point.

An easy way to implement this:

`function D()` in `pkg/js/helpers.js` should "throw" if the domain name already is defined.

captncraig:
I'm wondering if there is any valid use case for this at all. What if I want to do slightly different configs to different providers? Maybe I'm stretching here. Two possible situations:

Perhaps I want to use proxider X as my primary, but have provider Y set up as a standby with a slightly different config for emergency purposes:

```
D("example.com", REG_NAMECOM, DnsProvider(CF), 
      A("whatever", "1.2.3.4"),
      ALIAS("ack", "foo.com.")
)

D("example.com", REG_NONE, DnsProvider(NS1), 
      A("whatever", "1.2.3.4"),
      ALIAS("ack", "emergencyDownPage.com.")
)
```

Actually not sure if there are valid use cases there. But this one may be more realistic:

```
// main domain served from provider X
D("example.com", REG_NAMECOM, DnsProvider(CF), 
      A("whatever", "1.2.3.4"),
      NS("a", "ns1.providerY.com.")
)

// Delegated subdomain served from provider Y
D("example.com", REG_NONE, DnsProvider(NS1), 
      A("a", "1.2.3.4"),
      A("b.a", "2.3.4.5")
)
```

My gut tells me more often than not it will be an error though.

tlimoncelli:
I'd rather prevent the error than handle a use-case that nobody can point to in reality.

If someone does come up with a use-case, we can add some kind of "DUP_OK" meta flag.

tlimoncelli:
@chewrocca You seem to be awesome at JavaScript. Would you be interested in taking a try at this?  I think a simple "error if a domain is defined the second time" would be sufficient.

captncraig:
[...]

PR #247, opened 2017-10-25
throw an error if a domain name is declared more than once

pmoroney:
Fixes: https://github.com/StackExchange/dnscontrol/issues/195

Note: I had to remove the following test: pkg/js/parse_tests/005-multipleDomains.js
I was unable to figure out why that test was added since it looks like it is pre-github.


jeanpaulgalea:
As a user of dnscontrol, being able to declare multiple `D()` for the same FQDN is useful in certain cases.

```
var DNS_BIND = NewDnsProvider('bind', 'BIND');
var DNS_GCLOUD = NewDnsProvider('gcloud', 'GLOUD');

EXAMPLE_ORG = [
  A('@','127.0.0.1'),
  MX('@',10,'mx.example.org.'),
]

D('example.org', REG_NONE, DNS_BIND,
  EXAMPLE_ORG,
  NAMESERVER('ns1.example.org.'),
  NAMESERVER('ns2.example.org.')
);

D('example.org', REG_NONE, DNS_GCLOUD,
  EXAMPLE_ORG
);
```
`example.org` is deployed to BIND and GCP DNS, with the only difference being the `NS` records.

This is useful when testing out new providers, such as moving from an in-house BIND/NSD setup to Anycast.

Another use-case is having infrastructure doing `AXFR` from a hidden master, to build monitoring, backup, and other config from the BIND zone, while using GCP DNS as an authoritative server for public resolving.

tlimoncelli:
Suggestion: Permit duplicates if metadata exists to permit it.

`{ duplicates_ok: "true" }`

tlimoncelli:
PS. There may be easier ways to do the examples in @jeanpaulgalea 's comment.

The EXAMPLE_ORG example can be done easier with default_ns (https://stackexchange.github.io/dnscontrol/providers/bind)

If you don't want to include NS records from a particular provider, you can do: `DnsProvider("name",0)`  See https://stackexchange.github.io/dnscontrol/js

In the last example, I think you could just list two DnsProvider()s for the same domain, but I'd have to understand your use-case better.

PR #248, opened 2017-10-25
docs, providers: Minor fix to gcloud

thorduri:


geek1011:
Good catch. I missed that during my review of the docs.

PR #253, opened 2017-10-26
Downcase DNS names

TomOnTime:
This still needs a  lot of testing.


captncraig:
I wonder if we can save having to touch every provider by having the diff code downcase both sides before starting.

TomOnTime:
@captncraig  I  considered that.  There are pros and cons both  ways.

This is all going to change when we implement the NO_PURGE and record-level "no touch" stuff. So, I think this is sufficient. (assuming all the tests pass)

tlimoncelli:
All tests pass for:

BIND
ROUTE53
GCLOUD
NAMEDOTCOM
CLOUDFLAREAPI
DIGITALOCEAN
GANDI

We can merge this if you are ok with putting the Downcase call in a more central place.

Issue #251, opened 2017-10-25
Source filenames should be consistent 


tlimoncelli:
When I need to make a change to all providers I usually use a glob such as `providers/*/*Providers.go` however this misses: google namecheap ns1 softlayer vultr

We should make all the providers have a similar filename structure.  Either `providers/token/tokenProvider.go` or `providers/token/main.go`

At the same time, we should fix the fact that some providers (activedir cloudflare google) have TOKEN names that do not match their subdirectory name. (i.e. GCLOUD vs. google/).

tlimoncelli:
If I don't hear any complaints, I'll probably implement this using `providers/$token/main.go` next week.

PR #261, opened 2017-11-01
Source filenames should be consistent.

tlimoncelli:
* Fixes https://github.com/StackExchange/dnscontrol/issues/251


tlimoncelli:
@geek1011 makes a good point.  I've corrected the few that were not in the current pattern (name/nameProvider.go).  At least that fixes the immediate issue.

PR #267, opened 2017-11-14
Add _acme-challenge as an allowed record name

koesie10:
I am using [acme-dns](https://github.com/joohoi/acme-dns) which requires me to add a CNAME record on `_acme-challenge` to another domain name. This pull request removes the warning created when doing that.

### DIFF
--- a/pkg/normalize/validate.go
+++ b/pkg/normalize/validate.go
@@ -85,7 +85,7 @@ func validateRecordTypes(rec *models.RecordConfig, domain string, pTypes []strin
 
 // underscores in names are often used erroneously. They are valid for dns records, but invalid for urls.
 // here we list common records expected to have underscores. Anything else containing an underscore will print a warning.
-var labelUnderscores = []string{"_domainkey", "_dmarc", "_amazonses"}
+var labelUnderscores = []string{"_domainkey", "_dmarc", "_amazonses", "_acme-challenge"}
 
 //these record types may contain underscores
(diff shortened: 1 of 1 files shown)


captncraig:
That is an interesting project. If I understand correctly, you CNAME `_acme-challenge.mymaindomain` to some other domain that this server thing handles? That is a pretty novel approach.

FYI, I've been working on a setup for dnscontrol to be able to issue certs using the dns challenge. I've been kinda waiting to see if LE gets their wildcard support out soon. Thanks for the update!

koesie10:
That is correct. I create a subdomain manually via `acme-dns`. Then, I just add `CNAME('_acme-challenge', '679c0af2-3602-406d-b3b5-6766c8b58277.auth.example.org.'),` in `dnscontrol.js`, along with

```
NS('auth', 'ns1.auth.example.org.'),
NS('auth', 'ns2.auth.example.org.'),
A('ns1.auth', '1.1.1.1'),
A('ns2.auth', '1.1.1.1'),
```

I use [koesie10/acme-dns-certbot-hook](https://github.com/koesie10/acme-dns-certbot-hook) to then handle the validation automatically with Certbot:

```
certbot certonly --manual --manual-auth-hook '/etc/acme-dns/acme-dns-certbot-hook -config /etc/acme-dns/acme_dns.json' --preferred-challenges dns -d example.org
```

If issuing certificates becomes possible using dnscontrol, would dnscontrol need to be run nightly to check for certificate expiration and renew certificates if needed?

PR #268, opened 2017-11-14
Linode provider

### LINKED ISSUE #121: Linode Support
[Linode API Docs: DNS](https://www.linode.com/api/dns)

koesie10:
Fixes #121

All tests pass, except for 25 because Linode performs subdomain validation which does not allow the 企业 TLD.


pmoroney:
By subdomain validation,  do you mean they check to see if the ööö.企业. record exists?

koesie10:
I'm sorry, I meant TLD validation, so the 企业 TLD is not in their systems and thus not allowed to be added as a target for a CNAME record.

pmoroney:
I reached out to Linode about the TLD.
Also I think the checks for the PR are failing because you need to run `go generate` and then commit those changes too.

providers/linode/linodeProvider.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+
+		var serviceName, protocol string = result[1], strings.ToLower(result[2])
+
+		req.Protocol = protocol
+		req.Service = serviceName
+		req.Name = ""
+	case "CNAME":
+		req.Target = fixTarget(req.Target, dc.Name)
+	}
```
tlimoncelli: Switch statements should list all possible rtypes so that when a new rtype is added, we fail fast.  Look at https://github.com/StackExchange/dnscontrol/blob/master/models/dns.go#L90 for an example.

providers/linode/linodeProvider.go
```diff
@@ -0,0 +1,303 @@
...
+func fixTarget(target, domain string) string {
+	// Linode always wants a fully qualified target name
+	if target[len(target)-1] == '.' {
+		return target[:len(target)-1]
+	} else {
+		return fmt.Sprintf("%s.%s", target, domain)
+	}
+}
+
+func fixTTL(ttl uint32) uint32 {
```
tlimoncelli: If the code modifies TTLs, the docs should make note of that.

providers/linode/linodeProvider_test.go
```diff
@@ -0,0 +1,22 @@
+package linode
+
+import (
+	"testing"
+)
+
+func TestFixTTL(t *testing.T) {
```
tlimoncelli: Please add a unit test for {299, ???} (i.e. that way you are testing, n, n-1, and n+1)

(by the way... thank you so much for having unit tests. So few people write them and they are very important for keeping the code maintainable)

tlimoncelli:
Thanks for the contribution!

PR #274, opened 2017-11-23
GANDI: Registrar support

philhug:
Adds support to the existing gandi provider to support changing nameservers.

### DIFF
--- a/docs/_includes/matrix.html
+++ b/docs/_includes/matrix.html
@@ -89,8 +89,8 @@
 		<td class="success">
 			<i class="fa fa-check text-success" aria-hidden="true"></i>
 		</td>
-		<td class="danger">
-			<i class="fa fa-times text-danger" aria-hidden="true"></i>
+		<td class="success">
+			<i class="fa fa-check text-success" aria-hidden="true"></i>
 		</td>
 		<td class="danger">
 			<i class="fa fa-times text-danger" aria-hidden="true"></i>
--- a/providers/gandi/gandiProvider.go
+++ b/providers/gandi/gandiProvider.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"sort"
 
 	"github.com/StackExchange/dnscontrol/models"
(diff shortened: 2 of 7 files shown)


philhug:
This will require the following PR to be merged before it can build:
https://github.com/prasmussen/gandi-api/pull/13

captncraig:
So this still uses the old xml api right? So current api keys people are using will be compatible?

philhug:
@captncraig correct, registrar features are not available on new API yet

captncraig:
Eww. that makes me think twice about wanting to implement the new api format then if you have to manage two different sets of credentials. 

Anyway. This looks good. Thanks!

philhug:
@captncraig there aren't any tests yet for registrar support, correct?

captncraig:
No, there are not. Maybe there could be, but it seems a little less straightforward to me for some reason.

philhug:
I think the following test procedure could make sense:
set domain servers to dns1.xyz.com,dns2.xyz.com
update to dns2.xyz.com,dns1.xyz.com (no change should happen)
update to dns1.xyz.com,dns2.xyz.com (no change should happen)
update to dns1.xyz.com,dns2.xyz.com,dns3.xyz.com (dns3 added)
update to dns1.xyz.com,dns2.xyz.com (dns3 removed)
update to dns2.xyz.com,dns1.xyz.com (again no change)

what do you think?

captncraig:
I think it is a good idea. I guess my biggest hangups were all pretty much that it generally requires a dedicated domain you never care about, but that's kinda true of the dns provider tests as well. 

Since we never do live queries, we never really care about waiting for the actual records to propagate to the tld nameservers or anything like that. If you wanted to implement something like that it would certainly be a good addition.

pmoroney:
I think it is a good idea as well. Most registries will do checks to verify that the nameservers have been registered by the nameserver's registrant, some will verify that they respond on port 53, and a few will verify that they are hosting the zone for the domain being updated.
So for testing purposes it might be best to use nameservers that wildcard for all domains.
The only ones I know of off the top of my head are for parking services, like ns1.sedoparking.com

PR #277, opened 2017-11-29
NS1: Add SRV record support

bashgeek:
Title says it all: Added support for SRV records for NS1 provider. Tested and works for us.


tlimoncelli:
Thanks!

Issue #91, opened 2017-04-18
Multiple strings in RR value (from convertzone) cause parse failure


philpennock:
Many RR-types which take strings, take N strings, for `N >= 1`.  Eg, the `TXT` RR-type.

`convertzone` took a zonefile which had such RR values and made dnsconfig directives which had two sequential strings with no operator between them, which led to a JS parse failure.

In this case, the cause is DKIM keys in TXT records.  The below is actual data for the `spodhuis.org` zone.

I read all the docs I could find before trying the tool and saw nothing describing how to manual multiple strings as part of one RR value.  The output from `convertzone` is not accepted as input into `dnscontrol`.

```
% dnscontrol
2017/04/17 20:55:23 main.go:56: Error executing javasscript in (dnsconfig.js): (anonymous): Line 42:280 Unexpected string (and 3 more errors)

% sed -n 42p < dnsconfig.js
	TXT('d201611._domainkey', "v=DKIM1; k=rsa; p=MIG7MA0GCSqGSIb3DQEBAQUAA4GpADCBpQKBnQDJZk8JRPxvefSR/6CWRqgiGQvvtGuFmvIaUuOyVys2TYE61h/8rC6se00NKzN1hU+gJwfw8FKk8l+JhJ/znd9KdIJreu41KzIFIxeWsdYpcbwvTGYzfdbWifqNrgFa4l/D6Nicreh2/hT5aI121167qoZ6fgBRWFZ5vKXJoRHtGoqOX8qCLBBGUDz7ZaujVfI" "KQ1oU26fnLa2etlkCAwEAAQ==", TTL(7200)),
```

tlimoncelli:
Hi Philip!  Long time no see!

TXT records are worrisome for me.  We implemented TXT() to only accept 1 string because we know all the DNS provider APIs support that.  Some don't support multiple strings (those scoundrels, huh?).  We figured that we'd cross the "multiple strings in TXT records" bridge when we get to it.

We generally try to have the DSL prevent users from doing dumb things. So, we often are more restrictive than the RFCs because we don't let people do things that are legal but rare enough that they are usually typos (see https://stackexchange.github.io/dnscontrol/why-the-dot).  For example, you'll notice that TXT records are JavaScript quoted. Our idea is that you should put the string you want as it should appear, and any escaping/etc. should be done behind the scenes.

[...]

PR #294, opened 2018-01-05
convertzone: Handle TXT records with multiple strings

tlimoncelli:
* Handle TXT records with multiple strings in "dsl mode"
* Check for the right number of command line args.


igg:
This doesn't appear to work with gcloud, which does support multiple strings for TXT, at least according to this [stackoverflow question](https://stackoverflow.com/questions/36688278/unable-to-setup-dkim-txt-value-as-dns-record)

tlimoncelli:
Can you file an issue with enough info that we can reproduce the issue?

igg:
This was also mentioned in issue #91.
Added a new issue #407

PR #296, opened 2018-01-08
Vet and Lint the entire system

tlimoncelli:


build/generate/featureMatrix.go
```diff
@@ -93,11 +93,15 @@ func generateFeatureMatrix() error {
 	return ioutil.WriteFile("docs/_includes/matrix.html", buf.Bytes(), 0644)
 }
 
+// FeatureDef describes features.
```
captncraig: Types in this file can remain unexported. No need to comment.

PR #298, opened 2018-01-10
Namecom v4 api

pmoroney:
Hey Everyone,

We just released the newest version of our API.
Old API Tokens should still work with the new API, but we also have a new management page to create new tokens, so new users have an easier path to sign up.
We also released a Go client library for the API, and so I ported over the namedotcom provider to the new library.
I have done extensive testing, and the dnscontrol integration tests helped in our design and testing of the new API.

The new documentation is here: https://www.name.com/api-docs


tlimoncelli:
Thanks, Pat!  I can't wait to test this out!

tlimoncelli:
Pat, thanks for the awesome PR!  I took a peek at the Go client library and its really awesome too!

Thanks for the great contribution!

PR #299, opened 2018-01-11
add TXTMulti capability to the namedotcom provider

pmoroney:
I believe we support the TXTMulti capability, these changes should allow it.
Let me know if the resulting DNS entries don't look right in production.


pmoroney:
Just realized I forgot to escape any quotes in the multitxt when sent to the API. I'll add that and an integration test when I get back to a computer.

Issue #183, opened 2017-08-14
Add an IGNORE() DSL directive


onlyhavecans:
Occasionally I have a record or two which is either not a supported type by dnscontrol, or I otherwise would not want it managed by dnscontrol. It would be nice to have a `IGNORE(name)` directive to allow dnscontrol to skip deleting or modifying the records

examples for why;
* DynamicDNS records controlled externally
* URL records from DNSimple & NameCheap

tlimoncelli:
Let me make sure I understand the feature.  Basically IGNORE("label") would tell DNSControl not to touch a DNS record in that domain whose name is "label". The contents would be created by some external mechanism. (i.e. someone has inserted the record via a web UI or other mechanism)

captncraig:
This sounds a lot like the (poorly documented) `NO_PURGE` option.

```
D("example.com", .... , NO_PURGE,
  A("foo","1.2.3.4")
)
```

Will cause dnscontrol to manage the records you define, but not delete others. Is that close to what you are asking for?

onlyhavecans:
@tlimoncelli your summary is correct. That is the desired behavior. I'd like to define the record as being handled externally so it is not tampered with.

@captncraig that's very close, but I'd like to be able to define specific records not just a blanket behavior for all records not defined.

For now I could probably use NO_PURGE to do this but I'd still like to allow for record cleaning.

gbrayut:
We did have something like that for the Cloudflare provider called [ignored_labels](https://github.com/StackExchange/dnscontrol/search?utf8=%E2%9C%93&q=ignored_labels&type=) but it probably is a good idea to create a better standard approach for how to do this across all providers.

tlimoncelli:
The cleanest solution is to implement "special" records in the providers. There is now a mechanism to list capabilities of a provider (for example, some providers are marked as supporting SRV records). URL records from DNSimple & NameCheap could be implemented fairly easily as a "REDIRECT()" record that does the right thing depending on the provider.

onlyhavecans:
At some point in the future I'd like to try my hand at adding URL record support much like the existing ALIAS records assuming someone else doesn't get to it first.

tlimoncelli:
REDIRECT(): @onlyhavecans Nobody is currently working on it.  I did recently update the doc on how to add new record types (https://stackexchange.github.io/dnscontrol/adding-new-rtypes) so this seems like an excellent opportunity for you.

[...]

PR #300, opened 2018-01-13
Add the IGNORE(name) directive (#183)

masterzen:
IGNORE is like NO_PURGE but for a spefic record instead of the whole zone. 

This is very useful for instance if you have a zone where only some records are managed externally from `dnscontrol` (for instance using [kubernetes external dns](https://github.com/kubernetes-incubator/external-dns)).

Adding `IGNORE("foo")` in the zone will make `dnscontrol` not trying to manage the "foo" record (and especially not deleting it). `dnscontrol` will error out if the "foo" record is both ignored and
managed in dnscontrol.

This can be seen as a generic Cloudflare's `ignored_labels`. This changeset also adds a warning deprecation when using the Cloudflare's `ignored_labels` metadata.


masterzen:
@tlimoncelli 
I've just pushed a new version with the rename.

Thanks!

Issue #239, opened 2017-10-17
Add support for Route 53 ALIAS record type


koenrh:
I'm looking to implement the Route 53 `ALIAS` record type. As you mentioned in the comments, Amazon's implementation of this record is a bit different from the implementation of other providers.

For an `ALIAS` record at Route 53 you not only need to specify the **name** and the **target**, but also the **target record type**. Which means we need to find a way to capture that in the DSL.

I think there are two options:

```
// 1. A custom "record type".

D("foo.com" "bar",
  R53_ALIAS("www", "A", "baz.cloudfront.net."),
  R53_ALIAS("www", "AAAA", "baz.cloudfront.net.")
);
```

```
// 2. A modifier for the existing ALIAS record type.

D("foo.com" "bar",
  ALIAS("www", "baz.cloudfront.net.", R53_RS_TYPE("A"),
  ALIAS("www", "baz.cloudfront.net.", R53_RS_TYPE("AAAA")
);
```

I'm curious to hear what you think is the right way to go.

These are the record types you could use in a Route 53 `ALIAS` record: `A`, `CNAME`, `AAAA`, `MX`, `TXT`, `PTR`, `SPF`, `NAPTR`, and `CAA`

pmoroney:
If you don't need to support the non A and AAAA records, it might be possible to implement the ALIAS record type for route53 that works identically to the other ALIAS implementations by creating an R53 ALIAS for both A and AAAA types.

koenrh:
That might be even a better solution. It would be nice to have the ALIAS records behave more or less the same across all provides. The only challenge I see is that in Route 53's case we would need to "synthesise" `A` and `AAAA` locally, where other providers do this server-side. It's not undoable, but I'm not sure whether that's something we should implement in `dnscontrol`.

tlimoncelli:
I am not familiar with AWS's aliases so this may be ignorant but...

dnscontrol knows if an alias is pointing at an A or AAAA (or both)... at least in the case where the destination record is in a domain that dnscontrol manages.  Does that help?

If it can't automatically determine the destination type, dnscontrol could look for metadata ({ "aws_alias_type": "AAAA"}).  If there is no metadata it could error out with a message explaining the situation.  This error would generally only happen when changing providers, and could give a link to appropriate instruction about what to do.

captncraig:
[...]

PR #301, opened 2018-01-14
ROUTE53: Support Route53's ALIAS record type (#239)

masterzen:
This PR implements the Route53 version of the ALIAS record type.

Route53 ALIAS doesn't behave like a regular ALIAS, and is much more limited as its target can only be some specific AWS resources (ELB, S3 buckets, CloudFront distributions, etc...) or another record in the same zone.

According to #239, this change adds a new directive R53_ALIAS which implements this specific alias. This record type can only be used with the Route53 provider.

This directive usage looks like this:
```js
D("example.com", REGISTRAR, DnsProvider("ROUTE53"),
R53_ALIAS("foo1", "A", "bar") // record in same zone
R53_ALIAS("foo2", "A",
  "blahblah.elasticloadbalancing.us-west-1.amazonaws.com",
   R53_ZONE('Z368ELLRRE2KJ0')) // ELB in us-west-1

```

Unfortunately, Route53 API requires indicating the hosted zone id where the target is defined (those are listed in AWS documentation, see the R53_ALIAS documentation for links).
Theoretically it can be auto-discovered by looking at the target dns name in which the region and type of AWS resource can be inferred, and then lookup in a static map. Or we can use the AWS API to find the proper resource and get its hosted zone id, which unfortunately would require more permissions (and in this case the user would need to specify the resource id as target instead of the dns name which might not be very practical).
 I thought it was kind of complicated so I resorted to the simple solution of letting the user indicate the zone id. The easy case is when referring to another record in the same zone, as we know already know the zone id, in which case it is not necessary to specify the zone id (that's the reason of the `R53_ZONE` modifier instead of a required parameters of the `R53_ALIAS` directive).

Note that unlike the regular ALIAS, the R53_ALIAS requires indicating the type of the resource, which can vary depending on the target. For a target pointing to another record in the same zone, it should be this record type, but for other it should either be a A or AAAA.

Please review, I'm fully open to suggestions and fixes.


tlimoncelli:
Wow! I'm so happy that you took the time to write R53_ALIAS! People have been asking for this for ages but we (stackoverflow) don't use AWS so we don't have the knowledge of how to implement it.

I've made a couple very small comments.   Please also add this to the feature matrix (ontrol/build/generate/featureMatrix.g)  (The need to add that was only recently documented https://github.com/StackExchange/dnscontrol/commit/79983f34938c20a8cdf3062d1f110086286a114f)

Thanks!

masterzen:
@tlimoncelli I've pushed a new version of the PR containing the `featureMatrix.go` changes (along with some documentation formatting fixes).

tlimoncelli:
I guess I should have merged them in the other order.  :-(

Could you rebase, "go generate" and check it in?

masterzen:
@tlimoncelli 

I don't think it was possible to merge both without a conflict as both PRs touched `helpers.js` (and thus `static.go`).

Anyway, I've just pushed a rebased version (including generation of `static.go`).

Thanks!

PR #304, opened 2018-01-18
Add gandi live DNS api provider

tjamet:
As discussed in https://github.com/tjamet/gandi-api/pull/2 here is a PR to add the gandi live DNS provider
I have updated the initial changes I made to have a more testable code and actually have some automated tests for the mapping logic.

There are still things I am not completely sure like the calls around `dc.CombineCAAs()` (the base code is inherited from the current gandi provider) and I did not have time to test it against production system yet.

I have added a test dependency to `github.com/stretchr/testify` being used to this assertion library but I am happy to move to any other assertion library if needed, to have a coherent testing harness.

I hope tests are readable enough to explain what are the conversions that are done by the library

The main idea when applying the correction is to create a new empty zone with the same characteristics as the current one, add all required records and finally migrate the domain to use this zone.

Happy to edit this PR with any feedback or idea

Signed-off-by: Thibault Jamet <tjamet@users.noreply.github.com>


tjamet:
Also missing google/uuid, would need some help to vendor it as well

tlimoncelli:
Is it possible to have one provider name (“GANDI”) which determines if the
domain is on the old or new API and automagically does the right thing?
-- 
Sent from Gmail Mobile

tjamet:
it seems an interesting idea but a bit more difficult to implement, I will think about it and try to propose something interesting

tlimoncelli:
Hi!  I was away at a conference for a  bit so I didn't have time to reply.

Does the same API key work for both APIs?  If not.... My suggestions is to simply try LiveDNS and if that fails, try the other. Record that  info so we don't do it for every API call.  Eventually everyone will move to LiveDNS and we can clean this up.

Slightly cleaner would be: Download the list of zones using LiveDNS and then again using the old protocol, and match up domains to which protocol to use.

tlimoncelli:
Sorry, I don't think I answered your questions properly. Here's a second try.

1.  Having two providers (GANDI and GANDI-LIVEDNS) is just fine.  They require different API keys.

2.  Using testify is fine.  I generally "role my own" but anything that encourages more testing is a good thing. 

Please only  write tests for algorithms, not APIs.   I see that you've created a fake zone manager so you can run tests against it.   That's wonderful and heroic, but it is a lot of time for little value.  Use our integration tests against the real service (with a test domain).

I used to spend hours writing these amazing test harnesses that created fake versions of API services... only to have the service change but my tests kept working.

A better test would be: given a list of records, does the expected zonefile get created?
(Whether the mechanism for uploading the new zone and activating it works or  not... that's what integration tests are for.)

3.  dc.CombineSRVs rewrites the RecordConfig so that instead of separate priority, port, weight, and  target fields, all that is combined into the .Target.  It makes it easier to do validation with everything split  out, but doing updates is helped by having one big string.  This is kind of janky, and I've added .Contents() which outputs the string instead. We'll probably refactor this in the future as the abstraction is getting leaky.

4. "The main idea when applying the correction is to create a new empty zone..." yes, that's what we do with the old Gandi provider.  If you look at protocol.go, the createGandiZone function gets the current zoneID, copies it, updates the copy, then activates it.

Best,
Tom

tjamet:
Hi Tom,

Sorry for the delay, so as requested, I have removed the client mocking part, keeping the tests of conversion logic only.
Not sure to get what to change regarding `dc.CombineSRVs`? should it be removed from `GetDomainCorrections` ?

Let me know if there is anything else to update

Thibault

tjamet:
although, not sure about how to run tests with the test domain, shall a test account be created? it looks like we would need credentials to do so no?

tlimoncelli:
You can run the integration tests on a particular provider by doing something like this:

```
export GANDI_KEY='the-key'
export GANDI_DOMAIN='example.com'
cd dnscontrol/integrationTest
go test -v -verbose -provider GANDI
```

Change the-key to your API key and replace example.com with  a  domain that you own (one that you don't mind being modified... because the test will delete all records).

Tom

tjamet:
Thanks for the info, I added the configuration to run the integration test and got it to work locally
the new version required a unique zone name, the one that is set is identifying the tool, the zone and the time the zone was created to be unique and add traçability

let me know if you have any other feedback I should consider

Thibault

tjamet:
hi @tlimoncelli please let me know if there is anything else to change on the PR

Best,

Thibault

tlimoncelli:
Sorry for the lack of reply. I thought I had written an update. (I just had a little surgery and I'll blame the pain medication :-) )

That said...

The code looks good so far.  I'm getting a weird error. Any suggestions?

```
tlimwork2:~/gitthings/tal-dns$ dnscontrol push -domains  example.tld
******************** Domain: example.tld
----- DNS Provider: gandi-live...1 correction
#1: Setting dns records for example.tld:
A example.tld 174.136.107.195 259200
AAAA example.tld 2607:f2f8:a9c0::3 259200
CNAME beta.example.tld example.tld. 259200
CNAME ftp.example.tld example.tld. 259200
CNAME www.example.tld example.tld. 259200
1 jj2.whatexit.org.
TXT example.tld v=spf1 mx -all 259200
FAILURE! Unexpected http code 400 on URL https://dns.api.gandi.net/api/v5/zones/16c9020a-0abb-11e8-9cd3-00163ec31f40/domains/example.tld. expecting 201
----- Registrar: ThirdParty...0 corrections
Done. 1 corrections.
Completed with errors
tlimwork2:~/gitthings/tal-dns$ 
```

tjamet:
no worries, it happens to everybody

well this is weird, like if the zone was well created but did not manage to be associated with the domain, does it always behave this way for you?

tlimoncelli:
Yes, it happens 100% of the time (though I  only have 1 domain  on the new API. All my other domains are on the old API).

tjamet:
I just created the same kind of config and got:

```
var namecom = NewRegistrar("none", "NONE");
var gandilive = NewDnsProvider("gandi", "GANDI-LIVEDNS")

D("jamet.io", namecom, DnsProvider(gandilive),
  A("@", "127.0.0.1", TTL(259200)),
  AAAA("@", "::1", TTL(259200)),
  CNAME("beta","@", TTL(259200)),
  CNAME("ftp","@", TTL(259200)),
  CNAME("www","@", TTL(259200)),
  TXT("@", "v=spf1 mx -all", TTL(259200))
)
```


```
******************** Domain: jamet.io
----- Getting nameservers from: gandi
----- DNS Provider: gandi...1 correction
#1: Setting dns records for jamet.io:
A jamet.io 127.0.0.1 259200
AAAA jamet.io ::1 259200
CNAME beta.jamet.io jamet.io. 259200
CNAME ftp.jamet.io jamet.io. 259200
CNAME www.jamet.io jamet.io. 259200
TXT jamet.io v=spf1 mx -all 259200
SUCCESS!
Done. 1 corrections.
```

tjamet:
I purshased this domain directly with api v5 / liveDNS.
I remember having to migrate my domains after enabling the live DNS API, which could take a while, could this explain the 400 ?

tlimoncelli:
The test domain I was using (name changed in the above example) was created natively on livedns.... but the process was odd. Maybe it wasn't set up correctly.

I just converted another domain to livedns.  I get the same error but I didn't wait the recommended 12-24 hours.  Lets see if I get that error tomorrow.

tjamet:
no worries, let me know if the problem persists

tjamet:
Hi @tlimoncelli did you have the chance to try again?

tlimoncelli:
I finally found some time to work on this yesterday.

The short version: The error I'm getting is because my domains are owned by a different organization than the API key is for.  I've opened up a support ticket because I can't figure out how things got in this situation.

Longer version:

I made some improvements to the code to (1) catch up to the big refactoring that was merged in master, (2) expose error messages at a lower level:

https://github.com/StackExchange/dnscontrol/tree/tjamet-gandi/livedns

If you diff against that branch, you'll see some of the changes need to be made in the vendored code.

It has only been 18 hours since I reported the problem to Gandi.  I'll poke them in a little bit.

tlimoncelli:
I got a reply:

> For the moment the API V5 is not complete (only the part to manage
> LiveDNS is active), you will have to use the API v4 to manage your
> domains. Since the V5 is not finalized you will not been able to manage
> the organization or get informations via this API.

tlimoncelli:
I'm confused.  Is the code working for you? ("dnscontrol push" as well as "preview"?) . If it works for you, but not for me, that would be odd.

tlimoncelli:
By the way... the error I'm getting is:

`{"location": "body", "name": "fqdn", "description": "This zone does not share the same organization as the given domain"}]}
`

tlimoncelli:
Gandi emailed me to say that they've escalated the issue.

tlimoncelli:
I got a reply!

It turns out we need to send a sharing_id

> You have a personal account 'tomlim', which has the API key.
> 
> There is also an organization, 'PLTS', which your personal account has permissions for.
> 
> If you want to create an object (such as a Zone) under the organization 'PLTS', you need to pass a sharing_id in your API call. Otherwise the object is created under your personal account 'tomlim'.
> 
> The sharing_id can be obtained via the browser URL string when in the 'PLTS' organization.
> 
> b21bd6d2-5383-11e7-8261-00163e61ef31
> 
> POST https://dns.api.gandi.net/api/v5/zone?sharing_id=b21bd6d2-5383-11e7-8261-00163e61ef31
> 
> We'll update our documentation to include these details. Sorry about the confusion.
> 
>

tlimoncelli:
I got it working!

It turns out you need to create the zone with the sharing_id set, then everything else takes care of itself.

Here's a branch with everything working:

https://github.com/StackExchange/dnscontrol/tree/tjamet-gandi/livedns

NOTE that I made changes in vendor/github.com/prasmussen/gandi-api/...

I have to fix some unit tests.  I'll do that this weekend (maybe next week) then send a PR to prasmussen.

Tom

tlimoncelli:
I merged the fork of this PR
https://github.com/StackExchange/dnscontrol/pull/320

@tjamet  thank you so much for your code and patience.  I'm super excited that we were able to work together and get this merged into dnscontrol!  Thank you so much!

tjamet:
Thank you @tlimoncelli great that you managed to make it working for organisations as well!
Thank you for your feedbacks, great to contribute!

vincentbernat:
I have a domain I try to manage with dnscontrol and get 403 when pushing data (no problem with preview). This is the only domain not attached to my personal organization. From my understanding, operations should be done using a sharing ID. It's unclear how to pass it to and if I need to pass it at all. Looking at the code, it seems the sharing ID is not something a user should set but it should be retrieved automatically by the `InfoByUUID` method. Clearly, this doesn't seem to work for me. Any clue?

tlimoncelli:
@vincentbernat Wow, you now know more about this than I do.  Do you have a recommendation or a way to reproduce the issue?

vincentbernat:
I have mailed the support and I am waiting for an answer.

You need a domain owned by another control but that you have rights to modify. In my case, I also didn't created the domain with dnscontrol. I have enabled LiveDNS through the Gandi web interface (using my own account). I don't know if that matters or not.

vincentbernat:
So, I got feedback from support and I didn't get anything else than what is said in the documentation. @tlimoncelli, in https://github.com/StackExchange/dnscontrol/pull/304#issuecomment-368148135, you say you fixed the problem with the code in https://github.com/prasmussen/gandi-api/pull/17, but how do you pass the sharing ID through dnscontrol? From the code, it seems you expect `/zones/XXXX` to return the sharing ID. The documentation says this is not the case. Could you confirm if you get a sharing ID in your case?

tlimoncelli:
That's a good question!  My recollection is that yes, the sharing ID came back from that call, so we used that. I think we saw it work on one domain, so we merged the code and didn't give it much thought.  Maybe it was a fluke, or maybe we imagined it! :-)

"Copy the ID from a website" is not what I would consider an API best practice. :-(

Anyway...

We could put the sharing Id in creds.json file.  Managing multiple organizations would be done by having one entry per org. Thoughts?

vincentbernat:
So, the sharing ID only works with domains which are part of an organization, not a personal account. This is not my case and it explains why it doesn't work. This is confusing as I am able to extract information through the `/domain` endpoints, but I cannot through the `/zone` endpoints. I've just generated API keys for the user owning the zone I want to manage and use that directly. In the end, I can't really say how this should work.

tlimoncelli:
What could we put in the docs?

vincentbernat:
I have honestly no idea.

tlimoncelli:
No worries. Their entire v5 API has been nothing but frustration. They really botched the upgrade.

Here's my suggestion: Open a new bug with the steps to reproduce the issue, include any theories you have, or info you got from support.  Let's take this from first principles and see where it goes.

tlimoncelli:
Can we close this?  The GANDILIVE driver was merged.  I think this refers to older code.

jsoref:
@tlimoncelli fwiw, it's `roll your own`  https://www.urbandictionary.com/define.php?term=roll%20your%20own (TIL where it came from)

Issue #306, opened 2018-01-24
Namedotcom: Deletes and Creates records


captncraig:
Pinging @pmoroney if you have any ideas. I'll be debugging in a bit, but you may be able to zero in on it faster than I get to it.

While running our zones, we started getting uneccessary corrections when running a zone with name.com as a dns provider. Some output:

```
#1: DELETE A isstackoverflowdownforeveryoneorjustme.com. 75.126.101.254 ttl=300
SUCCESS!
#2: DELETE A *.isstackoverflowdownforeveryoneorjustme.com. 75.126.101.254 ttl=300
SUCCESS!
DELETE TXT isstackoverflowdownforeveryoneorjustme.com. "google-site-verification=u9iyMz-mJRdqVH9GLvBiG7EgqzZ6h6LQzPjLpilPa6M" ttl=300
SUCCESS!
#4: DELETE TXT isstackoverflowdownforeveryoneorjustme.com. "v=spf1 -all" ttl=300
SUCCESS!
#5: CREATE TXT isstackoverflowdownforeveryoneorjustme.com "v=spf1 -all" ttl=300
SUCCESS!
#6: CREATE TXT isstackoverflowdownforeveryoneorjustme.com "google-site-verification=u9iyMz-mJRdqVH9GLvBiG7EgqzZ6h6LQzPjLpilPa6M" ttl=300
SUCCESS!
#7: CREATE A isstackoverflowdownforeveryoneorjustme.com 75.126.101.254 ttl=300
SUCCESS!
#8: CREATE A *.isstackoverflowdownforeveryoneorjustme.com 75.126.101.254 ttl=300
SUCCESS!
```
Looks to me like it has to do with the domains being read in with a trailing dot, but not restored that way or vice versa.

pmoroney:
That is the case, I will fix that real quick.

PR #307, opened 2018-01-24
trim trailing '.' from FQDN

pmoroney:
fixes https://github.com/StackExchange/dnscontrol/issues/306


captncraig:
Wow. thanks for taking care of this so quickly,

PR #309, opened 2018-01-26
NEW PROVIDER: OctoDNS

tlimoncelli:
(not ready for merge, but ready for review)

This "provider" generates YAML files suitable for the OctoDNS system.

The OctoDNS system uses YAML files with very loose structure rules making it extremely difficult to parse or generate the files properly in a statically typed language.  I isolated all the interface{} shenanigans into a module called octoyaml to keep the rest of the code clean.

OctoDNS's validator is extremely strict about label order, attribute order, etc.  The code generates validated config files for our DNS zones.


tlimoncelli:
@captncraig  This is ready for your review.

convertzone now reads OctoDNS files, which means we can advertise that we convert bidirectionally.

PR #315, opened 2018-02-04
encode and decode name.com txt records

pmoroney:
This fixes the tests. The integration tests work now too.

There were some small omissions and a double encoding at one point.
I also always forget about raw string literals, so I changed the strings to be more readable as well.

Also, I'll add this to our documentation on Monday, but for single TXT records, we don't require the double quotes at the beginning and end, and in that case it is a raw string so double quotes can exist anywhere inside the string on their own. For multiTXTs we surround them by double quotes and escape any double quotes inside with a backslash.

### DIFF
--- a/providers/namedotcom/records.go
+++ b/providers/namedotcom/records.go
@@ -97,13 +97,7 @@ func toRecord(r *namecom.Record, origin string) *models.RecordConfig {
 	rc.SetLabelFQDN(fqdn, origin)
 	switch rtype := r.Type; rtype { // #rtype_variations
 	case "TXT":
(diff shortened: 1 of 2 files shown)


tlimoncelli:
Dude you are amazing!  (and you fixed it on the weekend, for which I am even more thankful!)

Thank you so much!

Tom

P.S.  If you  want to copy the encode/decode functions into `github.com/namedotcom/go/namecom` to make them more universally  available please do.

pmoroney:
No problem :)
I was already thinking about moving the encode/decode into the client library since they seem so useful. I might have some time to get to that this week.

Issue #318, opened 2018-02-07
CAA_CRITICAL sets wrong flag into CAA record


ismaelperal:
According to RFC [6844](https://tools.ietf.org/html/rfc6844#section-5.1): Bit 0, Issuer Critical Flag:  If the value is set to '1', the  critical flag is asserted and the property MUST be understood if the CAA record is to be correctly processed by a certificate issuer. Where bit 0 match with the decimal value "128"

tlimoncelli:
Fixed.

PR #319, opened 2018-02-07
Update helpers.js CAA_CRITICAL flag=128 (#318)

ismaelperal:
CAA flag "Issuer Critical Flag" sets first bit (bit 0) to 1 where bit 0 is the 8th bit in the flag, so it's doing the change by left shift 7 positions the value 1.

#318 - RFC [6844](https://tools.ietf.org/html/rfc6844#section-5.1)


captncraig:
It seems there is a test failing. The value in https://github.com/StackExchange/dnscontrol/blob/master/pkg/js/parse_tests/014-caa.json needs to be updated.

ismaelperal:
@captncraig thanks, I wasn't aware of this tests. I've updated it.

Now I don't know the reason why it's failing:

after -> go run build/validate/validate.go
"ERROR: The following files are modified after go generate:
pkg/js/static.go"

Did I miss something else?

Thank you!

tlimoncelli:
That  just means  you need to  run "go generate".   Any changes to helpers.js require a run of "go generate"  to embed helpers.js in the binary.

tlimoncelli:
By the way... I don't use CAA (yet) so when I created the integration tests, I was just guessing about what would be important. If you could look at https://github.com/StackExchange/dnscontrol/blob/master/integrationTest/integration_test.go#L410 and suggest improvements, please let fee free to  add/change them.

ismaelperal:
@tlimoncelli Thanks, I will!

I think something is wrong in my development environment, sorry for the inconvenience and thanks for your help

tlimoncelli:
I think you'll be fine if you rebase then to "go generate".  Let me know if you need assistance.

PR #330, opened 2018-03-08
skip traffic flow policy records

domelu:


tlimoncelli:
Please run `gofmt -w    providers/route53/route53Provider.go`

domelu:
Done

tlimoncelli:
Awesome!  Thank you for your contribution!  (that's especially important since it is a feature we don't use therefore the bug is invisible to us)

PR #331, opened 2018-03-08
Fix #312 - rewrite IGNORE to use GetLabel

### LINKED ISSUE #312: IGNORE doesn't work with providers that don't set RecordConfig.Name
I just discovered that some (most?) providers never sets `RecordConfig.Name` for existing record, unlike what I did in the OVH provider (coincidentally I'm only testing with this provider).

Unfortunately, the #300 PR adding the IGNORE directive just test on the `Name`, not the `NameFQDN`. I'm going to push a bug fix ASAP for this issue.

But a quick question, shouldn't all provider fill the `Name` field along with the `NameFQDN` field?
The comment in https://github.com/StackExchange/dnscontrol/blob/master/models/dns.go#L61 as I read it seems to imply it should always exist.

masterzen:
As discussed in #312, after the refactor to generalize the use of `GetLabel`, this change makes IGNORE also use `GetLabel`.

Tested on provider OVH and Route 53.

### DIFF
--- a/providers/diff/diff.go
+++ b/providers/diff/diff.go
@@ -6,7 +6,6 @@ import (
 	"sort"
 
 	"github.com/StackExchange/dnscontrol/models"
-	"github.com/miekg/dns/dnsutil"
 )
 
 // Correlation stores a difference between two domains.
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Thanks for the contribution!

Issue #323, opened 2018-02-27
Major refactoring complete. Please test with your provider.


tlimoncelli:
Hey folks!

I've reworked and greatly simplified a lot of code.  The integration test work for all the providers that I could test.  However I don't have test accounts on all the providers.  Therefore, I'd appreciate it if everyone could try the latest code and let me know if your integration tests run cleanly.

In particular...

* digital ocean @Deraen
* dnsimple @aeden
* Linode @koesie10
* namecheap @captncraig
* ns1 @captncraig
* OVH @masterzen
* Vultr @geek1011

Thanks!
Tom

aeden:
This is already on master?

masterzen:
@aeden I assume @tlimoncelli is referring to this commit de4455942bf38dc6b49e03b56748b00be5c062cf which is in master.

tlimoncelli:
Yes, the master branch.

masterzen:
@tlimoncelli all integration tests pass for the OVH provider.

Deraen:
@tlimoncelli integration tests pass for Digitalocean

onlyhavecans:
This issue is new post refactor https://github.com/StackExchange/dnscontrol/issues/324
I am testing on behalf of DNSimple

tlimoncelli:
Update...

- [X] digital ocean @Deraen
- [ ] dnsimple @aeden
- [ ] Linode @koesie10
- [ ] namecheap @captncraig
- [ ] ns1 @captncraig
- [X] OVH @masterzen
- [ ] Vultr @geek1011

TomOnTime:
Friendly ping to @koesie10 , @captncraig , @geek1011

geek1011:
@TomOnTime sorry, missed the notification. I get too many notifications from GitHub so I usually miss important ones. I'll test this sometime today or tomorrow.

TomOnTime:
@geek1011 no worries!  (and no rush)

geek1011:
Works for my config. Will test a few other edge cases later.

tlimoncelli:
@geek1011 FYI: I've updated the [integration test documentation](https://github.com/StackExchange/dnscontrol/blob/master/integrationTest/readme.md) recently.

geek1011:
@tlimoncelli Looks great!

tIimonceIIi:
@geek1011 Who cares. Nobody asked your opinion.

TomOnTime:
WTF?  I didn't write that.

PR #333, opened 2018-03-08
Fix Linode provider

koesie10:
In response to #323.

I've tested it with the Linode provider, and due to commit https://github.com/StackExchange/dnscontrol/commit/19ca7600726ba8251e4038bc6c94e004e55697de, the order of the integration tests have changed. So, the failing test's number has changed from 25 to 27 in `providers.json`.

Moreover, the integration test for TXT records does not pass anymore because fixTxt has been removed (https://github.com/StackExchange/dnscontrol/commit/de4455942bf38dc6b49e03b56748b00be5c062cf#diff-cbc2e007a10f561a01f36d383d88aa80R259), which the provider relied on.

All other integration tests pass.

This fixes all those problems, but I haven't been able to test whether `fixTxt` breaks other providers.

### DIFF
--- a/integrationTest/providers.json
+++ b/integrationTest/providers.json
@@ -46,7 +46,7 @@
     "COMMENT": "25: Linode's hostname validation does not allow the target domain TLD",
     "token": "$LINODE_TOKEN",
     "domain": "$LINODE_DOMAIN",
-    "knownFailures": "25"
+    "knownFailures": "27"
   },
   "NS1": {
     "domain": "$NS1_DOMAIN",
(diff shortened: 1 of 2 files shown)


captncraig:
Feels to me like the linode provider itself should be populating the appropriate fields rather than relying on the core routine to fix it up after the fact.

koesie10:
I would agree, but this was the previous behaviour, so multiple providers might rely on this behaviour being there. Therefore, I added it to the core routine.

tlimoncelli:
True, but we decided it was unreasonable to ask every provider to get this right.  We created .SetTargetTXT*() so they don't have to worry about the details.  Also, if the struct changes, it puts the responsibility on us to fix the Set*/Get* functions.  Letting providers do the wrong thing and fixing it later with fixTXT() was a bad decision on my part and I wanted to fix this.

Soon we'll be unexporting .Target and .Name to enforce all of this.

How about something like this? (I haven't tested it)
https://github.com/StackExchange/dnscontrol/tree/tlim_linode

Tom

koesie10:
Alright, thanks for the draft, I did test it and the two lines commented out were needed due to the way the Linode API works. I've also changed other obvious references to `.Target` and `.Name` to use the getters/setters.

tlimoncelli:
Thanks for the submission!

PR #337, opened 2018-03-19
Refactor in preparation to unexport RecordConfig.{Name,NameFQDN,Target}

tlimoncelli:


captncraig:
The one thing that makes unexporting those really tricky is that the json deserializes into those. I have a few convoluted solutions to that problem, but none I really love.

tlimoncelli:
I'm no where near ready to unexport them (the branch name is a bit misleading). I just want to be prepared for when we do.  

I believe this is the most we can do without actually un-exporting those fields.

PR #342, opened 2018-03-21
Fix #341 do not erase all records labels

### LINKED ISSUE #341: Current master always set records as apex records
This issue prevents any record modifications (only apex records can be modified), worst it can rewrite all records to apex records.

As discovered in #339, when running dnscontrol, dnscontrol always wants to modify all zone records.

This is because all records coming from `dnsconfig.js` after a call to `DomainConfig.PunyCode()` have no `Label` nor `LabelFQDN` anymore. I think all providers call `PunyCode`.

This is because `PunyCode` rebuilds the `Label` from the `LabelFQDN` which itself has never been set (it used to be set in [`NormalizeAndValidateConfig`]( https://github.com/StackExchange/dnscontrol/blob/master/pkg/normalize/validate.go#L324).

masterzen:
This just puts back the code as it was before until a correct and better fix is found :)

### DIFF
--- a/pkg/normalize/validate.go
+++ b/pkg/normalize/validate.go
@@ -321,7 +321,7 @@ func NormalizeAndValidateConfig(config *models.DNSConfig) (errs []error) {
 			}
 
 			// Populate FQDN:
-			//rec.NameFQDN = dnsutil.AddOrigin(rec.GetLabel(), domain.Name)
+			rec.SetLabel(rec.GetLabel(), domain.Name)
 		}
 	}
 


pkg/normalize/validate.go
```diff
@@ -321,7 +321,7 @@ func NormalizeAndValidateConfig(config *models.DNSConfig) (errs []error) {
 			}
 
 			// Populate FQDN:
-			//rec.NameFQDN = dnsutil.AddOrigin(rec.GetLabel(), domain.Name)
+			rec.NameFQDN = dnsutil.AddOrigin(rec.GetLabel(), domain.Name)
```
tlimoncelli: Please use rec.SetLabel() as it will set both .Name and .NameFQDN at the same time.

masterzen:
@tlimoncelli I've updated the PR so that it uses `SetLabel`.

PR #347, opened 2018-03-22
BIND: Fix panic on CNAME

tlimoncelli:


necrogami:
Just tested this and given my previous config this fixed the panic however none of the hosts are showing in my file. Do you want me to create a new issue?

PR #358, opened 2018-04-26
fix names for internal transformed records

captncraig:
Transformed records were not given the correct fqdn.

For our transform to `.internal`, and src record `foo.stackexchange.com`, the transformed record is not getting the `.internal` added onto the end, so the fqdn is not inside the same zone.

This PR:

1. Fixes the bug in inport transform that creates the wrong fqdn.
2. Makes the AD provider use the standard diff messages that include fqdn.
3. Add a validation pass to make sure EVERY record has an fqdn that ends i the correct domain name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ certs/
 spfcache*
 adzonedump.*
 zones/
+stack.sh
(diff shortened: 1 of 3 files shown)


tlimoncelli:
LGTM

PR #368, opened 2018-06-20
Cloudflare: Fix parsing of priority field in record (fixes #367)

### LINKED ISSUE #367: Cloudflare: Error when unmarshaling priority field for cfRecord
Cloudflare now seems to be returning a string instead of an int for the priority field of a cfRecord. I'm not sure if this is a mistake in the Cloudflare API or not, but I think dnscontrol should use `json.Number` to allow for this kind of thing.

Here is the error: `Error getting corrections: Error fetching record list from cloudflare: json: cannot unmarshal string into Go struct field cfRecord.priority of type uint16`

And here is a sample piece of the json: `... {"id":"...","type":"MX","name":"...","content":"mxb.mailgun.org","proxiable":false,"proxied":false,"ttl":120,"priority":"10","locked":false,"zone_id":"...","zone_name":"...","modified_on":"2018-04-18T23:52:15.186916Z","created_on":"2018-04-18T23:52:15.186916Z","meta":{"auto_added":false,"managed_by_apps":false,"managed_by_argo_tunnel":false}} ...`

geek1011:
This fixes #367 by making `cfRecord` use a `json.Number` (to allow numbers to be parsed from strings) when parsing the Cloudflare API response.

### DIFF
--- a/providers/cloudflare/cloudflareProvider.go
+++ b/providers/cloudflare/cloudflareProvider.go
@@ -355,20 +355,20 @@ type cfRecData struct {
 }
 
 type cfRecord struct {
-	ID         string     `json:"id"`
-	Type       string     `json:"type"`
-	Name       string     `json:"name"`
-	Content    string     `json:"content"`
-	Proxiable  bool       `json:"proxiable"`
-	Proxied    bool       `json:"proxied"`
-	TTL        uint32     `json:"ttl"`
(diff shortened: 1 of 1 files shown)


tlimoncelli:
I can't figure out why gogen is failing.

Hey, @captncraig, could you take a look?

geek1011:
Any news?

PR #369, opened 2018-06-23
sort returned nameservers

onlyhavecans:
The API isn't sorting the name servers so even when they match they would report being different. 
This will prevent it trying to update them every time

I tested this locally on my configuations


geek1011:
Works for me.

Issue #352, opened 2018-03-24
SOA RR Missing


necrogami:
You can generate a SOA record via the provider for Bind however it is currently not documented.

necrogami:
actually it appears that you cannot generate a SOA directly. This feels off as an SOA record is actually a resource record and should be the same of the likes of A, AAAA, PTR and NS however the only way to generate an SOA record for bind is using the bind default_soa option which i feel is off. 

Also bind file's dont generate an $ORIGIN line like they do a $TTL line.

tlimoncelli:
Yes, the SOA record needs some love.   All the API-driven providers auto-generate the SOA, so they don't need it. BIND needs an SOA record, so we just take it from the existing zone file but will populate it with the default_soa values if it isn't found.  If you are interested in adding an `SOA()` function, [there are instructions here](https://stackexchange.github.io/dnscontrol/adding-new-rtypes) for adding new record types.

However let's first take a step back and ask, "What problem are you trying to solve?"   I'd like to understand how the current functionality is negatively affecting your DNS operations.

Similarly for not generating the $ORIGIN... BIND defaults $ORIGIN to be the name of the zone, which is specified outside the file as part of the named.conf file.  What problem would be solved by generating the $ORIGIN line in the file?

necrogami:
The system i'm building won't have filenames for the dnsconfig.js file It's being auto generated and i don't like having to set the SOA details in the bind config when it's a resource record not a bind specific record. Also setting the $ORIGIN is considered best practice from everything i've seen no you don't need it but when setting other values like SOA nameserver and email it can pull the value from the $ORIGIN i will look into generating the SOA record.

tlimoncelli:
$ORIGIN doesn't default to the filename. It defaults to the zone name.

For example if you have a zonefile named "default.zone" and use it multiple times, BIND does the right thing:

````
  zone "mydomain.com"    {    type master;   file "zones/master/default.zone";  };
  zone "otherdomain.com" {    type master;   file "zones/master/default.zone";  };
  zone "yourdomain.com"  {    type master;   file "zones/master/default.zone";  };
```

With a $ORIGIN in the file, that won't work.  I'd rather keep the flexibility that is gained by leaving it out.

tlimoncelli:
[...]

PR #373, opened 2018-07-18
New Provider HEXONET

papakai:
Ready to merge to get 1st release out :)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -5,6 +5,7 @@ providers/digitalocean @Deraen
 providers/dnsimple @aeden
 providers/gandi @TomOnTime
 # providers/gcloud
+providers/hexonet @papakai
 providers/linode @koesie10
 providers/namecheap @captncraig
 # providers/namedotcom
--- a/docs/_providers/hexonet.md
+++ b/docs/_providers/hexonet.md
@@ -1,0 +1,92 @@
+---
+name: HEXONET
+title: HEXONET Provider
+layout: default
+jsId: HEXONET
+---
+# HEXONET Provider
+
+HEXONET is a leading developer and operator of domain names and DNS platforms.
+Individual, service provider and registrars around the globe choose HEXONET for
+domains and DNS because of our advanced technology, operational performance and
+up-time, and most importantly for DNS expertise. DnsControl with HEXONET's DNS
+marries DNS automation with an industry-leading DNS platform that supports DNSSEC,
+PremiumDNS via Anycast Network, and nearly all of DnsControl's listed provider features.
+
+## Configuration
+
+Please provide your HEXONET login data in your credentials file `creds.json` as follows:
+
+{% highlight json %}
+{
+  "hexonet": {
+    "apilogin": "your-hexonet-account-id",
+    "apipassword": "your-hexonet-account-password",
+    "apientity": "LIVE", // for the LIVE system; use "OTE" for the OT&E system
+    "ipaddress": "172.31.3.16", // provide here your outgoing ip address
+    "debugmode": "0", // set it to "1" to get debug output of the communication with our Backend System API
+  }
+}
+{% endhighlight %}
+
+Here a working example for our OT&E System:
+
+{% highlight json %}
+{
+  "hexonet": {
+    "apilogin": "test.user",
+    "apipassword": "test.passw0rd",
+    "apientity": "OTE",
+    "debugmode": "0",
+  }
+}
+{% endhighlight %}
+
+## Usage
+
+Here an example DNS Configuration `dnsconfig.js` using our provider module.
+Even though it shows how you use us as Domain Registrar AND DNS Provider, we don't force you to do that.
+You are free to decide if you want to use both of our provider technology or just one of them.
+
+{% highlight javascript %}
+// Providers:
+var REG_HX = NewRegistrar('hexonet', 'HEXONET');
+var DNS_HX = NewDnsProvider('hexonet', 'HEXONET');
+
+// Set Default TTL for all RR to reflect our Backend API Default
+// If you use additional DNS Providers, configure a default TTL
+// per domain using the domain modifyer DefaultTTL instead.
+// also check this issue for [NAMESERVER TTL](https://github.com/StackExchange/dnscontrol/issues/176).
+DEFAULTS(
+    {"ns_ttl":"3600"},
+    DefaultTTL(3600)
+);
+
+// Domains:
+D('abhoster.com', REG_HX, DnsProvider(DNS_HX),
+    NAMESERVER('ns1.ispapi.net'),
+    NAMESERVER('ns2.ispapi.net'),
+    NAMESERVER('ns3.ispapi.net'),
+    NAMESERVER('ns4.ispapi.net'),
+    A('elk1', '10.190.234.178'),
+    A('test', '56.123.54.12')
+);
+{% endhighlight %}
+
+## Metadata
+
+This provider does not recognize any special metadata fields unique to HEXONET.
+
+## New domains
+
+If a dnszone does not exist in your HEXONET account, DNSControl will *not* automatically add it with the `dnscontrol push` or `dnscontrol preview` command. You'll need to do that via the control panel manually or using the command `dnscontrol create-domains`.
+This is because it could lead to unwanted costs on customer-side that we want to avoid.
+
+## Debug Mode
+
+As shown in the configuration examples above, this can be activated on demand and it can be used to check the API commands send to our system.
+In general this is thought for our purpose to have an easy way to dive into issues. But if you're interested what's going on, feel free to activate it.
+
+## IP Filter
+
+In case you have ip filter settings made for you HEXONET account, please provide your outgoing ip address as shown in the configuration examples above.
--- a/providers/hexonet/hexonetProvider.go
+++ b/providers/hexonet/hexonetProvider.go
@@ -1,0 +1,69 @@
+// Package hexonet implements a registrar that uses the hexonet api to set name servers. It will self register it's providers when imported.
+package hexonet
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/StackExchange/dnscontrol/providers"
+	hxcl "github.com/hexonet/go-sdk/client"
+)
+
+// HXClient describes a connection to the hexonet API.
+type HXClient struct {
+	APILogin    string
+	APIPassword string
+	APIEntity   string
+	client      *hxcl.Client
+}
+
+var features = providers.DocumentationNotes{
+	providers.CanUseAlias:            providers.Cannot("Using ALIAS is possible through our extended DNS (X-DNS) service. Feel free to get in touch with us."),
+	providers.CanUseCAA:              providers.Can(),
+	providers.CanUsePTR:              providers.Can(),
+	providers.CanUseRoute53Alias:     providers.Cannot("Using ALIAS is possible through our extended DNS (X-DNS) service. Feel free to get in touch with us."),
+	providers.CanUseSRV:              providers.Can(),
+	providers.CanUseTLSA:             providers.Can(),
+	providers.CanUseTXTMulti:         providers.Can(),
+	providers.CantUseNOPURGE:         providers.Can(),
+	providers.DocCreateDomains:       providers.Can(),
+	providers.DocDualHost:            providers.Can(),
+	providers.DocOfficiallySupported: providers.Cannot("Actively maintained provider module."),
+}
+
+func newProvider(conf map[string]string) (*HXClient, error) {
+	api := &HXClient{
+		client: hxcl.NewClient(),
+	}
+	api.APILogin, api.APIPassword, api.APIEntity = conf["apilogin"], conf["apipassword"], conf["apientity"]
+	if conf["debugmode"] == "1" {
+		api.client.EnableDebugMode()
+	}
+	if len(conf["ipaddress"]) > 0 {
+		api.client.SetIPAddress(conf["ipaddress"])
+	}
+	if api.APIEntity != "OTE" && api.APIEntity != "LIVE" {
+		return nil, fmt.Errorf("wrong api system entity used. use \"OTE\" for OT&E system or \"LIVE\" for Live system")
+	}
+	if api.APIEntity == "OTE" {
+		api.client.UseOTESystem()
+	}
+	if api.APILogin == "" || api.APIPassword == "" {
+		return nil, fmt.Errorf("missing login credentials apilogin or apipassword")
+	}
+	api.client.SetCredentials(api.APILogin, api.APIPassword, "")
+	return api, nil
+}
+
+func newReg(conf map[string]string) (providers.Registrar, error) {
+	return newProvider(conf)
+}
+
+func newDsp(conf map[string]string, meta json.RawMessage) (providers.DNSServiceProvider, error) {
+	return newProvider(conf)
+}
+
+func init() {
+	providers.RegisterRegistrarType("HEXONET", newReg)
+	providers.RegisterDomainServiceProviderType("HEXONET", newDsp, features)
+}
--- a/providers/hexonet/nameservers.go
+++ b/providers/hexonet/nameservers.go
@@ -1,0 +1,100 @@
+package hexonet
+
+import (
+	"fmt"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/StackExchange/dnscontrol/models"
+)
+
+var defaultNameservers = []*models.Nameserver{
+	{Name: "ns1.ispapi.net"},
+	{Name: "ns2.ispapi.net"},
+	{Name: "ns3.ispapi.net"},
+}
+
+var nsRegex = regexp.MustCompile(`ns([1-3]{1})[0-9]+\.ispapi\.net`)
+
+// GetNameservers gets the nameservers set on a domain.
+func (n *HXClient) GetNameservers(domain string) ([]*models.Nameserver, error) {
+	// This is an interesting edge case. hexonet expects you to SET the nameservers to ns[1-3].ispapi.net,
+	// but it will internally set it to (ns1xyz|ns2uvw|ns3asd).ispapi.net, where xyz/uvw/asd is a uniqueish number.
+	// In order to avoid endless loops, we will use the unique nameservers if present, or else the generic ones if not.
+	nss, err := n.getNameserversRaw(domain)
+	if err != nil {
+		return nil, err
+	}
+	toUse := []string{
+		defaultNameservers[0].Name,
+		defaultNameservers[1].Name,
+		defaultNameservers[2].Name,
+	}
+	for _, ns := range nss {
+		if matches := nsRegex.FindStringSubmatch(ns); len(matches) == 2 && len(matches[1]) == 1 {
+			idx := matches[1][0] - '1' // regex ensures proper range
+			toUse[idx] = matches[0]
+		}
+	}
+	return models.StringsToNameservers(toUse), nil
+}
+
+func (n *HXClient) getNameserversRaw(domain string) ([]string, error) {
+	r := n.client.Request(map[string]string{
+		"COMMAND": "StatusDomain",
+		"DOMAIN":  domain,
+	})
+	code := r.Code()
+	if code != 200 {
+		return nil, n.GetHXApiError("Could not get status for domain", domain, r)
+	}
+	ns := r.GetColumn("NAMESERVER")
+	sort.Strings(ns)
+	return ns, nil
+}
+
+// GetRegistrarCorrections gathers corrections that would being n to match dc.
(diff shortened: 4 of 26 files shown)


tlimoncelli:
Looks good so far!  Let me know how we may be of assistance.

papakai:
Sorry, I am pretty busy.  I am using the below configuration in a test zone file

```
var REG_HX = NewRegistrar('hexonet', 'HEXONET');
var DNS_HX = NewDnsProvider('hexonet', 'HEXONET', {
    'default_soa': {
        'master': 'ns1.ispapi.net.',
        'mbox': 'hostmaster.ispapi.net.',
        'refresh': 3600,
        'retry': 600,
        'expire': 604800,
        'minttl': 1440,
    },
    'default_ns': [
        'ns1.ispapi.net.',
        'ns2.ispapi.net.',
        'ns3.ispapi.net.'
    ]
});
```

Here the questions that came up up to now:
1) HEXONET is both DNS and Domain Registrar. Thus I assume that EnsureDomainExists, GetRegistrarCorrections and GetDomainCorrections have to be implemented in the registrar module?
That's at least what I tried to cover based on the things I recognized in other provider modules..
2) Where have I to check if a DNS Zone exists?  I assume in GetDomainCorrections, right?
3) Wouldn't it make sense to have a method EnsureDNSZoneExists (or EnsureZoneExists) to cover that functionality in the way you're doing it for the domain (EnsureDomainExistis)?
4) How can I access such a DNSProvider default configurations (provided above in config file) so that I can use it for creating the dns zone?
5) Some TLDs require that a DNSZone is already working before Domains can be successfully registered. Are such cases covered in dnscontrol somehow? if you need examples, let me know.
6) We would love to also automatically delete a created dns zone again in case the domain registration did NOT work to avoid that such a zone generates unnecessary costs on customer-side. Is it possible to cover such a case through dnscontrol?
7) How to update an existing SOA record at later time? (especially the mbox config)
8) We would love to be one of the offically supported registrars. What do we have to cover / to care about to achieve this at the end? At least we could realize OT&E system integration tests that could be run automatically. Something else? Or do you decided case by case?

Sorry for so many questions, any help -> very appreciated.

tlimoncelli:
> Here the questions that came up up to now:
> 1) HEXONET is both DNS and Domain Registrar. Thus I assume that EnsureDomainExists, GetRegistrarCorrections and GetDomainCorrections have to be implemented in the registrar module?
> That's at least what I tried to cover based on the things I recognized in other provider modules..

Yes, but they don't need to be in the same file.  Some modules split out the registrar stuff into a separate file.

To be clear... to be a Registrar, implement all the methods in `type Registrar interface` (defined in models/provider.go).  To be a DNS provider, implement the methods in `type DNSProvider interface`.  A "registrar" is just something that can change the NS records for a domain. It can't actually register, buy, sell, etc. domains.  There is a "dnscontrol create-domains" command which is for telling a DNS provider to recognize that a (pre-existing) domain should exist.

If there are confusing points, please either add what you learned to https://stackexchange.github.io/dnscontrol/writing-providers or suggest what you'd like clarified and I'll update the file.

> 2) Where have I to check if a DNS Zone exists?  I assume in GetDomainCorrections, right?

EnsureDomainExists() is a method of the `type DomainCreator interface`. It is only used
for the create-zone command.

> 3) Wouldn't it make sense to have a method EnsureDNSZoneExists (or EnsureZoneExists) to cover that functionality in the way you're doing it for the domain (EnsureDomainExists)?

I don't understand the question. What are you trying to achieve?

> 4) How can I access such a DNSProvider default configurations (provided above in config file) so that I can use it for creating the dns zone?

Providers hardcode default data.  for example `providers/digitalocean/digitaloceanProvider.go`

> 5) Some TLDs require that a DNSZone is already working before Domains can be successfully registered. Are such cases covered in dnscontrol somehow? if you need examples, let me know.

No, we don't have to deal with this since we're not going registration or domain buying, just updating the NS records on the parent domain.

> 6) We would love to also automatically delete a created dns zone again in case the domain registration did NOT work to avoid that such a zone generates unnecessary costs on customer-side. Is it possible to cover such a case through dnscontrol?

Again, not an issue.

> 7) How to update an existing SOA record at later time? (especially the mbox config)

This is an outstanding bug.  There is currently no way to do this.  It only affects the BIND driver, which has an easy work-around (hand-edit the zone file). The other providers simply auto-generate the SOA and don't let the users control it at all. Considering that nobody but spammers use the mbox data, we're ok with this.  Fixing this is being discussed in https://github.com/StackExchange/dnscontrol/issues/352

> 8) We would love to be one of the offically supported registrars. What do we have to cover / to care about to achieve this at the end? At least we could realize OT&E system integration tests that could be run automatically. Something else? Or do you decided case by case?

The main point of being "official" is that we will block a new release if the provider isn't working. Unofficial providers, if they are broken, will not block a new release.  Currently all the official providers are ones that Stack Overflow depends on in production, therefore we're willing to fix anything that is broken to get a release out.  We would need a commitment from a vendor to fix bugs within a certain SLA to make those official. We haven't worked out how to do that.

Our internal CI/CD system runs the integration tests for official providers, plus any that we have a good relationship and can get a no-cost account for testing (OT&E is fine).  We'll add you to our internal CI/CD system that runs automated tests.  At least we'll know when things break and can report it to you.

> Sorry for so many questions, any help -> very appreciated.

Glad to help!  I'm excited that you are interested in providing such good support to dnscontrol users!

I haven't had much time to work on dnscontrol lately.  It is basically feature-complete for our needs right now.

Tom

papakai:
Thanks a lot @tlimoncelli for this feedback, very helpful.

Indeed, I was misleaded by documentation or even by naming conventions of e.g. EnsureDomainExists that this is the place where a domain registar could check if a domain exists. if not I understood this to be the place to trigger a registration. This needs some review imo - I'll try to cover this after my module is ready.

This affects my Question 3 where you had problems to follow me. I was searching the method for creating the zone as I though EnsureDomainExists is used to register the domain in case it does not exist. That's why I was wondering why no EnsureZoneExists method is available as it would make sense then. I hope you were able to follow me now :)

>Providers hardcode default data. for example providers/digitalocean/digitaloceanProvider.go
I've also hardcoded default data like nameservers, our Backend System applies SOA defaults anyway in case we create the zone but do not provide a SOA RR. But that's not my question :)
What if the customer provides a default_soa and default_ns configuration through his dns config file, is that auto-considered in dnscontrol through an SOA RR update later on or have I to consider it when creating the zone? if I have to care about that when creating the zone -> how do I get access to that configuration in EnsureDomainExists? I hope I could explain my thoughts now in a better way :)
Out of your response to Question 7, I assume that it is bad practice to have such an configuration? As you point to the fact that most of your supported providers don't let users control the SOA.
Our system honestly allows it, but implementing this in dnscontrol isn't a must then imo. Getting the module working is the primary task. Caring about specifics like this, might be an optional target for the future.

As you see, you could clarify a lot of my questions and misunderstandings. again ty a lot!

papakai:
Is there an existing module that covers such OT&E tests so that I could have an eye on how this is realized the best way? or do you have any suggestions?

papakai:
> DomainCreator should be implemented by providers that have the ability to add domains to an account. the create-domains command can be run to ensure all domains are present before running preview/push. That's where EnsureDomainExists appears.

This is what gave me the impression that not-existing domains need to be registered.

tlimoncelli:
> Is there an existing module that covers such OT&E tests...

For OT&E we just run the integration tests with the OT&E credentials.  Step 7 (Integration Test) in https://stackexchange.github.io/dnscontrol/writing-providers explains how to run the integration tests.

papakai:
@tlimoncelli 
when I am in my own fork, I can't run the integration test.
I think this is because I am in namespace github.com/hexonet/dnscontrol and includes point to github.com/dnscontrol/...
I don't want to change these include paths as this breaks the PR finally. Any suggestions on that matter?

```bash
go test -v -verbose -provider HEXONET
=== RUN   TestDNSProviders
--- FAIL: TestDNSProviders (0.00s)
        integration_test.go:46: DSP type HEXONET not declared
=== RUN   TestDualProviders
--- FAIL: TestDualProviders (0.00s)
        integration_test.go:46: DSP type HEXONET not declared
FAIL
exit status 1
FAIL    github.com/hexonet/dnscontrol/integrationTest   0.008s
```

papakai:
@captncraig can you please go into more details how to do this exactly?

captncraig:
1. Clone to `$GOPATH/src/github.com/StackExchange/dnscontrol
2. In that repo: `git remote add mine git@github.com:hexonet/dnscontrol.git`
3. `git push -u mine hexonetProvider`

captncraig:
That is for providers that allow you to create and host arbitrary dns names, but not necessarily register them.

We implement this for route53, cloudflare, and some others that let you host dns zones whether or not you actually own it or have the root nameservers configured properly.

Other providers do not implement it (name.com, namecheap), because they do not allow arbitrary user-created zones, and require you to buy a domain through them. 

This is not a required feature to implement.

papakai:
@captncraig sounds like we could support this. how to implement this? can we support here?

papakai:
@tlimoncelli @captncraig the current status of our PR passes the integration tests. We would love to get this released in short and as pointed out as official supported provider. How do can we proceed here?

Even though I have still open questions, it doesn't block a release. Things can be updated with a follow-up release.

papakai:
I've extended our API Connector library with a debug mode. I've updated the vendor dependency accordingly. The debug mode can be activated by environment setting when executing the integration tests. It shows details of the communication with our Backend system API.

papakai:
Hey guys,

even though we don't want to pressure: is there an update on this PR?

Thanks so much!
Kai

tlimoncelli:
FYI: I'm on vacation until Aug 13.

captncraig:
I'm not ignoring you either. Glad to see your interest. I'll respond to all
your issues when I'm in the office this afternoon.

papakai:
@tlimoncelli thanks for your update. I wish you a well regeneration. :)

papakai:
FYI: I am out of office on August 13th and 14th. August 15th is public holiday in Germany.

captncraig:
Ok, I'll be reviewing today, and hopefully get this merged asap. Sorry about the delays, but we've had a bunch of vacation get in the way. Thanks.

papakai:
I've cleaned up the warnings so that all checks passed, ready to merge. ;o)

papakai:
@captncraig something I can support with? open questions?
At least you would need the ot&e credentials for integration tests ;o)
Let me know how I've to share that with you guys.

docs/_providers/hexonet.md
```diff
@@ -0,0 +1,87 @@
+---
+name: HEXONET
+title: HEXONET Provider
+layout: default
+jsId: HEXONET
+---
+# HEXONET Provider
+
+HEXONET is a leading developer and operator of domain names and DNS platforms.  Thousands of individuals, service providers and registrars globally choose HEXONET for domains and DNS for our advanced technology, operational performance and up-time, and most importantly for unparalleled expertise in DNS. DnsControl with HEXONET's DNS marries DNS automation with an industry-leading DNS platform that supports DNSSEC, PremiumDNS via Anycast Network, and nearly all of DnsControl's listed provider features.
```
tlimoncelli: Please limit lines to 80 cols (makes future diffs smaller).

Also... would you mind removing the sentence "Thousands ... expertise in DNS" as it seems like a marketing pitch.

papakai: Hey @tlimoncelli,

it comes from our marketing department, I will talk to them if we can drop that sentence out.
Maybe just leaving "Thousands" out already removes that marketing pitch impression.
What you think?

OWNERS
```diff
@@ -5,6 +5,7 @@ providers/digitalocean @Deraen
 providers/dnsimple @aeden
 providers/gandi @TomOnTime
 # providers/gcloud
+# providers/hexonet
```
captncraig: If you don't mind putting your name here, it will help you be a recommended reviewer for future issues  / prs referencing this code.

providers/hexonet/hexonetProvider.go
```diff
@@ -0,0 +1,77 @@
+// Package hexonet implements a registrar that uses the hexonet api to set name servers. It will self register it's providers when imported.
+package hexonet
+
+// HEXONET is a leading developer and operator of domain names and DNS platforms.
```
captncraig: I don't think this comment s necessary.

providers/hexonet/hexonetProvider.go
```diff
@@ -0,0 +1,77 @@
...
+	providers.CanUseCAA:              providers.Can(),
+	providers.CanUsePTR:              providers.Can(),
+	providers.CanUseRoute53Alias:     providers.Cannot(),
+	providers.CanUseSRV:              providers.Can(),
+	providers.CanUseTLSA:             providers.Can(),
+	providers.CanUseTXTMulti:         providers.Can(),
+	providers.CantUseNOPURGE:         providers.Can(),
+	providers.DocCreateDomains:       providers.Can(),
+	providers.DocDualHost:            providers.Can(),
+	providers.DocOfficiallySupported: providers.Can(),
```
captncraig: "Officially supported" is a designation that means the maintainers (Stack Exchange) actively use and guarantee support for a provider. The hexonet provider will be "community supported", as we do not have the expertise to debug it thoroughly. But you do :)

providers/hexonet/hexonetProvider.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func newProvider(conf map[string]string) (*HXClient, error) {
+	api := &HXClient{
+		client: hxcl.NewClient(),
+	}
+	api.APILogin, api.APIPassword, api.APIEntity = conf["apilogin"], conf["apipassword"], conf["apientity"]
```
captncraig: Perhaps we should default "apientity" to the live system to save anyone needing to know the magic "54cd" identifier?

Perhaps even switch it to a "use OTE" type flag and handle the entity ids internally only?

providers/hexonet/nameservers.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (n *HXClient) updateNameservers(ns []string, domain string) func() error {
+	return func() error {
+		cmd := map[string]string{
+			"COMMAND": "ModifyDomain",
+			"DOMAIN":  domain,
+		}
+		for idx, ns := range ns {
+			cmd["NAMESERVER"+strconv.Itoa(idx)] = ns
```
captncraig: rather than strconv and concatenation, I think I'd prefer `cmd[fmt.Sprintf("NAMESERVER%d", idx]`

providers/hexonet/hexonetProvider.go
```diff
@@ -0,0 +1,77 @@
...
+	if conf["debugmode"] == "1" {
+		api.client.EnableDebugMode()
+	}
+	if len(conf["ipaddress"]) > 0 {
+		api.client.SetIPAddress(conf["ipaddress"])
+	}
+	if api.APIEntity != "1234" && api.APIEntity != "54cd" {
+		return nil, errors.Errorf("wrong api system entity used. use \"1234\" for OT&E system or \"54cd\" for Live system")
```
captncraig: Prefer `fmt.Errorf` to the `errors` package directly. Mostly for consistency across the code base.

providers/hexonet/nameservers.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		response := n.client.Request(cmd)
+		code := response.Code()
+		if code != 200 {
+			return errors.New(strconv.Itoa(code) + " " + response.Description())
```
captncraig: `fmt.Errorf` makes this a little cleaner.

providers/hexonet/records.go
```diff
@@ -0,0 +1,277 @@
...
+		return nil, err
+	}
+	actual := make([]*models.RecordConfig, len(records))
+	for i, r := range records {
+		actual[i] = toRecord(r, dc.Name)
+	}
+
+	for _, rec := range dc.Records {
+		if rec.Type == "ALIAS" {
+			rec.Type = "CNAME" // TODO as we fully support ALIAS through our XDNS system
```
captncraig: Are these just plain cnames, or do your nameservers do some additional flattening of cnames? We use cnames for cloudflare, for example, in place of alias, because they transparently flatten them and serve as A records. Does hexonet do something like that? If not, then we should not claim to support alias records as they are semantically different from cnames.

providers/hexonet/records.go
```diff
@@ -0,0 +1,277 @@
...
+	// Print a list of changes. Generate an actual change that is the zone
+	changes := false
+	params := map[string]string{}
+	delrridx := 0
+	addrridx := 0
+	for _, cre := range create {
+		changes = true
+		fmt.Fprintln(buf, cre)
+		rec := cre.Desired
+		params["ADDRR"+strconv.Itoa(addrridx)] = n.createRecordString(rec, dc.Name)
```
captncraig: All of these params would probably look better to me with fmt.Sprintf.

providers/hexonet/hexonetProvider.go
tlimoncelli: "officially supported" providers block a release if they don't work. "community supported" is best effort... we'll ship a release if it is broken and haven't heard back from the author.

providers/hexonet/hexonetProvider.go
papakai: Yes, that how we already cover it in our connector library `go-sdk`. By default live system will be used. Just in case customer wants to use OT&E system, method api.client.UseOTESystem() needs to be called.
The intention of the methods UseOTESystem and UseLIVESystem was exactly to not force people to use this very technical entity values. My integration here was a quickshot with a planned later rewrite - that I've forgotten to apply. Thanks for pointing on that.

providers/hexonet/hexonetProvider.go
papakai: Ok, I will forward this - I was asked to get us as officially supported. If that's not possible atm, I can't change it :)

providers/hexonet/hexonetProvider.go
papakai: cleaned up

OWNERS
papakai: I don't mind - it was just part of gettubg "officially" supported.

providers/hexonet/nameservers.go
papakai: makes sense

providers/hexonet/hexonetProvider.go
papakai: makes sense

providers/hexonet/nameservers.go
papakai: makes sense

providers/hexonet/records.go
papakai: You were right, I've discussed this with our CTO. Falling back to CNAME is for sure wrong.
For ther 1st release, I've decided to leave out ALIAS support.
Our system supports ALIAS RR in correct way, but it creates in addition A Records or AAAA Records (copied from target zone) which will get updated from time to time.
So creating a ALIAS RR automatically comes with further assigned records. Not sure how dnscontrol can handle this. Is it possible to apply such additional and automatically created RR also to customer's dnsconfig file so that this stays transparent?

The easiest and working way would be to simply filter these A or AAAA RR out that are bound to that ALIAS RR, but I've to ensure that I can do this without running into further issues.

docs/_providers/hexonet.md
papakai: I've updated the text. Is that version ok for you?

papakai:
Thanks for the review, I've applied all changes. Let me know if there's still something you want to have changed.

providers/hexonet/records.go
captncraig: Are you saying that if I make an ALIAS record, your system will create the underlying A/AAAA records, and actually return them form future api requests in addition to the original ALIAS record? That seems... odd. My naive expectation would probably be that those records get created only on the nameserver side of things, and remain hidden to the api consumer if possible. If they are indeed returned, then yes, they would probably need to be filtered out if identifiable in order to eliminate infinite update loop scenarios if running dnscontrol multiple times. We hope we can establish enough consistency so that if you ever run 'dnscontrol push' twice in a row, you should expect zero corrections on the second run.

papakai: Looks like I was faced with an bug in our API. These A RRs shouldn't get returned. I will work on ALIAS feature in the next days. But as pointed out: this doesnt block 1st release.

papakai: Launch of ALIAS is postponed for next weeks, no ETA for that backend system bugfix. So won't be part of the initial release. All parts of our module do already consider the change accordingly - no support for ALIAS. We plan this for the next release.

Thanks guys

providers/hexonet/hexonetProvider.go
papakai: got already changed to community supported.

papakai:
Can we get this merged this week? I am asking because I am weeks offline in holiday starting on September 1st.
TIA Kai

tlimoncelli:
That shouldn't be a problem. Craig is out today but should get to this by Wednesday.

Issue #258, opened 2017-11-01
error with convertzone build


MattTunny:
Hey guys,

I'm still new at Go, I can build the main dnscontrol no problems but when i try build ./cmd/convertzone it fails.

following doco: https://github.com/StackExchange/dnscontrol/blob/master/cmd/convertzone/README.md

```
$ cd cmd/convertzone
$ go build
$ cp convertzone ~/bin/.
```

give me this:
```
main.go:27:2: cannot find package "github.com/miekg/dns" in any of:
        /usr/local/go/src/github.com/miekg/dns (from $GOROOT)
        /Users/matthew.tunny/go/src/github.com/miekg/dns (from $GOPATH)
main.go:28:2: cannot find package "github.com/miekg/dns/dnsutil" in any of:
        /usr/local/go/src/github.com/miekg/dns/dnsutil (from $GOROOT)
        /Users/matthew.tunny/go/src/github.com/miekg/dns/dnsutil (from $GOPATH)
main.go:29:2: cannot find package "github.com/pkg/errors" in any of:
        /usr/local/go/src/github.com/pkg/errors (from $GOROOT)
        /Users/matthew.tunny/go/src/github.com/pkg/errors (from $GOPATH)
```

its probably something i'm doing wrong?

MattTunny:
found the issue: https://golang.org/doc/code.html

i was working from my normal github directory, had to be in
`/Users/matthew.tunny/go/src/github.com/StackExchange/dnscontrol/cmd/convertzone`

captncraig:
Yeah, the convertzone utility should not be separate like that anymore. We refactored the main cli, and we mean to move it into the main dnscontrol as a sub-command, but haven't gotten around to it yet. See #204

Issue #204, opened 2017-09-15
ConvertZone should be moved into a dedicated subcommand


captncraig:
Should probably be moved from cmd to commands.

And docs adjusted.

No rush, but it would be nice to have i the main binary.

pierrehenrymuller:
Until convertzone is reintegrated into dnscontrol control, is it possible to put the compiled binary in the releases and in the docker image?

I can't compile it either like #258 and it's not present in the docker image, so I can't evaluate dnscontrol for zone management.

Thank you.

pierrehenrymuller:
compilation is ok now

signalwerk:
I also struggle to get convertzone to run. I would love to see it in de docker build

tlimoncelli:
Did f1572c4b11690178c97a091a9b168feba0089d1c fix this?

signalwerk:
@tlimoncelli for my case yes. ((it's my commit to the dockerfile))

PR #377, opened 2018-07-25
Convert convertzone to a command

tiennou:
First time Go writer (one of those again ^^) ! I've just wanted to give a spin to `dnscontrol`, and the migration pointed me toward #258 / #204, so I've bit the bullet and fixed it.

Note that I've run it against one of my zones and it output an empty `D()`, so maybe I've broken it (I admit I didn't even check what would have done the original 😜).

I'll gladly accept style nitpicking (for obvious reasons), if I'm not matching Go-style.

I'm not quite sure what to do with the associated `README.md`, so I've just moved it with the rest of the docs, and I'll try to come back and cleanup/relink/clarify.


tlimoncelli:
Now that `dnscontrol get-zones` exists, there is no need for convertzone.  Closing.

PR #378, opened 2018-07-26
Include PTR types in SPF Builder

kjacobsen:
SPF builder doesn't support PTR types.

I think PTR counts as a DNS lookup, but happy for more opinions on the matter.


kjacobsen:
Hi,

Don't want to rush anyone, but can we look at getting this merged?

captncraig:
Sure. Thanks for the pr.

PR #386, opened 2018-08-07
Support AWS Tokens in credentials.NewStaticCredentials

ashleyhull-versent:


ashleyhull-versent:
Fixed those build errors.

PR #395, opened 2018-08-25
Create directories with execute permissions so they can be opened

arafferty:
`get-certs` currently fails with a permission denied after it creates new directories because MkdirAll is using 0644 instead of 0755. This PR resolves that.

### DIFF
--- a/pkg/acme/acme.go
+++ b/pkg/acme/acme.go
@@ -77,7 +77,7 @@ func (c *certManager) IssueOrRenewCert(cfg *CertConfig, renewUnder int, verbose 
 	}
 
 	log.Printf("Checking certificate [%s]", cfg.CertName)
-	if err := os.MkdirAll(filepath.Dir(c.certFile(cfg.CertName, "json")), perms); err != nil {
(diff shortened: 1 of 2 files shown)


captncraig:
What platform are you on? Yes, using 755 will solve all permissions issues,
but certificate directories really should be as restricted as possible. I
kinda feel like 0600 is an even better choice. What need is there for
execute permission on anything?

>

arafferty:
Ubuntu 16.04 on ext4. Same behavior on OS X. My bet is that your tests are being run with root permissions (docker? Windows?). The execute bit has to be set on directories in order for users to be able to enter and access files inside that directory (https://superuser.com/questions/168578/why-must-a-folder-be-executable). root obviously gets to ignore this detail.

What this means is that the MkdirAll call creates the certs directory but then subsequently fails when run as a non-root user when it tries to do anything inside of that directory (I haven't checked but my guess is that it tries to list the contents of the directory to see if it needs to create the next directory in the tree and fails).

0700 seems like a perfectly reasonable setting so that non-root users can still interact with the directory. I was just copying the generally permissive setting on the files themselves. While I'm fiddling with permissions should I also change it so that the files themselves are created as 0600?

tlimoncelli:
0700 seems reasonable to me.

PR #399, opened 2018-08-30
Correctly group R53_ALIAS records during IncrementalDiff.

ebardsley:
This mirrors logic in the R53 provider, and deals with spurious corrections in
the case that aliases exist for both A and AAAA records, which is common.

This also centralizes logic around keys to the models package and adds tests.


captncraig:
Great improvements all around. Thanks.

PR #409, opened 2018-09-29
update go version to 1.10

f110:
Can not build by go version under 1.9.
because using the strings.Builder by library.
https://github.com/StackExchange/dnscontrol/blob/36228949e9b6192f517708ea99d06481e7b18513/vendor/github.com/hexonet/go-sdk/client/socketcfg/socketcfg.go#L62

So update to 1.10 of docker image and README.md.


tlimoncelli:
Good catch! Thanks!

Issue #412, opened 2018-10-12
Can't preview a specific domain with OVH provider


tiennou:
I stumbled on the `--domains` argument of `preview`, which could be helpful in keeping the output sane, but it seem to be causing some kind of failure when using the OVH provider.

I'm getting this output with `dnscontrol preview --domains one-of-my.domains` :
```
******************** Domain: one-of-my.domains
----- DNS Provider: ovh...ERROR
Error getting corrections: Domain one-of-my.domains not found in your ovh account
Done. 0 corrections.
Completed with errors
```

While a normal `dnscontrol preview` results in the following :
```
[gobs of output for all domains]
******************** Domain: one-of-my.domains
----- DNS Provider: ovh...21 corrections
[21 DNS corrections]
----- Registrar: ovh...1 correction
#1: Change Nameservers from 'dns102.ovh.net,ns102.ovh.net' to 'ns1.one-of-my.domains,ns2.one-of-my.domains'
Done. 153 corrections.
Completed with errors
```

captncraig:
That seems... odd. Are you running the exact same commands except for `push` and `preview`? Its like the same code... Hmm.

tiennou:
It seems to be pretty consistent. I just noticed that I had some not-enabled zones in OVH (which is a known caveat that they have to be manually enabled), but even after enabling all zones, it still fails when a specific domain is "selected".

tiennou:
Also, `push` errors in the same way (though I tested against a not-changed zone).

tlimoncelli:
OVH is a community-supported provider.  Maybe @masterzen can look into this?

masterzen:
Indeed, I'm going to look at it @tlimoncelli.

masterzen:
@tiennou , I'm sorry I was unable to reproduce the issue with the latest master. I have 2 test domains out of 5 defined in my `dnsconfig.js`. Those domains have  items to modify at ovh, and `--domains` is working correctly, selecting the correct domains to either preview or apply.

What version of `dnscontrol` did you use?
What could be different in your setup?

One possibility is that your consummer key authorizations doesn't allow this domain to be modified or listed.

Looking at the ovh provider code, this error is raised when the `--domains` domain is not listed in the zone list accessible by your account as returned by the [list zone API call](https://api.ovh.com/console/#/domain/zone#GET).
Can you manually try this call in the ovh api console (see the previous link)  and check what it returns?

tlimoncelli:
[...]

PR #417, opened 2018-10-15
Populate ovh zones cache as early as possible (#412)

masterzen:
We were caching the zones in GetNameservers which wasn’t a good idea, because this one might not be called if not name servers are selected for this ovh domain. 
In this case, GetDomainCorrections would fail with an unknown domain.
Instead the idea is to populate the zones cache as soon as the provider is initialized.

### DIFF
--- a/docs/_includes/matrix.html
+++ b/docs/_includes/matrix.html
@@ -443,7 +443,9 @@
 		<td><i class="fa fa-minus dim"></i></td>
 		<td><i class="fa fa-minus dim"></i></td>
 		<td><i class="fa fa-minus dim"></i></td>
-		<td><i class="fa fa-minus dim"></i></td>
+		<td class="success">
+			<i class="fa fa-check text-success" aria-hidden="true"></i>
+		</td>
 		<td class="success">
 			<i class="fa fa-check text-success" aria-hidden="true"></i>
 		</td>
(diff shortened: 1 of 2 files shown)


masterzen:
Hmm, no idea why the build fails, I'll have a look.

tiennou:
See #416 😉.

tiennou:
Results are in !

```
dnscontrol preview --domains my-domain.com,our-domain.org,someone-elses-domain.com
******************** Domain: my-domain.com
----- DNS Provider: ovh...0 corrections
----- Registrar: ovh...0 corrections
******************** Domain: our-domain.com
----- DNS Provider: ovh...0 corrections
----- Registrar: ovh...0 corrections
Done. 0 corrections.
```

So this fixes #412. Thanks for the insight @masterzen !

masterzen:
Grrr, I don't get why the `gogen` status says `exit 1`, as it is ran by Travis and travis is green telling that the generate phase exited with 0.

When I run `go generate` locally I also get a 0 exit code, and no modified files as reported by `git diff --name-only` (which is what `validate.go` is using to check there are modified files).

@captncraig would you have an idea of the issue?

masterzen:
OK, this is finally green. I think the status were coming from travis building my fork of `dnscontrol` (and failing). 

@captncraig @tlimoncelli , feel free to review and merge to fix #412.

Thanks!

PR #429, opened 2018-12-14
Apply brand-preferred case

nikolay:


tlimoncelli:
This looks 90% good.  The changes to `*.md` and comments look safe.  I'm not comfortable making changes to code itself.  Could you revert `pkg/js/helpers.js`.  Also, the COPYRIGHT file comes from Cloudflare itself, so I'd rather not modify it. Please revert it too.

tlimoncelli:
Thanks for the corrections!

PR #430, opened 2018-12-19
Added Cloudflare options for target account

zBart:
This pull request adds two settings that can be set in the creds.json that allow you to specify the account where new domains should be added to. If the fields are left empty (the current situation) new domains end up in the account of the user the credentials belong to.

We bumped into this because the user that dnscontrol uses to push changes has access to multiple Cloudflare accounts and new domains we pushed ended up in the wrong account.


tlimoncelli:
Thanks!  Sounds like a highly useful feature!

PR #438, opened 2019-01-22
Return error for pending corrections in preview

### LINKED ISSUE #437: Option to return non-zero when preview detects changes
We are migrating to DNSControl and we would like to have a way to detect when there are differences between the current Zone and the one defined in DNSControl. This is, at least for now, to detect manual modifications done during interventions that would not be reflected in the configuration.

My idea would be to add a flag to `dnscontrol preview` so it returns a non-zero code when there are corrections detected.

Is this something desirable/acceptable? Is there already a way to have this workflow? What I'm currently doing is parsing the output of the tool to identify if there are changes to be done but a return-code approach would be much cleaner, in my opinion.

miguelbernadi:
Return a non-zero error code if there are corrections for preview
step. This is especially useful for automated alerting and scripting
around changes to the zone not reflected in the configuration.

Fixes #437

### DIFF
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -165,6 +165,9 @@ DomainLoop:
 	if anyErrors {
 		return errors.Errorf("Completed with errors")
 	}
+	if totalCorrections != 0 {
+		return errors.Errorf("Corrections pending")
+	}
 	return nil
 }
 


commands/previewPush.go
```diff
@@ -161,10 +161,13 @@ DomainLoop:
 		fmt.Fprintf(os.Stderr, "##teamcity[buildStatus status='SUCCESS' text='%d corrections']", totalCorrections)
 	}
 	notifier.Done()
-	out.Printf("Done. %d corrections.\n", totalCorrections)
```
captncraig: I feel like we need to still print the same thing in the 0 corrections case. There may be automations depending on that text. They can switch to status code now, but here we are.

I don't mind if the output looks a little redundant, or you could add a little more logic to make it sane. But I do fear something will be looking for that exact output.

miguelbernadi: I'll change it back.

miguelbernadi: Fixed.

miguelbernadi:
Any news about this?

ashleyhull-versent:
Hey, this broke my deployment in my situation where I want to preview to confirm 'syntax' errors and not config changes.

```
go get github.com/StackExchange/dnscontrol

dnscontrol preview

if [ $? -eq 0 ]; then
    echo "Succeeded. Pushing changes"
    dnscontrol push
else
    echo "preview failure. Fix error to continue"
    exit 1
fi
```

marcosnr:
It broke my pipeline too, seing #437 I agree is a good idea, but making it default without a flag to keep the old behaviour is not a good practice...

captncraig:
Reverted in master for now.

I do also see the definitely good use case for having a non-zero exit code to identify changes definitively without text-parsing.

I propose a new flag `--exit-code` for preview and push to activate tat behaviour on demand. Does that seem like an ok middle ground?

ashleyhull-versent:
Maybe an additional command? 'dnscontrol sync' that'll error if there is any change?

miguelbernadi:
I'm willing to implement the flag once it is decided how it should look like. I also think the `dnscontrol sync` proposed [here](https://github.com/StackExchange/dnscontrol/pull/438#issuecomment-461287200) is a nice approach.

tlimoncelli:
Ugh. I apologize for this.  I was the one that suggested we didn't need a flag for it.

tlimoncelli:
What makes this complex is there are many permutations:

1. Syntax check vs preview vs push
2. success or errors
3. changes or no changes (or # of changes)

Currently the status result simply means (2).

Encoding (2) and (3) in one status code isn't possible without breaking existing code (as we discovered this week.. my bad).

Puppet has `--detailed-exitcodes` which makes each bit of the status code mean something different:

```
* --detailed-exitcodes:
  Provide transaction information via exit codes. If this is enabled, an exit
  code of '2' means there were changes, an exit code of '4' means there were
  failures during the transaction, and an exit code of '6' means there were both
  changes and failures.
```

Thoughts?

tlimoncelli:
@miguelbernadi I've started a new bugid https://github.com/StackExchange/dnscontrol/issues/447

PR #441, opened 2019-01-28
Document require() and make it support relative paths

### LINKED ISSUE #440: require(...) is relative to the process working directory instead of the file it is called from
Let's say I have this file structure

```txt
.
├── dnsconfig.js
└── k8s
    ├── clusters
    │   └── my-cluster.js
    ├── clusters.js
    └── util.js
```

With the following file contents:

```javascript
// ./dnsconfig.js
require('./k8s/clusters.js')

// ...
```

```javascript
// ./k8s/clusters.js
require('./clusters/my-cluster.js')

// ...
```

```javascript
// ./k8s/clusters/my-cluster.js
require('../util.js')

// ...
```

`dnsconfig` will fail to push the zone:

```bash
# dnscontrol push --config ./dnsconfig.js
requiring: ./k8s/clusters.js
requiring: ./clusters/my-cluster.js
Executing javascript in ./dnsconfig.js: Error: Error: open ./clusters/my-cluster.js: no such file or directory
```

nlowe:
This brings `require(...)` more in-line with its equivalent in node.js by allowing imports relative to the file currently being parsed. I've made some assumptions in how the parsing works in that only one block is being parsed at a time. If there's a better way to do this with the javascript engine you're using I'm all ears.

I've also added some documentation for `require(...)`. Please let me know if it belongs in a different place.

Fixes #440

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ spfcache*
 adzonedump.*
 zones/
 stack.sh
+.idea/
--- a/commands/printIR.go
+++ b/commands/printIR.go
@@ -3,7 +3,6 @@ package commands
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 
 	"github.com/StackExchange/dnscontrol/models"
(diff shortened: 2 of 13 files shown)


tlimoncelli:
Wow, that was fast!  The code looks good!

I do have one suggestion.  Rather than using filepath.Abs(), use filepath.Clean().  This would have the benefit of not calling getcwd() which is slow and sometimes unreliable (it is probably more reliable now, but back in the day it didn't used to be).

That said, I may not understand your algorithm entirely.  However all the tests pass when I gave it a try. See https://github.com/StackExchange/dnscontrol/tree/nlowe-feat/GH-440-relative-require

nlowe:
Sure, that makes sense, I can adjust that tomorrow morning.

Regarding the implementation, what can I help clarify?

tlimoncelli:
I was digging deeper into the code to understand it better (thanks to your tests and docs, I was able to figure out my confusion).  What I couldn't figure out was why build a stack manually when require() is called recursively.  Local variables are on the execution stack, and that's all that should be needed.

I reworked your code to be "stackless" (well, to use the execution stack) instead.  It looks like this: https://github.com/StackExchange/dnscontrol/pull/443

Thoughts?

tlimoncelli:
Ok. I'll merge https://github.com/StackExchange/dnscontrol/pull/443 instead. Thanks for the documentation and the tests.  I enjoyed collaborating with you!

PR #448, opened 2019-02-12
Delegation Sets and retries

rileyberton:
- Support DelegationSet for Route53 (create-domains only)
- Retry Route53 operations which fail for rate limits under large numbers of domains
- Support for name_server_set for GCloud (create-domains only)
- Docs for both

### DIFF
--- a/docs/_providers/gcloud.md
+++ b/docs/_providers/gcloud.md
@@ -23,12 +23,13 @@ For Google cloud authentication, DNSControl requires a JSON 'Service Account Key
         "auth_uri": "https://accounts.google.com/o/oauth2/auth",
         "token_uri": "https://accounts.google.com/o/oauth2/token",
         "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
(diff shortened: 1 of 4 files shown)


tlimoncelli:
This looks good so far.  Is there a way to add a test for this code, possibly in `integrationTest/integration_test.go`?

rileyberton:
Will look into the test

rileyberton:
Test is a bit complicated as delegation sets only apply on creation.  Any test domain would only be usable once (AFA delegation set was concerned).

I could have the test create a delegation set, create a random domain in it and then check it was in the right set and delete the random domain if that is preferred..

thoughts?

tlimoncelli:
I don't know enough about the subject to have many thoughts :-(

I also realize that you mentioned that this feature has to be manually enabled.  This means the test would fail for most people.

So, I'm ok without adding a test now.  If the API becomes unstable in the future I might loop back to you and ask for tests to be added.

tlimoncelli:
Thanks for contributing to the project!

Issue #447, opened 2019-02-08
Report if changes were made using exit codes


tlimoncelli:
CC: miguelbernadi

There is a desire to be able to report back whether changes happened using an exit code.  To retain backwards compatibility, this needs to be activated using a flag.

My proposal is to model the flag after Puppet's `--detailed-exitcodes` flag:

```
* --detailed-exitcodes:
  Provide transaction information via exit codes. If this is enabled, an exit
  code of '2' means there were changes, an exit code of '4' means there were
  failures during the transaction, and an exit code of '6' means there were both
  changes and failures.
```

In other words, the 2^1 bit means "changes" and the 2^3 bit means "failure". Yeah binary!!

I think this flag should apply to `preview` and `push`.  In the case of `preview`, the flag indicates changes would be made.

miguelbernadi:
I like the `--expect-no-changes` approach as flag for `preview`.

tlimoncelli:
I agree with @captncraig 's analysis.  `--expect-no-changes` works for me too!

miguelbernadi:
Then I'll try to add the flag sometime this week and open a PR for it.

miguelbernadi:
PR is up for discussion.

PR #449, opened 2019-02-13
Add --expect-no-changes flag to preview

miguelbernadi:
We want to offer a return-code oriented interface to detect changes
pending to be applied. The preview command, by default, returns 0
regardless of pending changes.

We add a flag, `--expect-no-changes`, that will return non-zero if
there are changes pending application, and 0 if there are none.

Fixes: #447


tlimoncelli:
LGTM
That said, I'll let @captncraig do the merge. The last time I changed out exit codes worked I messed it up, so this time I'll get a second pair of eyes.

PR #463, opened 2019-04-17
Add glob pattern/wildcard support for IGNORE directive

koesie10:
This will add glob pattern and wildcard support for the IGNORE directive, allowing more use cases where DNS records are managed by external systems.

For example, you can now do `IGNORE("_acme-challenge.*)` to ignore all _acme-challenge records on the top level domain or `IGNORE("_acme-challenge.**)` to ignore all _acme-challenge records on the domain, as requested [here](https://github.com/StackExchange/dnscontrol/issues/183#issuecomment-391304329).

This could be a breaking change for  a file like the following:

```
D("example.com",
  IGNORE("*.foo"),
  A("baz.foo", "1.2.3.4")
);
```

Before, it would ignore the literal `*.foo` record, but now it would ignore all subdomains of `foo` and give an error on `baz.foo`, since it is not possible to manage an ignored record. It can be changed to be the same as the previous behaviour by changing the file to the following:

```
D("example.com",
  IGNORE("\\*.foo"),
  A("baz.foo", "1.2.3.4")
);
```

### DIFF
--- a/docs/_functions/domain/IGNORE.md
+++ b/docs/_functions/domain/IGNORE.md
@@ -24,4 +24,17 @@ D("example.com",
 {%endhighlight%}
 {% include endExample.html %}
 
+IGNORE also supports glob patterns in the style of the [gobwas/glob](https://github.com/gobwas/glob) library. All of 
+the following patterns will work:
+
+* `IGNORE("*.foo")` will ignore all records in the style of `bar.foo`, but will not ignore records using a double
+subdomain, such as `foo.bar.foo`.
+* `IGNORE("**.foo")` will ignore all subdomains of `foo`, including double subdomains.
+* `IGNORE("?oo")` will ignore all records of three symbols ending in `oo`, for example `foo` and `zoo`. It will
+not match `.`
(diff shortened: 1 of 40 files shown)


tlimoncelli:
Wow!  That's an impressive PR and a pretty cool feature!

I'll review this in the next few days.

tlimoncelli:
Hi!  I finally got around to looking at this. It all looks good. I do have one question:

Question: Why use an external library when Go has https://golang.org/pkg/path/filepath/#Match built in?

koesie10:
[github.com/gobwas/glob](https://github.com/gobwas/glob) is used instead of `filepath.Match` because `filepath.Match` only supports the file separator as the separator in the pattern. Therefore, the `*` operator would match the complete domain, while a user might want to only match the part between two separators (`.`). `github.com/gobwas/glob` also supports more patterns than `filepath.Match`, such as pattern alternatives `{...}` and excluding certain characters using `[!...]`.

tlimoncelli:
Ok, I hadn't realized the gobwas library had so many additional features.  Thanks for the PR!

Issue #75, opened 2017-04-12
Support the DS record type


haraldkoch:
Issues I see up front:

1) not all DNS providers support all record types.

2) The object model (e.g. RecordConfig) doesn't have fields for the extra data; it only understands Priority (for MX and SRV records). I'm not convinced you can treat the entire record as opaque data; some DNS providers want text strings instead of protocol numbers for the 'algorithm' and 'digest type' fields of a DS record, for example.

captncraig:
Yeah, we have an open issue #56 to support SRV records that is impacted by this too. Our general philosophy is that the RecordConfig should define a standard format for records, whatever that is, and it is up to the providers to adapt that to their own needs. We generally prefer to split fields into their own things, but if fields are not there, it likely means we don't support that record type **yet**. 

We will be adding top level fields for weight and port for SRV records, for example.

Your first point is a bit trickier. Generally we have been handling that by assuming all providers can handle all record types, and making it the responsibility of the provider implementation to print a warning or error if it must skip records it can't handle. 

This will all get more complicated as we move into more specialized custom things like ALIAS records, that have much less consistent adoption.

captncraig:
And DNSSec is a whole other beast that is going to take some effort to work in. API support for DNS providers and registrars is all pretty spotty at best.

haraldkoch:
@captncraig - I wouldn't want to tackle all of DNSSec in this tool, but DS records are a requirement for zone delegation. I have BIND automatic signing turned on for the rest of dnssec, but I still need to place DS records in my zone files, as well as configuring them into registars that support doing so.

Because of the RecordConfig fields, is it worth splitting this issue into separate issues for each record type?

captncraig:
Yes, we should make a seperate issue for each record type we wish to support. DS records are probably a domain level thing for the registrar to deal with. Probably would add a list of `DSRecord` structs to the domain, and have the registrar add them. And DNS providers that support dnssec can fill it like we do with nameservers. I think only one of the registrars we currently support have any api for DS records at all though :(

haraldkoch:
Ugh. DS records are complicated because they're a property of a zone, but they need to be added to the parent zone (just like NS records). Which means that to delegate from a zone that I control to another zone, I'd expect to write something like:

```
D("cfrq.net", registrar, DnsProvider(bind),
 ...
  DS("chk", 17314, 8, 1, "0B2D95FC3693F5C880CD2CDCF14BFBCF9FBD8274"),
[...]

Issue #31, opened 2017-02-01
DNSSEC Discussion


captncraig:
Just started thinking about this and wanted to write down notes for the future.

One thing I think would be interesting for DNSControl to manage would be DNSSEC records. This really has two parts:

1. DNS Provider needs to implement it. This involves managing keys, generating RRSIG records and so forth.
2. Registrar needs to register the DS record(s) with the tld nameservers.

I would love for dnscontrol to be able to activate DNSSEC for a domain with minimal effort required.

A few issues I foresee:

- Provider support is fairly limited. 

  Of the registrars we support, gandi has an api. name.com has functionality in the web ui, but no api. route53 as a registrar supports dnssec, but not as a dns provider. 

  As far as DNS providers go, gcloud has alpha support, gandi appears to support it, cloudflare has some sort of support, and all others seem to not. We could certainly do it for the bind provider with some work.

-  Multiple providers

  DNSSEC would definitely not work unless *all* providers for a zone support it. If they do, I believe they can all use a separate key, and we can add multiple DS records via the registrar.

 Most cloud providers provide only a "just turn it on" model where they maintain all the keys and just give you the DS info to give the registrar.

- Key management

Bind may or may not be worth implementing anything for. Maybe the keys live on the actual server, and part of the process of deploying zonefiles is using dnssec-signzone or something.

aeden:
We have API support over at DNSimple for signing zones as well as managing DS records when zones are signed elsewhere. It's true that support is spotty at registrars, especially with API support, but if you want to try to support it at some point, you can use us as your guinea pig.

You definitely want to avoid key management if you can. Key rotation needs to be handled carefully to ensure you do not lose DNS resolution when rotating out old keys.

Happy to answer questions about it if you have them, as our implementation just came out recently so it is front and center in my mind.

captncraig:
DS record support is gonna be added in #469 when it merges. After that I should have all the pieces I need to do a more automatic dnssec setup process with dnscontrol. It will be fairly analogous to how we handle nameservers currently.

If dnssec is "enabled" for a domain, we will:

1. Make a preliminary pass over all providers for that domain (which will need to implement some interface to provide DS info), and collect the DS record(s) we need to submit.
2. Make sure we have all of the discovered DS records added to the domain, in addition to any manually specified DS records already in the config.
3. Process dns provider records as usual.
4. Process registrars as usual.
5. Call special method on registrar to set up DS delegation with all of the previously discovered DS records.

[...]

PR #469, opened 2019-05-05
Added DS record type

kochie:
Added support for Delegation Signer record. Should partially address #31 and fully address #75. I've added support for Cloudflare at the moment and I'm hoping to add support for all compatible providers soon.

Current providers with DS enabled
- [x] Cloudflare
- [ ] Gandi
- [x] dnsimple

Hoping to get some feedback and verification that the data types are compatible with the [RFC](https://www.rfc-archive.org/getrfc.php?rfc=4034). 

Thanks!


captncraig:
Wow, thanks for working on this. I'm a bit confused about what situations you would be specifying the DS records directly though.

Looking at the cloudflare console, it looks like the default workflow is to take the DS record they have created and use for signing, and pass that to the upstream registrar. Most other providers I have looked at seem to have this kind of model where you "activate" it, then need to register the DS records upstream. You would never have access to the private keys that the provider uses is this case.

In what situation would you manually need to specify the DS records? Are you manually signing records yourself? Are you adding DS records from another provider for consistency? 

I feel like I'm missing something, since it is definitely something you can do in their web console, but I'm not clear why.

My imagined design for dnssec has been each compatible provider implements some `GetDSRecords` api call, and compatible registrars would have a corresponding `SetDSRecords` process. But last I checked (a while ago admittedly), none of the registrars I use had such an api. I imagine some may now.

captncraig:
Ok, after writing a little code I can see how it may be good to specify DS records directly, in case you need to manually enter them or a provider doesn't have a read api or something.

DS seem analagous to how we handle NS records perhaps. We try to automatically infer which nameservers to use from your providers and register them appropriately, but you can also directly specify nameserver records if you need something special. DS may need a little magic and a little flexibility. 

Please share how you envision using all this as well.

integrationTest/integration_test.go
```diff
@@ -601,6 +610,15 @@ func makeTests(t *testing.T) []*TestCase {
 		)
 	}
 
+	// DS
+	if !providers.ProviderHasCabability(*providerToRun, providers.CanUseRoute53Alias) {
```
captncraig: This looks like it was started and not finished. I'd love to have some integration tests to implement it for more providers.

pkg/js/parse_tests/024-ds.js
```diff
@@ -0,0 +1,4 @@
+D("foo.com","none",
+    DS("foo.com", 1000, 13, 2, "AABBCCDDEEFF"),
```
captncraig: We should use `@` instead of `foo.com`, even in tests.

pkg/normalize/validate.go
```diff
@@ -302,7 +304,7 @@ func NormalizeAndValidateConfig(config *models.DNSConfig) (errs []error) {
 			}
 
 			// Canonicalize Targets.
-			if rec.Type == "CNAME" || rec.Type == "MX" || rec.Type == "NAPTR" || rec.Type == "NS" || rec.Type == "SRV" {
+			if rec.Type == "CNAME" || rec.Type == "DS" || rec.Type == "MX" || rec.Type == "NAPTR" || rec.Type == "NS" || rec.Type == "SRV" {
```
captncraig: I don't think ds should use this case. There is no hostname in the target fields.

kochie:
Hey @captncraig, thanks for the feedback, I'll update the PR with the fixes. 

Just thought I'd give my perspective on adding the record. Your 100% right in that I rarely need to specify a DS record manually, however there is one use case which will probably start to become more commonplace, and that's transferring domains to new registrars. 

Basically I had a few domains with Namecheap that I transferred over to Cloudflare which automatically scanned and added a lot of junk records (>500) that Namecheap had included, since I also had DNSSEC enabled it copied over the DS record, which causes dnscontrol to panic when it read the record.

Basically my motivation for this PR was to allow dnscontrol to delete unneeded DS records that are imported when transferring domains using an external system.

💯 agree as well on your idea for how DNSSEC should be applied with dnscontrol, it would be great to simply assign a flag or something when defining a new domain in the config.

pkg/normalize/validate.go
kochie: Cool, can you think of any other validation checks for DS?

tlimoncelli:
Please rebase (`git pull --rebase --ff-only --prune`) and push. Thanks!

pkg/normalize/validate.go
tlimoncelli: I can't think of any.  We generally don't check the individual fields for fear of locking us into specific constants, etc.

patschi:
Any updates on this @kochie?

I would like to have DS support as well, specially for OVH in my case. So once this is merged I can start looking into getting this added for OVH too. :)

tlimoncelli:
Please rebase and run "go generate". Thanks!

tlimoncelli:
Please rebase and re-run "go generate".  I'll merge this as soon after. 

Thanks and sorry for the delay!

kochie:
It's still giving a gogen error, is this expected?

tlimoncelli:
Did you commit the file that gogen updated?

```
$ go version
go version go1.12.4 darwin/amd64
$ go generate
$ git status
On branch kochie-feature/ds-record-type
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   docs/_includes/matrix.html

no changes added to commit (use "git add" and/or "git commit -a")
$
```

It looks like you need to run `git commit docs/_includes/matrix.html` then `git push`

tlimoncelli:
Closing due to age.  Please re-open if you would like to continue.

McNetic:
I rebased and updated the original pull at https://github.com/StackExchange/dnscontrol/pull/746

I'm not a go developer and did not review the original work besides from what was necessary to get it to work. I was however able to create DS records with the desec provider.

tlimoncelli:
Thanks!  I'll take a look this week.

PR #474, opened 2019-05-17
Adds json loading support to require

ozkar99:
This commits adds support for require to load arbitrary "json" files.

This is useful for moving configuration/variables to their own file, and having them as `json` makes it much easier to manipulate using existing tooling.

Rationale:
-------------------------

In our use cases we have come regularly to the (anti) pattern of importing data files into variables like:

domains.js:
```json
ENABLED_DOMAINS = [
  "domain1.com",
  "domain2.com",
]
```

And then later on `dnsconfig.js` issuing a `require './domains.js'` and using the `ENABLED_DOMAINS` variable, for varied use cases (like looping through it to configure ips).

However, the `js` format becomes a problem if you want to manipulate the `domains` file, as this means, that you have to written custom parsers taking into account the variable declaration part.

If we add `json` loading, we can use any existing tooling to: read, write or modify these files much more easily.

Sample "check" Output:
-----------------------------

ip-map.json:
```json
{
  "192.168.1.1": "host1",
  "192.168.1.2": "host2",
  "192.168.1.3": "host3",
  "192.168.1.4": "host4",
  "192.168.1.5": "host5"
}
```

dnsconfig.js:
```javascript
var IP_MAP = require('./ip-map.json')

function reverseIPMap() {
	var ret = {};
	for (var ip in IP_MAP) {
		ret[IP_MAP[ip]] = ip;
	}
	return ret;
}

var HOSTNAME_MAP = reverseIPMap();

console.log(HOSTNAME_MAP['host3'])
```

dnscontrol check output:
```
192.168.1.3
No errors.
```


pkg/js/parse_tests/domain-ip-map.json
```diff
@@ -0,0 +1,4 @@
+{
+  "foo.com": "1.2.3.4",
+  "bar.com": "1.2.3.5"
```
tlimoncelli: Please make the two values very different (like... 1.1.1.1 and 5.5.5.5) so that visually they are very different.

tlimoncelli:
Very cool!  I can see how useful this will be.

Can you update the docs?

ozkar99:
For sure! i was about to start work on them.

ozkar99:
Added documentation, please let me know what you think, might work better to merge the example in the first example.

tlimoncelli:
This looks really good. However I wonder if we should have it be a separate function (`requirejson()` or maybe `importjson()`)  rather than doing something magical based on the filename.  I'm not sure about this but I thought I'd put it out there.

ozkar99:
Funny that you mention that, as that was my first intention having a `requireJson` function, but then decided to go more akin to the semantics of node.js's require: https://nodejs.org/api/modules.html#modules_require_id

It is also worth mentioning that node.js only supports 3 file extensions for `require`: **js**, **node** and **json** (https://nodejs.org/api/modules.html#modules_file_modules) out of wich, **js** and **node** behave the same, so even tho i neither really, really like the `if/else` based on the filepath, its not as it will continue to grow in the future, as there are only 2 possible branches (json/js+node):

  - **json**: get the value from the file.
 - **js/node**: evaluate the code on the file.

And the benefit is having the same ergonomics as that of node.js.

That being said, i don't feel particularly strong about not moving this functionality to a different function if you are not happy (you are the mantainer of the project after all), i am just explaining my rationale on trying to keep same ergonomics with node.js counterpart.

tlimoncelli:
I'm ambivalent.  If @captncraig doesn't object, we'll merge it.

tlimoncelli:
Craig is good with it.  Thanks for such an interesting feature!

ozkar99:
Nice, thank you for being and touching and keeping discussion up!

PR #476, opened 2019-05-17
Added support for specific TTL for SPF_BUILDER

### LINKED ISSUE #475: TTL for SPF_BUILDER?
Thanks for this very helpful tool! I'm currently working on getting dnscontrol implemented in my private IT environment, including CI.

While I was able to cover all my needs so far, I wasn't able to find any way how to specify a custom TTL for generated SPF records using `SPF_BUILDER`. Is there any way to do so?

patschi:
Closes #475


pkg/js/helpers.js
```diff
@@ -658,6 +659,10 @@ function SPF_BUILDER(value) {
     if (!value.raw) {
         value.raw = '_rawspf';
     }
+    if (!value.ttl) {
+        value.ttl = defaultArgs['defaultTTL'];
```
captncraig: I think we should just not set a ttl on the final txt record if ttl is not set. That feels more consistent.

patschi: If I leave `value.ttl` undefined (in case not specified in `dnsconfig.js`), it fails because of:
```
ERROR: Unknown modifier type
```

That's why I thought simply using the defaultTTL (which will be used anyway) will be the best way. The other solution might be wrapping a `IF-Else` around both `r.push(TXT(...))` lines.

captncraig: I think its probably best to have the if/else blocks tbh. we generally prefer letting the default ttls get bound as late as possible.

patschi: Understood. Changed!

patschi:
@captncraig Do you still require any changes? I've been using it since a bit before pushing this PR and for me it's working just fine so far.

PR #481, opened 2019-05-19
General docs improvements

patschi:
* Basic `IP()` explanation
* Mentioning that OVH also works with `CAA` records
* Mentioning the use of cloudflare proxy mode with `DEFAULTS()` (IMHO good to know)
* Fixed "URL" at `iodef` at `CAA` docs
* Small changed example of having default TTL
* Fixed `CAA_BUILDER` link on home site
* Fixed that PRs/issues where displayed under the wrong headlines on providers page


tlimoncelli:
Good catches!  Thanks!

PR #486, opened 2019-05-19
OVH: Fixed registrar ns correction

patschi:
This PR addresses a wrong behavior when the domain takes use of the additional OVH anycast service (Anycast can be ordered separately for a yearly fee). If used, `fetchNS()` will return back *all* nameservers referenced with a domain, whenever in use or not - so in case of anycast-enabled domains: 2x normal dns server, 2x anycast dns servers. This causes non-anycast nameservers being always added.

If anycast is enabled, IMHO it does not really make sense to add non-anycast nameservers. This PR uses `fetchRegistrarNS()` to get the **currently in-use** registrar nameserver. This way it works as intended.

(Regarding changes in file `protocol.go`: This is simply a line-switch-change to be consistent with the order as OVH returns back the data from their API.)


tlimoncelli:
Thanks!

PR #496, opened 2019-05-24
CF: Ability to manage UniversalSSL

### LINKED ISSUE #490: CF: Forcing TTL of 1 sec, when 300 was configured?
While working on a tiny change in the cloudflare provider I've came across following code:

https://github.com/StackExchange/dnscontrol/blob/d782bb002bccac04d66321480725f3eb46044857/providers/cloudflare/cloudflareProvider.go#L240-L242

Does anyone know the exact use-case for this? Why should be the TTL forced to `1`, when it was set to `300 seconds` (5mins)? Does this has any particular reason?

*Edit*: I'm thinking about removing `|| rec.TTL == 300` as it doesn't make any sense to me and might cause unexpected behavior.

### LINKED ISSUE #494: CF: Controlling "Universal SSL" option
I'd like to have a option to control the "Universal SSL" option of Cloudflare, just like with the proxy setting. The reason is following: I'd like to use Cloudflare as a DNS provider for a few domains, but CF shouldn't issue any certificates for these.

While I'm able to manually disable that specific setting, it is still much more convenient to handle that in an automated way with dnscontrol, and also integrates in a nice way when using `dnscontrol create-domains`.

Are there any voices against adding something like `CF_UNIVERAL_SSL('off')`?

patschi:
Closes #494 
Closes #490 

Example output:
```
******************** Domain: domain.tld
----- Getting nameservers from: cf
----- DNS Provider: cf...1 correction
#1: Universal SSL will be disabled for this domain.
SUCCESS!
```


patschi:
@tlimoncelli @captncraig I've solved the merge conflict of the `static.go` file. Do you would like me to do any further changes, or is this PR fine?

tlimoncelli:
Thanks for the new feature!

PR #515, opened 2019-07-01
Update github.com/mjibson/esc

TomOnTime:
* Update github.com/mjibson/esc
* Fix generate.go to new signature for esc.Run


TomOnTime:
Question for @mjibson: I noticed the calling signature for esc.Run changed.  Now all old code needs a bunch of template code to create the io.Writer.  Wouldn't it make sense to detect a nil writer and create the file automatically?

mjibson:
Yes, that would make sense. Most people (I believe) use the CLI and not the API, so moving this logic didn't seem to matter, but yes it should just do that.

PR #526, opened 2019-07-05
Update acme and dns vendored modules

tlimoncelli:
fetch github.com/go-acme/lego
fetch github.com/miekg/dns
fetch github.com/miekg/dns/dnsutil
fetch golang.org/x/crypto/ed25519

remove github.com/xenolf/lego/acme
remove github.com/xenolf/lego/log
remove github.com/xenolf/lego/acmev2


TomOnTime:
@captncraig I fixed some of the errors but it looks like the API changed a lot.  I think I've taken this as far as I can go.

PR #529, opened 2019-07-13
Switch to official Vultr API library

geek1011:
This PR switches to the new official Vultr API library and does a bit of cleanup on the provider code.

closes #528

PR #532, opened 2019-07-15
Add vultr to integration suite

captncraig:


captncraig:
Merging despite failure bacause of #529

PR #533, opened 2019-07-15
Add Namecheap CAA support

gpittarelli:
Namecheap supports CAA now -- this PR flags it as such.

A small update was also needed in GetDomainCorrections to handle the different record response data. This also fixed an unrelated bug where TXT records for Namecheap were being corrected on every run of  `dnscontrol push` even if nothing actually changed.


tlimoncelli:
I'm catching up with old PRs.  Is this still ready for merge?

PR #542, opened 2019-07-31
Added CAA support to DigitalOcean provider

turbotoad:
Have done a fix and release in my fork so I can use with my domains but maybe something you want to incorporate into a future release?

Note that Go is one of the few programming languages I haven't used before but it seems to work OK with the few changes I've made!

Paul.


tlimoncelli:
This was fixed by a different PR.

PR #543, opened 2019-08-21
Azure provider

mhenderson-so:
Azure provider for DNS Control. Passes all unit tests when I run it against my test domain - happy to assist in configuring a test domain in Azure that we can run every build against.

This operates in the same way as the Route 53 provider. It requests all records for a given label to be sent on each update (rather than a normal bind zone where you list each record on its own).


captncraig:
🎱

captncraig:
Mark, you have write access to this repo. I'm gonna push this to an "azure" branch and make a new pr from there so I can add this to integration pipeline.

PR #544, opened 2019-08-22
Azure Provider

captncraig:
Azure provider for DNS Control. Passes all unit tests when I run it against my test domain - happy to assist in configuring a test domain in Azure that we can run every build against.

This operates in the same way as the Route 53 provider. It requests all records for a given label to be sent on each update (rather than a normal bind zone where you list each record on its own).


captncraig:
🎱

captncraig:
🎱

captncraig:
Closing in favor of #547

Issue #455, opened 2019-03-05
Generalize CF_REDIRECT for other providers


DrMurx:
I propose to deprecate/remove the `CF_` prefix of `CF_REDIRECT` and `CF_TEMP_REDIRECT` to allow other providers implementing redirects. 

While CloudFlare is currently the only provider for those redirect directives, making this a general directive would allow to create something like a `nginx` or `apache` provider which would generate configuration snippets for the respective webservers.

tlimoncelli:
Some providers support wildcards, others don't.  Can you suggest a syntax that would work universally?

DrMurx:
> Some providers support wildcards, others don't. Can you suggest a syntax that would work universally?

Why would that be an issue? It's `REDIRECT(from, to)` and if you use wildcards in `from` and the provider of choice doesn't support wildcards, you'll get an error. However, I'd only support a generic match-all wildcard `*`, not complex regex patterns.

tlimoncelli:
I see. Yes, the lowest common denominator would be `REDIRECT(from, to)`.  We'd have to add to the validator code any providers that don't permit wildcards.

Or, we could forbid wildcards since we want REDIRECT() to be fully generic.

tlimoncelli:
Proposal:

- A generic HTTP redirect function called `REDIRECT()` that providers can opt into via Capabilities.
- `REDIRECT()` provides a restricted set of features that can be expected on all providers that opt-in.
- It should default to 302 (temporary), with a `REDIRECT(from, to, OPT_PERMANENT)` for 301 (Suggestions on a better name?) 
- Providers should implement a fully-featured redirect named `${PROVIDER}_${VERB}` where `$VERB` reflects the feature as $PROVIDER names it. They should then implement REDIRECT() as a edifice to `${PROVIDER}_${VERB}`
- Wildcards are not supported. (We could add this in the future)

Thoughts?

I feel like `REDIRECT()` plus `OPT_*` for new features we add in the future is a good way to go.  I also considered using different names instead of options (`REDIRECT_301()`, `REDIRECT_302()`, `REDIRECT_WILDCARD()`, etc.) but my hope is to create one redirect infrastructure that can be extended with `OPT_` functions.  We can use the `AuditRecords()` function to allow providers to reject/accept options.

tlimoncelli:
This is stale. I'm closing this. Please re-open if needed. Thanks!

PR #549, opened 2019-10-02
WIP: Add basic alidns provider (Alibaba Cloud DNS)

timonwong:
TODO:
- [x] REDIRECT_URL (302 redirect) & FORWARD_URL (iframe)
- [x] Basic documentation
- [ ] Tests
- [ ] "Geo" DNS
- [ ] Weighted round roubin (https://www.alibabacloud.com/help/doc-detail/60182.htm)

Limitation:
The "geodns" feature actually does not work, for example:

```js
// The following won't work because checkDuplicates does not respect provider specific metadata:
// A test-geo 1.1.1.1 -- ✅
// A test-geo 1.1.1.2 -- ❌ (duplicates)
// A test-geo 1.1.1.2 -- ❌(duplicates)
D("example.com", REG_NONE, DnsProvider(ALIDNS),
    A("test-geo", "1.1.1.1", {"alidns_line": "default"}),    // The default path
    A("test-geo", "1.1.1.2", {"alidns_line": "os_asia_sg"}),   // Singapore
    A("test-geo", "1.1.1.2", {"alidns_line": "os_asia_hk"})   // Hongkong
)
```

Any ideas?


tlimoncelli:
Looks good so far!  Thanks for working on this!

captncraig:
Yeah, that duplicate check doesn't provide any good way to allow what you are trying to do with geo. It definitely should. I think some things need to be rethought on that front. The underlying diff algotithm can handle it, but not the preprocessing phase.

timonwong:
> Yeah, that duplicate check doesn't provide any good way to allow what you are trying to do with geo. It definitely should. I think some things need to be rethought on that front. The underlying diff algotithm can handle it, but not the preprocessing phase.

Yes, after some digging (differences between ns1, constellix, etc), I think the opinionated way defined in [octodns](https://github.com/github/octodns/blob/master/docs/dynamic_records.md#an-annotated-example) is a good start.

Another question: Can we standardize the vendor-specific redirect records? (URL, URL301, FRAME, CF_TEMP_REDIRECT, CF_REDIRECT, and now with REDIRECT_URL and FORWARD_URL..) #455

tlimoncelli:
Any news?

tlimoncelli:
Closing due to age.  Please re-open if you would like to continue.

PR #555, opened 2019-10-23
Support Cloudflare API tokens

### LINKED ISSUE #534: Cloudflare API Tokens not supported
Cloudflare has a newish (beta) API Tokens. It allows custom permissions per token. It seems like a perfect fit for use by DNSControl. I have tried and it looks like it is not currently compatible with DNSControl. 

It is used as a bearer token. 

```
curl -X GET "https://api.cloudflare.com/client/v4/zones/cd7d0123e3012345da9420df9514dad0" \ -H "Content-Type:application/json" \ -H "Authorization: Bearer 1234567893feefc5f0q5000bfo0c38d90bbeb"
```

BenoitKnecht:
Cloudflare API tokens are a new way to authenticate to Cloudflare API.
Unlike the Global API key, tokens can be given specific permissions to
only access parts of the API. See [1] for details.

[1] https://blog.cloudflare.com/api-tokens-general-availability/

This commit introduces a new credential for cloudflare called
`apitoken`, which is mutually exclusive with `apiuser` and `apikey`.

In order for DNSControl to work with this token, it should have the
right to read DNS zones and edit DNS records.

Closes #534

### DIFF
--- a/docs/_providers/cloudflare.md
+++ b/docs/_providers/cloudflare.md
@@ -11,7 +11,21 @@ jsId: CLOUDFLAREAPI
 * When using `SPF()` or the `SPF_BUILDER()` the records are converted to RecordType `TXT` as Cloudflare API fails otherwise. See more [here](https://github.com/StackExchange/dnscontrol/issues/446).
 
 ## Configuration
-In the credentials file you must provide your Cloudflare API username and access token:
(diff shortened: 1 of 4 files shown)


tlimoncelli:
Awesome!  Thanks for the contribution!

PR #556, opened 2019-10-23
Fix small typo in Cloudflare API tokens documentation

BenoitKnecht:
Remove trailing "rather" at the end of a sentence, introduced in #555.


BenoitKnecht:
@tlimoncelli Sorry about that, I only noticed it while checking out the rendered documentation.

PR #558, opened 2019-10-24
Convert to Go modules

### LINKED ISSUE #500: Idea: Moving from vendor to modules
I'm sure people have seen the [introduction of modules to Go](https://blog.golang.org/using-go-modules). Just wondering if this is something that dnscontrol will move over to in the near future?

I think the biggest advantage to using modules is no longer needing to commit dependencies to `vendor/` which can increase the size of the project and requires constant updating. Are there any reasons why it wouldn't be a good idea to move to modules?

BenoitKnecht:
The idea is to get the ball rolling on #500, and see if the pipeline is happy with the change or not.

Closes #500


captncraig:
Even with modules, we would like to keep the dependencies vendored. What does it look like after `go mod vendor`?

BenoitKnecht:
@captncraig The dependencies _are_ vendored with Go modules; if you clone the repo and run `go get` (or `go build`), the dependencies will be pinned to what's listed in `go.mod`.

It's only when you run `go get -u` that the versions listed in `go.mod` get modified; you can then review those changes before committing `go.mod`, so you never have your modules change version unexpectedly.

tlimoncelli:
I'm concerned that we might use a module that will disappear from Github and then we'll be in trouble.  Luckily `go mod vendor` should copy all the files locally.  Could you do that?  After that, the PR should be very small.

Maybe a year from now I'll be more comfortable with modules and we can delete the files from the repo and rely on github not going away.   But... ya know... give me some time to adjust. :-)

geek1011:
@tlimoncelli proxy.golang.org is enabled by default, and will prevent that scenario from happening. There are also other projects which do the same thing.

captncraig:
We still desire to have build dependencies checked in so the repo is buildable from a single clone. I'd like to merge this, and can triple check it once `go mod vendor` is run on it.

PR #559, opened 2019-10-27
LIVEDNS-GANDI: Fix ability to remove empty TXT record

rdalverny:
Before this change:
an empty TXT record on the DNS server, not present in the local definition, triggers a slice out of range index error in `models.SetTargetTXTs`.

This ensures that `SetTargetTXTs` is not called with an empty slice from the Gandi LiveDNS provider.

This specific case occurred in my case because of an empty _acme-challenge TXT record left by a separate LetsEncrypt call (so not present in my local definition file). Removing this record is appropriate; dnscontrol not crashing on this was not.

It seemed more appropriate to me to fix this from the calling code (Gandi's).
Maybe should it be rather in `SetTargetTXTs` to ensure in a more general case, that the slice is not empty before using it? Like this:

```diff
diff --git models/t_txt.go models/t_txt.go
@@ -15,7 +15,11 @@ func (rc *RecordConfig) SetTargetTXT(s string) error {
 
 // SetTargetTXTs sets the TXT fields when there are many strings.
 func (rc *RecordConfig) SetTargetTXTs(s []string) error {
-       rc.SetTarget(s[0])
+       if len(s) > 0 {
+               rc.SetTarget(s[0])
+       } else {
+               rc.SetTarget("")
+       }
```

### DIFF
--- a/providers/gandi/livedns.go
+++ b/providers/gandi/livedns.go
@@ -199,6 +199,9 @@ func (c *liveClient) recordConfigFromInfo(infos []*gandiliverecord.Info, origin 
 			for _, txt := range info.Values {
 				parsed = append(parsed, models.StripQuotes(txt))
 			}
+			if len(parsed) == 0 {
+				parsed = append(parsed, "")
+			}
 			err := rc.SetTargetTXTs(parsed)
 			if err != nil {
 				panic(errors.Wrapf(err, "recordConfigFromInfo=TXT failed"))


tlimoncelli:
I think this is something that should be fixed at the global level.

First, there should be an integration test for a null string:

```
    tc("Empty"),
    tc("Null string TXT", txt("", "empty")),
    tc("Empty"),
```

That shouldn't break any of the existing providers, though it is something we currently don't test.  I ran it on all the providers I could, and it didn't break anything.

Next, there should be a test to make sure that any attempt to have an empty list of strings works too:

```
      tc("Create TXTMulti empty",
        txtmulti("foo0", []string{}),
      ),
```

That  crashed  ROUTE53 (I stopped testing after that).  Needs investigation.

Here's my work so far (please steal from it, then I'll close that branch):
https://github.com/StackExchange/dnscontrol/tree/rdalverny-fix-gandi-empty-txt

On the other hand, I can be convinced that this is being fixed at the wrong layer because the TXT() function in javascript requires at least one target string.  Maybe the Lets Encrypt stuff is where we should fix it.

@captncraig: any thoughts?

rdalverny:
Thanks a lot for the feedback! I agree with the need for tests for this fix; I will start from this and propose a new change more global (hmmm, I am not sure you did push your changes to my branch, I didn't see any on it?).

As for the TXT() function requirement for at least one string; yes.
Only here the Gandi API did return one record with an empty slice (which caused the crash).
As to why, I need to reproduce it first as I noticed that the API kept returning me this bizarre record, while a direct `dig -t TXT acme-challenge.domain.tld` did not return anything.
I only "fixed it" in the Gandi API response by using my patch to force an update. 😛

The mere idea of this change was more "don't crash on this unexpected case, consider the record empty then" but I totally understand it yet needs a clearer case. I'll work on that.

tlimoncelli:
FYI: I forgot to push my latest changes to https://github.com/StackExchange/dnscontrol/tree/rdalverny-fix-gandi-empty-txt

Take another look

tlimoncelli:
I rewrote GANDI from scratch, and called it GANDI_V5.  It is in branch https://github.com/StackExchange/dnscontrol/tree/gandi2

Can you check to see if the same problem exists there?

tlimoncelli:
The new GANDI_V5 provider doesn't have this bug. I'm going to consider this fixed, since that provider is stable.

I'm adding tests for this here: https://github.com/StackExchange/dnscontrol/pull/597

PR #562, opened 2019-11-02
Fix decoding empty Cloudflare SRV target (plus fixes for a few other providers)

### LINKED ISSUE #561: Cloudflare not working with empty SRV record targets
This seems to have been working last week, but it now errors with `Error getting corrections: Error fetching record list from cloudflare: json: cannot unmarshal bool into Go struct field cfRecData.result.data.target of type string`.

For reference, here is the JSON record (domain names have been changed): `{"id":"5ed257182fdf61652819accebb407dad","type":"SRV","name":"_pop3._tcp.example.com","content":"0\t0","proxiable":false,"proxied":false,"ttl":120,"priority":0,"locked":false,"zone_id":"ece505641eb3d1700d1e5b5bcfae5252","zone_name":"example.com","modified_on":"2019-10-26T16:36:50.923811Z","created_on":"2019-10-26T16:36:50.923811Z","meta":{"auto_added":false,"managed_by_apps":false,"managed_by_argo_tunnel":false},"data":{"service":"_pop3","proto":"_tcp","name":"example.com","weight":0,"port":0,"target":false,"priority":0}}`. Note how target is `false`.

And, to quote RFC2782:

```
   Target
        The domain name of the target host.  There MUST be one or more
        address records for this name, the name MUST NOT be an alias (in
        the sense of RFC 1034 or RFC 2181).  Implementors are urged, but
        not required, to return the address record(s) in the Additional
        Data section.  Unless and until permitted by future standards
        action, name compression is not to be used for this field.

        A Target of "." means that the service is decidedly not
        available at this domain.
```

I'll submit a PR to fix this.

geek1011:
SRV records with empty (".") targets are now returned as false by the API, which breaks Unmarshaling it into a string.

fixes #561

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -458,6 +458,7 @@ func makeTests(t *testing.T) []*TestCase {
 			tc("Change Priority", srv("_sip._tcp", 52, 6, 7, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
 			tc("Change Weight", srv("_sip._tcp", 52, 62, 7, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
 			tc("Change Port", srv("_sip._tcp", 52, 62, 72, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
+			tc("Null Target", srv("_sip._tcp", 52, 62, 72, "foo.com."), srv("_sip._tcp", 15, 65, 75, ".")),
 		)
 	}
 
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -458,8 +458,12 @@ func makeTests(t *testing.T) []*TestCase {
 			tc("Change Priority", srv("_sip._tcp", 52, 6, 7, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
 			tc("Change Weight", srv("_sip._tcp", 52, 62, 7, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
 			tc("Change Port", srv("_sip._tcp", 52, 62, 72, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
-			tc("Null Target", srv("_sip._tcp", 52, 62, 72, "foo.com."), srv("_sip._tcp", 15, 65, 75, ".")),
 		)
+		if *providerToRun == "NAMEDOTCOM" {
+			t.Log("Skipping SRV Null Target test because provider does not support them")
+		} else {
+			tests = append(tests, tc("Null Target", srv("_sip._tcp", 52, 62, 72, "foo.com."), srv("_sip._tcp", 15, 65, 75, ".")))
+		}
 	}
 
 	// SSHFP
--- a/providers/cloudflare/cloudflareProvider.go
+++ b/providers/cloudflare/cloudflareProvider.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"net"
-	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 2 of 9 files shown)


integrationTest/integration_test.go
```diff
@@ -458,6 +458,7 @@ func makeTests(t *testing.T) []*TestCase {
 			tc("Change Priority", srv("_sip._tcp", 52, 6, 7, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
 			tc("Change Weight", srv("_sip._tcp", 52, 62, 7, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
 			tc("Change Port", srv("_sip._tcp", 52, 62, 72, "foo.com."), srv("_sip._tcp", 15, 65, 75, "foo4.com.")),
+			tc("Null Target", srv("_sip._tcp", 52, 62, 72, "foo.com."), srv("_sip._tcp", 15, 65, 75, ".")),
```
geek1011: This may break other providers if they don't handle empty SRV records properly.

providers/cloudflare/cloudflareProvider.go
```diff
@@ -422,23 +423,46 @@ func newCloudflare(m map[string]string, metadata json.RawMessage) (providers.DNS
...
+	Fingerprint   string      `json:"fingerprint"`   // SSHFP
+}
+
+func (c cfRecData) Target() string {
+	switch v := c.CFTarget.(type) {
+	case string:
+		return v
+	case bool:
+		if !v {
+			return "" // the "." is already added by nativeToRecord
```
geek1011: This is a slightly ugly hack, and will need to be kept in mind if ever using `Target()` in other places. To actually do this properly, nativeToRecord shouldn't add a `.` if it's already a FQDN. I'll fix this if it doesn't cause other side effects (I'm not too familiar with the cloudflare provider code yet).

geek1011: I've decided to leave this as-is, as the alternative would be to add the `.` directly in this function, which will either greatly increase the complexity of `Target()` and `GetTarget()` (checking for the trailing `.`, etc) or would cause them not to be the inverse of each other in all cases (which could introduce obscure bugs).

geek1011:
I've tested this, and it works for me. I haven't ran the integration tests, though.

integrationTest/integration_test.go
tlimoncelli: Did you see this in the wild?  It seems like a bad thing to do, even if it is legal.  Like, maybe we should add a new opinion that bans this kind of thing (not because it is illegal, but because it is ill-advised).

geek1011: Yes, I've seen it in the wild (both its usage and providers not supporting it, I don't know which you're referring to). For example, [Fastmail](https://www.fastmail.com/help/receive/domains-advanced.html) and [Gmail](https://dns.google.com/resolve?name=_pop3._tcp.gmail.com&type=SRV) both use this for pop3. I've also seen it for other services like unencrypted SIP ports on VoIP providers, but I don't recall exactly which ones at the moment. For providers not supporting it, I have heard Plesk didn't support it.

tlimoncelli: I shouldn't read Gihub from my phone. I totally missed that "." is a legal target (you had written about that above). My apologies.

geek1011: No worries.

captncraig:
🎱

geek1011:
So, it seems the following needs to be fixed:
- Type for sending modification to Cloudflare (decoding works, encoding is slightly incorrect)
- DigitalOcean has the double-dot issue (similar to what I was pointing out previously)
- NameDotCom doesn't support null targets
- DNSimple returns a null record in an nonstandard format (the target is omitted entirely)

I'll fix CF and DO later today.

geek1011:
Ok, I've pushed a rewrite of this PR which should make CF work properly in all cases. In addition, it's more resilient to future API changes and it makes the AddOrigin usage less hacky.

geek1011:
I've fixed DO, and I think I've fixed DNSimple (I can't test it as I don't have an account).

@captncraig, can you run the integration tests again?

@tlimoncelli, what do you want to do about NameDotCom? Would you prefer to skip the test for it, or to add another capability for null SRV targets?

captncraig:
If the provider fails on a technically correct case like this, generally I will handle it in the provider code and return an error. In this case I'd probably have namedotcom return a hard error, as opposed to printing a warning, since it isn't possible to create the requested config. Make sure the message is somewhat descriptive of what the problem is.

captncraig:
🎱

geek1011:
DNSImple now seems to be failing with `panic: unparsable record received from dnsimple: SRV value too big for uint16: strconv.ParseUint: parsing "75.": invalid syntax [recovered]`. I don't know why the 75 has a . after it. Can someone with access to the DNSimple API post the response for a SRV record with an empty target so I can have a look?

geek1011:
I've now fixed everything now except for the last DNSimple error, which I'll need to see the API output for.

tlimoncelli:
@aeden can you assist with testing DNSimple?

onlyhavecans:
I am looking into the DNSimple breakage and will try to have a patch tomorrow

geek1011:
@onlyhavecans, any update on this?

onlyhavecans:
Hi! This is not an API but a logic bug.  I opened a PR against your repo @geek1011

geek1011:
@captncraig, can you run integration tests again?

geek1011:
I'm finished with this PR. All tests pass now (GCloud failed in the last run because someone seems to have removed the domain from the test account, but it passed last time, and nothing has changed since then).

tlimoncelli:
Awesome!  Thanks for this contribution!

PR #572, opened 2019-12-21
NEW PROVIDER: GANDI_V5 (replaces GANDI)

tlimoncelli:
NOT READY FOR MERGE.  Ready for review.

Before we merge:

[] Upstream changes to github.com/tiramiseb/go-gandi-livedns
[] Verify Registrar functions work
[] Get buy-in on names for future decoupled zoneget methods


models/dns.go
```diff
@@ -59,6 +59,19 @@ func StringsToNameservers(nss []string) []*Nameserver {
...
 
+// NameserversToStrings constructs a list of lists from *Nameserver structs
+func NameserversToStrings(nss []*Nameserver) (s []string) {
+	for _, ns := range nss {
+		s = append(s, ns.Name)
+	}
+	return s
+}
+
+// FIXME(tal): In hindsight, the Nameserver struct is overkill. We
```
captncraig: Lets remove this comment since there is an issue to track this now.

models/provider.go
```diff
@@ -6,6 +6,15 @@ type DNSProvider interface {
 	GetDomainCorrections(dc *DomainConfig) ([]*Correction, error)
 }
 
+// DNSProvider will replace DNSProvider in 3.0.
```
captncraig: Remove comment again.

models/record.go
```diff
@@ -328,6 +338,20 @@ func (r Records) GroupedByLabel() ([]string, map[string]Records) {
 	return order, groups
 }
 
+// GroupedByFQDN returns a map of keys to records, grouped by FQDN.
+func (r Records) GroupedByFQDN() ([]string, map[string]Records) {
```
captncraig: This will group records of different types together. I'm wondering what use case there is for this? I haven't needed to do that before, so wondering if this is too specific to gandi to belong in the models package? Usually want to group by the name/type tuple.

providers/diff/diff.go
```diff
@@ -221,6 +221,32 @@ func (d *differ) ChangedGroups(existing []*models.RecordConfig) map[models.Recor
 	return changedKeys
 }
 
+// DebugKeyMapMap debug prints the results from ChangedGroups.
+func DebugKeyMapMap(note string, m map[models.RecordKey][]string) {
```
captncraig: This looks like debug code that should maybe be removed. The only usage is commented out.

providers/gandi_v5/gandi_v5Provider.go
```diff
@@ -0,0 +1,338 @@
...
+				}
+			}
+		}
+	}
+
+	return corrections, nil
+}
+
+// debugRecords prints a list of RecordConfig.
+func debugRecords(note string, recs []*models.RecordConfig) {
```
captncraig: we should remove commented out debug code.

providers/gandi_v5/gandi_v5Provider.go
```diff
@@ -0,0 +1,338 @@
...
+	fmt.Println("DEBUG:", note)
```
captncraig: and uncommented out debug code :)

models/dns.go
tlimoncelli: I couldn't find the issue, so I created one. https://github.com/StackExchange/dnscontrol/issues/577

I've updated the comment to link to the issue. (as is the Google Styleguide standard... if you list a FIXME(), 90% of the time there should be an issue filed)

models/provider.go
tlimoncelli: I've updated it to explain what's going on to future provider authors. PTAL

models/record.go
tlimoncelli: I think that if one provider needs it, multiple will probably need it.  Maybe it should be in the diff package?

providers/diff/diff.go
tlimoncelli: This was originally I found this useful enough that I moved it to the diff package.  I've made it activate in gandi_v5 if the debug flag is true.

providers/gandi_v5/gandi_v5Provider.go
tlimoncelli: This is now used.

PR #592, opened 2020-01-23
Digitalocean CAA-support

### LINKED ISSUE #588: Digitalocean CAA Records
They have support for CAA records: https://www.digitalocean.com/docs/networking/dns/how-to/create-caa-records/

> ERROR: Domain [redacted].com uses CAA records, but DNS provider type DIGITALOCEAN does not support them

Deraen:
Change the tests to skip test case with CAA `;` value for Digitalocean,
because that specific feature isn't supported.

Closes #588


tlimoncelli:
Thanks for fixing this!

PR #593, opened 2020-01-23
Mark dnscontrol as v2 in go.mod

tlimoncelli:


geek1011:
Sorry, those references were a mistake. I meant to reference #595, not #593.

PR #595, opened 2020-01-23
Switch to v2 go.mod, drop GOPATH, and fix Azure Pipelines

geek1011:
Also set GO111MODULE=on in build stuff to always use Go modules
even when in GOPATH.

closes #593

PR #599, opened 2020-01-24
Regenerate provider support matrix

geek1011:
Also see #595.


tlimoncelli:
Thanks for noticing that!

PR #603, opened 2020-01-27
DOC: Improve internetbs doc

tlimoncelli:
Rather than listing a GCLOUD, @captncraig reminded me that we can just list 0 dns providers and it would work.

CC: @pragmaton @captncraig


pragmaton:
I was not aware about this feature.  It's pleasure to see how good software design is paying off -  empty set is still set. :) Agree, this example is better.

PR #605, opened 2020-01-28
Fix issues from go vet

geek1011:
* CloudFlare provider is missing quotes on a JSON struct tag.
* ACME package has a unnecessary JSON struct tag.
* Route53 provider has an unreachable return.


geek1011:
@tlimoncelli

Issue #468, opened 2019-05-02
get-certs should not exit after first failed cert.


captncraig:
It should keep going and try to validate the rest of the certificates.

PR #611, opened 2020-01-29
Try all cert renewals before returning error.

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/468


tlimoncelli:
Would you consider reviewing this please?  @captncraig

Issue #610, opened 2020-01-29
Replace "convertzone" with a subcommand "get-zone"


tlimoncelli:
Convertzone should be replaced by a subcommand in dnscontrol which works for all providers.

(this will be a lot easier after the code refactoring scheduled for 3.0)

Suggested (draft) command format:

The goal is that this would not depend on `dnsconfig.js`.  This would allow it to be used when creating a `dnsconfig.js` file, or just doing various diagnostics.

```
dnscontrol get-zone PROVIDER CREDNAME ZONENAME FORMAT

* PROVIDER: BIND, NAMEDOTCOM
* CREDNAME: If credentials are needed, refer to this name in creds.json
* ZONENAME: the DNS domain or zone to request
* FORMAT: bind (BIND-style https://en.wikipedia.org/wiki/Zone_file), dsl (the DNScontrol dnsconfig.js language), tsv (tab separated), yaml (octodns-compatible format)
```

Examples:

```
dnscontrol get-zone BIND - example.com tsv

dnscontrol get-zone GANDI_V5 mygandi example.com bind
```

Issue #609, opened 2020-01-29
Add a "checkcreds" subcommand


tlimoncelli:
There should be a subcommand that does some kind of "no-op" function with the provider.  It would be used to help debug credential problems.  The output should be as verbose as possible.

Issue #576, opened 2020-01-02
Major Refactoring of DnsProvider Interface


captncraig:
Currently convertzone only works with certain providers. We would like it to work with all providers, and become part of dnscontrol as a “getzone” subcommand. This has been requested by many people.

The current provider interface is monolithic, and only provides a “GetCorrections” function.
Our plan is to separate this into “Get Existing Records” and “Determine needed Corrections” functions that can be used independently for different functionality. 

The main thing this will enable is a single, unified “convert zone” command that will work for any provider, and ease migration to dnscontrol greatly for more complicated zones.

Technically what will happen is providers change from providing 2 big functions and will instead provide 5 more granular functions.

```
// DNSProvider is an interface for DNS Provider plug-ins.
type DNSProvider interface {
  GetNameservers(domain string) ([]*Nameserver, error)
  GetDomainCorrections(dc *DomainConfig) ([]*Correction, error)
}

// DNSProvider will replace DNSProvider in 3.0.
// NOTE: This is a draft and will change during implementation.
type DNSProvider interface {
 GetNameservers(domain string) ([]*Nameserver, error)
 GetZoneRecords(domain string) (Records, error)
 PrepFoundRecords(recs Records) Records
 PrepDesiredRecords(dc *DomainConfig)
 GenerateDomainCorrections(dc *DomainConfig, existing Records) ([]*Correction, error)
}
```

tlimoncelli:
I was able to achieve the same goal without a major refactoring.  I simply added `GetZoneRecords(domain string) (Records, error)` to `DNSProvider`, then added a function to each provider that reported "not implemented".  I then made a few different providers work. The others can do the conversion at their own pace.

Issue #538, opened 2019-07-25
convertzone produces invalid SRV records


ad8-bdl:
A minor issue:
I have a zone file with a couple of SRV records; `convertzone` emits these as:
```
	SRV('_sip._tls', '100 1 443 sipdir.example.net.', TTL(3600)),
```

which is invalid; the record arguments should be separated as:

```
	SRV('_sip._tls', 100, 1, 443, 'sipdir.example.net.', TTL(3600)),
```

The simple workaround is to manually edit SRV records after running `convertzone`.

PR #613, opened 2020-01-30
Add "get-zone" command

tlimoncelli:
* Related to https://github.com/StackExchange/dnscontrol/issues/539
* Fixes https://github.com/StackExchange/dnscontrol/issues/538
* Fixes https://github.com/StackExchange/dnscontrol/issues/576
* Fixes https://github.com/StackExchange/dnscontrol/issues/609
* Fixes https://github.com/StackExchange/dnscontrol/issues/610

CC: @ad8-bdl


tlimoncelli:
This is ready for testing.  Can people give it a try?

captncraig:
I tried it! 

1. The help text `Arguments should be: zonename credskey providername (Ex: example.com r53 ROUTE53)` seems to have the parameter order reversed. Seems to be `credskey providername zonename`
2. That's as far as I got. Cloudflare is not implemented yet. I may implement and make a pr into this shortly.

A few general thoughts though from looking it over:

1. It would be really cool if we could enable converting **all** zones in your account in one go. For people who have dozens or more and don't want to run the tool multiple times and maybe miss one. That may require an additional "ListZones" or something on the interface though.
2. This is a really big change, so I get why doing it piecemeal is the easy path. I would like to see a end-product where all providers must implement the GetZoneRecord function, and do it consistently. I feel like the integration tests are a good vector to helping that, if we verify the result of GetZoneRecords after each changeset. I was also a little surprised Push/Preview don't use the new code, but maybe both of these are future steps. Maybe there is a mergeable middle ground before then.

captncraig:
Awesome, I got it working with #625.

A few new notes:

1. Cloudflare uses metadata for the "orange cloud" proxy configuration. Right now the existing function to get records does not recreate that, but instead the driver queries the "original" record field to learn what it needs to know. It gets worse, because the "read-only" side of things doesn't really know whether to use "on" or "full" (I don't even remember which does what anymore). But maybe using "on" is an ok default for get-zone for records that are already enabled? `GetZoneRecords` could maybe add "on" to records it finds need it? Assuming `GetZoneRecords` creates the proper metadata, the dsl printer should probably output it. 

2. TTLs are always horrible. Cloudflare uses 1 as a default and my dsl output was littered with `TTL(1)`. I first changed the cloudflare code to replace 1 with 300, but then noticed the default is configurable. So I kinda went with the reverse of another convention we have used: "If `GetZoneRecords` gives you a 0 ttl, use the default value.". Is that sensible? Its in the PR above.

tlimoncelli:
> I tried it!
> 
> 1. The help text `Arguments should be: zonename credskey providername (Ex: example.com r53 ROUTE53)` seems to have the parameter order reversed. Seems to be `credskey providername zonename`

Correct.  Will fix.

> 2. That's as far as I got. Cloudflare is not implemented yet. I may implement and make a pr into this shortly.

+1

> A few general thoughts though from looking it over:
> 
> 1. It would be really cool if we could enable converting **all** zones in your account in one go. For people who have dozens or more and don't want to run the tool multiple times and maybe miss one. That may require an additional "ListZones" or something on the interface though.

I saw your PR.  Looks good. I agree that a ZoneLister interface is the best way to go.

> 2. This is a really big change, so I get why doing it piecemeal is the easy path. I would like to see a end-product where all providers must implement the GetZoneRecord function, and do it consistently. I feel like the integration tests are a good vector to helping that, if we verify the result of GetZoneRecords after each changeset. I was also a little surprised Push/Preview don't use the new code, but maybe both of these are future steps. Maybe there is a mergeable middle ground before then.

I figured that it was easier to just add a function than require every provider to rewrite their GetDomainCorrections code.  I think that once people start complaining that "get-zone" doesn't work for a particular provider, the providers will step up.

tlimoncelli:
> Awesome, I got it working with #625.
> 
> A few new notes:
> 
> 1. Cloudflare uses metadata for the "orange cloud" proxy configuration. Right now the existing function to get records does not recreate that, but instead the driver queries the "original" record field to learn what it needs to know. It gets worse, because the "read-only" side of things doesn't really know whether to use "on" or "full" (I don't even remember which does what anymore). But maybe using "on" is an ok default for get-zone for records that are already enabled? `GetZoneRecords` could maybe add "on" to records it finds need it? Assuming `GetZoneRecords` creates the proper metadata, the dsl printer should probably output it.

Interesting!  I hadn't thought about that.  I'll digest that more later.

> 2. TTLs are always horrible. Cloudflare uses 1 as a default and my dsl output was littered with `TTL(1)`. I first changed the cloudflare code to replace 1 with 300, but then noticed the default is configurable. So I kinda went with the reverse of another convention we have used: "If `GetZoneRecords` gives you a 0 ttl, use the default value.". Is that sensible? Its in the PR above.

Looks good.  I'll play with it a little and see what works best.

PR #615, opened 2020-02-05
Fix AzureDNS query updating the wrong Domain

vatsalyagoel:
When having multiple zones in the same resource group, the zone references were being switched up and ended up updating the wrong record set


tlimoncelli:
Thank you  for finding and fixing this bug!

PR #617, opened 2020-02-07
Add _mta-sts to labels allowed to have an underscore

tomhughes:
Although `_mta-sts` is normally used with a `TXT` record it can also be used with a `CNAME` record to delegate the authority.

See sections 3.1 and 8.2 of [RFC8461](https://www.rfc-editor.org/rfc/rfc8461.txt) for details.


tlimoncelli:
Thanks!

PR #626, opened 2020-02-16
Get all zones for a provider at once

captncraig:
A quick PoC of fetching multiple zones at once. Basically:

1. The core `get-zone` logic now expects to be handling a list of domains at once. 
2. If a provider implements `providers.ZoneLister` by creating `func ListZones() ([]string, error)`, you can run get-zone with zoneName `all` to generate a single output with all domains filled in. Other providers still work without adding an implementation, just can't handle `all`.

Diff is pretty long since this piggybacks on #625, that one should be considered first.


tlimoncelli:
Great idea!

Maybe we should change the command name to "get-zones" (plural) and permit multiple zone names on the command line.

If there is more than one zone, the pretty option should include "$ORIGIN example.com." at the top of each zone.

Issue #628, opened 2020-02-18
More providers should implement "get-zones" 


tlimoncelli:
Commit 87ad01d194ddc05950e79382c838d348e11d3db2 added a new subcommand called "get-zones" which replaces the (mostly broken) convertzone command.  It is a stand-alone command that allows users to download zone records from a provider. It is useful for bootstrapping your dnsconfig.js configuration.

This subcommand works for any provider that implements a new function called `GetZoneRecords`.  I've added such a function to all providers that simply outputs "not implemented".  I'm asking that all providers replace this function with one that actually does the required task.  Initially I've implemented it for BIND, GANDI_V5, and @captncraig wrote one for CLOUDFLAREAPI.

The functional is fairly easy to implement since the code exists in `GetZoneCorrections()`.  One simply (?) extracts out that code into a separate function.

- [x] **ACTIVEDIRECTORY_PS @tlimoncelli
- [x] **AZURE_DNS @vatsalyagoel
- [X] **BIND @tlimoncelli
- [X] **CLOUDFLAREAPI @captncraig
- [x] CLOUDNS @pragmaton
- [x] DIGITALOCEAN @Deraen
- [x] DNSIMPLE @aeden
- [ ] EXOSCALE @pierre-emmanuelJ 
- [X] GANDI_V5  @tlimoncelli
- [x] **GCLOUD  @tlimoncelli
- [ ] HEXONET @papakai
- [ ] INTERNETBS @pragmaton
- [ ] LINODE @koesie10
- [ ] NAMECHEAP @captncraig
- [x] **NAMEDOTCOM  @tlimoncelli
- [ ] NS1 @captncraig
- [ ] OCTODNS @tlimoncelli
- [ ] OPENSRS @philhug 
- [x] OVH @masterzen
- [x] **ROUTE53  @tlimoncelli
- [ ] SOFTLAYER @jamielennox 
- [x] VULTR @geek1011

We'll consider this bug complete when both of these statements are true:

* At least 50% of all providers implement this feature
* 100% of the providers labeled `**` implement this feature

This feature blocks the 3.0 release.

aeden:
@tlimoncelli for DNSimple I'm going to ask @onlyhavecans to implement it.

papakai:
@tlimoncelli for HEXONET I'll dive into the review in mid of march. As I am very busy, I can't handle it earlier.

tlimoncelli:
[...]

PR #631, opened 2020-02-18
Implement Get Zone for Azure DNS

vatsalyagoel:
Refactored Azure DNS to implement GetZone and ListZones 

Actions #628

### DIFF
--- a/providers/azuredns/azureDnsProvider.go
+++ b/providers/azuredns/azureDnsProvider.go
@@ -62,23 +62,33 @@ var features = providers.DocumentationNotes{
 	providers.CanUseNAPTR:            providers.Cannot(),
 	providers.CanUseSSHFP:            providers.Cannot(),
 	providers.CanUseTLSA:             providers.Cannot(),
-	providers.CanGetZones:            providers.Unimplemented(),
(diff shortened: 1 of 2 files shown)


providers/azuredns/azureDnsProvider.go
```diff
@@ -111,39 +121,66 @@ func (a *azureDnsProvider) GetNameservers(domain string) ([]*models.Nameserver,
...
+	existingRecords, _, _, err := a.GetExistingRecords(domain)
 	if err != nil {
 		return nil, err
 	}
+	return existingRecords, nil
+}
 
-	var corrections []*models.Correction
-	zone, ok := a.zones[dc.Name]
+func (a *azureDnsProvider) GetExistingRecords(domain string) (models.Records, []*adns.RecordSet, string, error) {
```
tlimoncelli: Does `GetExistingRecords` need to be exported?  It seems like it could be `getExistingRecords`

vatsalyagoel: Yep. can be an internal method. Fixed :)

tlimoncelli:
Thanks for the fast work on this!

Issue #619, opened 2020-02-12
Update dependencies


tlimoncelli:
hey folks!  there are a lot of outdated dependencies being used.  If you own a provider, please take a moment to update the related module.

List out-of-date modules and update any that

```
go get -u github.com/psampaz/go-mod-outdated
go list -u -m -json all | go-mod-outdated -update -direct
```

To update a module, `get` it, then re-run the unit and integration tests.

```
go get -u
    or
go get -u module/path
```

Once the updates are complete, tidy up:

```
go mod tidy
```

When done, vendor the modules:

```
go mod vendor
git add vendor
git commit -m'vendor modules' vendor
```

```
+-----------------------------------+---------------------------------------------------+------------------------------------+--------+------------------+
|              MODULE               |                      VERSION                      |            NEW VERSION             | DIRECT | VALID TIMESTAMPS |
+-----------------------------------+---------------------------------------------------+------------------------------------+--------+------------------+
| github.com/Azure/azure-sdk-for-go | v32.6.0+incompatible                              | v39.1.0+incompatible               | true   | true             |
| github.com/aws/aws-sdk-go         | v1.19.39                                          | v1.29.0                            | true   | true             |
| github.com/billputer/go-namecheap | v0.0.0-20170915210158-0c7adb0710f8                | v0.0.0-20191113012015-80fb801c9a11 | true   | true             |
| github.com/digitalocean/godo      | v1.17.1-0.20190625192827-3ec0b9ce8cde             | v1.30.0                            | true   | true             |
| github.com/dnsimple/dnsimple-go   | v0.20.1-0.20181001130357-234ec949d37c             | v0.31.0                            | true   | true             |
| github.com/exoscale/egoscale      | v0.10.5                                           | v1.19.0                            | true   | true             |
| github.com/go-acme/lego           | v2.7.2-0.20190727050804-58d6d9f4767a+incompatible | v2.7.2+incompatible                | true   | true             |
| github.com/hexonet/go-sdk         | v1.2.1                                            | v2.2.3+incompatible                | true   | true             |
| github.com/miekg/dns              | v1.0.14                                           | v1.1.27                            | true   | true             |
| github.com/robertkrimen/otto      | v0.0.0-20180617131154-15f95af6e78d                | v0.0.0-20191219234010-c382bd3c16ff | true   | true             |
| github.com/softlayer/softlayer-go | v0.0.0-20170804160555-5e1c8cccc730                | v1.0.0                             | true   | true             |
[...]

PR #637, opened 2020-02-20
DNSimple: Implement GetZones and ListZones 

onlyhavecans:
- Implement GetZoneRecords Per #628 
- Update go mod & tidy Per #619 
- Update comments and notes
- Implement ListZones per documentation
- Add a missing error checking
- update `go generate` after updating capabilities

### DIFF
--- a/docs/writing-providers.md
+++ b/docs/writing-providers.md
@@ -92,14 +92,14 @@ Pick a similar provider as your base.  Providers basically fall
 into three general categories:
 
 * **zone:** The API requires you to upload the entire zone every time. (BIND).
-* **incremental-record:** The API lets you add/change/delete individual DNS records. (ACTIVEDIR, CLOUDFLARE, NAMEDOTCOM, GCLOUD, ROUTE53)
+* **incremental-record:** The API lets you add/change/delete individual DNS records. (ACTIVEDIR, CLOUDFLARE, DNSIMPLE, NAMEDOTCOM, GCLOUD, ROUTE53)
 * **incremental-label:** Like incremental-record, but if there are
   multiple records on a label (for example, example www.example.com
 has A and MX records), you have to replace all the records at that
 label. (GANDI_V5)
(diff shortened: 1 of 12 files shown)


providers/dnsimple/dnsimpleProvider.go
```diff
@@ -360,6 +377,38 @@ func (c *DnsimpleApi) updateRecordFunc(old *dnsimpleapi.ZoneRecord, rc *models.R
...
+// Returns all the zones in an account
+func (c *DnsimpleApi) ListZones() ([]string, error) {
+	client := c.getClient()
+	accountID, err := c.getAccountID()
+	if err != nil {
+		return nil, err
+	}
+
+	opts := &dnsimpleapi.ZoneListOptions{}
+	zs := []dnsimpleapi.Zone{}
```
tlimoncelli: zs is used to collect `dnsimpleapi.Zone`'s which is used to later collect just the `Name` field from each.  Suggestion: just collect the `Name` field.

i.e. This line would become: `zones := []string` and then line 396 would become a loop that collects the zone names.  The loop around line 406 would be unneeded.

docs/writing-providers.md
```diff
@@ -92,14 +92,14 @@ Pick a similar provider as your base.  Providers basically fall
 into three general categories:
 
 * **zone:** The API requires you to upload the entire zone every time. (BIND).
-* **incremental-record:** The API lets you add/change/delete individual DNS records. (ACTIVEDIR, CLOUDFLARE, NAMEDOTCOM, GCLOUD, ROUTE53)
+* **incremental-record:** The API lets you add/change/delete individual DNS records. (ACTIVEDIR, CLOUDFLARE, DNSIMPLE, NAMEDOTCOM, GCLOUD, ROUTE53)
```
tlimoncelli: Thank you for updating the docs!

providers/dnsimple/dnsimpleProvider.go
onlyhavecans: Thanks for the feedback! Updated 🎊

tlimoncelli:
Thanks for the PR!!

PR #639, opened 2020-02-21
dnsimple: bug-fix SSHFP, add multi TXT support

philpennock:
The default logic for encoding SSHFP records was dropping the key and
hash algorithms and just posting the content, the `Can` check didn't
stop attempts to use SSHFP.  So, implement SSHFP support.

DNSimple support multiple DNS strings in a TXT record, by representing
the payload as quoted strings already.  This doesn't appear to be
documented, but it does actually work.

### DIFF
--- a/providers/dnsimple/dnsimpleProvider.go
+++ b/providers/dnsimple/dnsimpleProvider.go
@@ -20,7 +20,9 @@ var features = providers.DocumentationNotes{
 	providers.CanUseAlias:            providers.Can(),
 	providers.CanUseCAA:              providers.Can(),
 	providers.CanUsePTR:              providers.Can(),
+	providers.CanUseSSHFP:            providers.Can(),
 	providers.CanUseSRV:              providers.Can(),
+	providers.CanUseTXTMulti:         providers.Can(),
 	providers.CanUseTLSA:             providers.Cannot(),
 	providers.DocCreateDomains:       providers.Cannot(),
 	providers.DocDualHost:            providers.Cannot("DNSimple does not allow sufficient control over the apex NS records"),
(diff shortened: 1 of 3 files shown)


philpennock:
NB: `OWNERS` lists @aeden as owning the DNSimple provider but GitHub won't let me ask him for a review.  So ... tagging him here with this message.  :)

philpennock:
Before this PR: adding an `SSHFP` record to a DNSimple hosted domain would preview, would push, but would set bogus data which DNSimple's API doesn't reject.  The very next `dnsimple preview` would then panic because the data from the API server was malformed.

This PR does nothing to clean up such existing bad records.  It avoids creating them, so if dnscontrol ever worked "more than once" in a zone, it should be fine.

providers/dnsimple/dnsimpleProvider.go
```diff
@@ -471,3 +481,21 @@ func getTargetRecordPriority(rc *models.RecordConfig) int {
...
+// I know that it takes multiple quoted strings just fine, and constructs the
+// DNS multiple quoted items.
+// I'm not 100% on the escaping, but since it's a JSON API, JSON escaping seems
+// reasonable.
+// I do know that DNSimple have their own checks, so anything too crazy will
+// get a "400 Validation failed" HTTP response.
+func quoteDNSString(unquoted string) string {
+	b, err := json.Marshal(unquoted)
+	if err != nil {
+		panic(fmt.Errorf("enable to marshal to JSON: %q", unquoted))
```
tlimoncelli: Perhaps you mean s/enable/unable/ ?

philpennock:
Integration test: I don't know.  I just remembered to contact DNSimple support (coincidentally just before seeing this) to find out what's going on with their Sandbox account setup process: I'm unable to complete it, so can't run integration tests for this provider.

tlimoncelli:
Thanks for finding and fixing that!

PR #640, opened 2020-02-21
Add AUTODNSSEC, implement for DNSimple

philpennock:
There are two models for DNSSEC in DNS management: either dnscontrol manages all the DNSSEC records and has to be invoked regularly for re-signing, or the provider manages DNSSEC automatically and dnscontrol is not involved beyond saying "yeah, do that".

This implements the latter, as a model, and for one provider.

Potentially we could tune/configure options for DNSSEC such as algorithm, but DNSimple don't expose that API so I haven't implemented it.

This minimal model should be something which maps into other providers cleanly.


philpennock:
NB: this is cumulative on top of #639 ... review that first, then only the last commit in this PR matters.

This addresses issue #31 with one model of implementing DNSSEC.  It works for me, I have this used for `synadia.net` and `synadia.cloud` via DNSimple now.

@aeden for the DNSimple side of things.

PS: I spent the longest time trying to figure out why I wasn't seeing the DomainConfig field being set, before realizing that I was using the wrong example to crib from, and the helpers.js code needed to reference the JSON name to be able to set the field.  I haven't touched it here because it's not part of my work, but I think that `PURGE` and `NO_PURGE` might be broken: they're accessing `d.KeepUnknown` and should probably be accessing `d.keepunknown` instead, to hit the JSON field?  Unless I missed some case insensitivity somewhere.

providers/dnsimple/dnsimpleProvider.go
```diff
@@ -199,6 +209,34 @@ func (c *DnsimpleApi) GetRegistrarCorrections(dc *models.DomainConfig) ([]*model
...
+	enabled, err := c.getDnssec(dc.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	if enabled && !dc.AutoDNSSEC {
+		return []*models.Correction{
+			{
+				Msg: "Disable DNSSEC",
+				F:   func() error { _, err := c.disableDnssec(dc.Name); return err },
```
tlimoncelli: Will dc.Name be the right contents by the time this function is called?  I think so, but it is safer to have `domain := dc.Name` early in the parent function, then refer to domain in the func()'s. 

I can be talked out of this. I'm not an expert in this area. I just know I've hit this in the past with other things.

philpennock: Yes it's fine: you're thinking of loops which mutate the value being bound because people aren't capturing the value for the loop.  In this case, there's no looping and if someone changes the domain name underneath us ... all bets are off.  In this case, I'm matching existing style of several matching functions in this file.

So I'd prefer to match existing style, if that's acceptable?

philpennock:
Hrm, I'm getting a build failure because the validate tool insists that `gofmt` needs to be run on a file which I didn't touch, and where `gofmt` makes no changes.

Also: I found out how to render the docs locally so investigated, and the new `AUTODNSSEC` row in the feature matrix is missing a check for the platform which actually has it, so I need to investigate why that failed (unless you know and it's an easy fix?)

philpennock:
Okay, found the problem: missing `CanAutoDNSSEC: providers.Can()` in the provider for dnsimple.  I remember writing that line, must've deleted it by accident.

So does this mean that "missing Can" is not the same as "Cannot"?  How do we mark this feature as "must be explicitly enabled in a given provider, there's no default assumption that the provider probably can" ?

philpennock:
Ahah, hold on: the loop in checkProviderCapabilities is broken because AUTODNSSEC isn't an actual record, so we never ask if if the provider can auto-dnssec.  So this PR is buggy still.  Give me a few ...

philpennock:
Better.  With the latest change, and the debug print uncommented, and AutoDNSSEC temporarily commented out of DNSimple, I get:
```
% dnscontrol preview 
  (checking if "DNSIMPLE" can "AUTODNSSEC" for domain "synadia.net")
  (checking if "DNSIMPLE" can "AUTODNSSEC" for domain "synadia.cloud")
2 Validation errors:
ERROR: Domain synadia.net uses AUTODNSSEC records, but DNS provider type DNSIMPLE does not support them
ERROR: Domain synadia.cloud uses AUTODNSSEC records, but DNS provider type DNSIMPLE does not support them
Exiting due to validation errors
```

I need a way to run integration tests to catch this properly.  :(  But in the meantime, this is working.

providers/dnsimple/dnsimpleProvider.go
tlimoncelli: Yes, leave it as is.

onlyhavecans:
@philpennock To run integration tests you can set up a free account at https://sandbox.dnsimple.com 

see details here for how our sandbox works and mock billing info
https://developer.dnsimple.com/sandbox/


to use the sandbox outside of the integration tests  make sure to set the baser in your creds.json like below
```
{
  "dnsimple":{
    "token": "your token",
    "baseurl": "https://api.sandbox.dnsimple.com"
  }
}
```

philpennock:
@onlyhavecans I had a ticket in because I got stuck, and the ticket resolution was that the documentation I was following needs to be updated, and I was pointed towards the process to move forward.  I'm playing with integration tests now.

philpennock:
Yeah, not going to get integration tests done in time for landing this, just got quota exceeded errors, for testing _in the sandbox_. *sigh*

onlyhavecans:
I can confirm integration tests currently pass in this branch.

@philpennock I have increased your API limits in Sandbox significantly. If you hit any more issues please email support@dnsimple.com and I would be happy to assist.

philpennock:
Okay, I have tests for the missing check, and it's not an integration test since we don't want to actually talk to the providers, we're self-checking our own logic.

It's two branches ahead of this one, but each branch is a small commit, so it should be an easy flow, I just don't want to fight GitHub for "I want to land this dependent commit across repos, don't make people review all the other stuff"; here's the git log:
```
* 0394235 (HEAD -> pdp/test-capabilities, ce-github/pdp/test-capabilities) Use AST to ensure provider capabilities are checked
* a486fe4 (ce-github/pdp/bindDNSSEC, pdp/bindDNSSEC) Implement AutoDNSSEC for Bind output
* f013aae (ce-github/pdp/dnssec-agnostic, pdp/dnssec-agnostic) Correct wire in the AUTODNSSEC validation step
* df3a42e Validation fix for master broken
* 11bf78c Fix missing CanAutoDNSSEC on provider
* 07cc584 Add AUTODNSSEC, implement for DNSimple
* 1ec696e Implement get-zones (#641)
* f5d6f80 BIND: Implement get-zones (#642)
* d221471 dnsimple: bug-fix SSHFP, add multi TXT support (#639)
* 7c0e025 ACTIVEDIRECTORY: Implement get-zones (#643)
* 386275f (ce-github/master) ROUTE53: Fully implement get-zones (#638)
```

Branch `pdp/bindDNSSEC` has the Bind provider _claim_ to implement AutoDNSSEC, so that people can easily write out a copy of a zonefile still, but actually it just writes a comment into the zonefile recording that this was requested.  It changes the zonefile printer to let us have zone comments.

Branch `pdp/test-capabilities` adds the AST parser/checker which forces every provider capability to either be tested, or be explicitly whitelisted inside the new test.  It's depths-of-Go self-integrity-checks for the project, I think it's useful.  It caught another couple of missing tests.

https://github.com/ConnectEverything/dnscontrol/pull/new/pdp/test-capabilities

What I'd like to do is land this PR, then file small PRs for each of the next two features.  I don't want to split them to be independent branches because each would have merge conflicts.

So:
1. Please merge this PR and let me know
2. Expect two PRs after that, one after another
3. Expect people to complain when they suddenly get told that their provider doesn't support a feature they're using, because it worked despite not being declared as such.  And hey, the feature matrix will suddenly become a lot more accurate. :-)

tlimoncelli:
sounds like a plan!  I'll merge this and await the next 2 before any releases happen.

PR #646, opened 2020-02-21
Tests: check that RRtype record creation does fail

philpennock:
Rather than skip tests for things we assert will fail, test that they
actually will fail.

This might have portability problems across providers, but it works for
the one I've been using.  Someone with a rig designed to test all the
different providers should try this out.  :)

### DIFF
--- a/docs/_includes/matrix.html
+++ b/docs/_includes/matrix.html
@@ -939,8 +939,8 @@
 		<td class="success">
 			<i class="fa fa-check text-success" aria-hidden="true"></i>
 		</td>
-		<td class="info">
-			<i class="fa fa-circle-o text-info" aria-hidden="true"></i>
+		<td class="success">
+			<i class="fa fa-check text-success" aria-hidden="true"></i>
 		</td>
 		<td class="info">
 			<i class="fa fa-circle-o text-info" aria-hidden="true"></i>
@@ -952,8 +952,8 @@
 		<td class="info">
 			<i class="fa fa-circle-o text-info" aria-hidden="true"></i>
 		</td>
-		<td class="info">
-			<i class="fa fa-circle-o text-info" aria-hidden="true"></i>
+		<td class="success">
+			<i class="fa fa-check text-success" aria-hidden="true"></i>
 		</td>
 		<td><i class="fa fa-minus dim"></i></td>
 		<td class="info">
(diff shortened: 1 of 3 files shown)


captncraig:
I think maybe the reason we should skip it is that "new" features should be "opt-in" for existing providers. Once they have verified their ability to handle a record type, we can use it.

Before then, we don't require them to do anything special to deal with those if they get them. We just trust the system not to pass those in to them. I'd rather not force each provider to have code for each record type we know it can't handle.

Maybe there is a bad case where we don't support a record type, but someone manually creates one in a provider or something. If that's the case we should implement it.

philpennock:
This one is icky.  I'm happy to abandon this PR, but to make sure the combination of factors is out there:

In #639 @tlimoncelli had asked for an integration test that would have caught a problem with SSHFP not being rendered right, when not claimed to be supported by a provider which didn't have it.  Because we weren't including `"SSHFP"` in the list of records to be validated, this lack of support by the provider was not being honored.  Note that it's #640 (not yet merged) which fixes this by including SSHFP in the validation steps.

So, I guess really we want a different new integration test, which goes something like: build a list of all RR types guarded by `Can` capabilities, and then iterate through trying each one in a simple zone, to make sure that it's rejected unless and until enabled, to ensure the validation steps are up-to-date with the capabilities listings.

I'm not going to have time to revisit that again soon, sorry, but I hope that my exploring the problem space and spelling out a (perhaps?) better check is useful.

tlimoncelli:
My request in #639 was more simple.  The test created and deleted a SSHFP record, but didn't observe that the arguments were not properly updated.  The current test framework just tests if a record was created and/or deleted. Sorry for not being clear.

I wonder if we would have noticed the bug if there had been a test that (1) created the SSHFP record, (2) changed one of the fields, (3) changed another field.  That may have noticed the bug by side-effect.

The full solution would be to add to the integration test system a way to specify "The new zone should look exactly like this: a, b, c, ..."

@captncraig has some ideas about how to totally rework the integration test system.  I'd like to collect more requirements before we endeavour to rework the system.  Other things are higher priority right now.

philpennock:
The existing tests would, I think, have caught it, if the provider was marked to support `SSHFP`.  Because we have a test which pushes the records, then confirms that there's no diff on a second push.  The problem is that the test was _skipped_ because the provider was marked as not supporting `SSHFP`, but the logic was used anyway because the feature guard to only try to pass the RRtype through if the provider supports it was missing.

The last commit on the https://github.com/ConnectEverything/dnscontrol/tree/pdp/test-capabilities branch should fix this.  Once #648 lands, I can submit a PR for that branch to pull in just that one commit.  I add a new test `TestCapabilitiesAreFiltered` which detects when we're not actually checking the `CanUseFoo` guards, so that we don't try to use untested features.

PR #648, opened 2020-02-22
BIND: Implement AutoDNSSEC

philpennock:
There's a philosophy issue here around what is the Bind output meant to do.  Since AFAIK we're not integrating into Bind's catalog zones or the like, we're just targeting the zonefiles, we're not in a position to do _anything_ relating to registrar options such as setting up DS glue.

So at one level, enabling AutoDNSSEC for Bind is a lie. But without this, folks can't target a Bind zone as a secondary provider for their domain, to get debug dumps of the zone output, because the checks for "Can" block it.  So I think this commit achieves a happy compromise: we write a comment into the Bind zonefile, indicating that DNSSEC was requested.

Actually: we add support for arbitrary zone comments to be written into a zonefile via a slightly ugly "can be `nil`" parameter.  We then write in a generation timestamp comment, and if AutoDNSSEC was requested we then write that in too.

### DIFF
--- a/commands/getZones.go
+++ b/commands/getZones.go
@@ -151,11 +151,11 @@ func GetZone(args GetZoneArgs) error {
 	for i, recs := range zoneRecs {
 		zoneName := zones[i]
 
-		z := prettyzone.PrettySort(recs, zoneName, 0)
+		z := prettyzone.PrettySort(recs, zoneName, 0, nil)
 		switch args.OutputFormat {
 		case "pretty":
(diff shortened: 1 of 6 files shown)


philpennock:
Fixed the commit title to conform to repo style, fixed which email address I'm using, lots of house-keeping force-pushes there, none of which changed the content tree, only the commit data.  Sorry about that.

tlimoncelli:
LGTM

I'm merging this.  Just FYI: I'm about to do a big update to the BIND provider.  The SOA handling was never very good, and is overly complex.  Expect a big PR soon.

PR #650, opened 2020-02-22
Tests: ensure provider capabilities are checked

philpennock:
Adds test: `TestCapabilitiesAreFiltered`

We have a number of records and pseudo-records which in theory can only be used with a given provider if that provider indicates support.  In practice, we've been missing the checks for that support and have been passing the records down anyway.  The advice comment in the providers/capabilities.go file to edit `checkProviderCapabilities()` has not been reliably followed.

We need an internal self-consistency test.  The constants are not directly exported or enumerable based solely on the package interfaces at run-time, but with source access for a test suite, we can use the `go/ast` and related interfaces to examine the code, extract all the constants from a given package, figure out which ones we want to be handled, and then insist that they're handled.

Before my recent work, we only checked:

    ALIAS PTR SRV CAA TLSA

After this commit, we check:

    ALIAS AUTODNSSEC CAA NAPTR PTR R53_ALIAS SSHFP SRV TLSA

I've added `AUTODNSSEC` as a new feature; `SSHFP` and `PTR` were caught in other recent commits from me; implementing this test caused me to have to add `NAPTR` and `R53_ALIAS`.  I whitelist `CanUseTXTMulti` as a special-case.

This should prevent regressions.  We will probably want to post publicly to warn people that if they're using SSHFP/PTR/NAPTR/R53_ALIAS then they should check the feature matrix and if they don't see their provider listed, to report is as "hey that actually works" so we can update the provider flags.  Bonus: our feature matrix will suddenly be more accurate.

### DIFF
--- a/docs/adding-new-rtypes.md
+++ b/docs/adding-new-rtypes.md
@@ -117,16 +117,9 @@ As you debug, if there are places that haven't been marked
 `#rtype_variations` that should be, add such a comment.
 Every time you do this, an angel gets its wings.
 
-The tests also verify that for every "capability"
-If you see an error such as this, the "MISSING" means that
-providers.CanUseCAA is not listed in the `providerCapabilityChecks`
-array in `pkg/normalize/validate.go`.  This is a
-
-```
---- FAIL: TestCapabilitiesAreFiltered (0.00s)
-    capabilities_test.go:66: ok: providers.CanUseAlias (0) is checked for with "ALIAS"
-    capabilities_test.go:68: MISSING: providers.CanUseCAA (1) is not checked by checkProviderCapabilities
-```
+The tests also verify that for every "capability" there is a
+validation. This is explained in Step 2 (search for
+`TestCapabilitiesAreFiltered` or `MISSING`)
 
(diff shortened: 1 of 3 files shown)


philpennock:
I thought I'd finally have a PR with no changes on my side after asking for review, but just noticed two messy items in the imports list for the new file.  :2nd_place_medal: 
Fixed.

tlimoncelli:
I need to take a moment to bask in the awesomeness of this PR.

Dude... you're parsing the source code to make sure the source code is doing the right thing.  Dude.

philpennock:
Yeah, Domain Specific Languages are so last decade. Domain Specific Linters in the testsuite are where we should take the 20s.

tlimoncelli:
I have some suggested comments in https://github.com/ConnectEverything/dnscontrol/pull/1/files

philpennock:
Just to note over here: the PR Tom filed was merged into the branch which is this PR, so I think this is ready for merging.

tlimoncelli:
Awesome PR, man!  Thanks!

Issue #583, opened 2020-01-13
Softlayer DNS for TXT always tries to modify and fails


saurabhguptag:
I see there is a bug for TXT records for soflayer DNS, Once the following config is push, dnscontrol always recognise that there is change in the record.

`   
    TXT("XYZ", "ABC", TTL(300)),
    TXT("XYZ1", "ABC1", TTL(300)),
`   

following is preview 

`
----- DNS Provider: softlayer...2 corrections
#1: MODIFY TXT xyz.testcbz.com: ( ttl=300) -> ("ABC" ttl=300)
#2: MODIFY TXT xyz1.testcbz.com: ( ttl=300) -> ("ABC1" ttl=300)
----- Registrar: none...0 corrections
`

following is the output when trying to push

`
----- DNS Provider: softlayer...2 corrections
#1: MODIFY TXT xyz.testcbz.com: ( ttl=300) -> ("ABC" ttl=300)
FAILURE! didn't find changes when I expect some
#2: MODIFY TXT xyz1.testcbz.com: ( ttl=300) -> ("ABC1" ttl=300)
FAILURE! didn't find changes when I expect some
----- Registrar: none...0 corrections
`

using docker version at latest version.

tlimoncelli:
Can @jamielennox please take a look?

saurabhguptag:
Is there any expected timeline for this?

tlimoncelli:
FYI: I don't have a test account on Softlayer.   @jamielennox maintains this provider.

jamielennox:
eek - sorry. It's probably been 18 months since I left that job and haven't had access since. I'm surprised (unless it's a softlayer bug - in which case i'm not surprised) that this is something that has only started being a problem now.

saurabhguptag:
I have done the fix could check and merge ?

https://github.com/StackExchange/dnscontrol/pull/659

tlimoncelli:
@saurabhguptag Thanks for the bug fix.  I'll merge that soon.

PR #659, opened 2020-02-27
Fixed Softlayer TXT Record existence Issue #583

saurabhguptag:
Appending record data into TxtStrings array


tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/583

PR #666, opened 2020-02-28
(#628) Add get-zones to the OVH provider

masterzen:
Commit 87ad01d added the very useful `get-zones` command, whichrequires providers to implement a new method `GetZoneRecords` (tracked in #628).

This changes make the OVH provider support this new feature.


tlimoncelli:
this is awesome!  thanks!

Issue #668, opened 2020-02-28
Softlayer Lets Encrypt Certificate issue 


saurabhguptag:
While Requesting for Lets Encrypt Certificate The application is creating a TXT record the name _acme-challenge and  TTL value of 0, But Softlayer supports TTL value between 60 and 604800 only.

So the Application fails to start creating the certificate and throws following error.

`acme: error presenting token: SoftLayer_Exception_Public: Please provide a TTL value between 60 and 604800. (HTTP 500)`

saurabhguptag:
Once Fix done, I successfully generated the Certificate.

PR #669, opened 2020-02-28
Fixed: Softlayer Lets Encrypt Certificate issue #668

saurabhguptag:
modifying TTL value to Minimum supported Value by softlayer if found less than it.


tlimoncelli:
Thanks!

PR #671, opened 2020-02-29
DNSIMPLE: support NAPTR

philpennock:
Fix the model to strip quotes from all the string parts; BIND integration tests still pass.  DNSIMPLE integration tests pass.

### DIFF
--- a/models/t_naptr.go
+++ b/models/t_naptr.go
@@ -44,5 +44,5 @@ func (rc *RecordConfig) SetTargetNAPTRString(s string) error {
 	if len(part) != 6 {
 		return fmt.Errorf("NAPTR value does not contain 6 fields: (%#v)", s)
 	}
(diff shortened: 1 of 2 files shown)


philpennock:
There were a bunch of unrelated changes when I used `go generate` so I left that out of this PR.  There's a test failure, but that too is unrelated to this change: `TestCheckLabel/_y3_CNAME` looks to be related to recent work.

If master gets cleaned up, with the matrix updated, then I'll add the delta for this change to the PR.

tlimoncelli:
Rebase and the generate/tests should be fine.

philpennock:
Yup, though it looks like `docs/_includes/matrix.html` might have been generated with this PR applied?

tlimoncelli:
Thanks for the PR!

PR #673, opened 2020-03-01
GANDI_V5: ALIAS, SSHFP, TLSA

philpennock:
I can't find a sandbox for Gandi to test with so haven't run integration tests, but I've tested these all with one of my domains (which is how I found the "ALIAS only at apex" restriction).


philpennock:
(I'm reusing my employer's fork, since that's where my other work was done so it's less confusing all around, and I don't think my employer will mind, but this is actually for use with some of my personal domains.)

tlimoncelli:
Looks great!  Thanks for finding these issues and fixing them!

P.S. I also use GANDI for personal domains :-)

PR #678, opened 2020-03-03
One way we could reorganize the integration tests

tlimoncelli:


captncraig:
I was thinking more along the lines of https://github.com/StackExchange/dnscontrol/compare/master...captncraig:integration?expand=1

Taking out the major sections of the big list of tests and making each a standalone test function. The utility function can implicitly add an empty at the start of each one and so forth. Filtering and exceptions and stuff can be done in line with how you were going, or inside individual test functions.

The major problem remaining is that of the provider-specific exceptions. I ripped those out in my branch. Intention was to have the provider give a "warning" somehow for cases it absolutely can't support and have the integration tests detect that. But I still am not sure what that would look like.

captncraig:
The other thing that could be done now to the tests is running additional validation on the results of `GetZoneRecords`.

Each test case could run something like:

1. Get and push corrections.
2. Call `GetZoneRecords` and do some deep compare with the desired records. Check for stuff like dots and fqdns and stuff like that.
3. Get corrections again and expect zero.

tlimoncelli:

Issue #491, opened 2019-05-20
GetNameservers() is inconsistent wrt "trailing dot" across providers


tlimoncelli:
You are probably reading this issue because of a warning told you to read this.

The warning is to help us find and fix an inconsistency in the code.  Some providers store a nameserver with a trailing dot, others don't.  Currently the code strips out the dot.  We would prefer if all providers act the same.

More specifically: The problem is with `models.DomainConfig.Nameservers`.  Some providers store a `FQDN` others store a `FQDN + "."` (known as a "trailing dot").

This inconsistency means that later code must deal with having (or not having) the trailing dot.  In fact, in `pkg/nameservers/nameservers.go` the dot is stripped out to compensate for all future code. However this covers up the inconsistency.  That's kind of messy and we'd prefer to clean this up.  (@tlimoncelli is to blame for this inconsistency as he wrote the original code; now he's trying to fix it.)

What we would prefer is if all providers simply stored the FQDN without the dot.

Plan:

* Before 3.0: The trailing dot is silently removed.
* Starting with 3.0: The trailing dot will be removed, but a warning that includes a link to this issue will be printed.
* Eventually: dnscontrol will error out if the trailing dot exists.

If you've seen the warning message, please consider sending a PR for your provider.

A checkbox in this list means the provider code has been checked and fixed:

- [x] `ACTIVEDIRECTORY_PS` @tlimoncelli
- [x] `AZURE_DNS` @vatsalyagoel
- [ ] `BIND` @tlimoncelli
- [x] `CLOUDNS` @pragmaton
- [X] `CLOUDFLAREAPI` @tlimoncelli
- [x] `DIGITALOCEAN` @Deraen
- [x] `DNSIMPLE` @aeden
- [x] `EXOSCALE` @pierre-emmanuelJ
- [X] `GANDI_V5` @TomOnTime
- [X] `GCLOUD` @tlimoncelli
- [ ] `HEXONET` @papakai
- [ ] `INTERNETBS` @pragmaton
- [ ] `LINODE` @koesie10
- [ ] `NAMECHEAP` @captncraig
- [X] `NAMEDOTCOM` @tlimoncelli
- [ ] `NS1` @captncraig
- [X] `ROUTE53` @tlimoncelli
- [x] `OCTODNS` @TomOnTime
- [ ] `OPENSRS` @pierre-emmanuelJ
- [x] `OVH` @masterzen
- [ ] `SOFTLAYER`@jamielennox
[...]

PR #679, opened 2020-03-03
(#491) convert the OVH provider to models.ToNameservers

masterzen:
As requested in #491, as OVH doesn't return dot-suffixed NS, therefore we can use `models.ToNameservers` in this provider (which incidentally this PR does).


tlimoncelli:
Thanks!  Its a small change but it will help me in the future.  Gracias!

PR #680, opened 2020-03-04
Add CanUseTXTMulti to GANDI_V5 provider

jasonbcox:
Encountered this while trying to push a long TXT record with GANDI_V5.  The provider accepts it but reformats it into a compatible multi-TXT record.  Unfortunately, this results in every subsequent `dnscontrol preview` to indicate that the TXT record does not match.  `dnscontrol push` will also update this record with every call.

With this patch, I'm able to push the TXT record as expected, and preview after push indicates no more changes as expected.

`cd integrationTest && go test` passes


tlimoncelli:
Thanks for finding that!

PR #681, opened 2020-03-04
ClouDNS: Get zone records implemented 

pragmaton:
* #628 get-zones implemented
* #491 convert to models.ToNameservers


tlimoncelli:
Thanks for the PR!

PR #682, opened 2020-03-05
brief sketch of more modular integration tests

tlimoncelli:


tlimoncelli:

PR #684, opened 2020-03-07
refactor into groups

tlimoncelli:


tlimoncelli:
Not hearing any complaints, I've decided to go with the most recent iteration: testgroup()s

I think it adds all the features we were looking for except the ability to break out tests into individual files.  I think that's future-proofing we won't need, or could add later.

PR #696, opened 2020-03-13
Correct Cloudflare redirect function documentation

### LINKED ISSUE #695: Warning in CF_REDIRECT()/CF_TEMP_REDIRECT() doc is incorrect (Cloudflare Page Rules)
Cloudflare provides a [Page Rules feature](https://support.cloudflare.com/hc/en-us/articles/218411427-Understanding-and-Configuring-Cloudflare-Page-Rules-Page-Rules-Tutorial-) that "trigger certain actions whenever a request matches one of the URL patterns you define". One of these is a "[Forwarding URL](https://support.cloudflare.com/hc/en-us/articles/200172286-Configuring-URL-forwarding-or-redirects-with-Cloudflare-Page-Rules)" that returns a HTTP 301 or 302 redirect to the client; other Page Rules control caching, the web app firewall, and other CDN features.

The dnscontrol [Cloudflare provider](https://stackexchange.github.io/dnscontrol/providers/cloudflare) supports Cloudflare's "Forwarding URL" PageRules (but no other types) by way of a CF_REDIRECT() function. This is quite convenient and helpful.

The [CF_REDIRECT()/CF_TEMP_REDIRECT() function doc](https://stackexchange.github.io/dnscontrol/js#CF_REDIRECT) says

> WARNING: If the domain has other pagerules in place, they may be deleted. At this time this feature is best used on bare domains that need to redirect to another domain, perhaps with wildcard substitutions.

Looking at [the code](https://github.com/StackExchange/dnscontrol/blob/825ba2d08119508e66295013e4efc15747660465/providers/cloudflare/rest.go#L402
) though, it seems that the cloudflare interface explicitly ignores any rules with an ID other than `forwarding_url`. Should the above WARNING actually say "... If the domain has other [Forwarding URL] pagerules in place ..." ?

I do have non-"Forwarding URL" page rules, I'm fine to not manage them via dnscontrol (for now!) but I'd like to know they're not going to be messed with.

In a nutshell, is the design intention:
  a) if you have ONLY "Forwarding URL" page rules, then you can manage them via dnscontrol; if you have other types of rules as well then you can't use dnscontrol for any page rules, full stop.
or
  b) you can manage your "Forwarding URL" page rules via dnscontrol, rules of any other type will be left alone

I read the warning in the doc as a), the code does b).

If b) is correct, then the warning is really just saying "if you use dnscontrol to manage these settings, then anything not defined in the domain will be removed" - which is exactly how the rest of dnscontrol works; so the warning seems superfluous?

In practice, running a preview with "foreign" Forwarding URL page rules in place raises a "DELETE PAGE_RULE" change, exactly as you'd want just like any regular DNS record.

ad8-bdl:
Closes #695

### DIFF
--- a/docs/_functions/domain/CF_REDIRECT.md
+++ b/docs/_functions/domain/CF_REDIRECT.md
@@ -5,12 +5,26 @@ parameters:
   - modifiers...
 ---
 
-`CF_REDIRECT` is the same as `CF_TEMP_REDIRECT` but generates a
-http 301 redirect (permanent redirect) instead of a temporary
-redirect.
+`CF_REDIRECT` uses Cloudflare-specific features ("Forwarding URL"
+Page Rules) to generate a HTTP 301 permanent redirect.
+
+If _any_ `CF_REDIRECT` or `CF_TEMP_REDIRECT` functions are used
+then `dnscontrol` will manage _all_ "Forwarding URL" type Page Rules
+for the domain. Page Rule types other than "Forwarding URL” will
+be left alone.
 
 These redirects are cached by browsers forever, usually ignoring
 any TTLs or other cache invalidation techniques.   It should be
 used with great care.  We suggest using a `CF_TEMP_REDIRECT`
 initially, then changing to a `CF_REDIRECT` only after sufficient
 time has elapsed to prove this is what you really want.
+
+This example redirects the bare (aka apex, or naked) domain to www:
+
+{% include startExample.html %}
+{% highlight js %}
+D("foo.com", .... ,
+  CF_REDIRECT("mydomain.com/*", "https://www.mydomain.com/$1"),
+);
+{%endhighlight%}
+{% include endExample.html %}
--- a/docs/_functions/domain/CF_REDIRECT.md
+++ b/docs/_functions/domain/CF_REDIRECT.md
@@ -5,19 +5,23 @@ parameters:
   - modifiers...
 ---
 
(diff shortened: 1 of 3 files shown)


tlimoncelli:
Thanks! These edits are great!

My concern is that I'm paranoid that a bug in dnscontrol could lead to deleting page rules unintentionally.  Yes, the code ignores rules it doesn't understand, but that entire system hasn't received a lot of testing.  Because of the nature of working with an API we don't control, it is difficult to do extensive testing.

I'd like the text have some kind of warning.

ad8-bdl:
> I'd like the text have some kind of warning.

I find the existing warning to be ambiguous (ref. #695 for details; tl'dr = does it mean "_all_ your Page Rules are belong to us", or is it being a little imprecise with the terminology and really mean "all your Forwarding URL rules are mine, you keep the rest"; the code says the latter).

In neither reading do I take it to mean "this is an inadequately documented and tested interface, YMMV".

Should the warning, added to the changes already proposed here, be along the lines of
"WARNING: Cloudflare do not currently fully document the Page Rules API and this interface is not extensively tested."

or something else?

tlimoncelli:
I agree with your analysis and solution.  How about:

WARNING: Cloudflare does not currently fully document the Page Rules API and this interface is not extensively tested. Take precautions such as making backups and manually verifying `dnscontrol preview` output before running `dnscontrol push`. This is especially true when mixing pagerules that are managed by DNSControl and those that aren't.

tlimoncelli:
Awesome. Thank you for this contribution!

PR #703, opened 2020-03-22
Fix #702: Spurious null MX changes on CloudFlare

### LINKED ISSUE #702: Spurious changes to null MX on CloudFlare
I don't know if this affects any other backends, but if one creates a null MX (per RFC 7505) on the CloudFlare backend, it's mistakenly read as targeting `..` instead of `.`, which then appears not equal to the source record, causing a "change" to be previewed or pushed.

I traced the issue to the CloudFlare provider adding a `.` to the end of all MX record targets, even in the case when the target is just a `.`.

pjeby:
When the CloudFlare provider syncs, it adds an extraneous `.` to the retrieved record's target, if the target is just a `.` (e.g. for an RFC 7505 null MX of `MX('@', 0, '.')`).  This PR ensures that the `.` is only added to targets that don't already have one.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -610,7 +610,7 @@ func makeTests(t *testing.T) []*TestGroup {
 		),
 
 		testgroup("Null MX",
-			only("BIND", "CLOUDFLAREAPI", "GCLOUD", "ROUTE53"), // Not all providers support RFC 7505
+			not("AZURE_DNS", "GANDI_V5", "NAMEDOTCOM", "DIGITALOCEAN"), // These providers don't support RFC 7505
 			tc("Null MX", mx("@", 0, ".")),
 		),
 
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -610,7 +610,7 @@ func makeTests(t *testing.T) []*TestGroup {
 		),
 
 		testgroup("Null MX",
-			only("CLOUDFLAREAPI"), // Other providers might support this, CF definitely does
+			only("BIND", "CLOUDFLAREAPI", "GCLOUD", "ROUTE53"), // Not all providers support RFC 7505
 			tc("Null MX", mx("@", 0, ".")),
 		),
 
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -609,6 +609,11 @@ func makeTests(t *testing.T) []*TestGroup {
 			tc("Record pointing to @", mx("foo", 8, "**current-domain**")),
 		),
 
+		testgroup("Null MX",
+			only("CLOUDFLAREAPI"), // Other providers might support this, CF definitely does
+			tc("Null MX", mx("@", 0, ".")),
+		),
+
 		testgroup("NS",
 			not("DNSIMPLE", "EXOSCALE"),
 			// DNSIMPLE: Does not support NS records nor subdomains.
--- a/providers/cloudflare/cloudflareProvider.go
+++ b/providers/cloudflare/cloudflareProvider.go
@@ -541,7 +541,7 @@ type cfRecord struct {
 func (c *cfRecord) nativeToRecord(domain string) *models.RecordConfig {
 	// normalize cname,mx,ns records with dots to be consistent with our config format.
 	if c.Type == "CNAME" || c.Type == "MX" || c.Type == "NS" || c.Type == "SRV" {
-		c.Content = dnsutil.AddOrigin(c.Content+".", domain)
+		c.Content = dnsutil.AddOrigin(strings.TrimSuffix(c.Content, ".")+".", domain)
 	}
 
 	rc := &models.RecordConfig{
--- a/providers/cloudflare/cloudflareProvider.go
+++ b/providers/cloudflare/cloudflareProvider.go
@@ -541,7 +541,9 @@ type cfRecord struct {
 func (c *cfRecord) nativeToRecord(domain string) *models.RecordConfig {
 	// normalize cname,mx,ns records with dots to be consistent with our config format.
 	if c.Type == "CNAME" || c.Type == "MX" || c.Type == "NS" || c.Type == "SRV" {
-		c.Content = dnsutil.AddOrigin(strings.TrimSuffix(c.Content, ".")+".", domain)
+		if c.Content != "." {
+			c.Content = c.Content + "."
+		}
 	}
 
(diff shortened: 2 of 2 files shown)


tlimoncelli:
Maybe I don't understand the situation but I don't think that code will fix the problem.

Can you add an integration test that would fail in the current code but pass with your code?  I've created an example in https://github.com/StackExchange/dnscontrol/pull/704/files but the test fails for any provider I try it on, so perhaps I'm unclear on the topic.

Do many providers support RFC 7505?

pjeby:
Hi Tim.  The issue is that CloudFlare strips the trailing `.` from records returned by its API, *except* when the target is *just* a `.`.  This was fixed for SRV records in 70ce16ff, but MX records have a similar issue, in a different field.  (See [this code](/StackExchange/dnscontrol/blob/master/providers/cloudflare/cloudflareProvider.go#L520) for the equivalent SRV name translation.)

I don't know how many providers support RFC 7505, but CloudFlare *does*; the issue is simply that dnscontrol turns the `.` into `..` on read, which then leads to it issuing an unnecessary Change to set it back to `.`.

I've added an integration test that fails without the PR, and succeeds with it.  I marked it as only for Cloudflare, but of course you can see if it works on other providers, or whether they have the same problem.  DNSimple's provider looks similar enough to CF that it might have the same bug, depending on how their API works.

As to your test in #704, that record is not actually compliant with RFC 7505, which requires a zero priority and AFAIK is not supported for any name but `@`.  I would expect some providers to fail to validate it (or an actual RFC 7505 record), but the actual bug being fixed here is that CF returns a `.` and dnscontrol turns it into `..`, resulting in `Expected 0 corrections on second run, but found 1` in the integration test.  If you are seeing that error on other providers, it means there is a similar problem in that provider within dnscontrol; other errors would mean something else.

Finally, if you want to change how this fix is accomplished, the if-based approach would be something like:

```go
if c.Content != "." {
        c.Content = c.Content + "."
}
```

I simply stuck with the idempotent-functional pattern that was already in use here and elsewhere in dnscontrol (e.g. the similar handling for SRV records mentioned above).

tlimoncelli:
Ah, I see.  (I'm not familiar with the new "null MX" stuff. Very interesting).

I'd prefer the `if c.Content != "." {` version.  If you don't mind updating the PR, I'd appreciate it.

pjeby:
Done.  Should I also remove the `only("CLOUDFLAREAPI")` tag from the test?

tlimoncelli:
I suggest we leave it as is.  I've tested a couple providers and none of them support null MX.  If this becomes popular, we could add a capability so providers self-disclose their status.

In other news... I also think it might be useful to add a macro to the helpers.js file to make this easier to use such as a NULLMX alias? 

```
D("domain.tld", blah, blah,
     A("@", "1.2.3.4"),
     NULLMX,
)
```

I'm not good at picking names: `NULLMX`, `MXNULL`, `MXNULL()`?

pjeby:
Based on the naming of other things in helpers.js, I would say that `MX_NULL` most closely matches the existing naming convention (type of thing, `_`, specific value), though I personally think `NULL_MX` is more readable in context.  (That being said, I *think* you can have a null MX for a subdomain, in which case providing a predefined constant is of less use.)

Regarding the other providers, what errors, specifically, are you getting?  The BIND provider works fine, but I don't have accounts for any providers besides CloudFlare.  I assume some providers might complain about setting a `.`, but others might have the same problem as CloudFlare, in which case I can probably include the fix for them.  (Since e.g. dnsimple does a similar `.` concatenation thing.)

tlimoncelli:
I made similar changes as you did.  See https://github.com/StackExchange/dnscontrol/tree/b703_cfmx

On further testing:

These work:  BIND, CLOUDFLAREAPI, GCLOUD, ROUTE53

These do not:

```

AZURE_DNS:

=== RUN   TestDNSProviders/dnscontrol-azure.com/03:MX_null
    TestDNSProviders/dnscontrol-azure.com/03:MX_null: integration_test.go:179: CREATE MX foo.dnscontrol-azure.com 10 . ttl=300
    TestDNSProviders/dnscontrol-azure.com/03:MX_null: integration_test.go:183: dns.RecordSetsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="The domain name '.' is invalid."
=== RUN   TestDNSProviders/dnscontrol-azure.com/Post_cleanup


GANDI_V5:

=== RUN   TestDNSProviders/dnscontroltest-gandilivedns.com/03:MX_null
    TestDNSProviders/dnscontroltest-gandilivedns.com/03:MX_null: integration_test.go:179: CREATE MX foo.dnscontroltest-gandilivedns.com 10 . ttl=300
    TestDNSProviders/dnscontroltest-gandilivedns.com/03:MX_null: integration_test.go:183: {Code:0 Message: UUID: Object: Cause: Status: Errors:[]}: items: String does not match expected pattern

NAMEDOTCOM:

=== RUN   TestDNSProviders/dnscontrol-ndc.com/03:MX_null
    TestDNSProviders/dnscontrol-ndc.com/03:MX_null: integration_test.go:179: CREATE MX foo.dnscontrol-ndc.com 10 . ttl=300
    TestDNSProviders/dnscontrol-ndc.com/03:MX_null: integration_test.go:183: Invalid Argument: Parameter Value Error - Answer Invalid Host
=== RUN   TestDNSProviders/dnscontrol-ndc.com/Post_cleanup

DIGITALOCEAN:

=== RUN   TestDNSProviders/dnscontrol-do.com/03:Null_MX:Null_MX
    TestDNSProviders/dnscontrol-do.com/03:Null_MX:Null_MX: integration_test.go:179: CREATE MX dnscontrol-do.com 0 . ttl=300
    TestDNSProviders/dnscontrol-do.com/03:Null_MX:Null_MX: integration_test.go:183: POST https://api.digitalocean.com/v2/domains/dnscontrol-do.com/records: 422 record hostname is missing
=== RUN   TestDNSProviders/dnscontrol-do.com/Post_cleanup:Empty


```

pjeby:
Ok, I've added the other working providers to the testgroup.  Let me know if there's anything else you want me to add here.

tlimoncelli:
Let's flip it around and use not() instead of only().  The remaining providers might pass, and if they don't it will be apparent and they can submit a PR.

pjeby:
Ok, flipped.

tlimoncelli:
Thanks!

PR #725, opened 2020-04-26
NEW PROVIDER: deSEC

### LINKED ISSUE #724: Provider Request: deSEC
Hi!

The non-profit DNS Provider deSEC.de from Germany is now out of beta 🥳

Can you please implement their API https://desec.readthedocs.io?

Thanks!

D3luxee:
This should fix #724

### DIFF
--- a/docs/_providers/desec.md
+++ b/docs/_providers/desec.md
@@ -26,8 +26,8 @@ Example Javascript:
 var REG_NONE = NewRegistrar('none', 'NONE');    // No registrar.
 var deSEC = NewDnsProvider('desec', 'DESEC');  // deSEC
 
-D("example.tld", REG_NONE, DnsProvider(deSEC),
-    A("test","1.2.3.4")
+D('example.tld', REG_NONE, DnsProvider(deSEC),
+    A('test','1.2.3.4')
 );
 {% endhighlight %}
 
--- a/docs/_providers/desec.md
+++ b/docs/_providers/desec.md
@@ -1,0 +1,36 @@
+---
+name: deSEC
+title: deSEC Provider
+layout: default
+jsId: DESEC
+---
+# deSEC Provider
+## Configuration
+In your providers credentials file you must provide a deSEC account auth token:
+
+{% highlight json %}
+{
+  "desec": {
+    "auth-token": "your-deSEC-auth-token"
+  }
+}
+{% endhighlight %}
+
+## Metadata
+This provider does not recognize any special metadata fields unique to deSEC.
+
+## Usage
+Example Javascript:
+
+{% highlight js %}
+var REG_NONE = NewRegistrar('none', 'NONE');    // No registrar.
+var deSEC = NewDnsProvider('desec', 'DESEC');  // deSEC
+
+D("example.tld", REG_NONE, DnsProvider(deSEC),
+    A("test","1.2.3.4")
+);
+{% endhighlight %}
+
+## Activation
+DNSControl depends on a deSEC account auth token.
+This token can be obtained by logging in via the deSEC API: https://desec.readthedocs.io/en/latest/auth/account.html#log-in
--- a/providers/desec/convert.go
+++ b/providers/desec/convert.go
@@ -13,9 +13,9 @@ import (
 func nativeToRecords(n resourceRecord, origin string) (rcs []*models.RecordConfig) {
 
 	// deSEC returns all the values for a given label/rtype pair in each
-	// livedns.DomainRecord.  In other words, if there are multiple A
+	// resourceRecord.  In other words, if there are multiple A
 	// records for a label, all the IP addresses are listed in
-	// n.RrsetValues rather than having many livedns.DomainRecord's.
+	// n.Records rather than having many resourceRecord's.
 	// We must split them out into individual records, one for each value.
 	for _, value := range n.Records {
 		rc := &models.RecordConfig{
@@ -36,9 +36,9 @@ func nativeToRecords(n resourceRecord, origin string) (rcs []*models.RecordConfi
 }
 
 func recordsToNative(rcs []*models.RecordConfig, origin string) []resourceRecord {
-	// Take a list of RecordConfig and return an equivalent list of ZoneRecords.
-	// deSEC requires one ZoneRecord for each label:key tuple, therefore we
-	// might collapse many RecordConfig into one ZoneRecord.
+	// Take a list of RecordConfig and return an equivalent list of resourceRecord.
+	// deSEC requires one resourceRecord for each label:key tuple, therefore we
+	// might collapse many RecordConfig into one resourceRecord.
 
 	var keys = map[models.RecordKey]*resourceRecord{}
 	var zrs []resourceRecord
--- a/providers/desec/convert.go
+++ b/providers/desec/convert.go
@@ -1,0 +1,79 @@
+package desec
+
+// Convert the provider's native record description to models.RecordConfig.
+
+import (
+	"fmt"
+
+	"github.com/StackExchange/dnscontrol/v3/models"
+	"github.com/StackExchange/dnscontrol/v3/pkg/printer"
+)
+
+// nativeToRecord takes a DNS record from deSEC and returns a native RecordConfig struct.
+func nativeToRecords(n resourceRecord, origin string) (rcs []*models.RecordConfig) {
+
+	// deSEC returns all the values for a given label/rtype pair in each
+	// livedns.DomainRecord.  In other words, if there are multiple A
+	// records for a label, all the IP addresses are listed in
+	// n.RrsetValues rather than having many livedns.DomainRecord's.
+	// We must split them out into individual records, one for each value.
+	for _, value := range n.Records {
+		rc := &models.RecordConfig{
+			TTL:      n.TTL,
+			Original: n,
+		}
+		rc.SetLabel(n.Subname, origin)
+		switch rtype := n.Type; rtype {
+		default: //  "A", "AAAA", "CAA", "NS", "CNAME", "MX", "PTR", "SRV", "TXT"
+			if err := rc.PopulateFromString(rtype, value, origin); err != nil {
+				panic(fmt.Errorf("unparsable record received from deSEC: %w", err))
+			}
+		}
+		rcs = append(rcs, rc)
+	}
+
+	return rcs
+}
+
+func recordsToNative(rcs []*models.RecordConfig, origin string) []resourceRecord {
+	// Take a list of RecordConfig and return an equivalent list of ZoneRecords.
+	// deSEC requires one ZoneRecord for each label:key tuple, therefore we
+	// might collapse many RecordConfig into one ZoneRecord.
+
+	var keys = map[models.RecordKey]*resourceRecord{}
+	var zrs []resourceRecord
+
+	for _, r := range rcs {
+		label := r.GetLabel()
+		if label == "@" {
+			label = ""
+		}
+		key := r.Key()
+
+		if zr, ok := keys[key]; !ok {
+			// Allocate a new ZoneRecord:
+			zr := resourceRecord{
+				Type:    r.Type,
+				TTL:     r.TTL,
+				Subname: label,
+				Records: []string{r.GetTargetCombined()},
+			}
+			zrs = append(zrs, zr)
+			//keys[key] = &zr   // This didn't work.
+			keys[key] = &zrs[len(zrs)-1] // This does work. I don't know why.
+
+		} else {
+			zr.Records = append(zr.Records, r.GetTargetCombined())
+
+			if r.TTL != zr.TTL {
+				printer.Warnf("All TTLs for a rrset (%v) must be the same. Using smaller of %v and %v.\n", key, r.TTL, zr.TTL)
+				if r.TTL < zr.TTL {
+					zr.TTL = r.TTL
+				}
+			}
+
+		}
+	}
+
+	return zrs
+}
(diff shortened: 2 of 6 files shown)


providers/desec/desecProvider.go
```diff
@@ -0,0 +1,284 @@
...
+							F: func() error {
+								err := client.createRR(n, domain)
+								if err != nil {
+									return err
+								}
+								return nil
+							},
+						})
+				}
+			}*/
```
Knight1: remove?

providers/desec/desecProvider.go
```diff
@@ -0,0 +1,284 @@
...
+	providers.RegisterDomainServiceProviderType("DESEC", NewDeSec, features)
+}
+
+// Specific stuff below
+// Specific stuff below
+// Specific stuff below
+// Specific stuff below
+// Specific stuff below
+// Specific stuff below
+// Specific stuff below
```
Knight1: remove?

Knight1:
README is not updated.

providers/desec/desecProvider.go
```diff
@@ -0,0 +1,223 @@
...
+	return c, nil
+}
+
+var features = providers.DocumentationNotes{
+	providers.DocDualHost:            providers.Unimplemented(),
+	providers.DocOfficiallySupported: providers.Cannot(),
+	providers.DocCreateDomains:       providers.Can(),
+	providers.CanUseAlias:            providers.Cannot(),
+	providers.CanUseSRV:              providers.Can(),
+	providers.CanUseSSHFP:            providers.Cannot(),
```
nils-wisiol, line 42: deSEC supports SSHFP

providers/desec/desecProvider.go
```diff
@@ -0,0 +1,223 @@
...
+	recordsToKeep := make([]*models.RecordConfig, 0, len(dc.Records))
+	for _, rec := range dc.Records {
+		if rec.Type == "ALIAS" {
+			// deSEC does not permit ALIAS records, just ignore it
+			printer.Warnf("deSEC does not support alias records\n")
+			continue
+		}
+		if rec.TTL < 3600 {
+			if rec.Type != "NS" {
+				printer.Warnf("deSEC does not support ttls < 3600. Setting ttl of %s type %s from %d to 3600\n", rec.GetLabelFQDN(), rec.Type, rec.TTL)
```
nils-wisiol, line 127: deSEC supports lower TTLs if you ask nicely. Maybe there can be a hint in this message to contact support@desec.io?

providers/desec/desecProvider.go
```diff
@@ -0,0 +1,223 @@
...
+	providers.DocOfficiallySupported: providers.Cannot(),
+	providers.DocCreateDomains:       providers.Can(),
+	providers.CanUseAlias:            providers.Cannot(),
+	providers.CanUseSRV:              providers.Can(),
+	providers.CanUseSSHFP:            providers.Cannot(),
+	providers.CanUseCAA:              providers.Can(),
+	providers.CanUseTLSA:             providers.Can(),
+	providers.CanUsePTR:              providers.Unimplemented(),
+	providers.CanGetZones:            providers.Can(),
+	providers.CanAutoDNSSEC:          providers.Cannot(),
```
nils-wisiol, line 47: deSEC automatically signs, but not sure what AutoDESEC exactly means here.

D3luxee: As far as i understood this CanAutoDNSSEC this would require deSEC to be the registrar not only the authoritive nameserver.

providers/desec/desecProvider.go
D3luxee: Okay, this was something i was not sure about. I will update this.

providers/desec/desecProvider.go
D3luxee: This was already removed

Knight1:
Can we implement login with username and password instead of a token the user has to grab from the API?

providers/desec/desecProvider.go
D3luxee: I could give a hint in the message, but for dnscontrol it would not make sense.
Maybe i could implement the validation of the ttl based on the response minimum_ttl, this would allow users to utilize this exception.

Could you allow lower ttls for the zone d3luxee.de so i can test this feature?

D3luxee:
> Can we implement login with username and password instead of a token the user has to grab from the API?
tldr; No:

I've tested this / wanted to implement it in this way, but there are some issues.
Performing the login and logout is no problem in terms of rest api calls, but i did not find a suitable place to call the logout function to invalidate the token after usage.
We would need some Close function in the DNSProvider interface, but this would affect all other providers.

There was no good way to implement this, so i've decided to stick with the api token which is even better in terms of security.

providers/desec/desecProvider.go
nils-wisiol: Maybe rephrasing the message to "your deSEC account does not support" can give the user a better hint of what's going on?

I did update the minimum ttl for d3luxee.de.

D3luxee: I did implement this check based on the minimum_ttl yesterday and adjusted the message.
The current "uncommited" message looks like this:
`WARNING: Please contact support@desec.io if you need ttls < 3600. Setting ttl of autodiscover.d3luxee.de type A from 60 to 3600`

Do you have any suggestion for the message?

nils-wisiol: lgtm, thanks!

peterthomassen:
Regarding login: 2FA is on one of our top priorities on our roadmap. If dnscontrol handles the login, it would then probably break for users who enabled 2FA.

providers/desec/desecProvider.go
D3luxee: I've tested and pushed the latest changes, dnscontrol will now validate and correct against the minimum_ttl of the specific domain instead of the hardcoded 3600 seconds.

This handles exceptions made by you or changes on the default minimum_ttl.

D3luxee:
> Regarding login: 2FA is on one of our top priorities on our roadmap. If dnscontrol handles the login, it would then probably break for users who enabled 2FA.

Thats true, we should stick to the api token instead of implementing the login. Is the management of api tokens via web interface already on the roadmap? Maybe you could also prioritize this a bit?

peterthomassen:
Yes, it is very much on the roadmap (even before 2FA). We expect to have token management in the GUI in the next few weeks.

providers/desec/desecProvider.go
tlimoncelli: AutoDNSSEC means that the DNS provider does all the DNSSEC for you (including holding your private key).

For the first release it is best to disable things if you are unsure about them, then enable them in future releases as they are tested.

providers/desec/protocol.go
```diff
@@ -0,0 +1,215 @@
...
+	printer.Printf("If you want to use DNSSec please add the DS record at your registrar using one of the keys:\n")
+	printer.Printf("%+q", dm.Keys)
+	return nil
+}
+
+//upsertRR will create or override the RRSet with the provided resource record.
+func (c *api) upsertRR(rr []resourceRecord, domain string) error {
+	endpoint := fmt.Sprintf("/domains/%s/rrsets/", domain)
+	byt, _ := json.Marshal(rr)
+	if _, err := c.post(endpoint, "PATCH", byt); err != nil {
```
tlimoncelli, line 129: Should this be `PATCH`?  I thought it should be `PUT` but I could be wrong. (Do they both work if the label doesn't previously exist?)

providers/desec/protocol.go
```diff
@@ -0,0 +1,215 @@
...
+		return []byte{}, err
+	}
+
+	bodyString, _ := ioutil.ReadAll(resp.Body)
+	// Got error from API ?
+	if resp.StatusCode > 299 {
+		if resp.StatusCode == 429 && retrycnt < 5 {
+			retrycnt++
+			time.Sleep(500 * time.Millisecond)
+			goto retry
```
tlimoncelli, line 164: Handling retries that that is awesome. I need to copy that for digitalocean, which very much needs it.

providers/desec/protocol.go
```diff
@@ -0,0 +1,215 @@
...
+		return bodyString, fmt.Errorf("http status %d %s, the api does not provide more information", resp.StatusCode, resp.Status)
+	}
+	return bodyString, nil
+}
+
+func (c *api) post(endpoint, method string, payload []byte) ([]byte, error) {
+	retrycnt := 0
+retry:
+	client := &http.Client{}
+	req, err := http.NewRequest(method, "https://desec.io/api/v1"+endpoint, bytes.NewReader(payload))
```
tlimoncelli, line 180: Please make `https://desec.io/api/v1` a constant since it is used in multiple places.

providers/desec/desecProvider.go
peterthomassen: deSEC does handle all DNSSEC aspects, including holding the private key. That's one of the reasons why the project exists. :-)

providers/desec/protocol.go
peterthomassen: You can also look at the `Retry-After:` response header, which has the number of seconds you'll have to wait.

providers/desec/protocol.go
D3luxee: Done

providers/desec/convert.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+		}
+		rcs = append(rcs, rc)
+	}
+
+	return rcs
+}
+
+func recordsToNative(rcs []*models.RecordConfig, origin string) []resourceRecord {
+	// Take a list of RecordConfig and return an equivalent list of ZoneRecords.
```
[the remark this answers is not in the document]
D3luxee: Yes, thats some leftovers i fixed the comments with the latest commit.

providers/desec/convert.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/StackExchange/dnscontrol/v3/models"
+	"github.com/StackExchange/dnscontrol/v3/pkg/printer"
+)
+
+// nativeToRecord takes a DNS record from deSEC and returns a native RecordConfig struct.
+func nativeToRecords(n resourceRecord, origin string) (rcs []*models.RecordConfig) {
+
+	// deSEC returns all the values for a given label/rtype pair in each
+	// livedns.DomainRecord.  In other words, if there are multiple A
```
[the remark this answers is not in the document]
D3luxee: Yes, thats some leftovers i fixed the comments with the latest commit.

providers/desec/protocol.go
D3luxee: Thanks.

providers/desec/protocol.go
D3luxee: Both work almost in the same way, PATCH allowed me to leave out the ttl for deletions. But its better to use PUT and define the full rrsets.

docs/_providers/desec.md
```diff
@@ -0,0 +1,36 @@
...
+This provider does not recognize any special metadata fields unique to deSEC.
+
+## Usage
+Example Javascript:
+
+{% highlight js %}
+var REG_NONE = NewRegistrar('none', 'NONE');    // No registrar.
+var deSEC = NewDnsProvider('desec', 'DESEC');  // deSEC
+
+D("example.tld", REG_NONE, DnsProvider(deSEC),
```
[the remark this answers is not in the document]
D3luxee: Done

providers/desec/desecProvider.go
tlimoncelli: Oh!  That's why its called deSEC!  I get it now!

D3luxee: I'm still not sure if we talk about the same definition of AutoDNSSEC.
deSEC does sign the zone and provide the DS Records to their users, but each user has to transmit this DS Records to their registrar so that they are added on tld level.
And deSEC does sign those zones per default, you do not need to explicitly request it!

For example i got my neuland.tech domain registered at get.tech directly, i can add this domain to deSEC and they sign the zone. But i will have to add the DS Records using the control panel of get.techt manually.

Is this the definition / meaning of this AutoDNSSEC of dnscontrol?

tlimoncelli: I don't use the feature myself (yet!) but I think that's how most providers implement it.

D3luxee: If someone creates the domain / zone via dns control he gets the DS stuff printed, thats how i implemented it in dnscontrol. But its nothing they have to ask for, its the default behavior.

D3luxee:
I've started using deSEC and dnscontrol for my domains now, in my opinion this is ready now.
Finally i got DNSSEC and DANE for my mailserver again: https://dane.sys4.de/smtp/mail.neuland.tech

I'm using A, AAAA, CNAME, MX, CAA, TLSA, TXT and SRV records, and everything works without any issues.

@tlimoncelli @peterthomassen @nils-wisiol do you have any suggestions?

tlimoncelli:
I'm ready to merge.  @peterthomassen @nils-wisiol ?

peterthomassen:
lgtm!

peterthomassen:
Hmm... I just realized that deSEC is missing from the feature table at https://stackexchange.github.io/dnscontrol/provider-list. I had assumed this was autogenerated from the feature flags set in the code (such as AutoDNSSEC etc.). Can I do anything to extend the list?

D3luxee:
I think this file is autogenerated but requires someone to run go generate, i did not notice that its missing in the matrix.

@tlimoncelli can you generate the file?

nils-wisiol:
Thanks @D3luxee @tlimoncelli !

tlimoncelli:
Hi folks!  There are some documentation updates needed. See https://github.com/StackExchange/dnscontrol/issues/730

Issue #259, opened 2017-11-01
AXFR provider


captncraig:
I think I might have found a better way to manage some types of servers. It should work for AD, bind, maybe some cloud stuff out of the box. The general idea is:

- Get zone data with an AXFR request. WE can parse it using pretty much the exact code the bind provider uses.
- Calculate updates.
- Make changes using dns dynamic updates (RFC2136). 

Assuming the acls are correct, that may work well across a variety of infrastructure scenarios. For AD, we would need to figure out how to do the weird kerberos thing for authentication, but it should be better than random powershell stuff, and can run on linux.

tlimoncelli:
I've noted this  in https://stackexchange.github.io/dnscontrol/provider-list

If someone starts working on this, please re-open this issue.

PR #729, opened 2020-05-01
New provider: AXFR+DDNS (#259)

hnrgrgr:
This is a tentative implementation for a provider based on AXFR requests and Dynamic DNS updates as suggested in #259.

This is Work-In-Progress---it does not pass all the integration test yet---and this is my first try at go programming, so this MR is mostly for feedbacks.

### DIFF
--- a/docs/_providers/axfrddns.md
+++ b/docs/_providers/axfrddns.md
@@ -7,7 +7,7 @@ jsId: AXFRDDNS
 # AXFR+DDNS Provider
 
 This provider is able to work with any authoritative DNS server
-accepting AXFR requests (RFC5936) and Dynamic Update (RFC2136).
+accepting AXFR requests (RFC5936) and Dynamic Updates (RFC2136).
 
 It has been tested with [BIND](https://www.isc.org/bind/),
 [Knot](https://www.knot-dns.cz/), and
@@ -35,7 +35,7 @@ The `transfer-key` will be used to authenticate AXFR request, and the
 `update-key` will be used to authenticate the Dynamic Updates. Both
 keys are optional, and you could provide only one them.
 
-If distinct zones requires distinct keys, you might instantiate the
+If distinct zones require distinct keys, you might instantiate the
 provider multiple times:
 
 {% highlight javascript %}
@@ -103,6 +103,60 @@ port might be used.
 
 When no nameserver appears in the zone, and no default nameservers nor
 custom master are configured, the AXFR+DDNS provider will fail.
+
+## Server configuration examples
+
+### Bind9
+
+Here is a sample `named.conf` example for an authauritative server on
+zone `example.tld`. It accepts AXFR requests from anyone on private
+network `172.17.0.0/16`. It only accepts updates which are both
+authenticated with key `update-key-id` and originated from the the same
+private network.
+
+{% highlight %}
+options {
+
+	listen-on { any; };
+	listen-on-v6 { any; };
+
+	allow-query { any; };
+	allow-notify { none; };
+	allow-recursion { none; };
+	allow-transfer { none; };
+	allow-update { none; };
+	allow-query-cache { none; };
+
+};
+
+zone "example.tld" {
+  type master;
+  file "/etc/bind/db.example.tld";
+  allow-transfer { example-transfer; };
+  allow-update { example-update; };
+};
+
+## Allow transfer to anyone on our private network
+
+acl example-transfer {
+    172.17.0.0/16;
+};
+
+## Allow update only from authenticated client on our private network
+
+acl example-update {
+  ! {
+   !172.17.0.0/16;
+   any;
+  };
+  key update-key-id;
+};
+
+key update-key-id {
+  algorithm HMAC-SHA256;
+  secret "AnotherSecret=";
+};
+{% endhighlight %}
 
 ## FYI: get-zones
 
--- a/docs/_providers/axfrddns.md
+++ b/docs/_providers/axfrddns.md
@@ -6,12 +6,12 @@ jsId: AXFRDDNS
 ---
 # AXFR+DDNS Provider
 
-This provider uses the native DNS protocols. It uses the Zone Transfer
-Protocol (AXFR) (RFC5936) to retrieve the existing records and Dynamic
-Update (RFC2136) to make corrections. It can use TSIG (RFC2845) or
-ACL-based (IP) authentication.
+This provider uses the native DNS protocols. It uses the AXFR (RFC5936,
+Zone Transfer Protocol) to retrieve the existing records and DDNS
+(RFC2136, Dynamic Update) to make corrections. It can use TSIG (RFC2845) or
+IP-based authentication (ACLs).
 
-It is able to work with any properly configured, standards compliant,
+It is able to work with any standards-compliant
 authoritative DNS server. It has been tested with
 [BIND](https://www.isc.org/bind/), [Knot](https://www.knot-dns.cz/),
 and [Yadifa](https://www.yadifa.eu/home).
@@ -42,8 +42,8 @@ authentication for that function. Including both key is the most
 secure option. Omitting both keys defaults to IP-based ACLs for all
 operations, which is the least secure option.
 
-If distinct zones require distinct keys, you might instantiate the
-provider multiple times:
+If distinct zones require distinct keys, you will need to instantiate the
+provider once for each key:
 
 {% highlight javascript %}
 var AXFRDDNS_A = NewDnsProvider('axfrddns-a', 'AXFRDDNS'}
--- a/docs/_providers/axfrddns.md
+++ b/docs/_providers/axfrddns.md
@@ -1,0 +1,107 @@
+---
+name: AXFRDDNS
+title: AXFR+DDNS Provider
+layout: default
+jsId: AXFRDDNS
+---
+# AXFR+DDNS Provider
+
+This provider is able to work with any authoritative DNS server accepting AXFR requests (RFC5936) and Dynamic Update (RFC2136).
+
+It has been tested with [BIND](https://www.isc.org/bind/), [Knot](https://www.knot-dns.cz/), and [Yadifa](https://www.yadifa.eu/home).
+
+## Configuration
+
+### Authentication
+
+The AXFR+DDNS provider might work without anything in `creds.json` if the primary master of the zone accepts transfers and updates without TSIG authentication. But for non widely-open server, the authentication keys should be provided in `creds.json`.
+
+{% highlight json %}
+{
+    "axfrddns": {
+        "transfer-key": "hmac-sha256:transfer-key-id:Base64EncodedSecret=",
+        "update-key": "hmac-sha256:update-key-id:AnotherSecret="
+    }
+}
+{% endhighlight %}
+
+The `transfer-key` will be used to authenticate AXFR request, and the `update-key` will be used to authenticate the Dynamic Updates. Both keys are optional, and you could provide only one them.
+
+If distinct zones requires distinct keys, you might instantiate the provider multiple times:
+
+{% highlight javascript %}
+var AXFRDDNS_A = NewDnsProvider('axfrddns-a', 'AXFRDDNS'}
+var AXFRDDNS_B = NewDnsProvider('axfrddns-b', 'AXFRDDNS'}
+{% endhighlight %}
+
+And update `creds.json` accordingly:
+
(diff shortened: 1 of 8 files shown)


providers/axfrddns/axfrddnsProvider.go
```diff
@@ -0,0 +1,328 @@
...
+// GetDomainCorrections returns a list of corrections to update a domain.
+func (c *AxfrDdns) GetDomainCorrections(dc *models.DomainConfig) ([]*models.Correction, error) {
+	dc.Punycode()
+
+	foundRecords, err := c.GetZoneRecords(dc.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(foundRecords) >= 1 && foundRecords[0].Type == "SOA" {
```
tlimoncelli, line 237: Are we sure that the SOA will always be the first record?  (I presume so, otherwise the "S" wouldn't meant "start" but... thought I'd check)

docs/_providers/axfrddns.md
```diff
@@ -0,0 +1,107 @@
...
+        "transfer-key": "hmac-sha256:transfer-key-id:Base64EncodedSecret=",
+        "update-key": "hmac-sha256:update-key-id:AnotherSecret="
+    },
+    "axfrddns-b": {
+        "transfer-key": "hmac-sha512:transfer-key-id-B:SmallSecret=",
+        "update-key": "hmac-sha512:update-key-id-B:YetAnotherSecret="
+    }
+}
+{% endhighlight %}
+
```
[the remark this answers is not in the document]
hnrgrgr: Sure. Would you like a `named.conf` that matches the example with distinct keys for distinct domains ? Or just a simple `named.conf` that accept authenticated tranfer and update ?

I can provide both but for the former, rather than a single server requiring different credentials, I think the use case is more likely when domains are hosted on distinct servers. Each servers being managed by distinct provider and hence the distinct credential.

providers/axfrddns/axfrddnsProvider.go
hnrgrgr: Yes, this is a "MUST". See section 2.2 of rfc5936.

But, if you prefer, I can add some protective programming. In case of a buggy servers don't follow the RFC.

hnrgrgr: I added a comment in the code.

providers/axfrddns/axfrddnsProvider.go
```diff
@@ -0,0 +1,328 @@
...
+
+// GetNameservers returns the nameservers for a domain.
+func (c *AxfrDdns) GetNameservers(domain string) ([]*models.Nameserver, error) {
+	return c.nameservers, nil
+}
+
+// GetZoneRecords gets the records of a zone and returns them in RecordConfig format.
+func (c *AxfrDdns) GetZoneRecords(domain string) (models.Records, error) {
+
+	transfer := new(dns.Transfer)
```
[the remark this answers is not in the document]
hnrgrgr: Done.

providers/axfrddns/axfrddnsProvider.go
```diff
@@ -0,0 +1,328 @@
+package axfrddns
+
+/*
+
+axfrddns -
+  Fetch the zone with AXFR request to a given primary master, and
+  push DynamicDNS updates to the same server.
+
+  Both the AXFR request and the updates might be authentified with a
```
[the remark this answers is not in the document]
hnrgrgr: Fixed.

docs/_providers/axfrddns.md
```diff
@@ -0,0 +1,107 @@
+---
+name: AXFRDDNS
+title: AXFR+DDNS Provider
+layout: default
+jsId: AXFRDDNS
+---
+# AXFR+DDNS Provider
+
+This provider is able to work with any authoritative DNS server accepting AXFR requests (RFC5936) and Dynamic Update (RFC2136).
```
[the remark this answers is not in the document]
hnrgrgr: Done.

hnrgrgr:
This code now pass all the integration test. When run against a 'bind' server, the tested domain must not be `example.com`.

Bind refuses to insert NS records when is no associated A record in the zone, and the testsuite insert some records for `example.com`. By using a distinct domain, the NS records points out of the zone and bind cannot check the absence of A record. 

The tested domain must neither be `example.io`. Servers should refuses to remove the last NS record of the zone and the testsuite likes to fully clear the zone. So, for the testsuite, the provider is configured with a default NS `ns.example.io`.

docs/_providers/axfrddns.md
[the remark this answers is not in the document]
tlimoncelli: I think the audience is a person that has run BIND for a while, who wants to add updates via DNSControl.  I think seeing what needs to be added to one domain is sufficient. They can figure out multiple domains and multiple servers if the understand the simple case very well.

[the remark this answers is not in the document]
hnrgrgr: Ok, thanks. Done

docs/_providers/axfrddns.md
```diff
@@ -0,0 +1,130 @@
...
+name: AXFRDDNS
+title: AXFR+DDNS Provider
+layout: default
+jsId: AXFRDDNS
+---
+# AXFR+DDNS Provider
+
+This provider is able to work with any authoritative DNS server
+accepting AXFR requests (RFC5936) and Dynamic Update (RFC2136).
+
```
tlimoncelli, line 11: I hope you don't mind some rewrites. The editor in me has some suggested edits for this doc.

Suggested replacement:

This provider uses the native DNS protocols.  It uses the
Zone Transfer Protocol (AXFR) (RFC5936) to retrieve
the existing records and  Dynamic Update (RFC2136) to make
corrections.

It is able to work with any properly configured, standards compliant,
authoritative DNS server.  It has been tested with
[BIND](https://www.isc.org/bind/),
[Knot](https://www.knot-dns.cz/), and
[Yadifa](https://www.yadifa.eu/home).

docs/_providers/axfrddns.md
```diff
@@ -0,0 +1,130 @@
...
+[Yadifa](https://www.yadifa.eu/home).
+
+## Configuration
+
+### Authentication
+
+The AXFR+DDNS provider might work without anything in `creds.json` if
+the primary master of the zone accepts transfers and updates without
+TSIG authentication. But for non widely-open server, the
+authentication keys should be provided in `creds.json`.
```
tlimoncelli, line 23: Suggested rewrite: (Warning: I'm not an expert in how AXFR and DDNS work. I took many guesses when writing this.  Please correct any assumptions.)

Authentication information is included in the `creds.json` entry for the provider:

* `transfer-key`: If this exists, the value is used to authenticate AXFR transfers.
* `update-key`: If this exists, the value is used to authenticate DDNS updates.

If either key is missing, DNSControl defaults to IP-based ACL authentication for that function.  Including both key is the most secure option.  Omitting both keys defaults to IP-based ACLs for all operations, which is the least secure option.

docs/_providers/axfrddns.md
```diff
@@ -0,0 +1,130 @@
...
+{% highlight json %}
+{
+    "axfrddns": {
+        "transfer-key": "hmac-sha256:transfer-key-id:Base64EncodedSecret=",
+        "update-key": "hmac-sha256:update-key-id:AnotherSecret="
+    }
+}
+{% endhighlight %}
+
+The `transfer-key` will be used to authenticate AXFR request, and the
```
tlimoncelli, line 34: This paragraph is no longer needed.

docs/_providers/axfrddns.md
```diff
@@ -0,0 +1,130 @@
...
+{% highlight json %}
+{
+   nameservers = "ns1.example.tld,ns2.example.tld,ns3.example.tld,ns4.example.tld"
+}
+{% endhighlight %}
+
+### Primary master
+
+By default, the AXFR+DDNS provider will send the AXFR requests and the
+updates to the first nameserver of the zone, usually known as the
```
tlimoncelli, line 90: This reads as AXFR (requests and updates) but I think you mean AXFR (requests) and DDNS (updates).  If that's true, add "TSIG" before the word "updates".

docs/_providers/axfrddns.md
```diff
@@ -0,0 +1,130 @@
...
+port might be used.
+
+{% highlight json %}
+{
+   master = "10.20.30.40:5353"
+}
+{% endhighlight %}
+
+When no nameserver appears in the zone, and no default nameservers nor
+custom master are configured, the AXFR+DDNS provider will fail.
```
tlimoncelli, line 105: Is there an error message that is output?  I'm thinking that if someone gets that error and searches for it, they should find this paragraph.

tlimoncelli:
I had some editing suggestions for the docs. 

Some things left to do (from `docs/writing-providers.md`)

* Edit [docs/provider-list.md](https://github.com/StackExchange/dnscontrol/blob/master/docs/provider-list.md): Add the provider to the provider list.
* Edit [OWNERS](https://github.com/StackExchange/dnscontrol/blob/master/OWNERS): Add the directory name and your github id.

I'm really excited to receive this PR!  Being able to use DDNS is going to be a big win.  It might even make it easier for people using Windows Active Directory DNS to use DNSControl!

docs/_providers/axfrddns.md
hnrgrgr: Done.

docs/_providers/axfrddns.md
hnrgrgr: Done, thanks.

docs/_providers/axfrddns.md
hnrgrgr: The error message was still a `TODO`. It is now written... and I copy-pasted here.

docs/_providers/axfrddns.md
hnrgrgr: Thanks for the suggested edits !

tlimoncelli:
I cleaned up some text and fixed the following golint errors.  You can see all my suggested changes in https://github.com/StackExchange/dnscontrol/tree/hnrgrgr-axfddns

This was the lint output:

```
$ golint *.go
axfrddnsProvider.go:33:2: don't use underscores in Go names; const dnssec_dummy_label should be dnssecDummyLabel
axfrddnsProvider.go:34:2: don't use underscores in Go names; const dnssec_dummy_txt should be dnssecDummyTxt
axfrddnsProvider.go:96:11: should omit 2nd value from range; this loop is equivalent to `for key := range ...`
axfrddnsProvider.go:120:26: error strings should not be capitalized or end with punctuation or a newline
axfrddnsProvider.go:133:26: error strings should not be capitalized or end with punctuation or a newline
axfrddnsProvider.go:137:26: error strings should not be capitalized or end with punctuation or a newline
axfrddnsProvider.go:142:6: exported type Param should have comment or be unexported
axfrddnsProvider.go:146:6: exported type Key should have comment or be unexported
axfrddnsProvider.go:152:6: exported type AxfrDdns should have comment or be unexported
```

FYI: This explains the warning about the error strings:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

hnrgrgr:
Thanks for the edits and the link, I am slowly building my way into this programming language.

I cherry-picked your patch and removed the WIP.

tlimoncelli:
Awesome!  You're learning Go very quickly. You look like an expert already!

I'm ready to merge this.  Is there anything else that needs to be done on your side?

hnrgrgr:
Thanks, I appreciate your words! It is always fun to discover a new language... and DNSControl looked like the good opportunity to do so. I did probably micmicked the programming style a lot!
And it is a feature I need.

The PR looks complete on my side.

docs/_providers/axfrddns.md
```diff
@@ -0,0 +1,196 @@
...
+{
+    "axfrddns": {
+        "transfer-key": "hmac-sha256:transfer-key-id:Base64EncodedSecret=",
+        "update-key": "hmac-sha256:update-key-id:AnotherSecret="
+    }
+}
+{% endhighlight %}
+
+If either key is missing, DNSControl defaults to IP-based ACL
+authentication for that function. Including both key is the most
```
liftedkilt, line 41: Minor typo - should be keys on Line 41

hnrgrgr: Fixed, thanks.

hnrgrgr:
I rebased the PR.

tlimoncelli:
Thanks for this contribution!

PR #733, opened 2020-05-03
Update docs to resolve  #730

### LINKED ISSUE #730: Updates for deSEC
CC: @D3luxee @nils-wisiol @Knight1 @peterthomassen

Hi folks!  When I approved PR https://github.com/StackExchange/dnscontrol/pull/725 I forgot to check the items listed in `docs/writing-providers.md`. These should be easy to fix.

1. In `docs/writing-providers.md` near "Pick a similar provider as your base".  Please add this provider to the list for `incremental-label` or `incremental-label-type` as appropriate.  I think it is the former.

2. In `docs/provider-list.md` in `Providers with "contributor support"` please add DESEC and the maintainer's name.

3. Add the maintainer's name to `OWNERS`

Thanks!

D3luxee:
This will resolve the issues mentioned in  #730

### DIFF
--- a/docs/writing-providers.md
+++ b/docs/writing-providers.md
@@ -97,7 +97,7 @@ into three general categories:
   multiple records on a label (for example, example www.example.com
 has A and MX records), you have to replace all the records at that
 label. (GANDI_V5)
-* **incremental-label-type:** Like incremental-record, but updates to any records at a label have to be done by type.  For example, if a label (www.example.com) has many A and MX records, even the smallest change to one of the A records requires replacing all the A records. Any changes to the MX records requires replacing all the MX records.  If an A record is converted to a CNAME, one must remove all the A records in one call, and add the CNAME record with another call.  This is deceptively difficult to get right; if you have the voice between incremental-label-type and incremental-label, pick incremental-label.
+* **incremental-label-type:** Like incremental-record, but updates to any records at a label have to be done by type.  For example, if a label (www.example.com) has many A and MX records, even the smallest change to one of the A records requires replacing all the A records. Any changes to the MX records requires replacing all the MX records.  If an A record is converted to a CNAME, one must remove all the A records in one call, and add the CNAME record with another call.  This is deceptively difficult to get right; if you have the voice between incremental-label-type and incremental-label, pick incremental-label. (DESEC)
 
 TODO: Categorize NAMECHEAP
 
--- a/OWNERS
+++ b/OWNERS
@@ -3,6 +3,7 @@ providers/azuredns @vatsalyagoel
 providers/bind @tlimoncelli
 # providers/cloudflare
 providers/cloudns @pragmaton
+providers/desec @D3luxee
 providers/digitalocean @Deraen
 providers/dnsimple @aeden
 providers/gandi_v5 @TomOnTime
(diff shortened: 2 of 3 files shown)


docs/writing-providers.md
```diff
@@ -97,7 +97,7 @@ into three general categories:
...
+* **incremental-label-type:** Like incremental-record, but updates to any records at a label have to be done by type.  For example, if a label (www.example.com) has many A and MX records, even the smallest change to one of the A records requires replacing all the A records. Any changes to the MX records requires replacing all the MX records.  If an A record is converted to a CNAME, one must remove all the A records in one call, and add the CNAME record with another call.  This is deceptively difficult to get right; if you have the voice between incremental-label-type and incremental-label, pick incremental-label. (DESEC)
```
peterthomassen, line 100: > If an A record is converted to a CNAME, one must remove all the A records in one call, and add the CNAME record with another call.

That's not true for deSEC. You can make a [bulk request](https://desec.readthedocs.io/en/latest/dns/rrsets.html#bulk-operations) which removes `A` and sets `CNAME` atomically in the same operator.

D3luxee: Sure, but the other one isn't true either for deSEC.

> Like incremental-record, but if there are
  multiple records on a label (for example, example www.example.com
has A and MX records), you have to replace all the records at that
label. (GANDI_V5)

This document should help other developers to find similar implementations.
My implementation is based on label:type api calls, sure i merge multiple changes in one request. But `incremental-label-type` is the closest one to what i'm doing in my implementation.

peterthomassen: Sure, I didn't mean to criticize the categorization, and I fully agree!

I meant that maybe the description of the category can be changed. For example, the sentence I quoted above could be removed (because it does not apply to all providers in the category), and the next be adjusted. So instead of:

> If an A record is converted to a CNAME, one must remove all the A records in one call, and add the CNAME record with another call.  This is deceptively difficult to get right

... we would just have:

> This may be deceptively difficult to get right

Also, I'm not sure what the last sentence really means: I thought the dnscontrol provider implementations take care of these intricacies, so that from a user perspective, it doesn't matter at all. Why would the user then care about whether `incremental-label-type` is complicated internally? -- But I may very well be biased here, so perhaps I should leave that to others and just shut up. :-)

D3luxee: > I thought the dnscontrol provider implementations take care of these intricacies, so that from a user perspective, it doesn't matter at all. Why would the user then care about whether incremental-label-type is complicated internally?

This document (writing-providers.md) is meant for developers which want to develop another provider, so they have to take care about how complicated this might be.

> I meant that maybe the description of the category can be changed. For example, the sentence I quoted above could be removed (because it does not apply to all providers in the category), and the next be adjusted.

I did not want to change the existing descriptions, but maybe @tlimoncelli can say something about this.

peterthomassen: > This document (writing-providers.md) is meant for developers which want to develop another provider, so they have to take care about how complicated this might be.

Sorry! I was under the impression that this is going into https://stackexchange.github.io/dnscontrol/provider-list. Obviously, I'll take it back then!

tlimoncelli:
Thanks!

Issue #736, opened 2020-05-04
octodns: constant 4294967295 overflows int


decke:
I get a compile error on FreeBSD 12.1/i386 with latest Go 1.14:

providers/octodns/octoyaml/read.go:266:17: constant 4294967295 overflows int

Thanks!

tlimoncelli:
That's weird.  I do have access to a FreeBSD 11.x machine.  With go 1.13.4 it compiles fine. With go 1.14.1 the compiler outputs a lot of lines like:

```
/usr/local/go/src/runtime/internal/atomic/atomic_amd64x.go:18:6: Load redeclared in this block
	previous declaration at /usr/local/go/src/runtime/internal/atomic/atomic_amd64.go:16:24
```

I don't have access to a FreeBSD 12.x host at the moment.  Here are my suggestions:

1. Remove the provider.

I'm going to assume you aren't using the octodns provider (because I don't think anyone actually uses it ;-) ).  

`rm -rf providers/octodns` and remove the octodns line from `providers/_all/all.go`

2. Fix the code:

```
        case int:
                i := ttl.(int)
                if i < 0 || i > math.MaxUint32 {
                        return 0, fmt.Errorf("ttl won't fit in 32-bits (%d)", i)
                }
                return uint32(i), nil
```

The `if` doesn't make sense.  int is a signed integer type that is at least 32 bits in size.  Try this

```
                if i < 0 {
```

If that compiles, would you be willing to send a PR?

Thanks!
Tom

PR #738, opened 2020-05-07
Fix octodns: constant 4294967295 overflows int (Issue #736)

decke:
This fixes Issue #736 and was build tested with Go 1.14.2 on

- FreeBSD 12.1/i386
- FreeBSD 12.1/amd64
- FreeBSD 11.3/i386
- FreeBSD 11.3/amd64

https://ports.bluelife.at/builds/20200507-09:37:34.38360/


tlimoncelli:
Thanks!

PR #739, opened 2020-05-13
Update DNSimple Provider to api client 0.62

onlyhavecans:
- Adds Context support
- Updates supported integration tests
- Catches a previously discarded error
- Adds a UserAgent


tlimoncelli:
Looks great!  Thanks for bringing this provider up to date!

PR #748, opened 2020-05-28
New DNS provider PowerDNS

jpbede:
Added a new DNS provider: PowerDNS


providers/powerdns/powerdnsProvider.go
```diff
@@ -0,0 +1,269 @@
...
+		if len(messages) > 1 && i != len(messages)-1 {
+			msg += singleMsg + "\n    "
+		} else {
+			msg += singleMsg
+		}
+	}
+	return msg
+}
+
+// ensureDotSuffix ensures that the trailing dot is there
```
tlimoncelli, line 263: We might no longer need this function.

When we first wrote dnscontrol there was a lot of places where we stripped or added a `.` "just in case".  In all those places, it turned out we were adding complexity that wasn't needed.  The code is now very careful (for example) to make sure that:

* label.NameFQDN never ends in a "."
* label.GetTarget() always ends with a "." (unless it is an IP address, etc.)

That should simplify your code.

providers/powerdns/export_test.go
```diff
@@ -0,0 +1,5 @@
+package powerdns
```
tlimoncelli, line 1: I'm not clear about the function of this file.

providers/powerdns/powerdnsProvider.go
```diff
@@ -0,0 +1,269 @@
...
+	dc.Punycode()
+	models.PostProcessRecords(curRecords)
+
+	// create record diff by group
+	keysToUpdate := (diff.New(dc)).ChangedGroups(curRecords)
+	desiredRecords := dc.Records.GroupedByKey()
+
+	// create corrections by group
+	for label, msgs := range keysToUpdate {
+		labelName := ensureDotSuffix(label.NameFQDN)
```
tlimoncelli, line 158: This can simply be: `labelName := label.NameFQDN + "."`

You are guaranteed  NameFQDN will never end in a `.`

providers/powerdns/powerdnsProvider.go
```diff
@@ -0,0 +1,269 @@
...
+		TTL:      uint32(ttl),
+		Original: r,
+		Type:     rtype,
+	}
+	rc.SetLabel(name, domain)
+
+	content := r.Content
+	switch rtype {
+	case "CNAME", "NS":
+		return rc, rc.SetTarget(dnsutil.AddOrigin(ensureDotSuffix(content), domain))
```
tlimoncelli, line 232: I'm not sure I understand the logic here:

* The code adds a dot, then calls AddOrigin, which is a no-op if the string ends in a dot.

    rc.SetTarget(ensureDotSuffix(content))

Now that it is simplified, we might be able to go even more simple.

Does the API return content with a trailing dot sometimes but not others?  Old dnscontrol code used to be paranoid and handle everything both ways.  Eventually we found the complexity wasn't worth it.  APIs would never change such a fundamental thing. 

Unless you have proof the API sometimes returns with or without a trailing dot, this code could just be:

		return rc, rc.SetTarget(content + ".")

That said... test this.  I could always be wrong :-)

providers/powerdns/powerdnsProvider.go
```diff
@@ -0,0 +1,269 @@
...
+	case "CAA":
+		return rc, rc.SetTargetCAAString(content)
+	case "MX":
+		return rc, rc.SetTargetMXString(ensureDotSuffix(content))
```
tlimoncelli, line 236: Same logic as above.

    return rc, rc.SetTargetMXString(content + ".")

tlimoncelli:
Looking good!  I have some minor comments.  I may have more tomorrow.

providers/powerdns/powerdnsProvider.go
```diff
@@ -0,0 +1,269 @@
...
+	providers.DocCreateDomains:       providers.Can(),
+	providers.DocOfficiallySupported: providers.Cannot(),
+	providers.CanGetZones:            providers.Can(),
+}
+
+func init() {
+	providers.RegisterDomainServiceProviderType("POWERDNS", NewProvider, features)
+}
+
+// Provider represents the PowerDNS DNSServiceProvider.
```
tlimoncelli, line 34: go-lint says:

```
providers/powerdns/powerdnsProvider.go:34:1: comment on exported type PowerDNS should be of the form "PowerDNS ..." (with optional leading article)
```

providers/powerdns/export_test.go
jpbede: This was a proxy file for non exported functions used for testing. With this file i've exported the functions for the test file

providers/powerdns/powerdnsProvider.go
jpbede: I've tested it with a other version of PDNS, both returned it with a dot in the end. So i removed it.

jpbede:
Thanks for your feedback. I've implemented it now.

tlimoncelli:
Awesome!  Thanks for contributing a new provider!  This is very exciting!

PR #749, opened 2020-05-29
Added slack notifications

jpbede:
Added `slack` as a new notification connector

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -21,7 +21,7 @@ Notifications are set up in your credentials json file. They will use the `notif
         ...
   } ,
   "notifications":{
-      "bonfire_url": "https://chat.meta.stackexchange.com/feeds/rooms/123?key=xyz"
+      "slack_url": "https://hooks.slack.com/services/T00000000/B00000000/XXXXXXXXXXXXXXXXXXXXXXXX"
   }
 ```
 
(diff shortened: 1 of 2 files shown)


tlimoncelli:
This is awesome!  The code looks good. Could you please update docs/notifications.md too?

Thanks!

jpbede:
Sorry, forgotten that. This has been added.

pkg/notifications/slack.go
```diff
@@ -0,0 +1,52 @@
...
+		content = fmt.Sprintf(`**Preview: %s[%s] -** %s`, domain, provider, msg)
+	} else if err != nil {
+		content = fmt.Sprintf(`**ERROR running correction on %s[%s] -** (%s) Error: %s`, domain, provider, msg, err)
+	} else {
+		content = fmt.Sprintf(`Successfully ran correction for **%s[%s]** - %s`, domain, provider, msg)
+	}
+	s.messages = append(s.messages, content)
+}
+
+func (s *slackNotifier) Done() {
```
tlimoncelli, line 41: I noticed that you are collecting all the messages and then sending them at the end. That's fine.  However I was wondering if you tried sending them one at a time just to see how Slack displays them.

The reason I ask is that at Stackoverflow we once prevented an outage because we saw the notifications one-by-one.  The first few were deleting EVERYTHING.  Someone went into our CI/CD system and killed the job.  If we didn't have any notifications until the very end, nearly every DNS record related to stackoverflow.com would have been deleted.

jpbede: Good point, haven't thought about that case. Slack and also Mattermost should display them the same way as if i send them as a batch.

I will check that and change it to your suggestion.

docs/notifications.md
```diff
@@ -35,13 +35,19 @@ This is stack overflow's built in chat system. This is probably not useful for m
 
 Configure `bonfire_url` to be the full url including room and api key.
 
+### Slack/Mattermost
```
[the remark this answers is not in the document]
jpbede: Yep, good suggestion. I'm going to change it.

jpbede:
Added the suggestions.

tlimoncelli:
Looks great!  I can't wait to set this up in our environment!

PR #758, opened 2020-06-03
get-certs.md: correct flag names

ad8-bdl:


tlimoncelli:
Thank you for catching this!

Could you also add:

   --verbose           Enable detailed logging (deprecated: use the global -v flag) (default: false)
   --vault             Store certificates as secrets in hashicorp vault instead of on disk. (default: false)
   --vaultPath value   Path in vault to store certificates (default: "/secret/certs")
   --notify            set to true to send notifications to configured destinations (default: false)
   --only value        Only check a single cert. Provide cert name.

ad8-bdl:
done!

tlimoncelli:
Thanks!!

PR #760, opened 2020-06-09
Add support to GCLOUD provider for DS records

mhuxtable:
The gcloud provider seems to behave correctly with DS resource records,
and is able to read and write them properly. I've enabled this in the
provider options.

Please let me know of any special tests or actions you'd like me to take to
enable this to be merged. Thanks.


tlimoncelli:
Thanks for checking into that!  (and thanks for fixing the typo!)

PR #767, opened 2020-06-18
Linting

tlimoncelli:


SuperSandro2000:
Why did the vendor directory get removed? Is this an accident?

tlimoncelli:
We moved to "go modules" which doesn't require a vendor directory.  Google runs a proxy that caches everything indefinitely.

Is the lack of vendoring causing an issue for you?

SuperSandro2000:
It made it harder for me to update the nixos package to 3.3.0. https://github.com/NixOS/nixpkgs/pull/97394

The newer go.mod wasn't correctly parsed by vgo2nix so I also needed to fix some bugs in their first. At this point it does not matter to much for me anymore if you include the vendor again or not.

PR #780, opened 2020-07-07
go get -u gopkg.in/ns1/ns1-go.v2

tlimoncelli:
Hey @MaxRatmeyer!

I noticed that the external module for NS1 was using was very old.  Could you see if the integration tests work any better with this update?


MaxRatmeyer:
It looks like every test is failing under the error `400 Input validation failed (Value None for field '<obj>.filters' is not of type array)`

https://pastebin.com/8Uehkm4r

tlimoncelli:
Ah, it looks like the new module has a lot of breaking changes. Well, it was worth a  try!

Issue #775, opened 2020-07-05
NS1: Add support for ALIAS, PTR, and NAPTR


mratmeyer:
Hello,

According to the NS1 API Docs, it looks like they have support for ALIAS, PTR, and NAPTR records, which aren't currently available in DNSControl. Would it be possible to get these activated in DNSControl?

[https://ns1.com/api#putcreate-a-dns-record](url)

Thanks

tlimoncelli:
Seems reasonable!  This might be a good starter project if you are interested in diving into the code.  PTR and NAPTR might "just work" depending on how the code was written.  We may just never have tested it.

I don't have a test account with NS1, so I can't do this myseslf.

I'd be glad to walk you through the process.  The process is:

1.  Make sure the integration tests work as-is

```
export NS1_TOKEN="insert-your-token-here"
export NS1_DOMAIN="testdomain.tld"
cd integrationTest
go test -v -verbose -provider NS1
```

This just makes sure everything is working.

2.  Enable the features.

In `providers/ns1/ns1provider.go` look at the `var docNotes = providers.DocumentationNotes{` list and add the features to it.  Look at `providers/bind/bindProvider.go` for examples.

3. Re-run the integration tests.

My guess is that PTR and NAPTR might "just work".   ALIAS might require some more effort.


CC: @captncraig

MaxRatmeyer:
Hi, thanks so much for the instructions. I cloned the repo and ran the integration test, but without making any changes, I got this error.
`=== RUN   TestDualProviders
    TestDualProviders: integration_test.go:272: Clearing everything
    TestDualProviders: integration_test.go:278: Adding nameservers from another provider
    TestDualProviders: integration_test.go:266: #1: CREATE NS maxnet.work ns1.example.com. ttl=300
        CREATE NS maxnet.work ns2.example.com. ttl=300
    TestDualProviders: integration_test.go:281: Running again to ensure stability
--- PASS: TestDualProviders (0.80s)
FAIL
exit status 1
FAIL	github.com/StackExchange/dnscontrol/v3/integrationTest	25.090s`
I tried it with another domain in a different TLD as well and it ran into the same issue. It didn't seem like there were any problems in the test, it just abruptly cut off at the end.

tlimoncelli:
[...]

PR #782, opened 2020-07-10
NS1: Revert enabling TXTMulti

MaxRatmeyer:
#775 Reverted support for TXTMulti due to NS1 automatically converting TXTMulti records back to TXT


tlimoncelli:
Thanks for cleaning that up!

Issue #779, opened 2020-07-07
NS1 is failing integration tests


tlimoncelli:
Here's what it looks like https://pastebin.com/dDZx50tn

FYI: I don't have a test account with NS1, so I need someone to step up and look into these. I'd be glad to assist remotely!

MaxRatmeyer:
Hi,

I did a little more tinkering and I was able to fix all the issues with NS1 and it's now passing! I think it was just a problem with how the default record builder worked with MX records, so I added a special case in the NS1 provider file in the buildRecord function, similar to how the SRV record was built. 

Here is the test: https://pastebin.com/uT5tFqEw

I'll submit a PR in a bit!

tlimoncelli:
Can we close this issue?

Tom

P.S.  Thank you for stepping up and fixing this issue.  I'm really impressed by your debugging skills!

MaxRatmeyer:
Thank you! Yep, we're good to close it.

PR #783, opened 2020-07-14
Fixed MX records on NS1

MaxRatmeyer:
#779 Fix for MX records on NS1. NS1 is passing all integration tests now!

Integration test: https://pastebin.com/M4Qn0CKK


tlimoncelli:
Awesome!!

PR #786, opened 2020-07-17
Fixing PowerDNS domain creation

jpbede:
I did a mistake when checking if the domain is there or not.

The go powerdns api is raising a error when the domain is not there. I got a nil pointer exception cause the zone isn't there.
Now it is checking the returned error.


tlimoncelli:
Thanks for spotting this and supplying a fix!  W00t!!

Issue #446, opened 2019-02-02
CloudFlare SPF record


srfraser:
In a similar manner to #339, CloudFlare returns an SPF record, which crashes dnscontrol.

```******************** Domain: mydomain.org
----- Getting nameservers from: cloudflare
----- DNS Provider: cloudflare...panic: unparsable record received from cloudflare: Unknown rtype (SPF) when parsing (v=spf1 include:_spf.protonmail.ch mx ~all) domain=(mydomain.org)

goroutine 1 [running]:
github.com/StackExchange/dnscontrol/providers/cloudflare.(*cfRecord).nativeToRecord(0xc4205e4480, 0xc42060ecb0, 0xb, 0xc42061d360)
	/Users/tlimoncelli/src/github.com/StackExchange/dnscontrol/providers/cloudflare/cloudflareProvider.go:419 +0x504
github.com/StackExchange/dnscontrol/providers/cloudflare.(*CloudflareApi).getRecordsForDomain(0xc4202c1a70, 0xc420614220, 0x20, 0xc42060ecb0, 0xb, 0xc42054b601, 0x106d92c, 0x20ca950, 0xc42065a080, 0xc4200d4008)
	/Users/tlimoncelli/src/github.com/StackExchange/dnscontrol/providers/cloudflare/rest.go:70 +0x177
github.com/StackExchange/dnscontrol/providers/cloudflare.(*CloudflareApi).GetDomainCorrections(0xc4202c1a70, 0xc42030a140, 0xa, 0xc42030a100, 0x1, 0x0, 0x0)
	/Users/tlimoncelli/src/github.com/StackExchange/dnscontrol/providers/cloudflare/cloudflareProvider.go:106 +0x117
github.com/StackExchange/dnscontrol/commands.run(0x1ad4140, 0xc, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1ad2519, 0xa, 0x0, ...)
	/Users/tlimoncelli/src/github.com/StackExchange/dnscontrol/commands/previewPush.go:129 +0x3ce
github.com/StackExchange/dnscontrol/commands.Preview(0x1ad4140, 0xc, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1ad2519, 0xa, 0x0, ...)
	/Users/tlimoncelli/src/github.com/StackExchange/dnscontrol/commands/previewPush.go:81 +0x74
[...]

PR #787, opened 2020-07-17
Route53 uses a custom record type for SPF

TechplexEngineer:
Much like OVH Route53 when queried returns SPF records as their own
type. This small change transcribes the SPF type to TXT

See: https://github.com/StackExchange/dnscontrol/issues/446
See: https://github.com/StackExchange/dnscontrol/blob/644ba70e87bf21997d9d2f020216f90b3876359d/providers/ovh/ovhProvider.go#L169-L172


tlimoncelli:
Thanks for this!

TechplexEngineer:
Thanks for the quick merge!

PR #789, opened 2020-07-23
Build & attach release assets automatically

MaxHorstmann:
A workflow which automatically builds and attaches binaries to a release when it's published.

image


MaxHorstmann:
@tlimoncelli let's chat about this when you're back!

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Checkout repo
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15.3
```
atma-stackoverflow, line 26: For this, I think "at least 1.15" is sufficient; i.e. `^1.15`.  If .3 point release is required to build then what you have is appropriate.  If you want it locked to EXACTLY 1.15.3, it'd be `1.15.3`

MaxHorstmann: Agreed! https://github.com/StackExchange/dnscontrol/commit/bf7cb5bb03117a5d879383f4eab29564dc8c0f1a

MaxHorstmann:
Tested this one more time on master branch, works as expected:

image

[Release Docs](https://github.com/StackExchange/dnscontrol/blob/master/docs/release-engineering.md#step-6-publish-the-release) are updated to reflect the change.

@tlimoncelli @atma-stackoverflow

PR #794, opened 2020-07-30
SPF Optimizer: Enable the use of TXTMulti records to support longer SPF records

Crazybus:
I'm opening this up now to get a bit of feedback now that I have it in a
working state. I wanted to hold off polishing it up and writing
documentation until I heard what you folks thought of the
implementation.

Notes:

* This implements [RFC 4408][rfc] for the SPF optimizer. Allowing for
more SPF records to fit within the 10 lookups by using multiple strings.
* By default the max size of the TXT remains at 255. Meaning users will
still only get a single 255 length string unless they modify `txtLength`
(this needs a better name) and opt into this feature.
* The general recommendation when using multiple strings for TXT records
is to keep the size within a single UDP packet. It seems like the
maximum size for this depends on a bunch of factors that are sometimes
outside of your control. A similar tool has a [formula for estimating the
maximum allowed size][formula]. However I felt giving a user
configurable size would fit with the current configuration style that
dnscontrol has. Similar to how dnscontrol recommends only flattening a
record if absolutely needed, I can see this length being increased by
only enough to get you within 10 lookups.

Specific things I'm looking for feedback on:

* Is this the right place to convert the string into an array? It is
certainly the most convenient place available.
* Got any ideas for a better name instead of `txtLength`?
* Thoughts on using a struct for these tests? I wasn't easily able to
add in an extra paramter to the other tests without it getting messy. I
left the original tests untouched to make these changes easier to
remove but both tests could be potentially combined using a struct with
the parameters.

[rfc]: https://tools.ietf.org/html/rfc4408#section-3.1.3
[formula]: https://github.com/oasys/mkspf/blob/master/Overhead.md


tlimoncelli:
Wow, this is really a joy to read.  I was too lazy to do all that research, which is why the current implementation is so no-frills.  I appreciate you digging into this.

> Is this the right place to convert the string into an array? It is certainly the most convenient place available.

No complaints here!

> Got any ideas for a better name instead of txtLength?

txtMaxSize?  txtSize?   Or maybe it should be a bool such as `UseTXTMulti`?

> Thoughts on using a struct for these tests? I wasn't easily able to
> add in an extra paramter to the other tests without it getting messy. I
> left the original tests untouched to make these changes easier to
> remove but both tests could be potentially combined using a struct with
> the parameters.

Ideally the test input would be the output of SPF_BUILDER() and the test would see if the specific `TXT` records we expect get generated.  That way we are testing the system in a self-contained way.

`flattenSPFs()` isn't designed well for testing.  It calls `rec.Flatten()` but then, if splitting is enabled, it calls `rec.TXTSplit()`.  Each could be tested separately, but I think it would be better if there was a single testable function that does all the work. That would require a little refactoring but would be worth it.

Tom

Crazybus:
> Wow, this is really a joy to read. I was too lazy to do all that research, which is why the current implementation is so no-frills. I appreciate you digging into this.

Thanks :) And thanks so much for the speedy response! It sounds like I'm certainly on the right track so I'll start working on the finishing touches to get this ready for a proper review.

> txtMaxSize? txtSize? 

I really like the sound of `txtMaxSize`. I think `max` is a really important keyword to include since this indicates it being an upper limit to the size of each TXT record. Which is exactly what it is, my original `txtLength` doesn't really describe it as being some kind of upper limit. 

> Or maybe it should be a bool such as UseTXTMulti?

Maybe it could even be a combination? So it would look something like:

default: `UseTXTMulti: false`, `txtMaxSize: 255` (the current defaults)

`UseTXTMulti: true` sets the default `txtMaxSize` to 450 which seems to be a generally accepted "safe" max. With the user then being able to also set `txtMaxSize` to something lower or higher if they choose.

If I had to pick I would vote to keep just `txtMaxSize` and leave out the boolean. One thing that makes me hesitate with the boolean is that even the RFC is only comfortable saying that 450 characters "should" be fine. If there was a documented safe number then I would be totally on board with making it a boolean.

It also feels like the fact that multiple strings are enabled is just an implementation detail from a user point of view. If a user has a provider that doesn't support multiple strings they just won't be able to go past the default 255 limit. Which seems like something that just can be included as a note in the documentation

Thoughts? Seems like both options have their merits and I'm totally happy to go either way.

> Ideally the test input would be the output of SPF_BUILDER() and the test would see if the specific TXT records we expect get generated. That way we are testing the system in a self-contained way.
> flattenSPFs() isn't designed well for testing. It calls rec.Flatten() but then, if splitting is enabled, it calls rec.TXTSplit(). Each could be tested separately, but I think it would be better if there was a single testable function that does all the work. That would require a little refactoring but would be worth it.

Thanks for the tips! This sounds like it would be a really good improvement. There are a few other optimizations that I was considering implementing. So I might refactor the tests while doing that. For now my immediate goal is to get under the 10 DNS lookups :P

tlimoncelli:
`txtMaxSize` feels right. No need for a boolean.  Too much flexibility just creates confused users.  `txtMaxSize == 0` can mean "off" and doesn't require a second variable.  Don't be afraid to be opinionated: A little configurability (being able to turn a feature off) is a good.  Too much configurability is confusing.  What people actually want is something in the middle: its comforting to be told "enable this and we'll do the right thing".

I just realized that if you have 2 DNS providers and 1 doesn't support TXTMulti, you'll have to disable that feature for both providers (since you might get one TXT from one provider and the other TXT from other).  So, consider that in your code.

Lastly... refactoring to make things more testable is always a good thing.  My philosophy is that tests are good not just because it assures that today's code work: good test coverage makes it easier for future contributors to contribute without fear of breaking anything.

Crazybus:
Ok, this is ready for a proper review now :) PTAL!

> Too much flexibility just creates confused users. txtMaxSize == 0 can mean "off" and doesn't require a second variable. Don't be afraid to be opinionated: A little configurability (being able to turn a feature off) is a good. Too much configurability is confusing. What people actually want is something in the middle: its comforting to be told "enable this and we'll do the right thing".

I ended up sticking with the original logic since it "felt right". So the default will stick with 255 as the limit resulting in single TXT strings which is how things currently work. If someone increases this they can now set it to a recommended 450, or something even higher if they don't care about fitting into a single UDP packet. Let me know if you think this is too much configurability or not. Or if you have a better idea for how to handle it.

> I just realized that if you have 2 DNS providers and 1 doesn't support TXTMulti, you'll have to disable that feature for both providers (since you might get one TXT from one provider and the other TXT from other). So, consider that in your code.

Could you go into a bit more detail about this use case? And potentially how it would have to be implemented in the code? If I had to guess I would expect that this would mean adding some kind of `if !CanUseTXTMulti; txtMaxSize = 255` to forcefully disable it when the limit has been set higher for a provider that cannot do multiple TXT strings.

I poked around a bit and it seems like this might need a pretty big refactor to support. Since it seems like flattening of SPF records is handled at a global level here:

https://github.com/StackExchange/dnscontrol/blob/00b03de906b6d9c04203f3f542bfa83484d7c24f/pkg/normalize/validate.go#L364-L367

And that currently dnscontrol seems to just say "one provider doesn't support this so you can't do it" like with the handling of multiple TXT strings here:

https://github.com/StackExchange/dnscontrol/blob/00b03de906b6d9c04203f3f542bfa83484d7c24f/pkg/normalize/validate.go#L351-L357

https://github.com/StackExchange/dnscontrol/blob/00b03de906b6d9c04203f3f542bfa83484d7c24f/pkg/normalize/validate.go#L396-L400

I only looked a the code base for the first time yesterday, so maybe there is an easy way to handle this so I'm all ears!

My understanding is that a user would only enable this if they don't fit within the 10 DNS lookups. If they were only able to enable this for 1 of their 2 DNS providers then half of the requests would still be over that limit. I guess the argument could be made that only failing half of your SPF checks might be better than failing 100% of them though.

tlimoncelli:
As far as the multiple DNS provider situation: Now that I think about it, this situation would be very rare.  So rare, that I'm happy with the system just failing with a useful error message.  All the major DNS providers support TXTMulti.  If you want to have 2 dns providers and you are using one of the small players, then you aren't serious about reliability and I'm not going to go to great lengths to fix the problem (i.e. the existing error message is fine).

pkg/spflib/flatten_test.go
```diff
@@ -107,17 +109,91 @@ func TestSplit(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-			res := rec.TXTSplit("_spf%d.stackex.com", overhead)
-			if res["@"] != tst[2] {
-				t.Fatalf("Root record wrong. \nExp %s\ngot %s", tst[2], res["@"])
+			res := rec.TXTSplit("_spf%d.stackex.com", overhead, txtMaxSize)
```
tlimoncelli, line 112: Dumb question but...

I see that on my test domain the result was a TXT record like:

> _spf1      300   IN TXT   "v=spf1 ip4:209.85.128.0/17 ip4:216.58.192.0/19 ip4:216.239.32.0/19 ip6:2001:4860:4000::/36 ip6:2404:6800:4000::/36 ip6:2607:f8b0:4000::/36 ip6:2800:3f0:4000::/36 ip6:2a00:1450:4000::/36 ip6:2c0f:fb50:4000::/36 ip4:172.217.0.0/19 ip4:172.217.32.0/20 ip4:17" "2.217.128.0/19 ip4:172.217.160.0/20 ip4:172.217.192.0/19 ip4:172.253.56.0/21 ip4:172.253.112.0/20 ip4:108.177.96.0/19 ip4:35.191.0.0/16 ip4:130.211.0.0/22 include:_spf2.stackoverflow.com ~all"

I haven't read the RFCs but is splitting the string anywhere permitted or does it have to be on a space boundary?   i.e. I notice `ip4:17" "2.217.128.0/19` and that makes me wonder if it the split is permitted there.

There are other RFCs where the individual strings are joined with a space.: `strings.Join(list, "")` vs `strings.Join(list, " ")`

Crazybus: Not a dumb question at all. It caught me off guard as well since it just looks wrong. 

According to the RFC you can split anywhere and you don't add any extra white space. 

From: https://tools.ietf.org/html/rfc4408#section-3.1.3

> As defined in [RFC1035] sections 3.3.14 and 3.3, a single text DNS
> record (either TXT or SPF RR types) can be composed of more than one
> string.  If a published record contains multiple strings, then the
> record MUST be treated as if those strings are concatenated together
> without adding spaces.  For example:
> 
>    IN TXT "v=spf1 .... first" "second string..."
> 
> MUST be treated as equivalent to
> 
>    IN TXT "v=spf1 .... firstsecond string..."
> 
> SPF or TXT records containing multiple strings are useful in
> constructing records that would exceed the 255-byte maximum length of
> a string within a single TXT or SPF RR record.

I have been testing these changes with multiple strings against https://www.kitterman.com/spf/validate.html and a few other tools to double-check that they are valid.

If you want to check for yourself stspg-customer.com is an example of a provider that is using multiple strings that passes in kitterman.

```
❯ dig txt stspg-customer.com +short
"v=spf1 ip4:23.253.182.103 ip4:23.253.183.145 ip4:23.253.183.146 ip4:23.253.183.147 ip4:23.253.183.148 ip4:23.253.183.150 ip4:166.78.68.221 ip4:166.78.69.146 ip4:167.89.46.159 ip4:167.89.64.9 ip4:167.89.65.0 ip4:167.89.65.53 ip4:167.89.65.100 ip4:167.89.74" ".233 ip4:167.89.75.33 ip4:167.89.75.126 ip4:167.89.75.136 ip4:167.89.75.164 ip4:192.237.159.42 ip4:192.237.159.43 -all"
```

pkg/spflib/flatten.go
```diff
@@ -5,6 +5,28 @@ import (
 	"strings"
 )
 
+// Borrowed from https://stackoverflow.com/a/61469854/11477663
```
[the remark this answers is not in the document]
Crazybus: Thanks will fix it up!

pkg/spflib/flatten_test.go
tlimoncelli: Great!  Thanks for the explanation (translation: thank you for not making me read the RFC :-) )

PR #795, opened 2020-07-31
Make it possible to disable the raw SPF optimizer debug record

Crazybus:
Open to other configuration opens for how best to make this optional. Or
potentially making this an opt in configuration item which would be a
breaking change.

The main reason that someone would want to disable this is if their raw
SPF record goes over the 255 characters. This is potentially another
place that could get some multi string support. But as it is only used
for debugging purposes it seems like there should be a way to outright
disable it too.


tlimoncelli:
Sounds good!

PR #800, opened 2020-08-02
New features: DU() and getConfiguredDomains() to assist modifying domains

patschi:
**This implements two functions:**
* `getConfiguredDomains()` as a sort of helper function to get all current added/configured domains until the point of execution. Returning a string array of those domains. This could be used, e.g. at the end of the dnscontrol.js file, to do some special handling or dynamically set records (e.g. "domain2.tld_report._dmarc.domain1.tld"). (I think a dedicated function might be handy for the future, to have backwards-compatibility in case something gets changed in the future.)
* `DU()` is basically a changed `D()` function and allows changing/adding specific records at a later point during processing. Like above, like at the end of the script execution to add specific records for each previous processed domains.

For example:
```javascript
D('domain.tld', REG, DnsProvider(DNS),
	A('@', "127.0.0.1")
)
DU('domain.tld',
	A('@', "127.0.0.2")
)
```

Ends up in modifications like:
```
******************** Domain: domain.tld
----- Getting nameservers from: registrar
----- DNS Provider: registrar...3 corrections
#1: CREATE A domain.tld 127.0.0.1 ttl=43200
#2: CREATE A domain.tld 127.0.0.2 ttl=43200
#3: REFRESH zone domain.tld
----- Registrar: registrar...0 corrections
```

**Two more useful examples:**

Let's add a very important TXT() record to all domains we manage:
```javascript
var domains = getConfiguredDomains();
for(i = 0; i < domains.length; i++) {
	DU(domains[i],
		TXT('_important', 'BLA') // I know, not really creative.
	)
}
```

According to DMARC RFC you need to specify `domain2.tld._report.dmarc.domain1.tld` to allow `domain2.tld` to report to `domain1.tld`. So you need to configure quite a lot subdomains manually... or...
```javascript
var domains = getConfiguredDomains();
for(i = 0; i < domains.length; i++) {
	DU("domain1.tld",
		TXT(domains[i] + '._report._dmarc', 'v=DMARC1')
	);
}
```

Both examples end up with following modifications:
```
******************** Domain: domain1.tld
----- Getting nameservers from: registrar
----- DNS Provider: registrar...2 corrections
#1: CREATE TXT _important.domain1.tld "BLA" ttl=43200
#2: REFRESH zone domain1.tld
----- Registrar: registrar...0 corrections

******************** Domain: domain2.tld
----- Getting nameservers from: registrar
----- DNS Provider: registrar...5 corrections
#1: CREATE TXT _important.domain2.tld "BLA" ttl=43200
#2: CREATE TXT domain1.tld._report._dmarc.domain2.tld "v=DMARC1" ttl=43200
#3: CREATE TXT domain3.tld._report._dmarc.domain2.tld "v=DMARC1" ttl=43200
#4: CREATE TXT domain4.tld._report._dmarc.domain2.tld "v=DMARC1" ttl=43200
#5: REFRESH zone domain2.tld
----- Registrar: registrar...0 corrections
```


docs/_functions/global/DU.md
```diff
@@ -0,0 +1,31 @@
+---
+name: DU
```
tlimoncelli, line 2: Pretty cool addition!  I can see how this would be useful.  I would like to suggest a different name, however.  How about `ADD_TO()` or `D_ADD()` ?

docs/_functions/global/DU.md
```diff
@@ -0,0 +1,31 @@
...
+parameters:
+  - name
+  - modifiers...
+---
+
+`DU` allows updating existing domains added by `D`. It behaves the same way, however it is mandatory that the domain was already added. The first argument, the domain name,
```
tlimoncelli, line 8: Suggested text:

`DU` adds records (and metadata) to a domain. The domain must have previously been defined by `D()`. `DU` behaves the same as `D()` in all other ways: The first argument is the domain name. See the documentation of `D` for further details.

docs/_functions/global/getConfiguredDomains.md
```diff
@@ -0,0 +1,62 @@
+---
+name: getConfiguredDomains
+parameters:
+  - name
+  - modifiers...
+---
+
+`getConfiguredDomains` is a simple helper function to return back configured domains until the point where the function
```
tlimoncelli, line 8: Suggested description:

`getConfiguredDomains` is a helper function that returns the domain names configured at the time the function is called.  Calling this function early or later in `dnsconfig.js` may return different results.  Typical usage is to iterate over all domains at the end of your configuration file.

docs/_functions/global/DU.md
patschi: I'd go for a function name starting with `D` because 1) it's doing domain stuff, 2) it works pretty like `D()` does. But I think `D_ADD()` might be a bit too misleading, as it's not "adding a domain" (what I might think of based on the name). Usually I thought about `DU()` standing for "**D**omain **U**pdate". As alternatives: What about `D_UPD()` or `D_UPDATE()` instead?

tlimoncelli: Yes, I see your point.  It should start with `D`.  My issue with `DU` is that it is vague.  (of course, so is `D`).   I like `D_UPDATE` but I feel like that could be misinterpreted as "replace" (UPDATE in SQL is "replace if exists")

Just brainstorming here but ...

* `D_INSERT()`
* `D_MORE()`
* `D_CONTINUE()`

jadesprague: I like D_MORE() as it's clear what it does, that it doesn't replace D(), and that it's not performing an SQL-style UPDATE. (D_INSERT() sounds as if it could mean "insert domain", same objection as D_ADD(). D_CONTINUE is OK, but without being able to clearly articulate why it feels weird to me.

pearcec: Consider using REC() since it is quick to parse as a human when reading down through a config.  Or consider rewriting D() to handle the updates.

patschi: I also like `D_MORE()` actually. But @pearcec's idea is pretty good as well: Allowing `D()` to update. But not sure however if this could lead to confusion or unintended side effects.

patschi: What about: `DE()` or `DE_EXT()` or `D_EXTEND()`?

tlimoncelli:
I like `D_EXTEND()`

tlimoncelli:
Thanks for the cool new features!

PR #806, opened 2020-08-08
New feature: Add IGNORE_TARGET

mikenz:
Our zone files have a lot of CNAMEs used for DNS validation with AWS Certificate Manager (https://docs.aws.amazon.com/acm/latest/userguide/gs-acm-validate-dns.html). These records are managed by other processes and we would want these left untouched by dnscontrol.

It's not possible to match these by record name without risk of false positives. It is however possible to positively match them based on the target of the record.

I'm proposing an IGNORE_TARGET function to handly this scenario.


tlimoncelli:
Hi Mike!  Certainly an interesting feature!   We are getting many requests for ways to deal with specific records being managed by a 3rd party.

I'm not clear how this is different from IGNORE()?  Is it that it accepts globs?

tlimoncelli:
(Ignore my previous reply. I misunderstood the feature.)

Hi Mike! Certainly an interesting feature! We are getting many requests for ways to deal with specific records being managed by a 3rd party.

I'm concerned about adding such a broad feature which could be used in unexpected ways that we'd have to support in the future.

When AWS Certificate Manager looks for a validation record, is it always a CNAME?  Should we restrict this feature to be just for CNAMEs?  Maybe take a parameter indicating which rtype it applies to (and have it be an error to be anything other than CNAME for now)?

I'm concerned that IGNORE() becomes ambiguous if there is also a IGNORE_somethingelse().  We should add IGNORE_LABEL() to be an alias for IGNORE(), and have IGNORE() be deprecated. (or something like that)

Thoughts?

mikenz:
I'd certainly support renaming IGNORE to IGNORE_NAME ('name' is the term used in both Cloudflare and R53's interface so I would favor that over 'label'). 
Initially I assumed I would be able to use use IGNORE to match all the ACM validation CNAMES. It wasn't until I tried using it that I realised it only matched on the name. Renaming that function to be more specific seems like a good idea to me.

Passing in the rType and only supporting CNAME, I'll add that to this PR. Although I don't think it'd be too long before someone also wants A records.

mikenz:
FYI I'm not a Go dev, I'm expecting a bunch of code changes out of code review

pkg/diff/diff_test.go
```diff
@@ -260,7 +261,60 @@ func TestInvalidGlobIgnoredRecord(t *testing.T) {
...
+	}
+
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("should panic: invalid glob pattern for IGNORE_TARGET")
+		}
+	}()
+
+	checkLengthsFull(t, existing, desired, 0, 1, 0, 0, false, nil, []*models.IgnoreTarget{{Pattern: "[.www3", Type: "CNAME"}} )
+}
```
tlimoncelli, line 299: I want to thank you for testing the negative and the positive.  So good!

pkg/js/helpers.js
```diff
@@ -495,10 +496,18 @@ function format_tt(transform_table) {
 // IGNORE(name)
 function IGNORE(name) {
```
tlimoncelli, line 497: Please supply an IGNORE() and an IGNORE_NAME()

mikenz:
I think that's everything addressed from round 1 :-)

docs/_functions/domain/IGNORE.md
```diff
@@ -1,40 +0,0 @@
----
-name: IGNORE
----
-
```
tlimoncelli, line 4: Rather than deleting this file, how about just a brief note that it has been renamed to be IGNORE_NAME().  IGNORE() still works but has been deprecated, please use the new name, etc. etc.

docs/_functions/domain/IGNORE.md
```diff
@@ -2,39 +2,4 @@
...
-* `IGNORE("?oo")` will ignore all records of three symbols ending in `oo`, for example `foo` and `zoo`. It will
-not match `.`
-* `IGNORE("[abc]oo")` will ignore records `aoo`, `boo` and `coo`. `IGNORE("[a-c]oo")` is equivalent.
-* `IGNORE("[!abc]oo")` will ignore all three symbol records ending in `oo`, except for `aoo`, `boo`, `coo`. `IGNORE("[!a-c]oo")` is equivalent.
-* `IGNORE("{bar,[fz]oo}` will ignore `bar`, `foo` and `zoo`.
-* `IGNORE("\\*.foo")` will ignore the literal record `*.foo`.
-
-It is considered as an error to try to manage an ignored record.
\ No newline at end of file
+IGNORE has been renamed to `IGNORE_NAME`. IGNORE will continue to function, but its use is deprecated. Please update you configuration files to use `IGNORE_NAME`.
```
tlimoncelli, line 5: s/you/your/

tlimoncelli:
Looks ready for merge to me.  Let me know if you agree.

docs/_functions/domain/IGNORE.md
mikenz: fixed 🤦‍♂️

mikenz:
@tlimoncelli I believe this is ready to merge now. Thanks

tlimoncelli:
Thanks!  A lot of AWS users are going to be very happy!

PR #807, opened 2020-08-08
fix: only run SRV null target integration test when CanUseSRV is set

svenpeter42:
I'm currently trying to implement support for inwx.de and was surprised to see a SRV integration test fail when I hadn't added CanUseSRV yet.
I'm not sure when an empty SRV record is useful and it's not supported by inwx.de anyway, but I still think that the test should only run when the provider claims to support SRV records.


tlimoncelli:
Interesting!  I know that not() and only() can't be mixed but I've never tried mixing `requires()` with `not()`.  

Good catch!

Issue #254, opened 2017-10-30
Support provider: INWX


axel-rutz:
Internetworx is a berlin-located provider with good API support:

- https://www.inwx.de/en/offer/api
- https://github.com/andrexus/goinwx

axel-rutz:
I'm not firm at coding go but can provide testing.

tlimoncelli:
Hopefully we can find a volunteer.

By the way... we've made it very easy to add providers.  In fact, many people have created a provider as their first Go program.
https://everythingsysadmin.com/2017/08/go-get-up-to-speed.html

If you are interested, I'd be willing to mentor you here or in private email.

tlimoncelli:
I've noted this  in 2e61c835110911a34810678b1ca94807a45ca04d.  If someone starts working on this, please re-open this issue.  See https://stackexchange.github.io/dnscontrol/provider-list

svenpeter42:
I've started working on inwx support.

patschi:
Thanks to @svenpeter42 this can now be closed I guess, @tlimoncelli? (Because it's listed under "Providers" in the docs otherwise)

svenpeter42:
I think the issue is already closed but maybe I forgot to remove INWX from the list of requested providers, I can take a look later this week.

tlimoncelli:
I forgot tot remove the provider-request tag.  Should be all good now.

I'm having a busy week but I hope to do a new release next week.  Sadly the website docs are based on master, not the latest release.

PR #808, opened 2020-08-09
INWX provider

svenpeter42:
Adds registrar and dns provider support for inwx.de. First requested in #254 

ToDo:
- [x] use nrdcg/goinwx after my MR for that repo has been merged
- [x] Finish documentation
- [x] Go through all features and add providers.Cannot or providers.Unimplemented
- [x] add TOTP generation support
- [x] fix goinwx package for .de domains

### DIFF
--- a/providers/inwx/inwxProvider.go
+++ b/providers/inwx/inwxProvider.go
@@ -1,0 +1,289 @@
+package inwx
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"sort"
+
+	"github.com/StackExchange/dnscontrol/v3/pkg/diff"
+	"github.com/StackExchange/dnscontrol/v3/providers"
+	"github.com/StackExchange/dnscontrol/v3/models"
+
+	"github.com/svenpeter42/goinwx"
+)
+
+/*
+INWX Registrar and DNS provider
+
+Info required in `creds.json`:
+	- username
+	- password
+	- totp (TOPT code if 2FA is enabled)
+
+Additional settings available in `creds.json`:
+	- sandbox (set to 1 to use the sandbox API from INWX)
+
+*/
+
+type InwxApi struct {
+	client *goinwx.Client
+	sandbox bool
+}
+
+var InwxDefaultNs = []string{"ns.inwx.de","ns2.inwx.de","ns3.inwx.eu","ns4.inwx.com","ns5.inwx.net"}
+var InwxSandboxDefaultNs = []string{"ns.ote.inwx.de","ns2.ote.inwx.de"}
+
+var features = providers.DocumentationNotes{
+	providers.CanUseAlias:            providers.Cannot("INWX does not support the ALIAS or ANAME record type."),
+	providers.CanUseCAA:              providers.Can(),
+	providers.CanUsePTR:              providers.Can(),
+	providers.CanUseNAPTR:            providers.Can(),
+	providers.CanUseSRV:              providers.Can("SRV records with empty targets are not supported."),
+	providers.CanUseSSHFP:            providers.Can(),
+	providers.CanUseTLSA:             providers.Can(),
+	providers.CanAutoDNSSEC:          providers.Unimplemented("Supported by INWX but not implemented yet."),
+	providers.DocOfficiallySupported: providers.Cannot(),
+	providers.DocDualHost:            providers.Can(),
+	providers.DocCreateDomains:       providers.Unimplemented("Supported by INWX but not implemented yet."),
+	providers.CanGetZones:            providers.Can(),
+}
+
+func init() {
+	providers.RegisterRegistrarType("INWX", newInwxReg)
+	providers.RegisterDomainServiceProviderType("INWX", newInwxDsp, features)
+}
+
+func newInwx(m map[string]string) (*InwxApi, error) {
+	for key := range m {
+		switch key {
+		case "username",
+			"password",
+			"totp",
+			"sandbox":
+			continue
+		default:
+			return nil, fmt.Errorf("INWX: unknown key in `creds.json` (%s)\n", key)
+		}
+	}
+
+	if m["username"] == "" {
+		return nil, fmt.Errorf("INWX Username must be provided.")
+	}
+	if m["password"] == "" {
+		return nil, fmt.Errorf("INWX Password must be provided.")
+	}
+
+	var sandbox bool
+	if m["sandbox"] == "1" {
+		sandbox = true
+	} else {
+		sandbox = false
+	}
+
+	opts := &goinwx.ClientOptions{Sandbox: sandbox}
+	client := goinwx.NewClient(m["username"], m["password"], opts)
+
+	err := client.Account.Login()
+	if err != nil {
+		return nil, fmt.Errorf("Unable to login to INWX")
+	}
+
+	if m["totp"] != "" {
+		err := client.Account.Unlock(m["totp"])
+		if err != nil {
+			return nil, fmt.Errorf("Could not unlock INWX account - TOTP is probably invalid.")
+		}
+	}
+
+	api := &InwxApi{client: client, sandbox: sandbox}
+
+	return api, nil
+}
+
+func newInwxReg(m map[string]string) (providers.Registrar, error) {
+	return newInwx(m)
+}
+
+func newInwxDsp(m map[string]string, metadata json.RawMessage) (providers.DNSServiceProvider, error) {
+	return newInwx(m)
+}
+
+func makeNameserverRecordRequest(domain string, rec *models.RecordConfig) (*goinwx.NameserverRecordRequest) {
+	req := &goinwx.NameserverRecordRequest{
+		Domain: domain,
+		Type: rec.Type,
+		Content: rec.GetTargetField(),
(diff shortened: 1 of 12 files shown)


providers/inwx/inwxProvider.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/svenpeter42/goinwx"
+)
+
+/*
+INWX Registrar and DNS provider
+
+Info required in `creds.json`:
+	- username
+	- password
+	- totp (TOPT code if 2FA is enabled)
```
tlimoncelli, line 22: TOPT?  You'll have to edit creds.json every 60 seconds?

providers/inwx/inwxProvider.go
```diff
@@ -0,0 +1,292 @@
...
+		case "username",
+			"password",
+			"totp",
+			"sandbox",
+			"domain":
+			continue
+		default:
+			return nil, fmt.Errorf("INWX: unknown key in `creds.json` (%s)\n", key)
+		}
+	}
```
tlimoncelli, line 72: I think this should be a warning, not an error.  The reason is that sometimes people include a key like "comment" or "owner".  On the other hand, if someone misspells `passwrod` it would be good to get a warning.

providers/inwx/inwxProvider.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/StackExchange/dnscontrol/v3/pkg/diff"
+	"github.com/StackExchange/dnscontrol/v3/providers"
+
+	"github.com/svenpeter42/goinwx"
+)
+
+/*
+INWX Registrar and DNS provider
+
+Info required in `creds.json`:
```
tlimoncelli, line 19: The comment mentions `domain`

providers/inwx/inwxProvider.go
```diff
@@ -0,0 +1,292 @@
...
+
+	if m["username"] == "" {
+		return nil, fmt.Errorf("INWX Username must be provided.")
+	}
+	if m["password"] == "" {
+		return nil, fmt.Errorf("INWX Password must be provided.")
+	}
+
+	var sandbox bool
+	if m["sandbox"] == "1" {
```
tlimoncelli, line 82: Suggestion: `sandbox = m["sandbox"] == "1"`

providers/inwx/inwxProvider.go
```diff
@@ -0,0 +1,292 @@
...
+
+func makeNameserverRecordRequest(domain string, rec *models.RecordConfig) *goinwx.NameserverRecordRequest {
+	req := &goinwx.NameserverRecordRequest{
+		Domain:  domain,
+		Type:    rec.Type,
+		Content: rec.GetTargetField(),
+		Name:    rec.GetLabel(),
+		Ttl:     int(rec.TTL),
+	}
+	targetWithoutDot := strings.TrimRight(rec.GetTargetField(), ".")
```
tlimoncelli, line 124: Humor me and try this to see if it works:

    t := rec.GetTargetField()
    targetWithoutDot :=t[:len(t)-2]

If that doesn't work, TrimSuffix instead of TrimRight.

Why?  (1) TrimRight() will remove multiple dots.  (2) GetTargetField() should be consistent in always returning a trailing dot.

providers/inwx/inwxProvider.go
```diff
@@ -0,0 +1,292 @@
...
+		if record.Type == "SOA" {
+			continue
+		}
+
+		/* INWX is a little bit special for CNAME,NS,MX and SRV records:
+		   The API will not accept any target with a final dot but will
+		   instead always add this final dot internally.
+		*/
+		if record.Type == "CNAME" || record.Type == "MX" || record.Type == "NS" || record.Type == "SRV" {
+			if record.Content != "." {
```
tlimoncelli, line 226: The comment doesn't mention in what situation where the name is "."

providers/inwx/inwxProvider.go
svenpeter42: nice, that makes the code much easier to read!

providers/inwx/inwxProvider.go
svenpeter42: good point, fixed.

providers/inwx/inwxProvider.go
svenpeter42: Unfortunately INWX doesn't support creating scoped API keys like e.g. Cloudflare does. It's only possible to login with the normal account details which also allow actions like deleting or transferring out domains (even via a single API call).
Additionally, only TOTP is supported as two factor authentication.

I don't change my creds.json every 60 seconds but instead use a env variable:
```
	"inwx": {
		"username": "username",
		"password": "password", 
		"totp": "$INWX_TOTP",
	},
``` 
and run `INWX_TOTP=12345 dnscontrol preview`. I'll add this to the documentation as an example later.

I can additionally add `totp-key` where the secret key could be provided and then generate the TOTP based on the current time as an alternative. That's probably more useful when using dnscontrol as part of a CI/CD pipeline.

providers/inwx/inwxProvider.go
svenpeter42: Good point, I've replaced it with `[:len(t)-1]`

providers/inwx/inwxProvider.go
svenpeter42: Good catch, I forgot this code from an early version where I was trying to figure out why these records didn't work.

providers/inwx/inwxProvider.go
svenpeter42: yes, domain is used in the integration tests. i've added a comment.

providers/inwx/inwxProvider.go
svenpeter42: hrm, i've looked at the other provides and only found a similar check in AXFRDDNS: maybe there shouldn't even be a warning. Further down I already make sure that username and password are present and that should be enough to notice if e.g. password was misspelled. 

What are your thoughts on this? I would remove the warning to keep the behavior consistent with (almost) all other providers.

providers/inwx/inwxProvider.go
tlimoncelli: Great way to handle that!!

providers/inwx/inwxProvider.go
tlimoncelli: On second thought... remove it.  If someone does use (for example) "comment" they'd get a warning in every run and it would be super annoying.

tlimoncelli:
Looking good so far!

tlimoncelli:
Looks good!  Ready to merge?

svenpeter42:
> Looks good! Ready to merge?

Yup! Thanks for nice review and the [guide](https://stackexchange.github.io/dnscontrol/writing-providers). My main languages are C/C++/Python and I was quite surprised how fast I was able to get up to speed with Go :)

patschi:
Apologize the off-topic: About 1.5 weeks ago I was chatting with people in a public Telegram group I'm thinking about to move a few domains to INWX, but I do really miss an implementation for dnscontrol and that this is a showstopper. About 2 days later this PR was raised. This is spooky!

So... Thanks a lot @svenpeter42! :)

svenpeter42:
Hah, nice. And spooky! I'm glad to hear this is already going to be useful for someone else :)

tlimoncelli:
Thanks for the contribution!!

tlimoncelli:
Sorry for the late reply but... I just noticed that we forgot to run `go lint ./...`

```
providers/inwx/inwxProvider.go:37:6: type InwxApi should be InwxAPI
providers/inwx/inwxProvider.go:37:6: type name will be used as inwx.InwxApi by other packages, and that stutters; consider calling this Api
providers/inwx/inwxProvider.go:84:25: error strings should not be capitalized or end with punctuation or a newline
providers/inwx/inwxProvider.go:108:22: error strings should not be capitalized or end with punctuation or a newline
providers/inwx/inwxProvider.go:111:21: error strings should not be capitalized or end with punctuation or a newline
providers/inwx/inwxProvider.go:124:26: error strings should not be capitalized or end with punctuation or a newline
providers/inwx/inwxProvider.go:127:26: error strings should not be capitalized or end with punctuation or a newline
providers/inwx/inwxProvider.go:130:26: error strings should not be capitalized or end with punctuation or a newline
providers/inwx/inwxProvider.go:257:9: if block ends with a return statement, so drop this else and outdent its block
providers/linode/linodeProvider.go:47:6: type name will be used as linode.LinodeAPI by other packages, and that stutters; consider calling this API
```

svenpeter42:
Ugh.. I remember running it before adding function comments but I must've forgotten to run it again afterwards. I'll create a PR to fix the issues later this week.

Slightly related: What do you think about adding the linter to an automated CI system integrated with PRs so that it's much harder to forget this in the future?
I've only briefly looked at the dnscontrol setup and there's at least a .travis.yml and an Azure pipeline setup and an (experimental) PR for github actions but none of them seem to be automatically running for PRs.

tlimoncelli:
Thanks! Next week is fine.

Adding it to the CI is a great idea!  We're in the process of moving to github actions and that sounds like a good thing to add.  I've added it to my list.

tlimoncelli:
I'll open separate issues for both of these later today.

PR #812, opened 2020-08-13
Notifications: Microsoft Teams

mikenz:
Add support to send notifications to MS Teams

image

### DIFF
--- a/pkg/notifications/teams.go
+++ b/pkg/notifications/teams.go
@@ -1,0 +1,52 @@
+package notifications
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+func init() {
+	initers = append(initers, func(cfg map[string]string) Notifier {
+		url, ok := cfg["slack_url"]
+		if !ok {
+			return nil
+		}
+
+		notifier := &slackNotifier{
+			URL: url,
+		}
+		return notifier
+	})
+}
+
+// teamsNotifier sends notifications to teams or mattermost
+type teamsNotifier struct {
+	URL string
+}
+
+func (s *teamsNotifier) Notify(domain, provider, msg string, err error, preview bool) {
+	var payload struct {
+		Username string `json:"username"`
+		Text     string `json:"text"`
+	}
+	payload.Username = "DnsControl"
+
+	// Format changes as 'preformated' text
+	msg = strings.ReplaceAll(msg, "\n", "\n    ")
+
+	if preview {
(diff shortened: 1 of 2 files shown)


mikenz:
IRL:
image

pkg/notifications/teams.go
```diff
@@ -0,0 +1,50 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+func init() {
+	initers = append(initers, func(cfg map[string]string) Notifier {
+		if url, ok := cfg["teams_url"]; ok {
```
tlimoncelli, line 13: Typical Go style is to return nil early:

    		if url, ok := cfg["teams_url"]; !ok {
    		    return nil
    		}
    		notifier := &teamsNotifier{

...

pkg/notifications/teams.go
```diff
@@ -0,0 +1,50 @@
...
+
+	if preview {
+		payload.Text = fmt.Sprintf("**DnsControl Preview %s**\n%s", domain, msg)
+	} else if err != nil {
+		payload.Text = fmt.Sprintf("**DnsControl Error Making Changes %s**\n%s\nError: %s", domain, msg, err)
+	} else {
+		payload.Text = fmt.Sprintf("**DnsControl Successfully Changed %s**\n%s", domain, msg)
+	}
+
+	json, _ := json.Marshal(payload)
```
tlimoncelli, line 46: Please check the error and return it upstream.

pkg/notifications/teams.go
```diff
@@ -0,0 +1,50 @@
...
+	http.Post(s.URL, "text/json", bytes.NewReader(json))
```
tlimoncelli, line 47: I think `http.Post` returns an error. Please check.

pkg/notifications/teams.go
mikenz: If I return it upstream I'm going to have to change the interface definition as it doesn't currently allow for this. I guess this is going to break anyone else's custom notify types, are you okay with that? 

If so, `Notify(domain, provider string, message string, err error, preview bool) (err error)`, preferences on naming for the two err's?

tlimoncelli: Oops... I misread the signature.  (I should get more sleep :-) )

I'll file a bug to improve this in the future.

pkg/notifications/teams.go
mikenz: Fixed and branch squashed

PR #816, opened 2020-08-17
Fix for Azure DNS Provider with >100 Zones

### LINKED ISSUE #792: Azure DNS Provider with >100 Zones
I'm attempting to manage a large (>500 domains) Azure DNS environment with dnscontrol.

When I attempt to manage an Azure DNS zone that isn't in the first page of the Azure API response within an Azure Resource Group, dnscontrol doesn't think the domain exists, even if it already does.

I think this is likely the function that needs to be changed to use an iterator: https://github.com/StackExchange/dnscontrol/blob/42f7568074698956f70b0bef0feff6b7516fdb53/providers/azuredns/azureDnsProvider.go#L75 with a similar fix to https://github.com/StackExchange/dnscontrol/issues/770

run1et:
Resolves #792

Before changes (all domains were created previously by running `dnscontrol create-domains`):
```text
******************** Domain: zone-100-7930ba393bb96a6fb30f62e75d593f46.com
----- Getting nameservers from: azuredns
----- DNS Provider: azuredns...1 correction
#1: CREATE A record-001.zone-100-7930ba393bb96a6fb30f62e75d593f46.com 1.1.1.1 ttl=300
SUCCESS!
******************** Domain: zone-101-469eb2223790512426e52607b9029f93.com
----- Getting nameservers from: azuredns
Domain zone-101-469eb2223790512426e52607b9029f93.com not found in you Azure account
```

After (first run):
```
******************** Domain: zone-199-7b34b3adb50bac550114bb963a1bac88.com
----- Getting nameservers from: azuredns
----- DNS Provider: azuredns...1 correction
#1: CREATE A record-001.zone-199-7b34b3adb50bac550114bb963a1bac88.com 1.1.1.1 ttl=300
SUCCESS!
******************** Domain: zone-200-2d994aaaf808423e05b2a888d7047973.com
----- Getting nameservers from: azuredns
----- DNS Provider: azuredns...1 correction
#1: CREATE A record-001.zone-200-2d994aaaf808423e05b2a888d7047973.com 1.1.1.1 ttl=300
SUCCESS!
Done. 100 corrections.
```

Second run:
```
******************** Domain: zone-198-b6ab0b0a145d7648a6af91c48225be1e.com
----- Getting nameservers from: azuredns
----- DNS Provider: azuredns...0 corrections
******************** Domain: zone-199-7b34b3adb50bac550114bb963a1bac88.com
----- Getting nameservers from: azuredns
----- DNS Provider: azuredns...0 corrections
******************** Domain: zone-200-2d994aaaf808423e05b2a888d7047973.com
----- Getting nameservers from: azuredns
----- DNS Provider: azuredns...0 corrections
Done. 0 corrections.
```

### DIFF
--- a/providers/azuredns/azureDnsProvider.go
+++ b/providers/azuredns/azureDnsProvider.go
@@ -73,6 +73,10 @@ func init() {
 }
 
 func (a *azureDNSProvider) getExistingZones() (*adns.ZoneListResult, error) {
+    // Please note — this function doesn't work with > 100 zones
+    // https://github.com/StackExchange/dnscontrol/issues/792
+    // Copied this code to getZones and ListZones and modified it for using a paging
+    // As a result getExistingZones is not used anymore
 	ctx, cancel := context.WithTimeout(context.Background(), 6000*time.Second)
 	defer cancel()
 	zonesIterator, zonesErr := a.zonesClient.ListByResourceGroupComplete(ctx, *a.resourceGroup, to.Int32Ptr(100))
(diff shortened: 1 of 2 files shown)


run1et:
I've tested this fix on the following data:
[dnsconfig.js.zip](https://github.com/StackExchange/dnscontrol/files/5084507/dnsconfig.js.zip)

providers/azuredns/azureDnsProvider.go
```diff
@@ -73,6 +73,10 @@ func init() {
 }
 
 func (a *azureDNSProvider) getExistingZones() (*adns.ZoneListResult, error) {
+    // Please note — this function doesn't work with > 100 zones
```
tlimoncelli, line 76: Can we just delete `getExistingZones()`?  (considering that it has a known but and it isn't used)

run1et: I would like to test a new (upcoming) Azure SDK version. It can probably fix a `Complete` calls. That's the only one reason to keep this function at least for some time.
But we can restore it at any moment and for clarity reason better to delete it.

tlimoncelli: Sounds good. Let me know when it is ready for merge.

run1et: It's ready (and tested on Azure) right now. And I'll test new Azure SDK in few days/weeks when it will be released.

tlimoncelli:
Thanks for fixing this!

Issue #802, opened 2020-08-03
DEFAULTS(DefaultTTL()) doesn't apply to SOA records for bind provider


pkorovin:
Here's sample dnsconfig.js:

```javascript
DEFAULTS(DefaultTTL('1h'), NAMESERVER_TTL('1h'));

var REG_LOCAL = NewRegistrar("none", "NONE");
var BIND = NewDnsProvider('bind', 'BIND', {
    'default_soa': {
        'master':  'ns1.example.com.',
        'mbox':    'hostmaster.example.com.',
        'refresh': 43200,
        'retry':   7200,
        'expire':  2419200,
        'minttl':  3600,
    }
})

D("example.com", REG_LOCAL, DnsProvider(BIND),
        NAMESERVER('ns1.example.com.'),
        NAMESERVER('ns2.example.com.'),
        MX('@', 10, 'mx.example.com.'),
        A('mx',     '10.10.10.10')
)
```

DEFAULTS(DefaultTTL(3600)) doesn't apply to bind provider generated SOA record, which is always set to TTL=300:

```console
$ dnscontrol preview
******************** Domain: example.com
----- Getting nameservers from: bind
----- DNS Provider: bind...1 correction
#1: GENERATE_ZONEFILE: 'example.com'. Changes:
MODIFY SOA example.com: (ns1.example.com. hostmaster.example.com. 43200 7200 2419200 3600 ttl=3600) -> (ns1.example.com. hostmaster.example.com. 43200 7200 2419200 3600 ttl=300)

----- Registrar: none...0 corrections
Done. 1 corrections.
```

Relavant lines from generated zonefile:
```dns-zone
$TTL 3600
; generated with dnscontrol 2020-08-03T20:06:39+03:00
@          300   IN SOA   ns1.example.com. hostmaster.example.com. 2020080301 43200 7200 2419200 3600
```

SOA record always takes the default TTL of 300.

Veratil:
@pkorovin can you double check to make sure TTL is respected now? Additionally you should be able to add a TTL to the `default_soa` struct.

tlimoncelli:
@pkorovin Could you see if the issue still exists in the `master` branch?

pkorovin:
@tlimoncelli Tom, the issue was fixed with 5d2790, thanks!

PR #820, opened 2020-08-19
Add TTL to BIND SoaInfo struct

Veratil:
I've been working on adding an SOA record type to BIND, but I saw #802 and figured I can pull that work out and see if it works as I did it.

Let me know if there's anything I missed or need to do different.

I've tested on my own with changing `DEFAULTS(DefaultTTL())` and it has properly caught changes.

### DIFF
--- a/providers/bind/bindProvider.go
+++ b/providers/bind/bindProvider.go
@@ -85,10 +85,11 @@ type SoaInfo struct {
 	Retry   uint32 `json:"retry"`
 	Expire  uint32 `json:"expire"`
 	Minttl  uint32 `json:"minttl"`
+	TTL     uint32 `json:"ttl,omitempty"`
 }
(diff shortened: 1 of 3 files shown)


tlimoncelli:
Looks good. Could you add an integration test?  Take a look at `testgroup("SRV"` since it tests changing individual fields works.

Veratil:
The SOA record seems a bit harder to test since every run of the integration tests is leftover with the SOA record last changed from the previous run. From an initial run without a zone file, things will run fine, but a second run or more following can fail since it doesn't start from a "clean" SOA record.

tlimoncelli:
Ah... yeah, I can see how that would happen. Thanks for fixing https://github.com/StackExchange/dnscontrol/issues/802

PR #825, opened 2020-08-26
fix golint issues in INWX provider

svenpeter42:
see comments in #808 by @tlimoncelli


tlimoncelli:
Thanks for taking care of that!

PR #830, opened 2020-08-29
Allow for Name's that start with _

misilot:


pkg/normalize/validate.go
```diff
@@ -127,17 +113,12 @@ func checkLabel(label string, rType string, target, domain string, meta map[stri
 			return nil
 		}
 	}
-	// Don't warn for CNAMEs if the target ends with acm-validations.aws
-	// See https://github.com/StackExchange/dnscontrol/issues/519
-	if strings.HasPrefix(label, "_") && rType == "CNAME" && strings.HasSuffix(target, ".acm-validations.aws.") {
+	// Don't warn for records that start with _
+	// See https://github.com/StackExchange/dnscontrol/issues/829
+	if strings.HasPrefix(label, "_") {
```
tlimoncelli, line 118: I'm getting warnings for `fd._domainkey.stackoverflow.com` and similar.  How about:

    if strings.HasPrefix(label, "_") || strings.Contains(label, "._") {

That should cover both cases.

misilot: I merged your changes from #833 into #830.

tlimoncelli:
Awesome! Thanks for the PR.  This has been something I've been meaning to add for a while.  Thanks1

PR #832, opened 2020-08-30
Docs: Add a page about macros and loops

tlimoncelli:


tlimoncelli:
CC: @misilot

Your message inspired me to write up this advice.  Could I get a review?

misilot:
This looks great! I was curious about the combined Google Apps reference I found somewhere, but you provided a better solutions for Alias domains in this that makes sense.

I would be interested to see an example of the type of information you put for the domain at the top, but I figure it is general information about the domain so if something were to go completely wrong it provides a starting point to recover type of thing as well as "why" we have this domain

mikenz:
Looks good! Interesting that we ended up with a very similar file layout with macros and helper functions at the top followed by domains.

PR #840, opened 2020-08-31
Read-only 'Registrar' provider that does an NS lookup to validate the expected NS records

mikenz:
Okay, so this might be a "just me" problem, but I have a bunch of domains stuck with registrars without APIs. 

While I could just use `NewRegistrar('none', 'NONE');` and manage them manually, I'd like to ensure the nameservers are still correct.

This provider is my solution to this problem, it uses DNS-over-HTTPS to query the NS records for the domain and ensure they matched the expected servers. It returns an error if it is asked to push NS changes (eg `FAILURE! DNS-over-HTTPS 'Registrar' is read only, changes must be applied to example.com manually`). You can configure the DOH provider - tested with Quad9, Cloudflare, Google Public DNS.

I'm not sure if there's interest in this being merged, if not I'm happy to keen it in our own fork.

### DIFF
--- a/docs/_providers/doh.md
+++ b/docs/_providers/doh.md
@@ -1,0 +1,37 @@
+---
+name: DNS-over-HTTPS
+title: DNS-over-HTTPS Provider
+layout: default
+jsId: DNSOVERHTTPS
+---
+# DNS-over-HTTPS Provider
+
+This is a read-only/monitoring "registrar". It does a DNS NS lookup to confirm the nameserver servers are correct. This "registrar" is unable to update the NS servers but will alert you if they are incorrect. A common use of this provider is when the domain is with a registrar that does not have an API.
+
+## Configuration
+The DNS-over-HTTPS provider does not require anything in `creds.json`. By default, it uses Google Public DNS however you may configure an alternative RFC 8484 DoH provider.
+
+{% highlight json %}
+{
+  "DNS-over-HTTPS": {
+    "host": "cloudflare-dns.com"
+  }
+}
+{% endhighlight %}
+
+Some common DoH providers are `cloudflare-dns.com` [Cloudflare](https://developers.cloudflare.com/1.1.1.1/dns-over-https), `9.9.9.9` [Quad9](https://www.quad9.net/about/), and `dns.google` [Google Public DNS](https://developers.google.com/speed/public-dns/docs/doh)
+
+## Metadata
+This provider does not recognize any special metadata fields unique to Internet.bs.
+
+## Usage
+Example Javascript:
+
+{% highlight js %}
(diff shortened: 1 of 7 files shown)


tlimoncelli:
Interesting idea!  

Maybe the name should be more about the purpose (verification) than the protocolo.  I'll give it some thought.

mikenz:
How we ended up naming it when we used it:

```
// REG_NONE to maintain NS records completely manually
// REG_MONITOR to monitor the NS servers via DNS but maintain them manually
var REG_MONITOR = NewRegistrar('DNS-over-HTTPS', 'DNSOVERHTTPS');
// REG_CSCGLOBAL to automatically maintain NS records with CSC Global
var REG_CSCGLOBAL = NewRegistrar('CSC Global', 'CSCGLOBAL');
```

If that guides your thinking.

docs/_providers/doh.md
```diff
@@ -0,0 +1,37 @@
...
+
+{% highlight json %}
+{
+  "DNS-over-HTTPS": {
+    "host": "cloudflare-dns.com"
+  }
+}
+{% endhighlight %}
+
+Some common options are `cloudflare-dns.com` [Cloudflare](https://developers.cloudflare.com/1.1.1.1/dns-over-https), `9.9.9.9` [Quad9](https://www.quad9.net/about/), and `dns.google` [Google Public DNS](https://developers.google.com/speed/public-dns/docs/doh)
```
tlimoncelli, line 22: How about a default?  It would be nice to not require a creds.json file if not needed (BIND and others).

providers/doh/dohProvider.go
```diff
@@ -0,0 +1,61 @@
...
+
+	expected := []string{}
+	for _, ns := range dc.Nameservers {
+		expected = append(expected, ns.Name)
+	}
+	sort.Strings(expected)
+	expectedNameservers := strings.Join(expected, ",")
+
+	if foundNameservers != expectedNameservers {
+		return []*models.Correction{
```
tlimoncelli, line 51: This would look more stylistic if the `!=` was changed to `==` and the logic was reversed.

providers/doh/api.go
```diff
@@ -0,0 +1,36 @@
...
+func (c *api) getNameservers(domain string) ([]string, error) {
+	resolver := doh.Resolver{
+		Host:  c.host,
+		Class: doh.IN,
+	}
+
+	// Perform a NS lookup
+	nss, _, err := resolver.LookupNS(domain)
+	if err != nil {
+		return nil, fmt.Errorf("Failed fetching nameservers list (DNS-over-HTTPS): %s", err)
```
tlimoncelli, line 23: error strings should not be capitalized

PR #841, opened 2020-08-31
print-ir: print validation warnings/errors to stderr

anlutro:
currently, any validation warnings are printed to stdout, which means commands like `dnscontrol print-ir | jq` fails because the output isn't valid JSON.

by writing these to stderr instead (by using `log.Printf` instead of `fmt.Printf`), the errors are still printed, but won't cause JSON parsing of stdout to fail.

### DIFF
--- a/commands/printIR.go
+++ b/commands/printIR.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Good catch!  Thanks for the fix!

PR #843, opened 2020-09-04
AZURE: Cleanup: Return Errorf value of Error instead of pure Error

### LINKED ISSUE #836: AZURE: Cleanup: Errorf is a pure function but its return value is ignored
Hi @run1et

I just noticed this: The line `fmt.Errorf(` returns an error value, which is never used.  Code like that appears in two places:

https://github.com/StackExchange/dnscontrol/blob/de308c095245ee9d8585626d7c3c536ad5209c3f/providers/azuredns/azureDnsProvider.go#L97

https://github.com/StackExchange/dnscontrol/blob/de308c095245ee9d8585626d7c3c536ad5209c3f/providers/azuredns/azureDnsProvider.go#L145

In both cases, I think the intention is:

    return nil, fmt.Errorf("ListZones: zonesErr: %w", zonesErr)

Could you take a look?  Thanks!

run1et:
Resolves #836


tlimoncelli:
Thanks!!

PR #855, opened 2020-09-13
INWX: Support creating domains

patschi:
This adds support of `create-domains` for INWX.

INWX has two ways for "adding" domains:
- Adding domain to INWX as registrar
- Adding domain to INWX as only nameserver. (While INWX prohibits using their nameserver for external domains, it's technically still possible)

This implementation checks and adds domains to the nameserver only and does not register/order domains in this approach.

### DIFF
--- a/providers/inwx/inwxProvider.go
+++ b/providers/inwx/inwxProvider.go
@@ -45,6 +45,9 @@ var InwxDefaultNs = []string{"ns.inwx.de", "ns2.inwx.de", "ns3.inwx.eu"}
 // InwxSandboxDefaultNs contains the default INWX nameservers in the sandbox / OTE.
 var InwxSandboxDefaultNs = []string{"ns.ote.inwx.de", "ns2.ote.inwx.de"}
 
+// InwxDomainIndex contains all domains saved in the INWX nameserver
+var InwxDomainIndex map[string]int
+
 // features is used to let dnscontrol know which features are supported by INWX.
 var features = providers.DocumentationNotes{
 	providers.CanUseAlias:            providers.Cannot("INWX does not support the ALIAS or ANAME record type."),
@@ -59,7 +62,7 @@ var features = providers.DocumentationNotes{
 	providers.CanAutoDNSSEC:          providers.Unimplemented("Supported by INWX but not implemented yet."),
 	providers.DocOfficiallySupported: providers.Cannot(),
 	providers.DocDualHost:            providers.Can(),
-	providers.DocCreateDomains:       providers.Unimplemented("Supported by INWX but not implemented yet."),
+	providers.DocCreateDomains:       providers.Can("Does only create domain in nameserver and does not order domain."),
 	providers.CanGetZones:            providers.Can(),
 	providers.CanUseAzureAlias:       providers.Cannot(),
 }
@@ -165,7 +168,8 @@ func makeNameserverRecordRequest(domain string, rec *models.RecordConfig) *goinw
 	}
 
 	switch rType := rec.Type; rType {
-	/* INWX is a little bit special for CNAME,NS,MX and SRV records:
+	/*
+	   INWX is a little bit special for CNAME,NS,MX and SRV records:
 	   The API will not accept any target with a final dot but will
 	   instead always add this final dot internally.
 	   Records with empty targets (i.e. records with target ".")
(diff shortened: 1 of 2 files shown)


svenpeter42:
Thanks a lot! I felt a little bit bad about not adding this feature.

providers/inwx/inwxProvider.go
```diff
@@ -33,18 +33,15 @@ Additional settings available in `creds.json`:
...
-
 // InwxDefaultNs contains the default INWX nameservers.
-var InwxDefaultNs = []string{"ns.inwx.de", "ns2.inwx.de", "ns3.inwx.eu"}
+var InwxProductionDefaultNs = []string{"ns.inwx.de", "ns2.inwx.de", "ns3.inwx.eu"}
 
 // InwxSandboxDefaultNs contains the default INWX nameservers in the sandbox / OTE.
 var InwxSandboxDefaultNs = []string{"ns.ote.inwx.de", "ns2.ote.inwx.de"}
 
+// InwxDomainIndex contains all domains saved in the INWX nameserver
+var InwxDomainIndex map[string]int
```
svenpeter42, line 43: I like the idea of caching the domains since dnscontrol will possibly call `EnsureDomainExists` a lot. 

What do you think about instead putting this into the `inwxAPI` struct? There might be some users who manage two separate INWX accounts with dnscontrol.

providers/inwx/inwxProvider.go
```diff
@@ -250,13 +254,18 @@ func (api *inwxAPI) GetDomainCorrections(dc *models.DomainConfig) ([]*models.Cor
 	return corrections, nil
 }
 
-// GetNameservers returns the default nameservers for INWX.
-func (api *inwxAPI) GetNameservers(domain string) ([]*models.Nameserver, error) {
+// getDefaultNameservers returns back string map with default nameservers  based on e.g. sandbox mode.
+func (api *inwxAPI) getDefaultNameservers() []string {
 	if api.sandbox {
-		return models.ToNameservers(InwxSandboxDefaultNs)
+		return InwxSandboxDefaultNs
```
svenpeter42, line 260: I originally had the code like this as well, but apparently golint prefers to not have the `else` clause if the `if` clause already contains a `return` statement :man_shrugging:

svenpeter42 (changes requested):
Just some minor comments from my side. Looks great overall!

providers/inwx/inwxProvider.go
patschi: A specific domain can only be registered/used in one INWX account, so even when multiple accounts are used, it should be good. But yes, it would be cleaner the way you suggested it.

patschi: I've reworked it.

providers/inwx/inwxProvider.go
patschi: tbh, I don't understand why `golint` should complain about that. It works just fine :D

tlimoncelli: Lint complains because it creates a situation where code is never executed.

```
func (api *inwxAPI) getDefaultNameservers() []string {
  if api.sandbox {
    return InwxSandboxDefaultNs
  } else {
    return InwxProductionDefaultNs
  }
  // Suppose the next developer adds this line:
  fmt.Printf("I can't figure out why this never executes!")
}
```

patschi: Ah, I see! Any suggestion how to solve it somehow else, or should we leave it as-is?

tlimoncelli: The Go style is, when possible, return errors early:

```go
func (api *inwxAPI) getDefaultNameservers() []string {
  if validate something is bad {
    return fmt.Errorf(...)
  }
  if more input validation {
    return fmt.Errorf(...)
  }
  // now code can assume clean input, which makes it easier to write code
  x := api.foo
}
```

Consider this code which is anti the style guide. It creates many levels of indentation which is hard to read and difficult to maintain.  The error message is far away from the code that found the error.  Also, adding new validation checks cause you to re-indent your code. Sure, gofmt does that for you, but the diff will be huge.

```go
func (api *inwxAPI) getDefaultNameservers() []string {
  if validate something is ok  {
    if validate something else is ok {
      if validate another thing {
        // all input is good. now I can write code here
        // that is clean.
      } else {
        return error from "another thing"
    } else {
      return error from "something else"
  } else {
    return error from "something is ok"
  }
}
```

patschi: So basically:
```go
func (api *inwxAPI) getDefaultNameservers() []string {
	if api.sandbox {
		return InwxSandboxDefaultNs
	}
	return InwxProductionDefaultNs
}
```
That's what I've committed now.

providers/inwx/inwxProvider.go
```diff
@@ -59,11 +53,18 @@ var features = providers.DocumentationNotes{
 	providers.CanAutoDNSSEC:          providers.Unimplemented("Supported by INWX but not implemented yet."),
 	providers.DocOfficiallySupported: providers.Cannot(),
 	providers.DocDualHost:            providers.Can(),
-	providers.DocCreateDomains:       providers.Unimplemented("Supported by INWX but not implemented yet."),
+	providers.DocCreateDomains:       providers.Can("Does only create domain in nameserver and does not order domain."),
```
tlimoncelli, line 56: Change to `providers.Can()`.  DocCreateDomains is not supposed to order the domain. It just adds it to the namespace.

providers/inwx/inwxProvider.go
```diff
@@ -33,14 +33,8 @@ Additional settings available in `creds.json`:
 
 */
 
-// inwxAPI is a thin wrapper around goinwx.Client.
-type inwxAPI struct {
-	client  *goinwx.Client
-	sandbox bool
-}
-
 // InwxDefaultNs contains the default INWX nameservers.
```
tlimoncelli, line 36: Comment should refer to InwxProductionDefaultNs

providers/inwx/inwxProvider.go
```diff
@@ -250,13 +252,17 @@ func (api *inwxAPI) GetDomainCorrections(dc *models.DomainConfig) ([]*models.Cor
 	return corrections, nil
 }
 
-// GetNameservers returns the default nameservers for INWX.
-func (api *inwxAPI) GetNameservers(domain string) ([]*models.Nameserver, error) {
+// getDefaultNameservers returns back string map with default nameservers  based on e.g. sandbox mode.
```
tlimoncelli, line 255: Remove the word "back"

providers/inwx/inwxProvider.go
tlimoncelli: Great!

patschi:
There we go. As always, thanks for the feedback @tlimoncelli! :)

tlimoncelli:
"Make Tom happy"?  There should be more people doing that ;-)

svenpeter42:
Can be merged from my side!

svenpeter42 (approved):

Issue #847, opened 2020-09-04
AUTODNSSEC should have an "ignore" option, which should be the default


tlimoncelli:
Rather than on/off, AUTODNSSEC should have 3 settings: on, off, don't change. The default would be "don't change".

See discussion in https://github.com/StackExchange/dnscontrol/issues/814

patschi:
Just for clarifications and bring in my ideas:
**Existing AUTODNSSEC**
- `AUTODNSSEC` should be deprecated, as it's not suitable for many situations and changing the behavior could result in a disaster for existing providers.
- `AUTODNSSEC` should default to `AUTODNSSEC_ON`, as this is basically currently what it does - for backwards-compatibility reasons.

**New AUTODNSSEC**
- Introducing two options might be easier: `AUTODNSSEC_ON` and `AUTODNSSEC_OFF`, which are exactly doing that: Forcing DNSSEC being on/off at supported providers.
- When none of both mentions options are provided, the logic shouldn't be triggered and basically result in "don't change". (This way NO specific `AUTODNSSEC_IGNORE` or whatsoever is required)

PR #856, opened 2020-09-16
PowerDNS: Implemented AUTODNSSEC

jpbede:
This PR implements AUTODNSSEC for PowerDNS

### DIFF
--- a/providers/powerdns/dnssec.go
+++ b/providers/powerdns/dnssec.go
@@ -1,0 +1,74 @@
+package powerdns
+
+import (
+	"context"
+	"github.com/StackExchange/dnscontrol/v3/models"
+	"github.com/mittwald/go-powerdns/apis/cryptokeys"
+)
+
+// getDNSSECCorrections returns corrections that update a domain's DNSSEC state.
+func (api *PowerDNS) getDNSSECCorrections(dc *models.DomainConfig) ([]*models.Correction, error) {
+	cryptokeys, getErr := api.client.Cryptokeys().ListCryptokeys(context.Background(), api.ServerName, dc.Name)
+	if getErr != nil {
+		return nil, getErr
+	}
+
+	// check if any of the avail. key is active and published
+	hasEnabledKey := false
+	keyID := 0
+
+	if len(cryptokeys) > 0 {
+		for _, cryptoKey := range cryptokeys {
+			if cryptoKey.Active && cryptoKey.Published {
+				hasEnabledKey = true
+				keyID = cryptoKey.ID
+				break
+			}
+		}
+	}
+
+	// dnssec is enabled, we want it to be disabled
+	if hasEnabledKey && !dc.AutoDNSSEC {
+		return []*models.Correction{
+			{
+				Msg: "Disable DNSSEC",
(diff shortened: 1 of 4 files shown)


providers/powerdns/dnssec.go
```diff
@@ -0,0 +1,74 @@
...
+// getDNSSECCorrections returns corrections that update a domain's DNSSEC state.
+func (api *PowerDNS) getDNSSECCorrections(dc *models.DomainConfig) ([]*models.Correction, error) {
+	cryptokeys, getErr := api.client.Cryptokeys().ListCryptokeys(context.Background(), api.ServerName, dc.Name)
+	if getErr != nil {
+		return nil, getErr
+	}
+
+	// check if any of the avail. key is active and published
+	hasEnabledKey := false
+	keyID := 0
```
tlimoncelli, line 18: Does the zero have any significance?  If not, I suggest: `var keyID int`

jpbede: No, fixed that. Thanks for the hint.

tlimoncelli:
Looks good so far.  Is this ready for ready for merge? (and final review)

jpbede:
Yep. This works for me, it disables/enables DNSSEC for my domains.

patschi:
Just to be sure: What's happening with existing PowerDNS configurations where `AUTODNSSEC` is not set? Wouldn't this cause DNSSEC to be disabled?

jpbede:
Yes, that is right.

patschi:
I'd see this as a as it could break or cause major issues with existing deployments. It might be a good idea delaying this change until AUTODNSSEC has been rewritten (see #847), wouldn't it @tlimoncelli? Just my two cents.

tlimoncelli:
I was assuming nobody would be using AUTODNSSEC thus #847 wasn't that urgent.  I now realize was a bad assumption.

None of the providers I have access to support AUTODNSSEC, which makes it difficult for me to write that code.  Can someone else take a stab at it?

patschi:
> Can someone else take a stab at it?

If my suggested way-to-go in #847 is fine, I **might** invest some time next week or so. However: no promises, as it depends a lot how motivated for doing Go-stuff :D

> The workaround is [...] you'll need to be explicit about whether AUTODNSSEC is on or off when you upgrade...

To be honest this change should not be included in a release with that "workaround" without reworking AUTODNSSEC previously. I can imagine this might cause quite some troubles for who's using PowerDNS...

tlimoncelli:
I have some time today to make a draft PR.  I'll need others to test it.

Luckily since Go is statically typed, I can just change this one field and the compiler will tell me what to change.

jpbede:
I can have a look and test the PR at the mid./end of the next week.

tlimoncelli:
I made this: https://github.com/StackExchange/dnscontrol/pull/863

I changed the name of AUTODNSSEC_ON and AUTODNSSEC_OFF to be explicit.  Including neither makes no changes.  (Anyone using the old "AUTODNSSEC" keyword will fail. I consider that a good thing since we've changed behavior).

Dumb question but.... Does anyone ever turn AUTODNSSEC off?  Once you turn it on, wouldn't you leave it on forever?  In that case, instead of that PR, we could just remove the code that disables it.

PR #861, opened 2020-09-19
Use otto.Compile() instead Run() for better error msg

patschi:
Will result in:
```text
executing javascript: Error: domain.tld.js: Line 13:1 Unexpected token ) (and 1 more errors)
```


patschi:
Okay, reverting back. Seems to behave OK when running it for one domain (`dnscontrol preview --domains X`), but not when running it for all domains (`dnscontrol preview`)...

PR #863, opened 2020-09-20
AUTODNSSEC -> AUTODNSSEC_ON

tlimoncelli:
CC @hnrgrgr @aeden @jpbede

I'm changing how AutoDNSSEC works.  In https://github.com/StackExchange/dnscontrol/pull/856 we discussed that there is now way to specify "leave AutoDNSSEC alone".  This means someone that has previously enabled AutoDNSSEC on either axfrddns, dnsimple, or powerdns would get a big surprise when they upgrade to the newest DNSControl.  If they didn't know to add AUTODNSSEC to their dnsconfig.js file, DNSControl would disable AutoDNSSEC.  That would be bad.

I changed AutoDNSSEC from being a bool (true/false) to being either "on", "off", or "" (ignore). 

Please check your provider to make sure I didn't break anything.  If you'd like to discuss this change please join the discussion on https://github.com/StackExchange/dnscontrol/pull/856.


patschi:
Might also be relevant: #847

patschi:
This removes the existing `AUTODNSSEC` switch, which might break existing setups - regarding backwards compatibility. What about using `AUTODNSSEC` automatically for "ignore" to not break existing setups?

tlimoncelli:
How about AUTODNSSEC being a macro that fails with an error like, "AUTODNSSEC is replaced with AUTODNSSEC_ON, please see http://link for more details".  At that link we'll explain how to transition.

patschi:
Maybe ignoring & output a warning about deprecation? This won't break existing setups then, when e.g. people use the dnscontrol docker image:latest in their CI. (Beside the fact blindly using latest-tag in CI is an awful idea anyway).

Until a specific release in the future users should have migrated away from `AUTODNSSEC` and it can be removed completely. My 2 cents.

tlimoncelli:
Warning added.


```
$ dnscontrol  check
WARNING: AUTODNSSEC is deprecated. It is now a no-op.  Please use AUTODNSSEC_ON or AUTODNSSEC_OFF. The default is to make no modifications. This message will disappear in a future release.
No errors.
```

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/847

tlimoncelli:
Ok, now it is ready for testing by @hnrgrgr @aeden @jpbede

jpbede:
@tlimoncelli I've tested it now. I've tested following cases:

- Domain with externally enabled DNSSEC, no AUTODNSSEC in config -> It kept untouched
- Domain with externally enabled DNSSEC, AUTODNSSEC in config -> Warning
- Existing Domain (in config) with AUTODNSSEC in config -> Warning
- Existing Domain (in config) with AUTODNSSEC_ON -> Enables DNSSEC
- Existing Domain (in config) with AUTODNSSEC_OFF -> Disables DNSSEC

hnrgrgr:
@tlimoncelli I tested AXFRDDNS, LGTM.

tlimoncelli:
Thanks, everyone!  I like this new syntax a lot better!

Issue #848, opened 2020-09-05
2FA problematic with INWX


patschi:
I'm currently trying to migrate to INWX and take use of the great implementation by @svenpeter42. In my setup I'm using the `totp-key` (encrypted, etc) option, which seems to basically generate the TOTP during runtime just fine (verified by stepping through the code).

But at some point it seems that the TOTP is not valid anymore for some reason, see below. Absolutely nothing changed between the executions, just a few seconds passed by. So it might be a timing issue. Was someone able to see the same behavior?

See:
```
$ date; dnscontrol preview --domains domain.tld; echo
Sat Sep  5 04:03:21 BST 2020
INWX: Could not unlock account: (2200) Authentication error. Reason: (INVALID) The confirmation code is invalid.

$ date; dnscontrol preview --domains domain.tld; echo
Sat Sep  5 04:03:27 BST 2020
INWX: Could not unlock account: (2200) Authentication error. Reason: (INVALID) The confirmation code is invalid.

$ date; dnscontrol preview --domains domain.tld; echo
Sat Sep  5 04:03:32 BST 2020
******************** Domain: domain.tld
----- Getting nameservers from: inwx
----- DNS Provider: inwx...0 corrections
----- Registrar: inwx...ERROR
Error getting corrections: Error 404: "This service does not exist" <------ means that the login worked fine however
Done. 0 corrections.
completed with errors
```

What I can further rule out:
- The TOTP codes generated from dnscontrol are working fine, as I can use them to login in the WebUI just fine. So the basic functionality works.
- I can rule out local time drifts on my machine: 1) TOTP codes generated from my local password manager works, 2) checked time and regularly syncing it with NTP servers.
- Without 2-Factor-Authentication all authentications work fine, so it seems to be TOTP-related.

svenpeter42:
I think I know what's happening. INWX will only accept a single TOTP code once. This means you can only login once every 30 seconds.

Do you have INWX setup as both the registrar and the DNS provider? If so, dnscontrol will probably try to login twice within a few seconds. The first login will work fine but the second login will try to use the same TOTP code and will then fail. I'll have to think if there's an easy way to reuse an already existing session to work around that issue.

patschi:
I assumed something like that too (but rather inwx overwriting session or so) - haven't mentioned that and edited that wrong in the output above, apologies. I've noticed this also as the code stepped through getOtp twice during runtime. 

[...]

PR #865, opened 2020-09-20
INWX: add additional documentation about 2FA limitations

svenpeter42:
see #848


tlimoncelli:
Thanks for the clarification! That's going to help a lot of new users!

Issue #858, opened 2020-09-16
extend D_EXTEND to handle subdomains


ad8-bdl:
I'm finding it beneficial to split out my DNS config to separate files, per sub-domain, using the `require` and `require_glob` functions. These subdomains are not delegated, so I'm using the `D_EXTEND` function to extend the main domain. I then have to manually include the subdomain in every single record, hopefully without typos or other errors.

It would be neat if there were a `D_SUBDOMAIN` function that was the same as `D_EXTEND` but accepted a single additional required argument of a path to be inserted between the RRs and domain name. e.g. something like

    D('example.com', REG_NONE,  DnsProvider(DNS_CLOUDFLARE), ...)

    D_EXTEND('example.com',
      A('x.subdomain', '10.1.1.1')
      A('y.subdomian', '10.1.1.2')
    );

    D_SUBDOMAIN('example.com', 'subdomain',
      A('x', '10.1.1.1')
      A('y', '10.1.1.2')
    );

tlimoncelli:
That's an interesting use-case!

Let me suggest an alternative syntax:

```
D('example.com', REG_NONE,  DnsProvider(DNS_CLOUDFLARE), 
    SUBDOMAIN('subdomain', [
        A('x', '10.1.1.1')
        A('y', '10.1.1.2')
    ])
);
```

That would work on both `D()` and `D_EXTEND()`.

The `SUBDOMAIN()` function could be written in Javascript and included in the dnsconfig.js file.  It would require a bit more Javascript know-how than I have.

tlimoncelli:
Question: Given the domain `.foo.tld` and subdomain `.bar`, what would `CNAME("a", "b")` be the equivalent of:

```
      CNAME("a.bar", "b.bar.foo.tld")
```

or

```
      CNAME("a.bar", "b.foo.tld")
```

ad8-bdl:
Implementing this as a 'Domain Modifier' within a `D` or `D_EXTEND` would work for me. I had a go at implementing this as JS but realised my JS skills are very atrophied, moreover and I decided that this approach was contrary to KISS / YSMFN (https://stackexchange.github.io/dnscontrol/code-tricks). Maybe there's a simpler approach than the hack I was creating.

As to the `CNAME`: I'd expect it to be consistent with existing behaviour: a fully qualified target (ends in `.`) is as-is, anything else is tacked on to "the domain" as it is at the time the CNAME is called. i.e. including the subdomain.

[...]

PR #870, opened 2020-09-24
extend D_EXTEND to handle subdomains

ad8-bdl:
Ref. #858


tlimoncelli:
I added some parse_tests to verify things work as expected on the edge cases I mentioned before.  You can merge in my changes from https://github.com/StackExchange/dnscontrol/pull/872

Currently those tests fail. Please check my work... it's certainly possible my test data is invalid.

That said, I think the problem is with `_getDomainObject()`.  I think the algorithm should be along the lines of

* Split foo.bar.com into ["foo", "bar", "com"]
* Search for "foo.bar.com"
* Search for "bar.com"
* Search for "com"
* Return null

ad8-bdl:
Will do - how do I run these tests?

________________________________

tlimoncelli:
> Will do - how do I run these tests?

    go test -count=1 pkg/js/*.go

or

    cd pkg/js && go test -count=1

ad8-bdl:
On `_getDomainObject`: oops, yes the previous approach returned on the first match, not necessarily the best match. Fixed.

`_getDomainObject` still uses a 'truncated string comparison' approach rather than explicitly breaking down the paths to arrays and doing a piece-wise comparison of those paths, as the string comparison is much simpler than effectively the same process on arrays.

mikenz:
I love this proposed change! It's going to make my config so much easier to follow.

ad8-bdl:
Sooo, re. CNAMEs and its siblings: making the target fully qualified in `helpers.js` (as 929cd4711d323c5ce24a0a42f301774cc93967fc does) is not a great approach, as it makes CNAMEs/etc behave differently depending on whether they are in a D() or D_EXTEND().

Instead it seems much better to propagate the information on the subdomain through to `validate:ValidateAndNormalizeConfig`, so that it can be properly assembled in the canonicalise step. Same for all of CNAME, MX, NAPTR, NS and SRV. (I tested all bar NAPTR on Cloudflare).

ad8-bdl:
Ok, so with that - I believe it's good to go.

tlimoncelli:
I've been busy with other stuff for a bit so I may be a little out of touch and/or confused.

I think https://github.com/StackExchange/dnscontrol/pull/872 needs to be merged into https://github.com/StackExchange/dnscontrol/pull/870 before it can be merged to master.

I'll be focusing on dnscontrol a lot the next few days.  Let me know how I can help!

ad8-bdl:
I hope I've got that right...

tlimoncelli:
We're so close!  Sadly I found a bug:

```
D("domain.tld", REG, DnsProvider(DNS),
  A("@", "127.0.0.1"),
  A("a", "127.0.0.2"),
  CNAME("b", "c")
);

D("sub.domain.tld", REG, DnsProvider(DNS),
  A("@", "127.0.1.1"),
  A("aa", "127.0.1.2"),
  CNAME("bb", "cc")
);


// Should extend domain.tld, but currently extends sub.domain.tld 
D_EXTEND("ssub.domain.tld",
  A("@", "127.0.0.7"),
  A("j", "127.0.0.8"),
  CNAME("k", "l")
);

```

tlimoncelli:
Here's a PR with the failing test: https://github.com/StackExchange/dnscontrol/pull/885

tlimoncelli:
Merged as part of https://github.com/StackExchange/dnscontrol/commit/b275286dae653b0c3193515897b087d227c3c66b

PR #872, opened 2020-09-25
Add tests and edit docs

tlimoncelli:


pkg/js/parse_tests/028-dextend.js
```diff
@@ -0,0 +1,22 @@
+var REG = NewRegistrar("Third-Party", "NONE");
+var CF = NewDnsProvider("Cloudflare", "CLOUDFLAREAPI");
+
+// Zone and subdomain Zone:
+D("foo.com", REG, DnsProvider(CF),
+  A("@", "10.1.1.1"),
+  A("www", "10.2.2.2")
+);
+D("bar.foo.com", REG, DnsProvider(CF),
```
ad8-bdl, line 9: D_EXTEND?

tlimoncelli: No, actually, the comment should be "Two zones with similar names".

tlimoncelli:
Any thoughts on _getDomainObject?

pkg/js/parse_tests/028-dextend.js
ad8-bdl: Ok; practically speaking, this doesn't work for Cloudflare at least as it won't allow a "delegated" domain to itself. i.e. the above config will give the following (where etanidua.com is the domain registered with cloudflare; i.e. s/foo.com/etanidua.com/):
    nameservers for bar.etanidua.com not found in cloudflare account

tlimoncelli: Yup, a lot of providers would be in a similar situation.  BIND and others support it.  I chose that particular example because it is a harder test of the best match algorithm. :)

Issue #849, opened 2020-09-06
Possible ClouDNS regression on 3.3.0


marcolussetti:
Running on go1.15 / Manjaro Linux 20 (normally running on Windows).

On 3.3.0:
![image](https://user-images.githubusercontent.com/15794684/92331432-6bbc8b80-f02b-11ea-92ae-21f0942bc2ca.png)

On 3.2.0:
![image](https://user-images.githubusercontent.com/15794684/92331465-a8888280-f02b-11ea-9b99-bd5b14fb7956.png)

If I comment out all of my ClouDNS domains (leaving Cloudflare only), we're back in business:
![image](https://user-images.githubusercontent.com/15794684/92331513-f8674980-f02b-11ea-92f2-24fb4d465375.png)

tlimoncelli:
CC @pragmaton

pragmaton:
Hm, I see.  Shame on me, be we have not upgrade yet  :)  Unfortunately, I'm a bit busy now and  will have time to check only  next week. @marcolussetti  feel free to ping me if you will not get feedback till next Friday.

df-a:
Same issue on MacOS Catalina. Subscribing.

marcolussetti:
Fixed in 3.4.2 by #875

PR #875, opened 2020-09-27
ClouDNS: Return API error only when body unmarshals to errorResponse

pmaene:
This patch resolves the regression reported in #849. Go's `json.Unmarshal` would return a `json.UnmarshalTypeError` error when attempting to unmarshal successful responses to `cloudns.errorResponse`. I changed the behaviour to only return an error to the caller when this unmarshal operation does not fail, but I'm happy to revise this if another approach is preferred.


tlimoncelli:
Thanks for fixing this!

pmaene:
No problem, it was an easy issue to resolve! Thanks for merging my fix.

PR #877, opened 2020-09-29
Update CNAME.md

ad8-bdl:
The current example
```
CNAME("def", "test.subdomain"), // def.example.com -> test.subdomain.example.com
```
is invalid (raises a validation error, "ERROR: in CNAME def.example.com: target (test.subdomain) must end with a (.)")

### DIFF
--- a/docs/_functions/domain/CNAME.md
+++ b/docs/_functions/domain/CNAME.md
@@ -17,7 +17,7 @@ Target should be a string representing the CNAME target. If it is a single label
 D("example.com", REGISTRAR, DnsProvider("R53"),
   CNAME("foo", "google.com."), // foo.example.com -> google.com
   CNAME("abc", "@"), // abc.example.com -> example.com
-  CNAME("def", "test.subdomain"), // def.example.com -> test.subdomain.example.com
+  CNAME("def", "test"), // def.example.com -> test.example.com
 );
 
 {%endhighlight%}


ad8-bdl:
> The code that adds the subdomain to the label is in helpers.js. Maybe adding the subdomain to the target should be there too?

I don't think that's correct; `helpers.js` passes on the target unchanged (CNAME is just an unordained recordBuilder instance, no transform/etc). Hacking helpers.js to add the (sub)domain was what started me down this path.

From what I can tell so far, ValidateAndNormalizeConfig first validates the target and then canonicalises it; the validation of a relative name with dots fails before it can get a chance to be canonicalised. I think the right approach here might be for CNAME to call checkTarget but skip the ContainsRune check.

tlimoncelli:
> > The code that adds the subdomain to the label is in helpers.js. Maybe adding the subdomain to the target should be there too?
> 
> I don't think that's correct; `helpers.js` passes on the target unchanged (CNAME is just an unordained recordBuilder instance, no transform/etc). Hacking helpers.js to add the (sub)domain was what started me down this path.
> 
> From what I can tell so far, ValidateAndNormalizeConfig first validates the target and then canonicalises it; the validation of a relative name with dots fails before it can get a chance to be canonicalised. I think the right approach here might be for CNAME to call checkTarget but skip the ContainsRune check.

Oh yes!   The canonicalization happens in ValidateAndNormalizeConfig.  Correct.  I was thinking that we would prefix the target with the subdomain in the JS, then let the canonicalization happen as normal. 

Either way should work.  You're deeper into the code at this point so I'll leave it to you.

Tom

tlimoncelli:
Here are some test cases I thought of.  I'm unsure what the expected outcome should be for the 2nd CNAME.

```
D_EXTEND("sub.example.tld",
    CNAME("a", "b"); // a.sub.example.tld -> b.sub.example.tld
    CNAME("a", "@"); // a.sub.example.tld -> sub.example.tld or example.tld ???
    CNAME("a", "sub.example.tld."); // a.sub.example.tld -> sub.example.tld
    CNAME("a", "x.y"); Error. Contains dot but doesn't end with dot.
    CNAME("a", "otherdomain.tld."); // a.sub.example.tld -> otherdomain.tld
    // Also test for MX, NS, ANAME, SRV.
    // Not sure if PTR needs any special treatment. Haven't thought about it much.
```

ad8-bdl:
I've come to the conclusion that the as-_documented_ behaviour for CNAME (and the other records that accept a label/name target) is correct: either a plain label - no dots, or a FQDN. i.e. the whole https://stackexchange.github.io/dnscontrol/why-the-dot business.

That is, the current doc has an error in the example, and that error is corrected with this PR.

As to managing with the D_EXTEND subdomain, that's a different topic, and as it happens I have a "right" approach for that too, but I'll follow up on the other PR (#870) on that.

tlimoncelli:
Sounds good!  Thanks for following me through my detour!

Issue #886, opened 2020-10-07
D_EXTEND does not always match correctly.


tlimoncelli:
If domain `do.tld` and `sub.do.tld` exist, then `D_EXTEND("ub.do.tld", ...` will extend `sub.do.tld`, which is incorrect.  This will be fixed before the next release.

PR #888, opened 2020-10-07
Fix D_EXTEND matching bug

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/886

* Change js_test.go to use a more standard JSON comparison
* Reformat some of the js/parse_test files
* Fix https://github.com/StackExchange/dnscontrol/issues/886


tlimoncelli:
CC: @ad8-bdl

tlimoncelli:
Thanks, @ad8-bdl !  I couldn't have done this without you!  I have limited Javascript skills but I was able to build on top of your code.

Issue #893, opened 2020-10-08
Hexonet provider no longer compiles


tlimoncelli:
CC: @papakai

Regular `go get github.com/StackExchange/dnscontrol` does not work because of an incompatibility with the provider's module.  Please submit a PR with the newest .  The default module to download for the provider has made incompatible changes.   We are getting many reports about this (https://github.com/StackExchange/dnscontrol/issues/890 is one example).

I can't fix this because I do not have an account on the provider's system.  I need a volunteer to do this.

Generally the fix looks like this:

```
env | grep ^GO
# "unset" any GO* environment variables
unset GOPATH GO15VENDOREXPERIMENT
go get -d github.com/StackExchange/dnscontrol
cd ~/go/src/github.com/StackExchange/dnscontrol/
go test ./...
go get -u github.com/hexonet/go-sdk
go test ./...       # this will fail. Please fix.
# Once that works, try the integration test:
export HEXONET_ENTITY="redacted.tld"
export HEXONET_UID="redacted" 
export HEXONET_PW="redacted"
# Edit providers.json and change "HEXONET "domain" to a domain you can overwrite.
cd integrationTest && go test -v -verbose -provider HEXONET
# Fix problems until all tests pass.
```

---

Here is a demonstration of the `go get` problem:

```
$ go get -v github.com/StackExchange/dnscontrol
github.com/StackExchange/dnscontrol/pkg/acme
github.com/StackExchange/dnscontrol/providers/hexonet
github.com/StackExchange/dnscontrol/providers/vultr
# github.com/StackExchange/dnscontrol/providers/hexonet
/Users/tlimoncelli/go/src/github.com/StackExchange/dnscontrol/providers/hexonet/domains.go:7:41: cannot use map[string]string literal (type map[string]string) as type map[string]interface {} in argument to n.client.Request
/Users/tlimoncelli/go/src/github.com/StackExchange/dnscontrol/providers/hexonet/domains.go:13:41: cannot use map[string]string literal (type map[string]string) as type map[string]interface {} in argument to n.client.Request
/Users/tlimoncelli/go/src/github.com/StackExchange/dnscontrol/providers/hexonet/nameservers.go:44:41: cannot use map[string]string literal (type map[string]string) as type map[string]interface {} in argument to n.client.Request
[...]

PR #898, opened 2020-10-08
HEXONET: Implement get-zones

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/893

* `go get -u github.com/hexonet/go-sdk`
* Fix code to match new API/signatures
* Fix broken integration tests
* Implement get-zones, check-creds


providers/hexonet/domains.go
```diff
@@ -24,3 +26,39 @@ func (n *HXClient) EnsureDomainExists(domain string) error {
 	}
 	return nil
 }
+
+// ListZones lists all the
+func (n *HXClient) ListZones() ([]string, error) {
+
+	// Basic
+
+	r := n.client.Request(map[string]interface{}{
```
papakai, line 35: nearly :-)

the main "problem" is that our API limits the rows returned by list queries by default (in case parameter LIMIT is not set). This is for security and performance reasons.

Check this method `RequestAllResponsePages` instead: https://github.com/hexonet/go-sdk/blob/master/apiclient/apiclient.go#L437

```go
r := n.client.RequestAllResponsePages(map[string]interface{}{
	"COMMAND": "QueryDNSZoneList",
        "FIRST": "0",
        "LIMIT": "1000"
})
//now you have an array of API responses available to loop over.
```

This also applies to the below Premium DNS query accordingly.

providers/hexonet/records.go
```diff
@@ -182,7 +186,7 @@ func (n *HXClient) updateZoneBy(params map[string]string, domain string) error {
 func (n *HXClient) getRecords(domain string) ([]*HXRecord, error) {
 	var records []*HXRecord
 	zone := domain + "."
-	cmd := map[string]string{
+	cmd := map[string]interface{}{
 		"COMMAND":  "QueryDNSZoneRRList",
```
papakai, line 190: Please be so kind as to use here also `RequestAllResponsePages` to query all data and loop over the responses.
Sorry for the circumstance and the effort you guys have. I am impressed how well and fast you went through the changes! Very very appreciated. I am unlucky that we have to enforce that way.

We are planning a new API behind the scenes that finally leads to less code in DNSControl and easier extensions if necessary. Finally you won't have to deal with commands, just a nice SDK communicating in json. It will take some time until we come there, but then! :-)

papakai (changes requested):

papakai:
One additional thing: Would you mind to add the use of method `SetUserAgent` in [here](https://github.com/StackExchange/dnscontrol/blob/f92e08c32d38ab63fbcc66d86295c625ee8069d9/providers/hexonet/hexonetProvider.go#L35)

e.g.

```go
api := &HXClient{
    client: hxcl.NewAPIClient(),
}
//load the current version of DNSControl
//version := DNSControl.version
//... and provide it as argument
api.client.SetUserAgent("DNSControl", version)
```

That would be great and would help us identifying from which software the request comes from. I have just no idea if the current software release in use of dnscontrol is accessible and how. Would be great to have that too.

Thanks in advance!

papakai:
this PR will fix #628 for HEXONET. :-)

papakai:
fyi: I created a blog post internally to let the whole Reseller Division of CentralNic Group PLC know what great guys you are at DNSControl. (they have acquired HEXONET a year ago)

providers/hexonet/records.go
tlimoncelli: Ok, I tried that but I have some bad news and some good news.

The bad news it that when I use `RequestAllResponsePages` in the same way that I used it for `GetZones()` I get this error:

```
    integration_test.go:173: runTests: Failed loading resource records for zone a-b-c-movies.com. [%!s(int=503) Invalid Parameter: FIRST]
```

Maybe I'm not doing it right?  I made nearly the exact same change as for GetZones.

The good news is that without any changes the code works for our integration test where we create/modify/delete zones of size 101, 601, and 1201.  The output of the tests is here: https://gist.github.com/tlimoncelli/a1869aee6a6f1e61379790473cded51d

I increased the testing to 9999 records and the entire create/modify/delete cycle worked just fine.  Maybe it is fine as-is?

papakai: Ah, my fault. That's probably a wrong behavior/incorrect implementation of command `QueryDNSZoneRRList` in our API. In general, all our list query commands are limited or they should be.
So using `RequestAllResponsePages` is valid for the two use cases around the `QueryDNSZoneList` command, but not for the command `QueryDNSZoneRRList`. So, use the `Request` method for `QueryDNSZoneRRList` and the `RequestAllResponsePages` method for `QueryDNSZoneRRList`.
Why? Well, looks like `QueryDNSZoneRRList` is not limited and returns all records needed, but `QueryDNSZoneList` is limited to return 10000 zones at max.

Sorry for the confusion. HTH.

tlimoncelli: I've tested it up to 15,000 records in a zone.  No problem so far.

tlimoncelli:
All tests pass.  We're not using "RequestAllResponsePages" everywhere but if the tests pass, I'm ok with it.  We can always improve things in the future.  Thanks for all your help, @papakai !

papakai:
Thanks again for your time spent, very appreciated @tlimoncelli.

Issue #891, opened 2020-10-08
Route53 MODIFY only modifies 1 record instead of all


mludvig:
I'm bringing a number of our zones under _dnscontrol_ and realised that some of them have records with a missing trailing dot. Although _Route53_ doesn't seem to mind and resolves them correctly _dnscontrol_ complains. However when I fix the records in `dnsconfig.js` in each run `dnscontrol` only fixes one record.

This is what was fetched by `dnscontrol get-zone`, the trailing dots are actually missing in our Route53 zone:

```
	CNAME('test1', 'test1.example.com'),
	CNAME('test2', 'test2.example.com'),
	CNAME('test3', 'test3.example.com'),
```

The same file can't be previewed and pushed: "ERROR: in CNAME test1.smarterdata.nz: target (test1.example.com) must end with a (.)" - that's ok. 

So I add the trailing dots, preview and push:

```
/dns # dnscontrol push --domains smarterdata.nz
******************** Domain: smarterdata.nz
----- Getting nameservers from: Route53
----- DNS Provider: Route53...1 correction
#1: 
MODIFY CNAME test2.smarterdata.nz: (test2.example.com ttl=300) -> (test2.example.com. ttl=300)
MODIFY CNAME test3.smarterdata.nz: (test3.example.com ttl=300) -> (test3.example.com. ttl=300)
MODIFY CNAME test1.smarterdata.nz: (test1.example.com ttl=300) -> (test1.example.com. ttl=300)
SUCCESS!
----- Registrar: DiscountDomains...0 corrections
Done. 1 corrections.
```

However by now only one record has been updated. It's not a timing issue, any amount of waiting doesn't matter. **I have to run push 3x to update them all.**

```
/dns # dnscontrol push --domains smarterdata.nz
******************** Domain: smarterdata.nz
----- Getting nameservers from: Route53
----- DNS Provider: Route53...1 correction
#1: 
MODIFY CNAME test2.smarterdata.nz: (test2.example.com ttl=300) -> (test2.example.com. ttl=300)
MODIFY CNAME test3.smarterdata.nz: (test3.example.com ttl=300) -> (test3.example.com. ttl=300)
SUCCESS!
----- Registrar: DiscountDomains...0 corrections
Done. 1 corrections.


/dns # dnscontrol push --domains smarterdata.nz
******************** Domain: smarterdata.nz
----- Getting nameservers from: Route53
----- DNS Provider: Route53...1 correction
#1: 
[...]

PR #901, opened 2020-10-09
ROUTE53: Document CNAME bug

tlimoncelli:
See https://github.com/StackExchange/dnscontrol/issues/891

### DIFF
--- a/docs/_providers/route53.md
+++ b/docs/_providers/route53.md
@@ -116,6 +116,9 @@ delegation set.  Note that you you only want the portion of the `Id` after the `
 > Delegation sets only apply during `create-domains` at the moment. Further work needs to be done to have them apply during `push`.
 
 ## Caveats
+
+### Route53 errors if it is not the DnsProvider
+
 This code may not function properly if a domain has R53 as a Registrar
 but not as a DnsProvider.  The situation is described in
 [PR#155](https://github.com/StackExchange/dnscontrol/pull/155).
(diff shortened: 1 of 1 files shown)


cslee00:
Have been working with AWS Support on what is likely the same issue - not with CNAMEs, but with CAA record that does not update as part of a batch, yet updates when done by itself.

```
▶︎ Yes, you are correct. The first API call did NOT push any changes to Route53. The reasoning for this is because the values were the same as what existed, however just in a different order.

▶︎ The second API call submitted DID change the records in R53 however, so why?
The team has explained that since it was the only change set in this API call, due to internal mechanisms it proceeded with the change even though the values were the same, but in a different order. This behavior more has to do with evaluating if any changes are needed during an UPSERT request.

Next Actions from the AWS side:
The team is going to discuss if this difference in behavior warrants changes to how these mechanisms operate.```

Our feedback: ```
To your points on the API call behaviour:

1) The first API call that succeeded (but not really): those aren't the same values as what existed.  They're in a different order.  Not only is this undocumented behaviour, R53 should not presume that order is insignificant to consumers of DNS records.

2) The second API call that _actually_ updated the record.  Seriously: different behaviour when it's only a single record?  Also not documented, nor does that make any practical sense.  Just update the damn record _every_ time.
```

AWS has acknowledged the janky behaviour and is working through assessing what remediation looks like.

Issue #890, opened 2020-10-08
Unable to "go get" - build fails


mludvig:
go version go1.15.2 linux/amd64 on Ubuntu 18.04

```
~ $ go get -v github.com/StackExchange/dnscontrol
github.com/StackExchange/dnscontrol/pkg/acme
github.com/StackExchange/dnscontrol/providers/hexonet
# github.com/StackExchange/dnscontrol/pkg/acme
go/src/github.com/StackExchange/dnscontrol/pkg/acme/acme.go:143:36: not enough arguments in call to client.Certificate.Renew
	have (certificate.Resource, bool, bool)
	want (certificate.Resource, bool, bool, string)
# github.com/StackExchange/dnscontrol/providers/hexonet
go/src/github.com/StackExchange/dnscontrol/providers/hexonet/domains.go:7:41: cannot use map[string]string literal (type map[string]string) as type map[string]interface {} in argument to n.client.Request
go/src/github.com/StackExchange/dnscontrol/providers/hexonet/domains.go:13:41: cannot use map[string]string literal (type map[string]string) as type map[string]interface {} in argument to n.client.Request
go/src/github.com/StackExchange/dnscontrol/providers/hexonet/nameservers.go:44:41: cannot use map[string]string literal (type map[string]string) as type map[string]interface {} in argument to n.client.Request
go/src/github.com/StackExchange/dnscontrol/providers/hexonet/nameservers.go:97:31: cannot use cmd (type map[string]string) as type map[string]interface {} in argument to n.client.Request
go/src/github.com/StackExchange/dnscontrol/providers/hexonet/records.go:175:23: cannot use cmd (type map[string]string) as type map[string]interface {} in argument to n.client.Request
go/src/github.com/StackExchange/dnscontrol/providers/hexonet/records.go:191:23: cannot use cmd (type map[string]string) as type map[string]interface {} in argument to n.client.Request
github.com/StackExchange/dnscontrol/providers/vultr
# github.com/StackExchange/dnscontrol/providers/vultr
go/src/github.com/StackExchange/dnscontrol/providers/vultr/vultrProvider.go:123:103: cannot use r.Priority (type *int) as type int in argument to api.client.DNSRecord.Create
go/src/github.com/StackExchange/dnscontrol/providers/vultr/vultrProvider.go:196:35: cannot convert r.Priority (type *int) to type uint16
[...]

PR #902, opened 2020-10-09
Go module build improvements

pgaskin:
- Add module info to version
- Add warning if built without Go modules
- Update source installation instructions for Go modules

see #890

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"runtime/debug"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -15,6 +16,9 @@ import (
 
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
+	if info, ok := debug.ReadBuildInfo(); !ok && info == nil {
+		fmt.Fprint(os.Stderr, "Warning: dnscontrol was built without Go modules. See https://github.com/StackExchange/dnscontrol#from-source for more information on how to build dnscontrol correctly.\n\n")
```
pgaskin, line 20: @tlimoncelli, what do you think of this message?

main.go
```diff
@@ -36,6 +40,9 @@ func versionString() string {
 	} else {
 		version = fmt.Sprintf("%s-dev", Version) // no SHA. '0.x.y-dev' indicates it is run from source without build script.
 	}
+	if info, ok := debug.ReadBuildInfo(); !ok && info == nil {
+		version += " (non-modules)"
```
pgaskin, line 44: Note that it's also possible to list all module versions if that's something you're interested in.

main.go
tlimoncelli: :+1:

main.go
tlimoncelli: We can add that later. I think this is good for now.

tlimoncelli (approved):

PR #918, opened 2020-10-28
Document: CLI variables

### LINKED ISSUE #916: DOCUMENT: How to use `-v var=value`
This could use a page in the docs folder.

jpbede:
This is my initial version of the CLI variables doc.

@tlimoncelli What is your opinion ?

Fixes #916


tlimoncelli:
Hey @liftedkilt and @rbelnap!

I think this feature will simplify your configuration.  Can you suggest a use-case for the docs?

tlimoncelli:
Just thinking out loud...

I think it would be good to have 2 examples. The first should be very simple, just to introduce the feature. The second should be a more complex example.

Simple case:

* An IP address that defaults one way, but in an emergency can be pointed to a failover site.  `-v webIP=99.99.99.99`

Complex case:

* A "second pass" that does something special:

    dnscontrol push && dnscontrol push -v foo=bar -domains=special.tld,special2.tld 

Maybe the second pass would enable a different DNS provider and do something the first provider can't do.  For example, the default case uses an ALIAS but the other case uses a AZURE_ALIAS.

Lastly, let's have a paragraph like in `docs/_functions/global/D_EXTEND.md` that warns people against complexity.  As Brian Kernighan said, "Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it."

jpbede:
> As Brian Kernighan said, "Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it."

Agreed :) 

Sure, i'm for ideas. This was my first draft and just a write down to work with.

rbelnap:
Well isn't that just fantastic!  Glad to see this will be on the way!  My use case is internal/external resolution, where bind is running internally, and handing out different addresses than publicly available servers.  The gist is:

```
if (view == "internal") {
    var host01 = "192.168.0.16";
    var host02 = "192.168.0.17";
    var host03 = "192.168.0.18";
} else {
    var host01 = "10.0.0.16";
    var host02 = "10.0.0.17";
    var host03 = "10.0.0.18";
}


D("example.org", registrar, DnsProvider(public), DnsProvider(bind),
  A('sitea', host01, TTL(1800)),
  A('siteb', host01, TTL(1800)),
  A('sitec', host02, TTL(1800)),
  A('sited', host02, TTL(1800))
)
```
Then the internal nameserver serves zones generated with:
`dnscontrol push -v view=internal --providers bind`

While external dns (the default case) is updated with just
`dnscontrol push`

assuming that creds.json includes:
```  
"bind":{
    "_exclude_from_defaults": "true",
  }
```

The internal nameserver will then resolve sitea.example.org to 192.168.0.16, while the public provider resolves it as 10.0.0.16 - while maintaining the repository as the single source of truth.

jpbede:
@tlimoncelli added @rbelnap's use case and added your hint that warns people against complexity.

tlimoncelli:
Suggestion: Commit this as-is and improve it over time.

jpbede:
Yea I would like that

PR #919, opened 2020-10-29
testing Go 1.14

atma-stackoverflow:
running `go run build/build.go` produces an error indicating go 1.14 is required, testing that


atma-stackoverflow:
this did not fix it, confirmed it is not based on Go version

PR #926, opened 2020-11-01
HETZNER: fix retry of POST/PUT requests -- rebuild request body

das7pad:
Followup for #904 

Previously for any retry the request body was already consumed and
 the server received an empty body.
Which in turn results in a response with a status code of 400 for the retied request.

This is best reviewed with white-space changes ignored: https://github.com/StackExchange/dnscontrol/pull/926/files?diff=unified&w=1


tlimoncelli:
Awesome!

PR #928, opened 2020-11-02
Run integration tests as part of build checks

MaxHorstmann:
Following up on https://github.com/StackExchange/dnscontrol/pull/922 and https://github.com/StackExchange/dnscontrol/pull/927, this also runs a set of integration tests as part of build checks.

Branch protection of `master` is in place, i.e. changes which break any of the integration tests can't be merged.

[Repo secrets](https://github.com/StackExchange/dnscontrol/settings/secrets) added as necessary.

PR #930, opened 2020-11-04
Run integration tests for additional DNS providers

MaxHorstmann:
Follow-up to https://github.com/StackExchange/dnscontrol/pull/928: add

* Azure DNS
* NAMEDOTCOM
* GCloud
* Cloudflare

([secrets](https://github.com/StackExchange/dnscontrol/settings/secrets))

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,26 +9,23 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-
     - name: Checkout repo
       uses: actions/checkout@v2
       with:
         fetch-depth: 0
-
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
         go-version: ^1.15
-
     - name: Build binaries
       run: go run build/build.go
-
     - name: Run unit tests
       run: go test ./...
 
-  integration-tests-bind:
-    runs-on: ubuntu-latest
-    needs: build
+  integration-tests:
+    strategy:
+      matrix:
+        provider: [BIND, AZURE_DNS, NAMEDOTCOM, ROUTE53, GCLOUD, CLOUDFLAREAPI]
     steps:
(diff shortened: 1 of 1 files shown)


tlimoncelli:
The integration tests can run in parallel (they all use different scratch DNS domains).  Can github actions do that?

MaxHorstmann:
GH can't run steps in parallel *yet* ([common request](https://github.community/t/steps-in-parallel/16343)). It can run jobs in paralllel. I'll check how hard it would be to refactor this into separate jobs.

MaxHorstmann:
@tlimoncelli @atma-stackoverflow this works great (faster than TC due to parallel execution!) and is another nice shippable piece. Mind taking a quick look?

.github/workflows/build.yml
```diff
@@ -5,34 +5,59 @@ on:
...
-      working-directory: integrationTest
-      run: go test -v -verbose -provider BIND
-
-    - name: Run integration test for ROUTE53 provider
+  integration-tests:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        provider: [BIND, AZURE_DNS, NAMEDOTCOM, ROUTE53, GCLOUD, CLOUDFLAREAPI]
```
tlimoncelli, line 30: Minor nit.  Sort the names of the providers and the list of env variables.

tlimoncelli (approved):

.github/workflows/build.yml
MaxHorstmann: ok! https://github.com/StackExchange/dnscontrol/commit/f145ae14386f92a04408851e5f5efaa725e898fe

I originally kept them in the same order as in TC so they would run in the same sequence. But they're running in parallel now, so there's no sequence and this is a tad more readable.

PR #936, opened 2020-11-08
HETZNER: better rate limit handling

das7pad:
This is a followup for https://github.com/StackExchange/dnscontrol/pull/904#issuecomment-720155114

- Hetzner is using a Proxy service 'kong' which broadcasts it limits
- honor 'Retry-After' of 429 responses
- delay requests per-se: see the amended docs for details

Signed-off-by: Jakob Ackermann <das7pad@outlook.com>


providers/hetzner/api.go
```diff
@@ -236,23 +281,55 @@ func (requestRateLimiter *requestRateLimiter) beforeRequest() {
...
-		// Future proof this provider, use exponential back-off.
-		requestRateLimiter.delay = requestRateLimiter.delay * 2
-		backoffType = "exponential"
+func (requestRateLimiter *requestRateLimiter) setDefaultDelay() {
+	// default to a rate-limit of 1 req/s -- the next response should update it.
+	requestRateLimiter.delay = time.Second
+}
+
+func (requestRateLimiter *requestRateLimiter) setOptimizeForRateLimitQuota(quota string) error {
+	quotaNormalized := strings.ToLower(quota)
```
tlimoncelli, line 290: Suggestion: The code would be more simple if you used ToLower'ed strings.  In other words, if you used `hour` instead of `Hour`, the case statement below could be replaced with just:

    requestRateLimiter.optimizeForRateLimitQuota = quoteNormalized

docs/_providers/hetzner.md
```diff
@@ -54,19 +54,60 @@ At this time you cannot update SOA records via DNSControl.
...
+Depending on how many requests you are planning to perform, you can adjust the
+ delay between requests in order to stay within your quota.
+
+The setting `optimize_for_rate_limit_quota` controls this behavior and accepts
+ a case-insensitive value of
+- `Hour`
+- `Minute`
+- `Second`
+
+The default for `optimize_for_rate_limit_quota` is `Second`.
```
tlimoncelli, line 69: I'm not sure I understand how the ratelimiting works but my concern is that the default should be something reasonable for the typical user.  It might be, I'm just unsure because I don't fully understand how it works.

tlimoncelli (commented):
Looks good so far. Just a few curiosity questions.

tlimoncelli:
I hope you aren't getting many emails about failing tests. The automated testing is new and it is little unstable right now. I hope to fix that soon. In the meanwhile, feel free to ignore the emails.

docs/_providers/hetzner.md
das7pad: I think the lowest rate-limit is a reasonable one to start with and when the user hits the next level we will print a warning for them to consider upping it. Burning though 1000 requests [1] in less than an hour which regular use seems quite unlikely given that we now creating and updating records in batches.

Starting with the highest rate-limit `hour` (1 request every 3.6 seconds) may annoy users who never approach the limit due to small/infrequent changes to their zones.

[1] As I understand it, Hetzner is rate-limiting with multiple sliding windows (second, minute, hour). In my case they upped the limits to 2 requests per sec, 120 per min, 2000 per hour. From my testing before they upped the limits, I think the defaults were 1 per sec, 60 per minute (a guess, based on the fact that I could sustain 1 req/s though out the entire test suite) and likely 1000 per hour (another guess given that the other rates were doubled for me). This also suggests that the second and minute rates are identical for all users.

providers/hetzner/api.go
das7pad: Thanks for the suggestion. I also found `strings.Title()` for casting it back when reading the header -- Hetzner is using the header name `X-Ratelimit-Limit-Hour`. I am going to retain the switch logic as I think it's a nice way of validating the user provided value and providing a fallback for the empty string. da730e103b05dd749104911627e4834928757777

tlimoncelli:
This all looks good to me.  Ready to merge?

(Please ignore the test failures, if there are any. The tests are very flaky right now.  I trust that you've run the integration tests for this code.)

das7pad:
> This all looks good to me. Ready to merge?

Yes, thanks for checking.

> (Please ignore the test failures, if there are any. The tests are very flaky right now. I trust that you've run the integration tests for this code.)

Aye :) The integration tests pass locally.

tlimoncelli:
Thanks for this!  I'm excited that so many of the providers now react to 429.

PR #963, opened 2020-11-24
Allow TXTMulti for provider Hetzner

afrimberger:
**Issue**
The Hetzner provider doesn't allow TXTMulti. Hence, creating DKIM TXT records with >255 bytes fails. 

**Solution**
In that PR I just enabled TXTMulti. For testing purposes I've successfully created, updated and verified a DKIM record against the Hetzner DNS API. Even the verification with a DKIM record checker succeeded. Therefore, I propose to enable TXTMulti for the provider.


tlimoncelli:
Thanks for fixing that.  You'd be surprised how many times we've found providers have a feature that works and just needs the capability enabled.

PR #971, opened 2020-11-28
INWX: Guard against single-quote TXT targets and split the integration test

### LINKED ISSUE #954: INWX: integration test "single_TXT:Create_TXT_with_single-quote" fails
looks a new integration test has been added since INWX support has been added and it seems to fail.
feel free to assign this to me. I will look into fixing it.

```
=== RUN   TestDNSProviders/testdomain.eu/08:single_TXT:Create_TXT_with_single-quote
    integration_test.go:180: CREATE TXT foo.testdomain.eu "blah`blah" ttl=300
    integration_test.go:199: Expected 0 corrections on second run, but found 1.
    integration_test.go:201: #0: MODIFY TXT foo.testdomain.eu: ("blahblah" ttl=300) -> ("blah`blah" ttl=300)
```

svenpeter42:
Fixes #954 by splitting out the single-quote TXT integration test and by adding an additional check to the INWX provider to fail if there are TXT records with single-quotes.


tlimoncelli:
Thanks!

PR #977, opened 2020-11-28
Gracefully skip tests if the secrets are not available

tlimoncelli:
Secrets are not propagated to an Action if the PR is from a forked repo.  The good news is that this prevents an outsiders from logging secrets.  The bad news it that outsiders always see their tests as failing.

With this change, every provider is tested but the tests abort early if the action doesn't have access to the secrets.  More specifically, the value "true" must be found in an env named `${PROVIDER}__CAN_SECRET` (for example `BIND__CAN_SECRET` or `HEXONET__CAN_SECRET`)

In this PR we:

* List the matrix of providers one-per-line to make it easier to add/remove provider names; as well as add comments.
* For each provider in the matrix, set an env named `${PROVIDER}__CAN_SECRET` to true if tests are to run.
* Add a step that determines if future steps should be skipped

You can see that TomOnTime/dnscontrol (a fork) was able to "bring my own secrets" here: https://github.com/TomOnTime/dnscontrol/pull/1/checks?check_run_id=1468101735

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,13 +52,13 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: ^1.15
-    - name: Run integration tests for ${{ matrix.provider }} provider
-      working-directory: integrationTest
-      run: if [ "$${{ matrix.provider }}__CAN_SECRET" = "true" ] ; then  go test -v -verbose -provider ${{ matrix.provider }} ; else echo SKIPPING ; fi
+    - name: Examining creds for ${{ matrix.provider }} provider
+      run: if [ "$${{ matrix.provider }}__CAN_SECRET" = "true" ] ; then echo "CAN_CONTINUE=true" >> $GITHUB_ENV ; fi
       env:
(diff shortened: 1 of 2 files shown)


MaxHorstmann:
@tlimoncelli nice! 

A thought: is "bring my own secrets" something we realistically expect to be used by outside collaborators? 

If yes, we would probably want to make the domain names (`GCLOUD_DOMAIN` etc) configurable, too, so people can run integration tests against heir own test domains. 
If no, I wonder if we just wanna disable the integration tests in forks altogether? Something like:

```
  integration-tests:
    runs-on: ubuntu-latest
    needs: build
    if: ${{ github.event.pull_request.head.repo.full_name == 'StackExchange/dnscontrol' }}
```

tlimoncelli:
> @tlimoncelli nice!
> 
> A thought: is "bring my own secrets" something we realistically expect to be used by outside collaborators?

I think it is.  Once I get this working and merged, I'll add docs for people that want to join in with their own tests.

I agree that domain names should be moved to be secrets. I'll do that in a separate PR.

PR #997, opened 2020-12-18
Fix deSEC long TXT bug

tlimoncelli:


providers/desec/desecProvider.go
```diff
@@ -48,6 +48,7 @@ var features = providers.DocumentationNotes{
 	providers.CanUsePTR:              providers.Can(),
 	providers.CanGetZones:            providers.Can(),
 	providers.CanAutoDNSSEC:          providers.Cannot(),
+	providers.CanUseTXTMulti:         providers.Can(),
```
bgort, line 51: This seems to be the only change required.  Tested and works fine.

tlimoncelli:
Sorry for the long delay. I don't know how this got lost.

I'm merging this. However, I should warn you that when https://github.com/StackExchange/dnscontrol/pull/1063 arrives, TXT records will be handled differently. You may need to make some changes at that time.

PR #1007, opened 2020-12-26
Support fetch external resources over HTTP

### LINKED ISSUE #994: Idea: support fetch external resources over HTTP
Fetching external resources and parsing them in dnsconfig.js will be very useful, for example get a list of nodes from your cloud provider and automatically set up DNS for all of them.

There's an extension for otto https://github.com/deoxxa/ottoext which implements an event loop, fetch and a few other things, and seems straightfoward to integrate.

xddxdd:
This PR adds [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) to dnscontrol, mentioned in issue #994, via this extension for otto [https://github.com/deoxxa/ottoext](https://github.com/deoxxa/ottoext).

In addition, a `PANIC(reason)` function is added to abort dnscontrol execution with an exit code, in case an error happened.

And a `FETCH()` wrapper function is added to automatically call `PANIC()` if a network error occurred.

Fixes #994

<details>
<summary>Example: successful request</summary>

dnscontrol.js:
```javascript
var DNS_BIND = NewDnsProvider('bind', 'BIND');

D('example.com', REG_NONE, DnsProvider(DNS_BIND), [
  A('@', '1.2.3.4'),
]);

console.log('Hello World');
FETCH('https://example.com').then(function(r) {
  return r.text();
}).then(function(t) {
  // Do something with response body
  D_EXTEND('example.com', [
    TXT('@', t.slice(0, 100)),
  ]);
});
console.log('Goodbye');
```

Generated zone:
```
$TTL 300
; generated with dnscontrol 2020-12-27T00:57:20+08:00
@                IN SOA   DEFAULT_NOT_SET. DEFAULT_NOT_SET. 2020122601 3600 600 604800 1440
                 IN A     1.2.3.4
                 IN TXT   "<!doctype html>\010<html>\010<head>\010    <title>Example Domain</title>\010\010    <meta charset=\"utf-8\" />\010    <m"
```
</details>

<details>
<summary>Example: failed request</summary>

dnscontrol.js:
```javascript
var DNS_BIND = NewDnsProvider('bind', 'BIND');

D('example.com', REG_NONE, DnsProvider(DNS_BIND), [
  A('@', '1.2.3.4'),
]);

console.log('Hello World');
// Example of failure: invalid URL
FETCH('httttttps://example.com').then(function(r) {
  return r.text();
}).then(function(t) {
  // Do something with response body
  D_EXTEND('example.com', [
    TXT('@', t.slice(0, 100)),
  ]);
});
console.log('Goodbye');
```

Dnscontrol output:
```
Hello World
Goodbye
Error: Get "httttttps://example.com": unsupported protocol scheme "httttttps"
```
</details>


xddxdd:
I'm unsure if I should add unit tests for the fetch functionality, since I didn't find individual tests for basic functions like `A()`. Please let me know if a unit test for this is necessary.

pkg/js/js.go
```diff
@@ -35,10 +42,25 @@ func ExecuteJavascript(file string, devMode bool, variables map[string]string) (
 	currentDirectory = filepath.Dir(file)
 
 	vm := otto.New()
+	l := loop.New(vm)
```
tlimoncelli, line 45: Please add comments explaining what these timers are for.

tlimoncelli (commented):
A.  It is fabulous that you can add such a big feature in so few lines of code. Good job!

I have 2 concerns:
1. There have been no updates to fknsrs.biz/p/ottoext in years.  Is it abandoned or just very stable?
2. This will encourage people to make dnsconfig.js configs that break when a third-party dependency is broken or down.  I think that the docs should include a warning against complexity.


B.  Please add docs (`docs/_functions/global/FETCH.md`)

C. Unit tests would be great:

> I'm unsure if I should add unit tests for the fetch functionality, since I didn't find individual tests for basic functions like `A()`. Please let me know if a unit test for this is necessary.

My knowledge of Javascript is limited.  If you can help me make unit tests, I would appreciate it!

xddxdd:
> There have been no updates to fknsrs.biz/p/ottoext in years.  Is it abandoned or just very stable?

The project seems abandoned, but from the testing the functionalities are stable. Nevertheless I can take over and maintain it.

For example, I just forked that project and added the functionality to specify request headers, and updated this PR to point to my fork. Please let me know if it's acceptable to you.

> Please add docs (`docs/_functions/global/FETCH.md`)

I've added docs for the two functions I added, FETCH and PANIC.

> This will encourage people to make dnsconfig.js configs that break when a third-party dependency is broken or down.  I think that the docs should include a warning against complexity.

A warning is included in the docs for FETCH function.

> Unit tests would be great:

On a second thought, the ottoext project already includes its own unit tests, so I doubt if adding separate unit tests for FETCH is necessary. If needed, I can reuse the unit tests from ottoext.

tlimoncelli:
Yes, using your fork is good.

I don't see the new docs.  Did you "git add" them? (or maybe it's a problem on my side?)

xddxdd:
Oh wait, I forgot to push the doc changes to remote.

pkg/js/js.go
```diff
@@ -35,10 +42,25 @@ func ExecuteJavascript(file string, devMode bool, variables map[string]string) (
...
+	if err := timers.Define(vm, l); err != nil {
+		return nil, err
+	}
+	if err := promise.Define(vm, l); err != nil {
+		return nil, err
+	}
+	if err := fetch.Define(vm, l); err != nil {
+		return nil, err
+	}
+	if err := process.Define(vm, flag.Args()); err != nil {
```
tlimoncelli, line 56: Why use github.com/jessevdk/go-flags when dnscontrol already processes flags using github.com/urfave/cli/v2 ?

xddxdd: I wasn't aware of github.com/urfave/cli/v2 being used, this is copy pasted from the unit test of ottoext.

Upon closer inspection, all it does is expose environment variables and call arguments to the Javascript environment. We probably don't need that.

By the way, to my understanding it's not using github.com/jessevdk/go-flags. It's using Go's builtin flag package.

xddxdd: This is removed for being unnecessary, and possibly causing credential leaks in some cases.

xddxdd:
Speaking of security, I just realized that a switch is needed to explicitly enable this functionality, along with some additional warnings.

Or, in the case where DnsControl runs in a CI, based on data from a git repo that accepts and tests PRs, I can easily turn the CI into a DDoS machine with some `while true do fetch(victim)` thing.

tlimoncelli:
Good point.  It would be good to enable this only with a flag.

xddxdd:
A flag `--allow-fetch` is added to enable fetch. Otherwise dnscontrol will exit with the following error if fetch is used:

```
executing dnsconfig.js: ReferenceError: 'fetch' is not defined
```

tlimoncelli (approved):
LGTM.  Ready for merge?

xddxdd:
Sure, good to go.

tlimoncelli:
Thanks!

P.S.   I'd love to hear about how people use this feature!

patschi:
Sorry for being late to the party: A usecase I can think of is querying the DKIM public key from my mailserver setup (using mailcow) from the API and add the records to the DNS zone. However this requires authenticating on the API via POST parameters, but when I see it correct the current `fetch()` implementation isn't capable of POST'ing values on the same request?

xddxdd:
@patschi Technically POSTing parameters is supported, but for now you'll have to handcraft the request body. For example, both snippers below POST "a=12345" and "b=54321".

This is the easier way of form-urlencoded, useful to pass a pair of API key/secret along:

```javascript
FETCH('http://localhost/test.php', {
  headers: {
    "Content-Type": "application/x-www-form-urlencoded",
  },
  method: "POST",
  body: 'a=12345&b=54321',
}).then(function(r) {
  return r.text();
}).then(function(t) {
  console.log(t);
});
```

This is the more complicated way of creating a multipart form-data.

```javascript
data = '--------------------------2c3e2cab183d59b3' + "\n"
     + 'Content-Disposition: form-data; name="a"' + "\n"
     + '' + "\n"
     + '12345' + "\n"
     + '--------------------------2c3e2cab183d59b3' + "\n"
     + 'Content-Disposition: form-data; name="b"' + "\n"
     + '' + "\n"
     + '54321' + "\n"
     + '--------------------------2c3e2cab183d59b3--' + "\n";

FETCH('http://localhost/test.php', {
  headers: {
    "Content-Type": "multipart/form-data; boundary=------------------------2c3e2cab183d59b3",
  },
  method: "POST",
  body: data,
}).then(function(r) {
  return r.text();
}).then(function(t) {
  console.log(t);
});
```

Issue #1003, opened 2020-12-23
Long TXT fails with OVH


lukaszczerpak:
Hi,

The new version throws an error for long txt that AUTOSPLIT is not enabled. After enabling it, another error is thrown "TXT records with multiple strings not supported by ovh". However, long TXT was working fine with OVH, even today after rolling back to the previous version of dnscontrol.

tlimoncelli:
CC: @masterzen

Hi there!

I'm sorry that a recent change has broken the OVH provider.  I'm CC:ing the maintainer.

The problem we were trying to solve is that some people would use a very long TXT record, then switch to a provider that didn't support long txt records.  The root cause was that certain providers were not marked as supporting TXTMulti explicitly, but instead were silently splitting up a long TXT record on the back-end.  The solution was to make this more explicit: Always require AUTOSPLIT or manually split TXT strings.

For OVH to work in this new system, you basically just need to enable TXTMulti for the provider, then modify the API to join the strings together and send them at one big chunk so that the old behavior (splitting them on the back-end) kicks in.

tlimoncelli:
P.S.  See https://github.com/StackExchange/dnscontrol/issues/996 for a similar issue with deSEC.

masterzen:
I'll have a look this week-end or early next week.

lukaszczerpak:
Thank you @tlimoncelli and @masterzen !

masterzen:
@tlimoncelli, I just added the `TXTMulti` capability to the OVH provider and the integration tests are passing. Is that enough or should I implement anything?

tlimoncelli:
That's it!   If the tests pass, you're good.  If someone finds a reason why that isn't sufficient, I add more tests to catch that situation if it happens again.

You aren't the first to find that the change was so small!    Basically this means their API uses a string that is understood by `PopulateFromString()`.

You might want to check to see other records work too.  CAA, PTR, SRV are the most commonly used types.

masterzen:
@lukaszczerpak, 

Can you try PR #1008? 
and let us know if that fixes your original issue?
Thanks !

lukaszczerpak:
Many thanks for looking into this @masterzen. I just compiled from master branch but unfortunately it didn't seem to work:

PREVIEW:

```
******************** Domain: xxx
----- Getting nameservers from: ovh
----- DNS Provider: ovh...2 corrections
[...]

PR #1008, opened 2020-12-28
(#1003) Make OVH TXTMulti capable

masterzen:
In fact OVH natively supports multi TXT, so there's really nothing to do here :)
I also excluded OVH from the Null MX integration test, because this provider doesn't yet support this RFC.


tlimoncelli:
Thanks!

PR #1018, opened 2021-01-13
Implement DS record support for ClouDNS

jamaltebi:
ClouDNS does supprt DS Record ( DS record can be add just for subdomain nor for root domain).

- Question: Can I add a DS record for a subdomain, if there are already other records for the same hostname, such as A, MX, TXT, etc.?

==> Answer: No, you can't. First and foremost, in order for you to be able to add a DS record for your subdomain, the delegation part of your subdomain must be in action. In simple words, the relevant NS records for your subdomain, the "delegators" so to say, must be added first. And to add the NS records, there must be no other records for that particular hostname.


jamaltebi:
@tlimoncelli  I tested the DS record for ClouDNS, It's working

docs/_providers/cloudns.md
```diff
@@ -23,7 +23,11 @@ Current version of provider doesn't support `sub-auth-user`.
 
 ## Records
 
-ClouDNS does not supprt DS Record. 
+ClouDNS does supprt DS Record ( DS record can be add just for subdomain nor for root domain).
+
+- Question: Can I add a DS record for a subdomain, if there are already other records for the same hostname, such as A, MX, TXT, etc.?
+
+==> Answer: No, you can't. First and foremost, in order for you to be able to add a DS record for your subdomain, the delegation part of your subdomain must be in action. In simple words, the relevant NS records for your subdomain, the "delegators" so to say, must be added first. And to add the NS records, there must be no other records for that particular hostname.
```
tlimoncelli, line 30: Suggested replacement: (its shorter and fits the style of the other providers)

ClouDNS does support DS Record on subdomains (not the apex domain itself).  

ClouDNS requires NS records exist for any DS records.  No other records for
the same label may exist (A, MX, TXT, etc.).  If DNSControl is adding NS and
DS records in the same update, the NS records will be inserted first.

integrationTest/integration_test.go
```diff
@@ -179,7 +179,7 @@ func makeChanges(t *testing.T, prv providers.DNSServiceProvider, dc *models.Doma
 		if len(corrections) == 0 && expectChanges {
 			t.Fatalf("Expected changes, but got none")
 		}
-		for _, c := range corrections {
+		for _, c := range corrections { // corre =cloudns.(*cloudnsProvider).GetDomainCor
```
tlimoncelli, line 182: I'm not sure what the comment means.  please clarify or remove comment.

integrationTest/integration_test.go
```diff
@@ -195,7 +195,7 @@ func makeChanges(t *testing.T, prv providers.DNSServiceProvider, dc *models.Doma
 		}
 
 		// run a second time and expect zero corrections
-		corrections, err = prv.GetDomainCorrections(dom2)
+		corrections, err = prv.GetDomainCorrections(dom2) // corre ==> invalid length
```
tlimoncelli, line 198: please clarify or remove comment.

integrationTest/integration_test.go
```diff
@@ -204,8 +204,8 @@ func makeChanges(t *testing.T, prv providers.DNSServiceProvider, dc *models.Doma
 			for i, c := range corrections {
 				t.Logf("#%d: %s", i, c.Msg)
 			}
-			t.FailNow()
-		}
+			t.FailNow() //FailNow marks the function as having failed and stops its execution by calling runtime.Goexit (which then runs all deferred calls in the current goroutine)
+		} //
```
tlimoncelli, line 208: please remove empty comment.

integrationTest/integration_test.go
```diff
@@ -997,6 +999,50 @@ func makeTests(t *testing.T) []*TestGroup {
...
+		testgroup("DS (children only) CLOUDNS",
+			requires(providers.CanUseDSForChildren),
+			only("CLOUDNS"),
+			// Use a valid digest value here, because GCLOUD (which implements this capability) verifies
+			// the value passed in is a valid digest. RFC 4034, s5.1.4 specifies SHA1 as the only digest
+			// algo at present, i.e. only hexadecimal values currently usable.
+			// Cloudns requires NS  Record before creating DS Record.
+			tc("create DS",
+				ns("child", "ns101.cloudns.net."),
+				ds("child", 35632, 13, 1, "1E07663FF507A40874B8605463DD41DE482079D6"),
```
tlimoncelli, line 1011: Does this test work if we list `ds(` before `ns(`?  The provider should do the right thing in either case.

providers/cloudns/api.go
```diff
@@ -177,12 +181,12 @@ func (c *cloudnsProvider) getRecords(id string) ([]domainRecord, error) {
 		return nil, fmt.Errorf("failed fetching record list from ClouDNS: %s", err)
 	}
 
-	var dr recordResponse
+	var dr recordResponse // existant records at provider
```
tlimoncelli, line 184: s/existant/existing/

providers/cloudns/api.go
```diff
@@ -177,12 +181,12 @@ func (c *cloudnsProvider) getRecords(id string) ([]domainRecord, error) {
...
 	json.Unmarshal(bodyString, &dr)
 
 	var records []domainRecord
 	for _, rec := range dr {
-		records = append(records, rec)
+		records = append(records, rec) // apend request to records
```
tlimoncelli, line 189: s/apend/append/

providers/cloudns/cloudnsProvider.go
```diff
@@ -48,6 +48,8 @@ var features = providers.DocumentationNotes{
 	providers.CanUseTLSA:             providers.Can(),
 	providers.CanUsePTR:              providers.Can(),
 	providers.CanGetZones:            providers.Can(),
+	//providers.CanUseDS:               providers.Can(),
+	providers.CanUseDSForChildren: providers.Can(),
```
tlimoncelli, line 52: Please sort this list.

providers/cloudns/cloudnsProvider.go
```diff
@@ -96,6 +98,7 @@ func (c *cloudnsProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*mode
 
 	differ := diff.New(dc)
 	_, create, del, modify, err := differ.IncrementalDiff(existingRecords)
+
```
tlimoncelli, line 101: Please do not add  blank lines without a reason.

providers/cloudns/cloudnsProvider.go
```diff
@@ -172,12 +175,14 @@ func (c *cloudnsProvider) EnsureDomainExists(domain string) error {
...
 }
 
+//parses the ClouDNS format into our standard RecordConfig
 func toRc(domain string, r *domainRecord) *models.RecordConfig {
 
 	ttl, _ := strconv.ParseUint(r.TTL, 10, 32)
 	priority, _ := strconv.ParseUint(r.Priority, 10, 32)
 	weight, _ := strconv.ParseUint(r.Weight, 10, 32)
 	port, _ := strconv.ParseUint(r.Port, 10, 32)
+	//
```
tlimoncelli, line 185: Please remove the empty `//`

providers/cloudns/cloudnsProvider.go
```diff
@@ -214,13 +219,24 @@ func toRc(domain string, r *domainRecord) *models.RecordConfig {
...
+	case "DS":
+		dsKeyTag, _ := strconv.ParseUint(r.DsKeyTag, 10, 32)
+		rc.DsKeyTag = uint16(dsKeyTag)
+		dsAlgorithm, _ := strconv.ParseUint(r.SshfpAlgorithm, 10, 32)
+		rc.DsAlgorithm = uint8(dsAlgorithm)
+		dsDigestType, _ := strconv.ParseUint(r.DsDigestType, 10, 32)
+		rc.DsDigestType = uint8(dsDigestType)
+		rc.DsDigest = r.Target
+		rc.SetTarget(r.Target)
+
```
tlimoncelli, line 231: Please remove blank line.

providers/cloudns/cloudnsProvider.go
```diff
@@ -214,13 +219,24 @@ func toRc(domain string, r *domainRecord) *models.RecordConfig {
...
 	default:
 		rc.SetTarget(r.Target)
 	}
 
 	return rc
 }
 
+//generates the update requests sent to the API (it takes a RecordConfig and turns it into the request).
```
tlimoncelli, line 239: Suggested comment:

// toReq takes a RecordConfig and turns it into the native format used by the API.

tlimoncelli (changes requested):

integrationTest/integration_test.go
jamaltebi: - NS record must be added before adding DS record
- DS record must be deleted before we can delete NS record

IT-Sumpfling: > Does this test work if we list `ds(` before `ns(`? The provider should do the right thing in either case.

Any hint how we can get the provider to "order" the entries in the diff?
As we are new in this, we did not find the right place to do this.

Because when deleting entries, DS needs to be deleted before NS, too.

tlimoncelli: Other providers make multiple passes over the records.  For example, the first pass only does DS records, the second pass does all records. 

Alternatively just move the DS records to the top of the list so they happen first.

IT-Sumpfling:
Perhaps wait with the review. I've just committed some changes which @jamaltebi will test.
These should
- fix the order of NS / DS entries (so no longer necessary to list NS first or to delete in 2 passes)
- implement a very simple rate-limit by delaying each request for 100ms
(ClouDNS seems to rate-limit the API to 10 req/second - but I haven't found a documentation from their side)

IT-Sumpfling:
> This should be fine for now. Another way to only delay after a 429 http status code. Look at `providers/desec/protocol.go` for an example.

Yes. But I did not yet get around to test if ClouDNS sends 429 or another status code ...

tlimoncelli:
>
>
> Yes. But I did not yet get around to test if ClouDNS sends 429 or another
> status code ...
>

To be clear: the current solution is sufficient for now.

>

integrationTest/integration_test.go
jamaltebi: with the last changes, DS record can be created even if we list `ds(` before `ns(`. The test work

IT-Sumpfling:
Thank you. Just to be sure - the branch / pull request should be ready for merge from our side.
integration tests run, order of ns/ds does not matter and your review-comments should be addressed.

tlimoncelli (approved):
Great job!!!

PR #1020, opened 2021-01-16
OCTODNS: Add SPF support

armandgrillet:
This type is identical to TXT but used [for other purposes](https://tools.ietf.org/html/rfc4408#section-3), it is [officially supported by OctoDNS](https://github.com/octodns/octodns/blob/19a68ec502306409b03c3df60ebf8fdcb637737f/docs/records.md#record-types).

I faced a panic from `dnscontrol` when using `ToRR()` on SPF records I got from an OctoDNS `yaml` due to the missing support, this fixes the issue.

Example with such a record config:
```
---
spf:
  ttl: 600
  type: SPF
  value: v=spf1 ip4:192.168.0.2/16-all
```

and this codebase:
```
func printRecords(recordsFile string) error {
	f, err := os.Open(recordsFile)
	if err != nil {
		return err
	}

	recordsList, err := octoyaml.ReadYaml(f, "stackexchange.com")
	if err != nil {
		return err
	}

	fmt.Println(recordsList[0])
}
```

Before:
```
%!v(PANIC=String method: ToRR: Unimplemented rtype SPF)
panic: ToRR: Unimplemented rtype SPF
```

After:
```
"v=spf1 ip4:192.168.0.2/16-all"
```

I have also checked that `dns.IsDuplicate(recordsList[0].ToRR(), dnsRRObtainedFromServerAnswer)` passes.


tlimoncelli:
Interesting!

I agree that an SPF record shouldn't crash the code.  However I thought SPF was deprecated and people were supposed to use TXT records instead.  Other providers silently convert SPF to TXT.

By the way... I'm curious about your use of the OctoDNS provider.  I honestly didn't know anyone was using it. I wrote it mostly as an experiment.   I'd be interested in knowing if you are using it in production.

armandgrillet:
You're right regarding [the deprecation](https://tools.ietf.org/html/rfc7208#section-3.1): 

> As a result, developers of SPF found it easier and more practical to use the TXT RR type for SPF records.

My use case was also an experiment where I wanted to parse OctoDNS resource record files using `dnscontrol`, I didn't find any other Go package offering something similar to `octoyaml.ReadYaml` which is very practical to go from a file to a list of records.

The `panic` when facing `SPF` records was the main problem I faced when using this provider, I wish I had more time to rewrite more of it to return `error` instead of panicking in other places but for this first PR I wanted to keep things contained so that reviewers have a better time.

tlimoncelli (approved):

tlimoncelli:
Ok, I'll accept this change.  It will be the only provider that uses SPF (instead of converting them to TXT records).

NB: I expect there will come a day where we remove SPF support or OCTODNS support (I don't think it is actually used by anyone).

armandgrillet:
Thanks, please let me know if I need to do anything else to get the PR merged.

armandgrillet:
`build / integration-tests (HEXONET) (pull_request)` fails for something that doesn't seem related to this PR and CI on `master` looks flaky, e.g. https://github.com/StackExchange/dnscontrol/runs/1732385587

Issue #419, opened 2018-11-28
Provider request: Oracle Cloud Infrastructure


Marcus-James-Adams:
More info at https://cloud.oracle.com/en_US/edge/dns/faq

tlimoncelli:
We've noted your request for this provider!  
 
As an open source project, we are glad to accept PRs that add new
providers. We encourage you to try adding this provider yourself.  The
process is well-documented here:
https://stackexchange.github.io/dnscontrol/writing-providers

Many people have reported that adding a provider was their first
experience writing Go.

tlimoncelli:
FYI: Now that it is tagged "provider-request", we close the issue.   I know that's a bit confusing, but it will remain on the "requested provider" list where people can find it.

PR #1021, opened 2021-01-16
NEW PROVIDER: Oracle Cloud

kallsyms:
This PR introduces a provider for Oracle Cloud DNS.

OCI DNS API Docs: https://docs.oracle.com/en-us/iaas/api/#/en/dns/20180115/

Few miscellaneous notes from development:
* This currently marks DS records as not supported, despite Oracle saying they are. I'm getting 500s from the API when running the integration tests. I can investigate this further if wanted.
* The API provides only full updates (i.e. replace the entire given rrset/fqdn/zone) or patch (specifying a batch of add/remove operations on the given rrset/rqdn/zone). This PR implements the provider with batched patches per zone.
* The patch API doesn't provide a way to update a single existing record, so this implementation batches a remove then an add operation (mimicking how the console does it).

Addresses #419.


tlimoncelli (commented):
Looks great so far!  I will have a closer look in a few days.

don't worry about adding DS in a separate PR.  It's common to do that.  Get the base working well then augment it.

providers/oracle/oracleProvider.go
```diff
@@ -0,0 +1,362 @@
...
+	defer cancel()
+
+	getResp, err := o.client.GetZone(ctx, dns.GetZoneRequest{
+		ZoneNameOrId:  &domain,
+		CompartmentId: &o.compartment,
+	})
+
+	if err == nil {
+		return nil
+	} else if err != nil && getResp.RawResponse.StatusCode != 404 {
```
tlimoncelli, line 97: Since the "if" has a return, the "else if" can be an "if".

providers/oracle/oracleProvider.go
```diff
@@ -0,0 +1,362 @@
...
+		return err
+	} else {
```
tlimoncelli, line 99: Since the prior `if` ends with a `return`, this doesn't need to be an `else`.

tlimoncelli (changes requested):
Great work!  There is very close to being ready to merge.  I just had a few small changes.

By the way... in intergration_test.go, Oracle isn't running the `pager601` tests.  Since the API is slow, and since page size is 100, this test isn't needed. That said, it might be useful to run this test just once to verify functionality (but don't include it in the PR).

kallsyms:
Changes addressed! Just tried the `pager601` test and it creates more than 100 records (so the batching works), but the test hits the 60s deadline before it finishes adding all of them.

providers/oracle/oracleProvider.go
```diff
@@ -0,0 +1,364 @@
...
+// EnsureDomainExists creates the domain if it does not exist.
+func (o *oracleProvider) EnsureDomainExists(domain string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	getResp, err := o.client.GetZone(ctx, dns.GetZoneRequest{
+		ZoneNameOrId:  &domain,
+		CompartmentId: &o.compartment,
+	})
+
```
tlimoncelli, line 94: please remove blank line, to group the error check with the related code.

providers/oracle/oracleProvider.go
```diff
@@ -0,0 +1,364 @@
...
+	if err == nil {
+		return nil
+	}
+
```
tlimoncelli, line 98: please remove blank line, to group the error check with the related code.

providers/oracle/oracleProvider.go
```diff
@@ -0,0 +1,364 @@
...
+	_, err = o.client.CreateZone(ctx, dns.CreateZoneRequest{
+		CreateZoneDetails: dns.CreateZoneDetails{
+			CompartmentId: &o.compartment,
+			Name:          &domain,
+			ZoneType:      dns.CreateZoneDetailsZoneTypePrimary,
+		},
+	})
+
```
tlimoncelli, line 110: please remove blank line, to group the error check with the related code.

tlimoncelli (commented):
Looks great!  Super minor change and we're all done!

(thanks for running the pager601 test.  Yeah, timeouts are expected. At least it worked up until the timeout!)

kallsyms:
Done!

PR #1030, opened 2021-01-22
feature: adding RPM/DEB package creation and upload as asset

atma-stackoverflow:
Using the fpm github action, this creates and uploads a RPM and DEB for each release


atma-stackoverflow:
See https://github.com/StackExchange/dnscontrol/releases/tag/v3.5.2 as a verification this works as expected

tlimoncelli (approved):

PR #1047, opened 2021-02-15
Add DMARC_BUILDER

### LINKED ISSUE #1022: DMARC_BUILDER
I think a function similar to `SPF_BUILDER` would be great to have for DMARC, too.

juliusrickert:
Fixes #1022.


tlimoncelli:
Please rebase (i.e. `get merge master`) and `go generate`

tlimoncelli:
I'm super looking forward to trying this out!

I'm going to merge this so I can try it out at Stack Overflow easier.  I won't close this until later.

docs/dmarc-builder.md
```diff
@@ -0,0 +1,89 @@
+---
+layout: default
+title: CAA Builder
```
kmutahar, line 3: Shouldn't this be DMARC builder?

juliusrickert: Yes, you are right. I thought I had corrected it, but apparently, I did not.

kmutahar: It happens.

Does this mean that after it is fixed the doc should be live on the site?

tlimoncelli: Thanks for pointing this out!

The docs are based on master.  I fixed this in master 1 minute ago, so the docs should update soon.

For future references, if you search https://stackexchange.github.io/dnscontrol/js for the word `dmarc_` you'll have found the docs, just in the wrong place.

Issue #1052, opened 2021-02-21
Reenginer TXT record handling


tlimoncelli:
TXT records should work as follows:

The user specifies either a string or a list of strings:

    TXT("label", "124abc"),
    TXT("another", ["one", "two", "three"]),

The provider either works, warns, or errors:

Warnings: (printed by check/preview/push)
* String length > 255 octets: Will be split along 255-octet boundaries.
* Provider does not support multiple strings.  Strings will be joined then re-split along 255-octet boundaries.

Errors: (check/preview/push will exit with error)
* Provider does not support TXT strings longer than 255 octets.
* Provider does not support multiple TXT strings. Attempt to join failed due to total length > 255 octets.

The warnings/errors must be output during check/preview/push.  Therefore, checking will be done by the provider module, not by doing the API call and seeing if it fails.

Providers that mishandle quoting and are unable to (for example) support a txt record with an embedded quote, should warn about this in check/preview/push but may rely on the actual API call to error.  Rationale: Sometimes these bugs get fixed at the server end.

Open issue: Should there be a way to not print the warnings?   (Meta flag such as TXT_NO_WARNING?  AUTOSPLIT?)

juliusrickert:
> Open issue: Should there be a way to not print the warnings? (Meta flag such as TXT_NO_WARNING? AUTOSPLIT?)

I think there should definitely be a way to suppress the warnings, one for each type of adjustment, maybe another one to suppress all (could be "created" by the users by forming an array?).

The suppression should work on a per-record and per-domain/default basis.

tlimoncelli:
Here's a thought: "The right way" to remove the warnings is to split the string at the appropriate boundaries so that the warning isn't needed.   Can we make that easier rather than invent some general "warning suppression system"?

We could provide functions that make doing that easy.  Something like:

    TXT("label", LONGTXT("my very long string")),
    TXT("other", REPACK255("my", "list", "which", "should", "be", "re-split at 255-octet intervals"),

This way (1) the warning goes away, (2) the user doesn't have to work too hard (splitting a string manually is a pita), and (3) if they change provider, the right thing will happen at that provider.

I don't like the naes LONGTXT andd REPACK255 but I'm sure we can come up with something better.

juliusrickert:
From my understanding, `LONGTXT` would split the string into (at most) 255-octet chunks?
At least for `LONGTXT`, it is too verbose, in my opinion. It is the behaviour I am expecting.

[...]

PR #1054, opened 2021-02-21
Rework how TXT records are handled

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/1052


tlimoncelli:
Fixing in https://github.com/StackExchange/dnscontrol/tree/tlim_target_and_txt instead

PR #1064, opened 2021-03-01
HEDNS: Fix CNAME, NS and PTR record handling of trailing .

rblenkinsopp:
Fixed handling of CNAME, NS and PTR record targets where HE.net DNS now automatically appends these.


tlimoncelli:
Thanks for fixing those!

PR #1075, opened 2021-03-02
HEDNS: Fix issue with domain listing on completely empty/new accounts

rblenkinsopp:
There is an issue where if the HE.net DNS account is completely fresh and new, the domains list will not be retrievable. Previously this would error incorrectly, when in fact there were just no domains to be listed. This is fixed by returning an empty domain list in this scenario.

This MR change also corrects a typo in the provider name.


rblenkinsopp:
@tlimoncelli This should also indicate the automated integration testing (with bring-your-own secrets) functions as intended on a fork 🙂

tlimoncelli:
Awesome!  Thanks for fixing those!

PR #1093, opened 2021-03-11
NEW PROVIDER: DNS Made Easy

vojtad:
This PR introduces a provider for DNS Made Easy.

DNS Made Easy API Docs: https://api-docs.dnsmadeeasy.com/

**Few notes:**
- ~Changing apex NS records is not supported.~
- System NS records cannot be changed. Custom NS records are supported even for root domain.
- Custom HTTPRED records used for HTTP redirection by DNS Made Easy are ignored and not supported.
- Global Traffic Director feature is not supported.
- DNS Made Easy limits request rate to 150 requests / 5 minutes. Backoff is implemented.
- Tested against DNS Made Easy sandbox environment.

**Known issues:**
- `TestDNSProviders/dnscontroltest-dnsme.com/10:complex_TXT:Create_TXT_with_double-quote` is failing. It looks like DNS Made Easy is removing all quotes from TXT record value. I've contacted their support. I am not able to create TXT record with double quote in their control panel either.

### DIFF
--- a/docs/_providers/dnsme.md
+++ b/docs/_providers/dnsme.md
@@ -44,7 +44,7 @@ D("example.tld", REG_NONE, DnsProvider(DNSME),
 {%endhighlight%}
 
 ## Activation
-You can generate you `api_key` and `secret_key` in [Control Panel](https://cp.dnsmadeeasy.com/) in Account Information in Config menu.
+You can generate your `api_key` and `secret_key` in [Control Panel](https://cp.dnsmadeeasy.com/) in Account Information in Config menu.
 
 API is only available for Business plan and higher plans.
 
@@ -59,4 +59,4 @@ Global Traffic Director feature is not supported.
 Set `DNSME_DEBUG` environment variable to dump all API calls made by this provider.
 
 ### Testing
-Set `sandbox` key to `"true"` (this must be string, not a boolean) in credentials JSON alongside `api_key` and `secret_key` to make all API calls against DNS Made Easy sandbox environment.
+Set `sandbox` key to any non-empty value in credentials JSON alongside `api_key` and `secret_key` to make all API calls against DNS Made Easy sandbox environment.
--- a/docs/_providers/dnsme.md
+++ b/docs/_providers/dnsme.md
@@ -1,0 +1,62 @@
+---
+name: DNSME
+title: DNS Made Simple Provider
+layout: default
+jsId: DNSME
+---
+# DNS Made Simple Provider
+
+## Configuration
+In your credentials file, you must provide your `api_key` and `secret_key`. More info about authentication can be found in [DNS Made Easy API docs](https://api-docs.dnsmadeeasy.com/).
+
+{% highlight json %}
+{
+  "dnsme": {
+    "api_key": "1c1a3c91-4770-4ce7-96f4-54c0eb0e457a",
+    "secret_key": "e2268cde-2ccd-4668-a518-8aa8757a65a0"
+  }
+}
+{% endhighlight %}
+
+## Records
+
+Changing apex NS records is not supported.
+
+ALIAS/ANAME records are supported.
+
+This provider does not support HTTPRED records.
+
+SPF records are ignored by this provider. Use TXT records instead.
+
+## Metadata
+This provider does not recognize any special metadata fields unique to DNS Made Easy.
+
+## Usage
+Example Javascript:
+
+{% highlight js %}
+var REG_NONE = NewRegistrar('none', 'NONE')
+var DNSME = NewDnsProvider("dnsme", "DNSME");
+
+D("example.tld", REG_NONE, DnsProvider(DNSME),
+    A("test","1.2.3.4")
+);
+{%endhighlight%}
+
+## Activation
+You can generate you `api_key` and `secret_key` in [Control Panel](https://cp.dnsmadeeasy.com/) in Account Information in Config menu.
+
+API is only available for Business plan and higher plans.
+
+## Caveats
+
+### Global Traffic Director
+Global Traffic Director feature is not supported.
+
+## Development
+
+### Debugging
+Set `DNSME_DEBUG` environment variable to dump all API calls made by this provider.
+
+### Testing
+Set `sandbox` key to `"true"` (this must be string, not a boolean) in credentials JSON alongside `api_key` and `secret_key` to make all API calls against DNS Made Easy sandbox environment.
--- a/docs/_providers/dnsme.md
+++ b/docs/_providers/dnsme.md
@@ -20,8 +20,6 @@ In your credentials file, you must provide your `api_key` and `secret_key`. More
 
 ## Records
 
-Changing apex NS records is not supported.
-
 ALIAS/ANAME records are supported.
 
 This provider does not support HTTPRED records.
--- a/integrationTest/providers.json
+++ b/integrationTest/providers.json
@@ -43,6 +43,12 @@
     "domain": "$DIGITALOCEAN_DOMAIN",
     "token": "$DIGITALOCEAN_TOKEN"
   },
+  "DNSME": {
+    "domain": "$DNSME_DOMAIN",
+    "sandbox": "true",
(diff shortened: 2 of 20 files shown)


integrationTest/providers.json
```diff
@@ -43,6 +43,12 @@
     "domain": "$DIGITALOCEAN_DOMAIN",
     "token": "$DIGITALOCEAN_TOKEN"
   },
+  "DNSME": {
```
tlimoncelli, line 46: Can we use a name that is longer than "ME"?  (for this and the folder name)

docs/_providers/dnsme.md
```diff
@@ -0,0 +1,62 @@
...
+### Global Traffic Director
+Global Traffic Director feature is not supported.
+
+## Development
+
+### Debugging
+Set `DNSME_DEBUG` environment variable to dump all API calls made by this provider.
+
+### Testing
+Set `sandbox` key to `"true"` (this must be string, not a boolean) in credentials JSON alongside `api_key` and `secret_key` to make all API calls against DNS Made Easy sandbox environment.
```
tlimoncelli, line 62: This seems like an easy mistake for humans to make.  How about if `sandbox` is non-null?

providers/dnsme/dnsmeProvider.go
```diff
@@ -0,0 +1,243 @@
...
+
+	var corrections []*models.Correction
+
+	var deleteRecordIds []int
+	deleteDescription := []string{"Batch deletion of records:"}
+	for _, m := range del {
+		originalRecordId := m.Existing.Original.(*recordResponseDataEntry).Id
+		deleteRecordIds = append(deleteRecordIds, originalRecordId)
+		deleteDescription = append(deleteDescription, m.String())
+
```
tlimoncelli, line 128: delete blank line

providers/dnsme/dnsmeProvider.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	if len(deleteRecordIds) > 0 {
+		corr := &models.Correction{
+			Msg: strings.Join(deleteDescription, "\n\t"),
+			F: func() error {
+				return api.deleteRecords(domain.Id, deleteRecordIds)
+			},
+		}
+
```
tlimoncelli, line 138: Remove blank line

providers/dnsme/dnsmeProvider.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, err
+	}
+
+	existingRecords := make([]*models.RecordConfig, 0, len(records))
+	for i := range records {
+		// Ignore HTTPRED and SPF records
+		if records[i].Type == "HTTPRED" || records[i].Type == "SPF" {
+			continue
+		}
+
```
tlimoncelli, line 224: Remove blank line

providers/dnsme/types.go
```diff
@@ -0,0 +1,189 @@
...
+	if record.Type == "MX" {
+		err = rc.SetTargetMX(uint16(record.MxLevel), record.Value)
+	} else if record.Type == "SRV" {
+		err = rc.SetTargetSRV(uint16(record.Priority), uint16(record.Weight), uint16(record.Port), record.Value)
+	} else if record.Type == "CAA" {
+		value, err := strconv.Unquote(record.Value)
+		if err != nil {
+			panic(err)
+		}
+
```
tlimoncelli, line 142: Remove blank line

tlimoncelli (commented):
Looks great so far!   I've written some minor feedback (I'll have a full look later this week).

I wouldn't worry about Create_TXT_with_double-quote for the initial release. Many providers do not permit double quotes.

integrationTest/providers.json
vojtad: I can rename DNSME to DNSMADEEASY everywhere. It looked too long to me at first, but maybe it's ok?

docs/_providers/dnsme.md
vojtad: Oh, my bad! There actually is a condition to compare `settings["sandbox"] != ""` instead of `"true"` string now. I forgot to update the docs when I changed it, sorry.

vojtad:
Thanks! I am looking forward to you review. This was actually my first real-world code I wrote in Go.

I have added a check for TXT with double quotes to `AuditRecords` and I have also fixed all `golint` warnings.

Please let me know what you think about renaming DNSME to DNSMADEEASY. I would rename all occurrences if we would like to change it to keep it consistent.

tlimoncelli:
> Thanks! I am looking forward to your review. This was actually my first real-world code I wrote in Go.

I couldn't tell it was your first time!  Looks excellent!

> I have added a check for TXT with double quotes to `AuditRecords` and I have also fixed all `golint` warnings.

👍 

> Please let me know what you think about renaming DNSME to DNSMADEEASY. I would rename all occurrences if we would like to change it to keep it consistent.

Yes, I agree that DNSMADEEASY is a better name. (Easier to understand by someone who isn't a customer of DNS Made Easy; don't be confused with EasyDNS; etc)   Likewise we should be consistent about the directory name, variables,  `*Provider.go`, etc.

Tom

cncallaghan:
@vojtad DNS Made Easy API does support GTD and has for some time, here is the link to the POST call https://api-docs.dnsmadeeasy.com/#1f3ac624-b30a-46a4-904d-c1e7ffc24994 the argument to modify would be "gtdLocation":"DEFAULT", we support GTD for the following regions: 
DEFAULT,
US_EAST,
US_WEST,
EUROPE,
ASIA_PAC,
OCEANIA,
SOUTH_AMERICA


We also do support NS records at the root/apex of a domain, here's an example body that goes through without issue, you would just leave the name field blank (same POST call as above):
{"name":"","type":"NS","value":"ns1.cncallaghan.net.","gtdLocation":"DEFAULT","ttl":86400}


Could you please elaborate on what you mean by "custom" HTTPRED records? All HTTP redirections and their respective fields that can be entered through our UI can also be entered through the API.

Your assessment of the API rate limit is correct, this can be increased for a fee through the DNSME control panel under "Purchase Services"

Thank you for bringing to our attention the issue with doublequotes in the middle of a TXT record, this is possible in our UI but not in our API, we will work on adding this functionality. 

If you ever run into issues with API calls or otherwise in our system, please don't hesitate to reach out to us at support@dnsmadeeasy.com, we're happy to assist!

vojtad:
@cncallaghan Thanks for the info.

I will look into supporting apex NS records. I've had some issues with API not returning "System NS Records" when I query for all records but they were still being added, probably some internal thing of dnscontrol. I will give it another try.

These "System NS Records" cannot be changed or deleted in any way, is that right? But there should be no issue with creating/changing/deleting custom NS records?

I know you (DNS Made Easy) support GTD. It is not supported by this provider for dnscontrol I implemented. It is a non-standard feature and it would require a lot more work.

Same goes for HTTPRED records. I meant they are custom from the point of DNS spec. They would also require more work which is not part of this initial provider implementation.

I might explore options for implementing GTD and HTTPRED in the future, but right now it is out of scope for me.

API rate limit is just fine. I just noted it for other to know about it and also to let them know I implemented retries with backoff delays.

tlimoncelli:
This all looks good so far!  I suggest we merge this and fix the things listed above (if any are remaining) in other PRs.

Before we do the merge, I'd like to verify that (1) you feel this is ready to merge, (2) the integration tests all pass.

vojtad:
Thanks! I think everything is ready to merge. I just re-ran the integration tests and they all passed except the one testing double quotes in TXT record.

I will be migrating our company's DNS management to dnscontrol and we will be using this provider. That should put it under real-word test pretty soon.

Just to summarize what can and cannot be implemented:
1) We can implement HTTPRED records and GTD. It should be possible without anything required from DNS Made Easy.
2) To enable creating TXT records with double-quotes we are waiting for DNS Made Easy to fix an issue on their end.
3) Changing system apex NS records is not supported, they are hard-coded by DNS Made Easy. However, custom apex NS records can be created/changed/deleted.

PR #1094, opened 2021-03-14
typo fix and Audit Records update for ClouDNS

riku22:
The following two points have been fixed.

- Fixed that "ClouDNS" was changed to "CloDNS" in the source code.
- Added a description to Audit Records that does not allow double quotes in TXT records.

I am new to Go language.
Therefore, there may be something strange.


tlimoncelli:
Thanks! Excellent work!

PR #1097, opened 2021-03-19
Add cache steps

cdhunt:


cdhunt:
The cache step does eliminate the redundant module downloads but doesn't noticeably improve the workflow runtime.

jauderho:
IIRC, there's a bug with the caching where the cache increases until it hits the max. See https://github.com/docker/build-push-action/issues/252#issuecomment-744400434

This is how I fixed it for my Docker builds. https://github.com/jauderho/dockerfiles/blob/main/.github/workflows/cloudflared.yml

I'll see if I can find some time to test out a similar fix.

Issue #1038, opened 2021-02-02
Implement Multiple String for ClouDNS


jamaltebi:
When I join the TXTStrings before sending to ClouDNS the String is added like :"one" "two" and  when receiving records ( at second run), the string is readed : "one/" "two/".
I have the same issue as  #996 (https://github.com/StackExchange/dnscontrol/issues/996#issuecomment-748247815) in ClouDNS with caracter `\"` . create succeds with join String. but modify not

tlimoncelli:
CC @pragmaton

tlimoncelli:
Changes related to this have been merged into `master`.  Can you try again to see if you can still reproduce the problem?

riku22:
Hello.

I ran an integration test.
However, I don't know if I can report it here, but it seems that there is a problem with the long string TXT record.
Note that adding a PTR record has failed because it is not supported in the zone used for testing.

```
=== RUN   TestDNSProviders
=== RUN   TestDNSProviders/dnscontrol-cloudns.com
=== RUN   TestDNSProviders/dnscontrol-cloudns.com/Clean_Slate:Empty
    integration_test.go:198: DELETE NS dnscontrol-cloudns.com ns1.example.com. ttl=300, ClouDNS ID: 231249036
    integration_test.go:198: DELETE NS dnscontrol-cloudns.com ns2.example.com. ttl=300, ClouDNS ID: 231249035
=== RUN   TestDNSProviders/dnscontrol-cloudns.com/00:GeneralACD:Create_an_A_record
    integration_test.go:198: CREATE A dnscontrol-cloudns.com 1.1.1.1 ttl=300
=== RUN   TestDNSProviders/dnscontrol-cloudns.com/00:GeneralACD:Change_it
    integration_test.go:198: MODIFY A dnscontrol-cloudns.com: (1.1.1.1 ttl=300) -> (1.2.3.4 ttl=300), ClouDNS ID: 231249255: 
=== RUN   TestDNSProviders/dnscontrol-cloudns.com/00:GeneralACD:Add_another
    integration_test.go:198: CREATE A www.dnscontrol-cloudns.com 1.2.3.4 ttl=300
=== RUN   TestDNSProviders/dnscontrol-cloudns.com/00:GeneralACD:Add_another(same_name)
    integration_test.go:198: CREATE A www.dnscontrol-cloudns.com 5.6.7.8 ttl=300
[...]

PR #1098, opened 2021-03-21
[ClouDNS] Fix for TXT records. 

pragmaton:
Better handling of TXT records. ClouDNS NOT allow multiple TXT records with same name but allow values longer the 255

Now TXT tests - pass

> go test -v -verbose -provider CLOUDNS -start 9 -end 13
> ...
>PASS


pragmaton:
May be related to #1038 1038

tlimoncelli:
Thanks for fixing that!

Issue #1078, opened 2021-03-04
Request: Get the TTL available in ClouDNS from the API


riku22:
Hello.

I use ClouDNS as my DNS provider.
Currently, the TTL value seems to be specified as a constant value, but it would be nice to get the available TTL from the API and use it.
[ClouDNS: Get the available TTL](https://www.cloudns.net/wiki/article/153/)
I'm sorry for my poor English, but I would appreciate it if you could consider it.

Best regards.

PR #1102, opened 2021-03-22
Fetch ClouDNS allowed TTL values from API. #1078

riku22:
Fetch ClouDNS allowed TTL values from API.
Fix the problem of # 1078.

### DIFF
--- a/providers/cloudns/api.go
+++ b/providers/cloudns/api.go
@@ -71,21 +71,7 @@ type domainRecord struct {
 
 type recordResponse map[string]domainRecord
 
-var allowedTTLValues = []uint32{
-	60,      // 1 minute
-	300,     // 5 minutes
-	900,     // 15 minutes
-	1800,    // 30 minutes
-	3600,    // 1 hour
-	21600,   // 6 hours
-	43200,   // 12 hours
-	86400,   // 1 day
-	172800,  // 2 days
-	259200,  // 3 days
-	604800,  // 1 week
-	1209600, // 2 weeks
-	2419200, // 4 weeks
-}
+var allowedTTLValues = []uint32{}
 
 func (c *cloudnsProvider) fetchAvailableNameservers() error {
 	c.nameserversNames = nil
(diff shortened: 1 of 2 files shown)


providers/cloudns/api.go
```diff
@@ -107,6 +93,21 @@ func (c *cloudnsProvider) fetchAvailableNameservers() error {
...
+	params := requestParams{
+		"domain-name": domain,
+	}
+
+	var bodyString, err = c.get("/dns/get-available-ttl.json", params)
+	if err != nil {
+		return fmt.Errorf("failed fetching available TTL values list from ClouDNS: %s", err)
+	}
+
+	json.Unmarshal(bodyString, &allowedTTLValues)
```
tlimoncelli, line 107: Return err result instead of nil?

providers/cloudns/api.go
```diff
@@ -107,6 +93,21 @@ func (c *cloudnsProvider) fetchAvailableNameservers() error {
 	return nil
 }
 
+func (c *cloudnsProvider) fetchAvailableTTLValues(domain string) error {
+	allowedTTLValues = nil
```
tlimoncelli, line 97: Is `fetchAvailableTTLValues` going to make this API call more than once?  Would it be sufficient to call it only if needed?

providers/cloudns/api.go
```diff
@@ -71,21 +71,7 @@ type domainRecord struct {
 
 type recordResponse map[string]domainRecord
 
-var allowedTTLValues = []uint32{
```
tlimoncelli, line 74: Using the API is good.  It is better than a constant.
Is the API returning any values that are not on this list?  For backwards compatibility it would be good to know.

riku22:
Hello.

1. The API must be called for each zone.
1. It's basically a constant value, but it seems that you can add a TTL by contacting support.
For example, 10 minutes is available for my account.

tlimoncelli:
hello!

"1. The API must be called for each zone."

Interesting!  Please add a comment in the code. Other people will find that useful.

riku22:
Added a comment.

tlimoncelli:
Thanks!

PR #1107, opened 2021-03-28
AuditRecord: removed one successful case

Rashad-j:
The test case of "recordaudit.TxtNoLongStrings(records)" no longer needed for the provider HEXONET


tlimoncelli:
@Rashad-j I'm super psyched that you were able to eliminate one of the checks!  Thanks for testing this and submitting a PR!

PR #1109, opened 2021-03-28
DESEC: Handle long TXT records better

tlimoncelli:


tmkis2:
in providers/desec/convert.go remove the unused "strings" from the import in line 7 then this will work exceptionally fine with desec as provider.

sfxn:
Hey @tlimoncelli seems you are 95% ready. Never used go but it seems the unused import at `providers/desec/convert.go:7` is now obsolete and should be removed.

PR #1115, opened 2021-03-31
Direct SOA record management

fuero:
This adds the SOA record macro to JS configs.
It introduces a feature lock (CanUseSOA) plus validation checks I thought reasonable.


tlimoncelli:
Looks good so far!

Please add this to the "matrix" variable in `dnscontrol/build/generate/featureMatrix.go` (see https://stackexchange.github.io/dnscontrol/adding-new-rtypes)

build/generate/featureMatrix.go
```diff
@@ -34,6 +34,7 @@ func generateFeatureMatrix() error {
 			{"CAA", "Provider can manage CAA records"},
 			{"PTR", "Provider supports adding PTR records for reverse lookup zones"},
 			{"NAPTR", "Provider can manage NAPTR records"},
+			{"SOA", "Provider can manage SOA records"},
```
fuero, line 37: @tlimoncelli I did - was I supposed to add the generated 'docs/_includes/matrix.html' file to the PR?

tlimoncelli: Ooops!  You certainly did!  My bad.

patschi:
Just a random thought: IIRC in case of INWX the serial number will be generated/managed on the provider-side and will be overwritten there. So I guess in dnscontrol it would be sufficient to simply specify any value then. Not sure if there should be any specific handling when serial number is managed server-side, or when only target and soambox are managed?

tlimoncelli:
I propose we don't let the user set the serial number and just manage it ourselves automatically.

My first thought was that if the user sets the serial number to 0 that indicates they want dnscontrol to magically take care of it.  BIND "takes care of it" by setting it to `max(YYYYMMMDD00, current+1)`.  If "current" is "unknown", assume it is zero.  This works surprisingly well.

What if the user **does** specify a (non-zero) serial number?  Now the user has to remember to increment the serial number every time they update dnsconfig.js.   Personally I'd hate to do that.  More likely I would simply forget to update it.  In fact, I used to forget all the time when I was hand-editing BIND zone files.  I would spend an hour trying to figure out why no updates happened, then realize my mistake, get frustrated, and eventually just give up and write a `sed` script to do it for me... then [HostDB](https://www.usenix.org/legacy/event/lisa05/tech/limoncelli.html), then [DNSControl](https://www.usenix.org/conference/srecon17americas/program/presentation/peterson).  See the pattern? ;-)

I can't imagine a scenario where we would want to make the user edit a serial number every time they modify dnsconfig.js.

So why not just always automatically manage the serial number?  (i.e. don't even ask the user for one)

fuero:
> I propose we don't let the user set the serial number and just manage it ourselves automatically.

Sounds like a good idea - I'll change the PR accordingly.

fuero:
I'd just like a heads up, is the code ok so far, or should I make changes?

tlimoncelli:
The code is ok so far!  

You should do a `go generate` to remake `pkg/js/static.go` (you'll hit that when you rebase).

You should also reformat the using gofmt:

```
gofmt -s -w $(find * -name \*.go -print)
```

fuero:
ok, have another look.
I used `gofmt -s -w $(git diff master --name-only | grep -Ee '\.go$')` to reformat the code.

tlimoncelli:
Is that 'Ok, merge into master" or "Ok, leave it as a separate branch until it's 100% done"?

fuero:
It's finished in my opinion.
If you are satisfied with it, merge it, otherwise advise changes please :-)

tlimoncelli:
Oops.  We need to add integration tests for SOA records in `dnscontrol/integrationTest/integration_test.go`

PR #1117, opened 2021-04-03
GANDI_V5 & OVH: Implement ZoneLister

nemunaire:
Hi!

It appears that those two providers I use, can implement the `ZoneLister` interface.

For OVH, as it already fetches the zones when creating the `DNSServiceProvider`, I choose to return them instead of doing the request again (as the code already makes the assumption this list doesn't change).


tlimoncelli:
Thanks!!

Issue #1116, opened 2021-04-02
Feature request: execute creds.json if it is executable


x70b1:
At the moment I think its a bit hard to secure the credentials in `creds.json` as it is required to have a plaintext file.
I have seen the blackbox hint.

Ansible offers a way to run a script to retrieve the required secrets. 
Example: `vault_password_file=pass.sh` in ansible.cfg
pass.sh can now contain the vault password, or runs if it is executable. This makes it easy to integrate different solutions.

Something similar would be really cool for `dnscontrol`.

tlimoncelli:
That's an interesting idea!

I understand the value of having a script that executes instead of reading the file.   Many open source projects use different techniques. What do you think of these?

1. The execute bit: 
    * `chmod a+x pass.sh ; dnscontrol push --creds=pass.sh`
2. Prefix with "!": 
    * `dnscontrol push --creds=!pass.sh`
3. Use a flag to indicate that creds are executable:
    * `dnscontrol push --creds=pass.sh --creds-execute`
4. Use a flag to indicate what to execute (which would override `--creds`):
    * `dnscontrol push --creds-program=pass.sh`

The first one seems unfriendly to Windows users, as the executable bit isn't used much.

They're all unfriendly to sites that need one config that works in both Windows and Linux (maybe Stack is the only place with that requirement?)

I'm interested in your feedback!

x70b1:
Thanks for that response! I use `dnscontrol` on linux only.

I would say: All 4 ideas are solutions for the existing problem/idea.

1. Would be cool as the implementation on linux should be very easy, no bigger changes. We only have to check the file. But I have no idea ho wo solve this on windows systems.
2. Would be cool, as it is a configurable way.
3. I would be against idea as it introduces a 2nd config param that need to be set. 
4. Also nice, but introduces a new config setting too.

I would say idea 2 solves the problem without introducing a new config setting, while a solution for windows could be implemented.

tlimoncelli:
Thanks to @jpbede and https://github.com/StackExchange/dnscontrol/pull/1119, now you can either prefix the name with `!` or set the executable bit!

Test it out and let us know how it works for ya!

Re-open this ticket if you have any problems.

x70b1:
@tlimoncelli I checked the executable bit. It works perfekt!

@jpbede Thanks for working on it!

PR #1119, opened 2021-04-12
Make creds file executable

jpbede:
Makes the creds file executable as requested in #1116. It checks the unix file permissions for the executable bit or it can be marked as executable with a "!" as in front of the file name.

### DIFF
--- a/providers/config/providerConfig.go
+++ b/providers/config/providerConfig.go
@@ -8,24 +8,42 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/DisposaBoy/JsonConfigReader"
 	"github.com/TomOnTime/utfutil"
 )
 
-// LoadProviderConfigs will open the specified file name, and parse its contents. It will replace environment variables it finds if any value matches $[A-Za-z_-0-9]+
+// LoadProviderConfigs will open or execute the specified file name, and parse its contents. It will replace environment variables it finds if any value matches $[A-Za-z_-0-9]+
 func LoadProviderConfigs(fname string) (map[string]map[string]string, error) {
 	var results = map[string]map[string]string{}
-	dat, err := utfutil.ReadFile(fname, utfutil.POSIX)
-	if err != nil {
-		// no creds file is ok. Bind requires nothing for example. Individual providers will error if things not found.
-		if os.IsNotExist(err) {
-			fmt.Printf("INFO: Config file %q does not exist. Skipping.\n", fname)
-			return results, nil
+
+	var dat []byte
+	var err error
+
+	filename := strings.TrimPrefix(fname, "!")
+	if stat, statErr := os.Stat(filename); statErr != nil {
+		return nil, statErr
+	} else if mode := stat.Mode(); mode&0111 == 0111 || strings.Contains(fname, "!") {
+		// found executable bit or was marked as executable by user
+		dat, err = executeCredsFile(filename)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


providers/config/providerConfig.go
```diff
@@ -8,24 +8,42 @@ import (
...
+
+	var dat []byte
+	var err error
+
+	filename := strings.TrimPrefix(fname, "!")
+	if stat, statErr := os.Stat(filename); statErr != nil {
+		return nil, statErr
+	} else if mode := stat.Mode(); mode&0111 == 0111 || strings.Contains(fname, "!") {
+		// found executable bit or was marked as executable by user
+		dat, err = executeCredsFile(filename)
```
tlimoncelli, line 30: This logic seems very complex.  It works, but the next person to read it might appreciate more simple logic.

Suggestion: (1) Abstract out reading the creds file into readCredsFile() like you did with executeCredsFile(), (2) abstract out whether or not the creds file is executable to a function isExecutable(), (3) simplify logic such as:

```
    if strings.HasPrefix(fname, "!") {
        dat, err = executeCredsFile(strings.TrimPrefix(fname, "!"))
        if err != nil { ... }
   } else if isExecutable(fname) {
        dat, err = executeCredsFile(fname)
        if err != nil { ... }
   } else {
        dat, err = utfutil.ReadFile(fname, utfutil.POSIX)
        if err != nil { ... }
   }
```

providers/config/providerConfig.go
```diff
@@ -38,6 +56,15 @@ func LoadProviderConfigs(fname string) (map[string]map[string]string, error) {
 	return results, nil
 }
 
+func executeCredsFile(filename string) ([]byte, error) {
+	pwd, err := os.Getwd()
```
tlimoncelli, line 60: Is prepending `os.Getwd()` really required?

It would be more simple to prepend "./"

```
        cmd := filename
        if ! strings.HasPrefix(fname, "/") {
                // if the path doesn't start with `/` make sure we aren't relying on $PATH.
               cmd = filepath(".", filename)
        }
	out, err := exec.Command(cmd).Output()
        return out
```

jpbede:
Thanks for review @tlimoncelli. Good points, implemented them. Much appreciated :)

providers/config/providerConfig.go
```diff
@@ -38,6 +49,38 @@ func LoadProviderConfigs(fname string) (map[string]map[string]string, error) {
...
+		return nil, fmt.Errorf("failed reading provider credentials file %v: %v", filename, err)
+	}
+	return dat, nil
+}
+
+func executeCredsFile(filename string) ([]byte, error) {
+	cmd := filename
+	if !strings.HasPrefix(filename, "/") {
+		// if the path doesn't start with `/` make sure we aren't relying on $PATH.
+		cmd = "./" + filename
```
tlimoncelli, line 78: Suggestion: `cmd = filepath.Join(".", filename)` 

(this way it will work on Windows)

jpbede: If i do this, it relies on $PATH:

```
$ dnscontrol check-creds --creds=creds.sh powerdns POWERDNS 
failed GetZone lpc("creds.sh"): exec: "creds.sh": executable file not found in $PATH
```

tlimoncelli: Dang it!  The filepath.Join() function tries to be smart and clean up afterwords.  Ugh.  How about:

    strings.Join(".", filepath.Separator, filename)

jpbede: Yes, this works like expected.

Issue #1128, opened 2021-04-21
R53_ALIAS to Cloudfront Distribution - InvalidChangeBatch


DrewTurner:
DNSControl-Linux version 3.8.0

When adding an A record for R53_ALIAS to a cloudfront distribution like so:

`R53_ALIAS('@', 'A', '38d2dc74ceaa.cloudfront.net', R53_ZONE('Z2FDTNDATAQYW2'))`

I receive the following error:
```
MODIFY R53_ALIAS example.com: (38d2dc74ceaa.cloudfront.net. atype=A zone_id=Z2FDTNDATAQYW2 ttl=300 type=A zone_id=Z2FDTNDATAQYW2) -> (38d2dc74ceaa.cloudfront.net atype=A zone_id=Z2FDTNDATAQYW2 ttl=300 type=A zone_id=Z2FDTNDATAQYW2)
FAILURE! InvalidChangeBatch: [The request contains an invalid set of changes for a resource record set 'A example.com.']
```
The docs indicate I need the cloudfront distribution to include the alternate domain name that matches the record and that is present.

Do I have the syntax of that R53 alias correct?

tlimoncelli:
CC: @masterzen @maclennann

tlimoncelli:
Should there be a `.` at the end of `'38d2dc74ceaa.cloudfront.net.'`  ?

I ask because when I look at the error message (and insert line breaks to make it easier to examine) I see one is missing:

```
MODIFY R53_ALIAS example.com: 
(38d2dc74ceaa.cloudfront.net. atype=A zone_id=Z2FDTNDATAQYW2 ttl=300 type=A zone_id=Z2FDTNDATAQYW2) -> (38d2dc74ceaa.cloudfront.net atype=A zone_id=Z2FDTNDATAQYW2 ttl=300 type=A zone_id=Z2FDTNDATAQYW2)
```

tlimoncelli:
Note to self: If adding the `.` fixes this, we should update `docs/_functions/domain/R53_ALIAS.md`'s examples (which don't have trailing dots) and document that this error sometimes means there is a missing dot.

DrewTurner:
@tlimoncelli - Tom thank you for your quick feedback.  Adding the . as the standard for these does indeed fix the issue.

PR #1130, opened 2021-04-22
ROUTE53: Docs should specify FQDN with dot.

tlimoncelli:
After helping with https://github.com/StackExchange/dnscontrol/issues/1128 I noticed that the examples don't include a "." at the end of FQDNs.  I think that lead to the confusion.

Here's a PR with corrected (I hope) examples.  Are they more correct?

CC @masterzen


masterzen (approved):
Indeed, the trailing `.` seems mandatory. 
But I couldn't find anywhere in the AWS Route53 documentation a mention for ALIAS records. The mention that the trailing dot is optional for records that take a domain as target, like `CNAME` (see https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/ResourceRecordTypes.html).
ALIAS are essentially `A` records, so I assume that this doesn't match the aforementioned rule :)

So I looked to the `R53_ALIAS` I have in my dnscontrol setup and indeed, all have a trailing dot. So my bad for this non working sample.

Approving :)

PR #1150, opened 2021-05-09
Enable support for CAA in ns1

costasd:
NS1 rest client already supports CAA records, let's enable the capability and handle CAA records.

### DIFF
--- a/providers/ns1/ns1provider.go
+++ b/providers/ns1/ns1provider.go
@@ -18,6 +18,7 @@ import (
 var docNotes = providers.DocumentationNotes{
 	providers.CanUseAlias:            providers.Can(),
 	providers.CanUsePTR:              providers.Can(),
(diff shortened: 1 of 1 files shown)


providers/ns1/ns1provider.go
```diff
@@ -18,6 +18,7 @@ import (
 var docNotes = providers.DocumentationNotes{
 	providers.CanUseAlias:            providers.Can(),
 	providers.CanUsePTR:              providers.Can(),
+	providers.CanUseCAA:              providers.Can(),
```
tlimoncelli, line 21: Please sort this list (including the DocDualHost).

tlimoncelli:
Yes, the SKIPPED ones are skipped for a reason.

As far as test 34, I'd look earlier in the output. It should give some clues about why it failed.

tlimoncelli (approved):

PR #1157, opened 2021-05-14
ns1: Implement EnsureDomainExists

costasd:
API support is there, let's use it.


costasd:
Unfortunately no integration tests exists for the `create-domains` operations to formally verify this PR's functionality (I guess because cleaning up domains after the suite has run isn't that straightforward), but I've checked with an account of mine and it seems to be working fine.

tlimoncelli:
Thanks for adding this feature!

PR #1159, opened 2021-05-15
Disable ClouDNS NAPTR support

riku22:
This change removes part of commit 9949cba.
ClouDNS does not support changing the NAPTR in the API, so I need to disable this change.
I apologize for any inconvenience caused.


tlimoncelli:
No worries.  The important thing is that you found the issue and reported it asap.  Thanks!

PR #1161, opened 2021-05-21
AXFRDDNS: fix doc: false advertisement of Get-Zones capability

nemunaire:
Hi!

The AXFRDDNS doesn't implement the `ListZones` function, but advertises `providers.CanGetZones: providers.Can()`.

If I understand correctly the `providers.CanGetZones` capability, it relies only on `ListZones` function/`ZoneLister` interface, so this PR just indicates in the doc that this provider doesn't support it.

AFAIK, you can't ask authoritative servers for domains they serve: so it can't be implemented.


tlimoncelli:
Agreed.

CC @hnrgrgr

PR #1166, opened 2021-06-01
Show actual URL to use instead of just "sandbox"

jamesgecargo:
Instead of stating "sandbox", sometimes incorrectly if sandbox is false, just output the actual URL that will be used.


providers/dnsmadeeasy/api.go
```diff
@@ -12,13 +12,13 @@ type dnsMadeEasyProvider struct {
...
 
 func newProvider(apiKey string, secretKey string, sandbox bool, debug bool) *dnsMadeEasyProvider {
-	fmt.Println("creating DNSMADEEASY provider for sandbox")
-
 	baseURL := baseURLV2_0
 	if sandbox {
 		baseURL = sandboxBaseURLV2_0
 	}
 
+	fmt.Println("Creating DNSMADEEASY provider for %q", baseURL)
```
tlimoncelli, line 20: Should be `fmt.Printf` and add a `\n` at the end of the format string.

jamesgecargo: Sorry about that, total go newbie here! I've updated this now.

tlimoncelli: You are most certainly not the first to make that typo.  :-)

PR #1179, opened 2021-06-23
MSDNS: Improve reliability of zone dump

tlimoncelli:
* Zone dump was 90M.  By moving from UTF16 to UTF8 we saved 50%
* Zone dump was capturing stdout, which is slow for big files. Now we write to a tmp file.
* Zone dump was outputting a lot of `Cim*` crap which we didn't need. File size now reduced 90%


devblackops (approved):
LGTM 👍

PR #1190, opened 2021-06-30
Add Dependabot ignores as requested

jauderho:
This seems to not throw any errors so it should be safe to merge. 

The ignored packages currently have no new releases that I am aware of so there is a limit to what we can test.


tlimoncelli (approved):

jauderho:
FWIW, it appears possible to configure allow/ignore so that patch releases trigger a pull request but not major/minor versions for packages that are semver compliant. 

Would that be useful?

PR #1199, opened 2021-07-02
Return error instead of panic when converting RR to RC

nemunaire:
Hi!

Today I came across a domain containing an unknown RR type and `dnscontrol` stopped early.

Here is a proposal to avoid `log.Fatal` and `panic` when calling RRtoRC. Most of the code base already support returning error, so there is not so much changes left.

In my case, it changes the output for unknown record from:

```
******************** Domain: yyyyyy.zzz
----- Getting nameservers from: xxx
----- DNS Provider: xxx...
2021/07/03 00:03:44 dnsrr.go:94: rrToRecord: Unimplemented zone record type=OPENPGPKEY (xxxxx...)
```

to:

```
******************** Domain: yyyyyy.zzz
----- Getting nameservers from: xxx
----- DNS Provider: xxx...
ERROR
Error getting corrections: rrToRecord: Unimplemented zone record type=OPENPGPKEY (xxxxxx....)

Done. 0 corrections.
completed with errors
```

What do you think?

### DIFF
--- a/models/dnsrr.go
+++ b/models/dnsrr.go
@@ -4,7 +4,6 @@ package models
 
 import (
 	"fmt"
-	"log"
 	"strings"
 
 	"github.com/miekg/dns"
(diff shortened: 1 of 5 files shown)


tlimoncelli:
Thanks!  This is definitely some tech debt that needed cleaning up!

PR #1202, opened 2021-07-06
Less frequently Dependabot runs

tlimoncelli:
AWS and Azure dependencies seem to have daily updates that usually don't affect this project.  Changing the dependabot settings so that security issues are reported immediately but other changes are reported monthly or weekly.


jauderho:
Agreed, the daily setting is a bit much and probably only useful during the initial phase.

I will note that Dependabot is smart enough to close an existing pull request and create a new one if a new version is released before the pull request is merged so there should always be only at most one pull request per dependency.

Issue #1177, opened 2021-06-20
deSEC Provider and Pagination


tmkis2:
Using deSEC will stop working as soon as you reach the 500 domain limit there. Pagination is required to get the whole domain list. @peterthomassen mentioned that the validity of a domain should be checked without getting the whole domain list.
In EnsureDomainExists() the whole domainlist is fetched and then checked against the domain name. Right solution would be to append the domainname to the endpoint and check the return code. 200 = domain exists and 404 domain does not exist.

https://github.com/StackExchange/dnscontrol/blob/master/providers/desec/desecProvider.go#L109

@tlimoncelli or @D3luxee any spare time to fix this? :)

Futhermore the whole domain list is being fetched to check if the token is valid. @peterthomassen suggests to make a simple request for example: GET /api/v1/auth/account/ - this would fetch some account information in which you can check if the Token is valid and also you would get the max number of domains in the reply.

https://github.com/StackExchange/dnscontrol/blob/master/providers/desec/desecProvider.go#L31
https://desec.readthedocs.io/en/latest/auth/account.html#retrieve-account-information

tlimoncelli:
I don't have a deSEC account, so @D3luxee will have to fix this.

D3luxee:
I will take a look at it in July, but do you really have over 500 domains or is this a theoretical issue? :D
I'm just curious.

tmkis2:
I have 499 at deSEC at the moment because I rely heavily on dnscontrol to manage them. My other domains at (mostly at hosting.de) should be transferred over to deSEC which makes a total of ~850 Domains. :)

D3luxee:
I started to work on this and already implemented the token validation using the /auth/account api endpoint.

I've also change the EnsureDomainExists() function to validate each individual domain instead of fetching the whole list. 
This means that dnscontrol will do one api call for each domain configured when ever you run `dnscontrol create-domains`, in your case this will be ~850 api calls which could let us run into the rate limit. 

But thats how dnscontrol works and i don't think we can get around this. (see https://github.com/StackExchange/dnscontrol/blob/master/commands/createDomains.go#L49)

tmkis2:
Thank you very much @D3luxee - great to hear about the progress. Because of the rate limits I already splitted domains in separate directories for all domains starting with 0-9, a, b, .... so there are only a few of them in each dnsconfig.js - using require for shared variables and credentials works fine for me.

D3luxee:
[...]

PR #1204, opened 2021-07-07
deSEC: Implements support for long / multistring txt records

### LINKED ISSUE #996: Long / MultiString TXT records fail with deSEC
deSEC provider doesn't allow TXT records with multiple strings, but the service splits long records automatically.

After #947 validation changes, dnscontrol fails validation for long TXT records with `ERROR: txt target >255 bytes and AUTOSPLIT not set`. Adding AUTOSPLIT to deSEC records (or splitting manually) causes dnscontrol to fail with `ERROR: TXT records with multiple strings not supported by desec`

Before, long TXT records worked, though the automatically split record caused unnecessary modifications.

D3luxee:
This PR will resolve #996 and #1177 the multi/long txt part is working but i suggest to wait with the merge until @tmkis2 confirmed that its working with over 500 domains since i do not have that many domains to test it.


providers/desec/desecProvider.go
```diff
@@ -23,13 +24,12 @@ Info required in `creds.json`:
...
 	c.creds.token = m["auth-token"]
+	c.domainIndex = map[string]uint32{}
 	if c.creds.token == "" {
 		return nil, fmt.Errorf("missing deSEC auth-token")
 	}
-
-	// Get a domain to validate authentication
-	if err := c.fetchDomainList(); err != nil {
-		return nil, err
+	if err := c.authenticate(); err != nil {
```
tlimoncelli, line 31: Is this going to authenticate once for each domain that calls `DnsProvider()`? 

(It might not... I would test this by setting up a dnsconfig.js with 2 or more domains, and put a Printf statement in this function to see if it prints multiple times)

Please check.

providers/desec/protocol.go
```diff
@@ -58,32 +58,42 @@ type dnssecKey struct {
...
 	err = json.Unmarshal(bodyString, &dr)
 	if err != nil {
 		return err
 	}
-	for _, domain := range dr {
-		//We store the min ttl in the domain index
-		//This will be used for validation and auto correction
-		c.domainIndex[domain.Name] = domain.MinimumTTL
-	}
+	c.domainIndex[dr.Name] = dr.MinimumTTL
```
tlimoncelli, line 88: Please add a comment why dr.MinimumTTL is being used here.

Better yet.... please document in `type desecProvider` what domainIndex represents.

providers/desec/desecProvider.go
D3luxee: This is in the initialization function of the provider and is only called once. I've also validated this with printf.

providers/desec/protocol.go
D3luxee: I've added some documentation.

D3luxee:
I think this can be merged when you are ready. This will only fix #996 the issue #1177 will be fixed in another PR.

providers/desec/protocol.go
```diff
@@ -175,6 +176,17 @@ retry:
...
+			waitfor := resp.Header.Get("Retry-After")
+			if waitfor != "" {
+				wait, err := strconv.ParseInt(waitfor, 10, 64)
+				if err == nil {
+					printer.Warnf("Rate limiting.. waiting for %s seconds", waitfor)
+					time.Sleep(time.Duration(wait) * time.Second)
+					goto retry
+				}
+			}
+			printer.Warnf("Rate limiting.. waiting for 500 milliseconds")
```
peterthomassen, line 189: the message is incorrect

providers/desec/protocol.go
```diff
@@ -222,6 +234,17 @@ retry:
...
```
peterthomassen, line 247: the message is incorrect

providers/desec/protocol.go
D3luxee: No its not since this is the fallback, if it can't get the Retry-After header it will wait for 500 milliseconds as it did before.
But if everything works as expected this code and message should never be triggered anymore.

providers/desec/protocol.go
D3luxee: No its not since this is the fallback, if it can't get the Retry-After header it will wait for 500 milliseconds as it did before.
But if everything works as expected this code and message should never be triggered anymore.

peterthomassen:
You may also want to add a cut-off for the wait time, let's say 120 seconds? (In some weird edge cases, it could happen that the daily limit is hit etc., and you probably don't want dnscontrol to stall.)

providers/desec/protocol.go
peterthomassen: My bad, I didn't look thoroughly enough. Sorry!

D3luxee:
> You may also want to add a cut-off for the wait time, let's say 120 seconds? (In some weird edge cases, it could happen that the daily limit is hit etc., and you probably don't want dnscontrol to stall.)

You mean if the Retry-After is like 86400 seconds because its in the daily limit? I'm not sure if thats what you mean :D

peterthomassen:
There is a general limit of 2000 requests per day per user (see docs). If you hit that within two hours (let's say -- but that's almost one request per second), you will have to wait 22 hours (79200 seconds).

I would assume that people don't want to wait for such long time. -- Does dnscontrol have a general timeout parameter? If so, you might want to error out if you can see from the `Retry-After` header that the timeout would be hit while waiting.

D3luxee:
This PR includes various improvements and resolves #996 and #1177 which @tmkis2 just confirmed.

I will now stop to add more commits to this so that @tlimoncelli can complete the review and merge if everything is fine.

tlimoncelli (approved):

tlimoncelli:
So much to love here!  Thanks for this!  (especially the rate limiting!)

PR #1208, opened 2021-07-11
deSEC implement pagination

D3luxee:
This PR will implement pagination for the deSEC provider (#1177), this is needed when ever the api could return over 500 elements.

### DIFF
--- a/providers/desec/protocol.go
+++ b/providers/desec/protocol.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/StackExchange/dnscontrol/v3/pkg/printer"
@@ -25,6 +26,7 @@ type desecProvider struct {
 		user     string
 		password string
 	}
+	mutex sync.Mutex
 }
 
 type domainObject struct {
@@ -73,6 +75,8 @@ func (c *desecProvider) authenticate() error {
 	return nil
 }
 func (c *desecProvider) initializeDomainIndex() error {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
 	if c.domainIndex != nil {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


providers/desec/protocol.go
```diff
@@ -71,37 +73,138 @@ func (c *desecProvider) authenticate() error {
 	}
 	return nil
 }
+func (c *desecProvider) initializeDomainIndex() error {
+	if c.domainIndexInitialized {
```
tlimoncelli, line 77: Suggestion: Protect domainIndex with a mutex.  It isn't needed now, but some day we'll call these concurrently.

providers/desec/protocol.go
```diff
@@ -16,9 +18,9 @@ const apiBase = "https://desec.io/api/v1"
 
 // Api layer for desec
 type desecProvider struct {
-	domainIndex      map[string]uint32 //stores the minimum ttl of each domain. (key = domain and value = ttl)
-	nameserversNames []string
-	creds            struct {
+	domainIndex            map[string]uint32 //stores the minimum ttl of each domain. (key = domain and value = ttl)
+	domainIndexInitialized bool              // if the domainIndex was fully initialized once
```
tlimoncelli, line 22: Do we need `domainIndexInitalized`?  Could we simply [check if it is non-nil?](https://stackoverflow.com/questions/31339249/check-if-a-map-is-initialised-in-golang)

D3luxee: I've changed it.

providers/desec/protocol.go
D3luxee: Done

tlimoncelli (approved):
Looks great!  I have no further comments.  Anything else or is it ready to merge?

D3luxee:
It can be merged

tlimoncelli:
Thanks for fixing this!  Big sites really appreciate it!

PR #1219, opened 2021-07-24
Add INCLUDE statement to include records from other domains

blackshadev:
Add the ability to include records from other domain.  This can be very handy for working with split horizon where you want to overwrite your public domain for an internal one, while still having (some) of these external records.

Example:
```
D("example.com!external", REGISTRAR, DnsProvider(R53),
  A("test", "8.8.8.8")
);

D("example.com!internal", REGISTRAR, DnsProvider(R53),
  INCLUDE("example.com!external"),
  A("home", "127.0.0.1")
);
```

### DIFF
--- a/docs/_functions/domain/INCLUDE.md
+++ b/docs/_functions/domain/INCLUDE.md
@@ -1,0 +1,23 @@
+---
+name: INCLUDE
+parameters:
+  - domain
+---
+
+Includes all records from a given domain
+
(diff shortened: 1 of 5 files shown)


tlimoncelli:
Wow!  That's brilliant!  I'll review this over the weekend.

tlimoncelli (approved):

PR #1220, opened 2021-07-25
SPF Flatten for non apex domains

### LINKED ISSUE #785: SPF_Builder's flattening only works on apex domain
SPF_BUILDER is essentially disables itself if it is run on a non-apex record.  This is a feature request to be able to place an SPF record on non-apex labels. 

Discovered during https://github.com/StackExchange/dnscontrol/issues/784

blackshadev:
SPF Flatten for non apex domains. Fixes #785.

The fix was relatively easy. Instead of only considering the TXT of the Apex, let's consider all TXT records in a domain.

### DIFF
--- a/models/record.go
+++ b/models/record.go
@@ -452,6 +452,16 @@ func (recs Records) FQDNMap() (m map[string]bool) {
 	return m
 }
 
+func (recs Records) GetByType(typeName string) Records {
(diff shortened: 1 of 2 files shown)


tlimoncelli:
Thanks for digging into this!

PR #1221, opened 2021-07-25
INWX: fix apex domains

### LINKED ISSUE #1218: inwx: domain entries for apex stopped working
I recently got a new domain and discovered that my dnscontrol setups stopped working. Here are a few examples:
```
D("example.net", REG_NONE, DnsProvider(DNS_INWX), 
MX( "@", 10, "nameofmx.example.com."),
TXT("@", "v=spf1 +mx -all"),
TXT("@", "v=DMARC1;p=reject;pct=100;sp=none;aspf=s;adkim=r;"),
CNAME("openpgpkey", "wkd.keys.openpgp.org."),
TXT("dkim._domainkey", "v=DKIM1; dkim_pubkey_data_strings")
):
```
dnscontrol check says everything fine
dnscontrol preview says everthing is fine
for unchanged data for existing domains, dnscontrol push is fine
for new domains, all the entries with the "@" key give me errors like this one

```
#1: CREATE MX example.net 10 nameofmx.example.com. ttl=86400
FAILURE! (2005) Parameter value syntax error
```
I guess there has been a change to the inwx api, so that entries for the apex ("@") stopped working.
The other entries, like fo the wkd and dkim key get pushed without errors, but the "@" entries do not.

Using an additional dummy BIND provider creates zone files that look ok to me.

I cannot really figure out what is behind this and would appreciate any help.

Regards,

svenpeter42:
INWX has changed their API to use an empty string () instead of
the common @. Fix that by converting between the two.

Fixes #1218

### DIFF
--- a/providers/inwx/inwxProvider.go
+++ b/providers/inwx/inwxProvider.go
@@ -161,11 +161,16 @@ func newInwxDsp(m map[string]string, metadata json.RawMessage) (providers.DNSSer
 func makeNameserverRecordRequest(domain string, rec *models.RecordConfig) *goinwx.NameserverRecordRequest {
 	content := rec.GetTargetField()
 
+	name := rec.GetLabel()
+	if name == "@" {
+		name = ""
+	}
+
 	req := &goinwx.NameserverRecordRequest{
 		Domain:  domain,
 		Type:    rec.Type,
 		Content: content,
-		Name:    rec.GetLabel(),
+		Name:    name,
 		TTL:     int(rec.TTL),
 	}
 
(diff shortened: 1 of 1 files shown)


tlimoncelli (approved):

tlimoncelli:
Thanks for the fix!

inwx-mba:
Hello Together and greetings from INWX!

I would like to apologize briefly, unfortunately an error has led to this problem. We unintentionally disabled the @ customization in the origin of the domain name. Unfortunately there was a problem in the communication to our developers.

Yesterday we were informed by a customer about this issue and my colleagues were able to fix it.

The @ is now accepted correctly again!

Thanks a lot for your quick efforts to change it in your software, too!

Best regards
Montgomery

PR #1238, opened 2021-08-15
Vultr - Adding quotes to non-spf TXT records

Aechrok:


tlimoncelli:
What if the txt includes a “ in it?--
Sent from Gmail Mobile

Aechrok:
They would be caught by https://github.com/StackExchange/dnscontrol/blob/master/providers/vultr/auditrecords.go (Line 14) -> https://github.com/StackExchange/dnscontrol/blob/master/pkg/recordaudit/txt.go (Line 45)

Attempting to escape a double or single quote into the TXT record to circumvent Vultr's quotation requirement hits that error as is, making the user unable to effectively make TXT records as they either hit the error that the string is required to be encased in quotes, or the error regarding double quotes.

After speaking to a contact at Vultr, this has been a backlog item with no current timeframe of patching.

Successful Operation:
`TXT("_acme-challenge.example","abcdefghijklmnopqrstuvwxyz", TTL(30)),`

Unsuccessful Operation:
`TXT("_acme-challenge.example","\"abcdefghijklmnopqrstuvwxyz\"", TTL(30)),`

Definitely open to suggestions for better implementation. "Measure twice, cut once" and all :)

tlimoncelli:
Ah! It's already caught byauditrecords.go!  Excellent!  In that case, there's nothing to do.  If there is an bug id that you can cite in auditrecords.go then do so. Otherwise, I guess we're good to go!

Aechrok:
Added a check to only make adjustments to non-spf records.
It feels dirty, but I didn't see another way of determining if SPF Builder was used or not.

The end result looks to be that SPF Builder is already adding the quotes, and then this function to correct TXT records to Vultr adds another set resulting in:
`Unable to add record: Quotes may only appear at the beginning and end of the data`

Aechrok:
Merged in #1239

PR #1243, opened 2021-08-30
CLOUDFLAREAPI: Initial support for CF_WORKER_ROUTE (without tests).

fdcastel:
Initial implementation of CF_WORKER_ROUTE.

Adds [Cloudflare Worker routes](https://developers.cloudflare.com/workers/platform/routes) support via new `CF_WORKER_ROUTE` function.

No tests, as I have yet to understand the tests infrastructure. But the code is working as I did some manual tests with a personal CF account (Create, Modify and Delete routes are working as expected).

Thanks to @tlimoncelli for guidance and incentive.

### DIFF
--- a/docs/_functions/domain/CF_WORKER_ROUTE.md
+++ b/docs/_functions/domain/CF_WORKER_ROUTE.md
@@ -1,0 +1,29 @@
+---
+name: CF_WORKER_ROUTE
+parameters:
+  - pattern
+  - script
+---
+
+`CF_WORKER_ROUTE` uses [Cloudflare Workers](https://developers.cloudflare.com/workers/) 
+API to setup [routes](https://developers.cloudflare.com/workers/platform/routes)
+for a given domain.
+
+If _any_ `CF_WORKER_ROUTE` function is used then `dnscontrol` will manage _all_ 
+Worker Routes for the domain.
+
+WARNING: This interface is not extensively tested. Take precautions such as making
+backups and manually verifying `dnscontrol preview` output before running
+`dnscontrol push`. This is especially true when mixing Worker Routes that are
+managed by DNSControl and those that aren't.
+
+This example assigns the patterns `api.foo.com/*` and `foo.com/api/*` to a `my-worker` script:
+
+{% include startExample.html %}
+{% highlight js %}
+D("foo.com", .... ,
+    CF_WORKER_ROUTE("api.foo.com/*", "my-worker"),
+    CF_WORKER_ROUTE("foo.com/api/*", "my-worker"),
+);
+{%endhighlight%}
+{% include endExample.html %}
--- a/docs/_providers/cloudflare.md
+++ b/docs/_providers/cloudflare.md
@@ -170,3 +170,20 @@ Notice a few details:
 2. The IP address in those A records may be mostly irrelevant, as cloudflare should handle all requests (assuming some page rule matches).
 3. Ordering matters for priority. CF_REDIRECT records will be added in the order they appear in your js. So put catch-alls at the bottom.
 4. if _any_ `CF_REDIRECT` or `CF_TEMP_REDIRECT` functions are used then `dnscontrol` will manage _all_ "Forwarding URL" type Page Rules for the domain. Page Rule types other than "Forwarding URL” will be left alone.
+
+## Worker routes
+The Cloudflare provider can manage Worker Routes for your domains. Simply use the `CF_WORKER_ROUTE` function passing the route pattern and the worker name:
+
+{% highlight js %}
+
+var CLOUDFLARE = NewDnsProvider('cloudflare','CLOUDFLAREAPI');
+
+D("foo.com", REG_NONE, DnsProvider(CLOUDFLARE),
+    // Assign the patterns `api.foo.com/*` and `foo.com/api/*` to `my-worker` script.
+    CF_WORKER_ROUTE("api.foo.com/*", "my-worker"),
+    CF_WORKER_ROUTE("foo.com/api/*", "my-worker"),
+);
+
+{%endhighlight%}
+
+Please notice that if _any_ `CF_WORKER_ROUTE` function is used then `dnscontrol` will manage _all_ Worker Routes for the domain.
--- a/models/record.go
+++ b/models/record.go
@@ -34,12 +34,14 @@ import (
 //     ALIAS
 //     CF_REDIRECT
 //     CF_TEMP_REDIRECT
+//     CF_WORKER_ROUTE
 //     FRAME
 //     IMPORT_TRANSFORM
 //     NAMESERVER
 //     NO_PURGE
 //     NS1_URLFWD
 //     PAGE_RULE
+//     WORKER_ROUTE
 //     PURGE
 //     URL
 //     URL301
@@ -512,7 +514,7 @@ func downcase(recs []*RecordConfig) {
 		case "ANAME", "CNAME", "DS", "MX", "NS", "PTR", "NAPTR", "SRV", "TLSA", "AKAMAICDN":
 			// These record types have a target that is case insensitive, so we downcase it.
 			r.target = strings.ToLower(r.target)
-		case "A", "AAAA", "ALIAS", "CAA", "IMPORT_TRANSFORM", "TXT", "SSHFP", "CF_REDIRECT", "CF_TEMP_REDIRECT":
+		case "A", "AAAA", "ALIAS", "CAA", "IMPORT_TRANSFORM", "TXT", "SSHFP", "CF_REDIRECT", "CF_TEMP_REDIRECT", "CF_WORKER_ROUTE":
 			// These record types have a target that is case sensitive, or is an IP address. We leave them alone.
 			// Do nothing.
 		case "SOA":
--- a/providers/cloudflare/cloudflareProvider.go
+++ b/providers/cloudflare/cloudflareProvider.go
@@ -60,6 +60,7 @@ func init() {
 	providers.RegisterDomainServiceProviderType("CLOUDFLAREAPI", fns, features)
 	providers.RegisterCustomRecordType("CF_REDIRECT", "CLOUDFLAREAPI", "")
 	providers.RegisterCustomRecordType("CF_TEMP_REDIRECT", "CLOUDFLAREAPI", "")
+	providers.RegisterCustomRecordType("CF_WORKER_ROUTE", "CLOUDFLAREAPI", "")
 }
 
 // cloudflareProvider is the handle for API calls.
@@ -191,6 +192,12 @@ func (c *cloudflareProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 		records = append(records, prs...)
 	}
 
+	wrs, err := c.getWorkerRoutes(id, dc.Name)
+	if err != nil {
+		return nil, err
+	}
+	records = append(records, wrs...)
+
 	for _, rec := range dc.Records {
 		if rec.Type == "ALIAS" {
 			rec.Type = "CNAME"
@@ -227,6 +234,11 @@ func (c *cloudflareProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 				Msg: d.String(),
 				F:   func() error { return c.deletePageRule(ex.Original.(*pageRule).ID, id) },
 			})
+		} else if ex.Type == "WORKER_ROUTE" {
+			corrections = append(corrections, &models.Correction{
+				Msg: d.String(),
+				F:   func() error { return c.deleteWorkerRoute(ex.Original.(*workerRoute).ID, id) },
+			})
 		} else {
 			corr := c.deleteRec(ex.Original.(*cfRecord), id)
 			// DS records must always have a corresponding NS record.
@@ -244,6 +256,11 @@ func (c *cloudflareProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 			corrections = append(corrections, &models.Correction{
 				Msg: d.String(),
 				F:   func() error { return c.createPageRule(id, des.GetTargetField()) },
+			})
+		} else if des.Type == "WORKER_ROUTE" {
+			corrections = append(corrections, &models.Correction{
+				Msg: d.String(),
+				F:   func() error { return c.createWorkerRoute(id, des.GetTargetField()) },
 			})
 		} else {
 			corr := c.createRec(des, id)
@@ -264,6 +281,11 @@ func (c *cloudflareProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 			corrections = append(corrections, &models.Correction{
 				Msg: d.String(),
 				F:   func() error { return c.updatePageRule(ex.Original.(*pageRule).ID, id, rec.GetTargetField()) },
+			})
+		} else if rec.Type == "WORKER_ROUTE" {
+			corrections = append(corrections, &models.Correction{
+				Msg: d.String(),
+				F:   func() error { return c.updateWorkerRoute(ex.Original.(*workerRoute).ID, id, rec.GetTargetField()) },
 			})
 		} else {
 			e := ex.Original.(*cfRecord)
@@ -422,6 +444,16 @@ func (c *cloudflareProvider) preprocessConfig(dc *models.DomainConfig) error {
 			currentPrPrio++
 			rec.TTL = 1
 			rec.Type = "PAGE_RULE"
+		}
+
+		// CF_WORKER_ROUTE record types. Encode target as $PATTERN,$SCRIPT
+		if rec.Type == "CF_WORKER_ROUTE" {
+			parts := strings.Split(rec.GetTargetField(), ",")
+			if len(parts) != 2 {
+				return fmt.Errorf("invalid data specified for cloudflare worker record")
+			}
+			rec.TTL = 1
+			rec.Type = "WORKER_ROUTE"
 		}
 	}
 
--- a/providers/cloudflare/rest.go
+++ b/providers/cloudflare/rest.go
@@ -14,12 +14,14 @@ import (
 )
 
 const (
-	baseURL           = "https://api.cloudflare.com/client/v4/"
-	zonesURL          = baseURL + "zones/"
-	recordsURL        = zonesURL + "%s/dns_records/"
-	pageRulesURL      = zonesURL + "%s/pagerules/"
-	singlePageRuleURL = pageRulesURL + "%s"
-	singleRecordURL   = recordsURL + "%s"
+	baseURL              = "https://api.cloudflare.com/client/v4/"
+	zonesURL             = baseURL + "zones/"
+	recordsURL           = zonesURL + "%s/dns_records/"
+	singleRecordURL      = recordsURL + "%s"
+	pageRulesURL         = zonesURL + "%s/pagerules/"
+	singlePageRuleURL    = pageRulesURL + "%s"
+	workerRoutesURL      = zonesURL + "%s/workers/routes/"
+	singleWorkerRouteURL = workerRoutesURL + "%s"
 )
 
 // get list of domains for account. Cache so the ids can be looked up from domain name
@@ -505,6 +507,80 @@ func (c *cloudflareProvider) sendPageRule(endpoint, method string, data string) 
 	return err
 }
 
+func (c *cloudflareProvider) getWorkerRoutes(id string, domain string) ([]*models.RecordConfig, error) {
+	url := fmt.Sprintf(workerRoutesURL, id)
+	data := workerRouteResponse{}
+	if err := c.get(url, &data); err != nil {
+		return nil, fmt.Errorf("failed fetching worker route list from cloudflare: %s", err)
+	}
+	if !data.Success {
+		return nil, fmt.Errorf("failed fetching worker route list cloudflare: %s", stringifyErrors(data.Errors))
+	}
+	recs := []*models.RecordConfig{}
+	for _, pr := range data.Result {
+		var thisPr = pr
+		r := &models.RecordConfig{
+			Type:     "WORKER_ROUTE",
+			Original: thisPr,
+			TTL:      1,
+		}
+		r.SetLabel("@", domain)
+		r.SetTarget(fmt.Sprintf("%s,%s", // $PATTERN,$SCRIPT
+			pr.Pattern,
+			pr.Script))
+		recs = append(recs, r)
+	}
+	return recs, nil
+}
+
+func (c *cloudflareProvider) deleteWorkerRoute(recordID, domainID string) error {
+	endpoint := fmt.Sprintf(singleWorkerRouteURL, domainID, recordID)
+	req, err := http.NewRequest("DELETE", endpoint, nil)
+	if err != nil {
+		return err
+	}
+	c.setHeaders(req)
+	_, err = handleActionResponse(http.DefaultClient.Do(req))
+	return err
+}
+
+func (c *cloudflareProvider) updateWorkerRoute(recordID, domainID string, target string) error {
+	if err := c.deleteWorkerRoute(recordID, domainID); err != nil {
+		return err
+	}
+	return c.createWorkerRoute(domainID, target)
+}
+
+func (c *cloudflareProvider) createWorkerRoute(domainID string, target string) error {
+	endpoint := fmt.Sprintf(workerRoutesURL, domainID)
+	return c.sendWorkerRoute(endpoint, "POST", target)
+}
+
+func (c *cloudflareProvider) sendWorkerRoute(endpoint, method string, data string) error {
+	// $PATTERN,$SCRIPT
+	parts := strings.Split(data, ",")
+	pattern := parts[0]
+	script := parts[1]
+	wr := &workerRoute{
+		Pattern: pattern,
+		Script:  script}
+
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	if err := enc.Encode(wr); err != nil {
+		return err
+	}
+	req, err := http.NewRequest(method, endpoint, buf)
+	if err != nil {
+		return err
+	}
+	c.setHeaders(req)
+	// Cloudflare API returns error code 10001 without Content-Type.
+	req.Header.Set("Content-Type", "application/json")
+	_, err = handleActionResponse(http.DefaultClient.Do(req))
+	return err
+}
+
 func stringifyErrors(errors []interface{}) string {
 	dat, err := json.Marshal(errors)
 	if err != nil {
@@ -563,6 +639,18 @@ type pageRuleAction struct {
 type pageRuleFwdInfo struct {
 	URL        string `json:"url"`
 	StatusCode int    `json:"status_code"`
+}
+
+type workerRouteResponse struct {
+	basicResponse
+	Result     []*workerRoute `json:"result"`
+	ResultInfo pagingInfo     `json:"result_info"`
+}
+
+type workerRoute struct {
+	ID      string `json:"id,omitempty"`
+	Pattern string `json:"pattern"`
+	Script  string `json:"script"`
 }
 
 type zoneResponse struct {
--- a/models/domain.go
+++ b/models/domain.go
@@ -84,7 +84,7 @@ func (dc *DomainConfig) Punycode() error {
 				return err
 			}
 			rec.SetTarget(t)
-		case "CF_REDIRECT", "CF_TEMP_REDIRECT":
+		case "CF_REDIRECT", "CF_TEMP_REDIRECT", "CF_WORKER_ROUTE":
 			rec.SetTarget(rec.GetTargetField())
 		case "A", "AAAA", "CAA", "DS", "NAPTR", "SOA", "SSHFP", "TXT", "TLSA", "AZURE_ALIAS":
 			// Nothing to do.
--- a/pkg/js/helpers.js
+++ b/pkg/js/helpers.js
@@ -716,7 +716,8 @@ function recordBuilder(type, opts) {
             // Handle D_EXTEND() with subdomains.
             if (d.subdomain &&
                 record.type != 'CF_REDIRECT' &&
-                record.type != 'CF_TEMP_REDIRECT') {
+                record.type != 'CF_TEMP_REDIRECT' &&
+                record.type != 'CF_WORKER_ROUTE') {
                 fqdn = [d.subdomain, d.name].join(".")
 
                 record.subdomain = d.subdomain;
@@ -849,6 +850,18 @@ var CF_TEMP_REDIRECT = recordBuilder('CF_TEMP_REDIRECT', {
         record.target = args.source + ',' + args.destination;
     },
 });
+
+var CF_WORKER_ROUTE = recordBuilder('CF_WORKER_ROUTE', {
+    args: [
+        ['pattern', _validateCloudflareRedirect],
+        ['script', _validateCloudflareRedirect],
+    ],
+    transform: function(record, args, modifiers) {
+        record.name = '@';
+        record.target = args.pattern + ',' + args.script;
+    },
+});
+
 
 var URL = recordBuilder('URL');
 var URL301 = recordBuilder('URL301');


fdcastel:
Added initial tests, similar to `CF_REDIRECT` ones.

Guidance required for the integration tests (I didn't find where the expected result should be).

docs/_functions/domain/CF_WORKER_ROUTE.md
```diff
@@ -0,0 +1,29 @@
+---
+name: CF_WORKER_ROUTE
+parameters:
+  - pattern
+  - script
+---
+
+`CF_WORKER_ROUTE` uses [Cloudflare Workers](https://developers.cloudflare.com/workers/)
```
tlimoncelli, line 8: s/uses/uses the/

docs/_functions/domain/CF_WORKER_ROUTE.md
```diff
@@ -0,0 +1,29 @@
...
+  - pattern
+  - script
+---
+
+`CF_WORKER_ROUTE` uses [Cloudflare Workers](https://developers.cloudflare.com/workers/) 
+API to setup [routes](https://developers.cloudflare.com/workers/platform/routes)
+for a given domain.
+
+If _any_ `CF_WORKER_ROUTE` function is used then `dnscontrol` will manage _all_ 
+Worker Routes for the domain.
```
tlimoncelli, line 13: If this means it will delete routes that were added via the web portal, be extremely clear:

```
for the domain. To be clear: this means it will delete existing routes that were created outside of DNSControl.
```

docs/_functions/domain/CF_WORKER_ROUTE.md
```diff
@@ -0,0 +1,29 @@
...
+
+WARNING: This interface is not extensively tested. Take precautions such as making
+backups and manually verifying `dnscontrol preview` output before running
+`dnscontrol push`. This is especially true when mixing Worker Routes that are
```
tlimoncelli, line 17: Can we mix worker routes that aren't managed by DNSControl?  If not, this last sentence should make that clear.

models/record.go
```diff
@@ -34,12 +34,14 @@ import (
...
 //     CF_REDIRECT
 //     CF_TEMP_REDIRECT
+//     CF_WORKER_ROUTE
 //     FRAME
 //     IMPORT_TRANSFORM
 //     NAMESERVER
 //     NO_PURGE
 //     NS1_URLFWD
 //     PAGE_RULE
+//     WORKER_ROUTE
```
tlimoncelli, line 44: Is there a "WORKER_ROUTE"?

providers/cloudflare/cloudflareProvider.go
```diff
@@ -227,6 +234,11 @@ func (c *cloudflareProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 				Msg: d.String(),
 				F:   func() error { return c.deletePageRule(ex.Original.(*pageRule).ID, id) },
 			})
+		} else if ex.Type == "WORKER_ROUTE" {
```
tlimoncelli, line 237: Let's consistently use `CF_WORKER_ROUTE` and not `WORKER_ROUTE` (or are they 2 different things?)

providers/cloudflare/cloudflareProvider.go
```diff
@@ -245,6 +257,11 @@ func (c *cloudflareProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 				Msg: d.String(),
 				F:   func() error { return c.createPageRule(id, des.GetTargetField()) },
 			})
+		} else if des.Type == "WORKER_ROUTE" {
```
tlimoncelli, line 260: CF_WORKER_ROUTE?

providers/cloudflare/cloudflareProvider.go
```diff
@@ -265,6 +282,11 @@ func (c *cloudflareProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 				Msg: d.String(),
 				F:   func() error { return c.updatePageRule(ex.Original.(*pageRule).ID, id, rec.GetTargetField()) },
 			})
+		} else if rec.Type == "WORKER_ROUTE" {
```
tlimoncelli, line 285: CF_WORKER_ROUTE?

providers/cloudflare/cloudflareProvider.go
```diff
@@ -423,6 +445,16 @@ func (c *cloudflareProvider) preprocessConfig(dc *models.DomainConfig) error {
...
 		}
+
+		// CF_WORKER_ROUTE record types. Encode target as $PATTERN,$SCRIPT
+		if rec.Type == "CF_WORKER_ROUTE" {
+			parts := strings.Split(rec.GetTargetField(), ",")
+			if len(parts) != 2 {
+				return fmt.Errorf("invalid data specified for cloudflare worker record")
+			}
+			rec.TTL = 1
+			rec.Type = "WORKER_ROUTE"
```
tlimoncelli, line 456: CF_WORKER_ROUTE

providers/cloudflare/rest.go
```diff
@@ -505,6 +507,80 @@ func (c *cloudflareProvider) sendPageRule(endpoint, method string, data string)
...
+}
+
+func (c *cloudflareProvider) createWorkerRoute(domainID string, target string) error {
+	endpoint := fmt.Sprintf(workerRoutesURL, domainID)
+	return c.sendWorkerRoute(endpoint, "POST", target)
+}
+
+func (c *cloudflareProvider) sendWorkerRoute(endpoint, method string, data string) error {
+	// $PATTERN,$SCRIPT
+	parts := strings.Split(data, ",")
```
tlimoncelli, line 561: return an error if `len(parts) != 2`

docs/_functions/domain/CF_WORKER_ROUTE.md
fdcastel: To be frank, this was copied from CF_REDIRECT. The same logic applies here.

I confess I don't like the original text too I don't think anybody would use DNS Control and manually set some routes (since any route manually included will be removed in next push). Or I am missing something?

models/record.go
fdcastel: Yes. `WORKER_ROUTE` is for `CF_WORKER_ROUTE` what a `PAGE_RULE` is for a `CF_REDIRECT`/`CF_TEMP_REDIRECT`.

providers/cloudflare/cloudflareProvider.go
fdcastel: 2 different things. See last comment.

docs/_functions/domain/CF_WORKER_ROUTE.md
fdcastel: I think I got it. The original text was written that way because there are OTHER kind of page rules not touched by DNSControl (only "Forwarding URL" rules are managed by it).

I'm removing this sentence.

docs/_providers/cloudflare.md
```diff
@@ -170,3 +170,20 @@ Notice a few details:
 2. The IP address in those A records may be mostly irrelevant, as cloudflare should handle all requests (assuming some page rule matches).
 3. Ordering matters for priority. CF_REDIRECT records will be added in the order they appear in your js. So put catch-alls at the bottom.
 4. if _any_ `CF_REDIRECT` or `CF_TEMP_REDIRECT` functions are used then `dnscontrol` will manage _all_ "Forwarding URL" type Page Rules for the domain. Page Rule types other than "Forwarding URL” will be left alone.
+
+## Worker routes
+The Cloudflare provider can manage Worker Routes for your domains. Simply use the `CF_WORKER_ROUTE` function passing the route pattern and the worker name:
```
tlimoncelli, line 175: Repeat the same paragraph as before:

```
If _any_ `CF_WORKER_ROUTE` function is used then `dnscontrol` will manage _all_ 
Worker Routes for the domain. To be clear: this means it will delete existing routes that
were created outside of DNSControl.
```

providers/cloudflare/rest.go
```diff
@@ -505,6 +507,83 @@ func (c *cloudflareProvider) sendPageRule(endpoint, method string, data string)
...
+	if len(parts) != 2 {
+		return fmt.Errorf("unexpected data: '%s' (expected: 'PATTERN,SCRIPT')", data)
+	}
+	pattern := parts[0]
+	script := parts[1]
+	wr := &workerRoute{
+		Pattern: pattern,
+		Script:  script}
```
tlimoncelli, line 569: Please format the last item like this so that diff's in the future will be smaller.

```
    Script: script,
}
```

tlimoncelli (commented):
Just a few more comments.

* Please rebase
* Please add a test to integrationTest/integration_test.go

fdcastel:
About the integration tests... I am somewhat lost. 😢 

How do we test this? How could I check if the test is passing or not (won't I need some API credentials to run it locally?). 

I'm surely missing something here 😅

tlimoncelli:
Here is how to add a simple integration test: https://github.com/StackExchange/dnscontrol/pull/1250

To run the integration tests you'll need a test domain:

```
export CLOUDFLAREAPI_DOMAIN='dnscontroltest-cf.com'      <<< CHANGE TO YOUR DOMAIN
export CLOUDFLAREAPI_TOKEN='REDACTED'
go test -v -verbose -provider CLOUDFLAREAPI
```

Even without that PR, I've found a small bug.  It looks like the code assumes that there will be a worker route list.

```
tlimwork5:~/git/dnscontrol/integrationTest$ go test -v -verbose -provider CLOUDFLAREAPI
=== RUN   TestDNSProviders
=== RUN   TestDNSProviders/dnscontroltest-cf.com
=== RUN   TestDNSProviders/dnscontroltest-cf.com/Clean_Slate:Empty
{"success":false,"errors":[{"code":10000,"message":"Authentication error"}]}

    integration_test.go:193: runTests: failed fetching worker route list from cloudflare: bad status code from cloudflare: 403 not 200
=== RUN   TestDNSProviders/dnscontroltest-cf.com/00:GeneralACD:Create_an_A_record
{"success":false,"errors":[{"code":10000,"message":"Authentication error"}]}

    integration_test.go:193: runTests: failed fetching worker route list from cloudflare: bad status code from cloudflare: 403 not 200
=== RUN   TestDNSProviders/dnscontroltest-cf.com/Post_cleanup:Empty
{"success":false,"errors":[{"code":10000,"message":"Authentication error"}]}
```

tlimoncelli:
FYI: The automated tests (https://github.com/StackExchange/dnscontrol/pull/1243/checks) are skipping CloudFlare because it can't find any credentials.  Github Actions uses the credentials from the fdcastel/dnscontrol repo and there are none.  Once it is merged to StackExchange/dnscontrol it will have access to the credentials in StackExchange/dnscontrol and the tests will run. (this is to protect someone from writing a PR that steals the creds.)

fdcastel:
I'm still missing something. Even after pulling your sample code and settings my env vars like:

```powershell
$env:CLOUDFLAREAPI_DOMAIN='my-cf-domain.com'
$env:CLOUDFLAREAPI_KEY='my-api-key'
$env:CLOUDFLAREAPI_USER='my-api-user'
```
A call to 
```powershell
go test -v -verbose -provider CLOUDFLAREAPI
```

gives me

```
?   	github.com/StackExchange/dnscontrol/v3	[no test files]
``` 

_Running on Windows 10 / Powershell_

fdcastel:
Just to be clear: I already read https://stackexchange.github.io/dnscontrol/byo-secrets and it seems clear to me. 

But I'm trying to run the tests locally on my machine before I put them into the CI. 

Is this possible or I misunderstood something? (i.e., the tests would only run inside Github CI engine?)

fdcastel:
Got it!

```powershell
go test -v ./integrationTest -verbose -provider CLOUDFLAREAPI  
```

Tests are running now 🙌 . Working on it.

fdcastel:
There is some way -- in the current test infrastructure -- to run additional code _before_ some `testgroup` starts to run?

Rationale: I have the following test:

```javascript
  testgroup("CF_WORKER_ROUTE",
    only("CLOUDFLAREAPI"),
    tc("simple", cfWorkerRoute("cnn.**current-domain-no-trailing**/*", "cnn-worker")),
    tc("changeScript", cfWorkerRoute("cnn.**current-domain-no-trailing**/*", "msnbc-worker")),
```

But before it, I need to call an additional Cloudflare api to _create_ `cnn-worker` and `msnbc-worker` workers. 

Without this, the API calls under test will fail with a

```json
[{"code":10019,"message":"workers.api.error.invalid_route_script_missing"}]
```

fdcastel:
✔️ Rebase done. 
✔️ Integration tests done.
✔️ Requested changes done.
❗ Integration tests *need two cloudflare workers* (named `cnn-worker` and `msnbc-worker`) on cloudflare account to pass.

tlimoncelli:
> There is some way -- in the current test infrastructure -- to run additional code _before_ some `testgroup` starts to run?
> 
> Rationale: I have the following test:
> 
> ```js
>   testgroup("CF_WORKER_ROUTE",
>     only("CLOUDFLAREAPI"),
>     tc("simple", cfWorkerRoute("cnn.**current-domain-no-trailing**/*", "cnn-worker")),
>     tc("changeScript", cfWorkerRoute("cnn.**current-domain-no-trailing**/*", "msnbc-worker")),
> ```
> 
> But before it, I need to call an additional Cloudflare api to _create_ `cnn-worker` and `msnbc-worker` workers.
> 
> Without this, the API calls under test will fail with a
> 
> ```json
> [{"code":10019,"message":"workers.api.error.invalid_route_script_missing"}]
> ```

There isn't a uniform way to do setup code.  However....

Around line 55 of integration_test.go we do some "just for CloudFlare" actions.  Maybe that could call a function that creates the workers if they don't already exist?  Please create them with names that won't overlap real workers... maybe dnsctrl_integrationtest_foo and dnsctrl_integrationtest_bar ?

Tom

fdcastel:
Minor update: Yet working on it. I should update this PR in a few days. Thanks, @tlimoncelli !

fdcastel:
_Sorry for the long delay. It has been a long week around here._

It's almost done. However, **I'm having a hard time making Cloudflare-specific API calls from `integration_test.go`.** It seems wrong to call provider-specific code here. I don't event have the context for Cloudflare account, creds, etc.

One solution would be to add a `PrepareTests()` in `DNSProvider` interface. However this would add a lot of empty boilerplate code for every other provider which doesn't need this. Not good, IMHO.

Could we use `preprocess_test.go` for the test workers creation? There is *any* guarantee that this code would run *before* the tests from `integration_test.go`?

fdcastel:
Pushed what I got so far. Looking for advice.

I added `TestPreprocess_CreateTestWorkers` in `preprocess_test.go` but it is not being called before the integration tests.

I also added two new env vars `CLOUDFLAREAPI_ACCOUNTID` and `CLOUDFLAREAPI_ACCOUNTNAME` which were already present in `creds.json` but without any way to set them via environment. 

AccountName is not used, actually. However, there is [a test in `cloudflareProvider.go`](https://github.com/StackExchange/dnscontrol/blob/master/providers/cloudflare/cloudflareProvider.go#L466) which enforces both vars to be set.

tlimoncelli:
Hi there!

Sorry for the delay.  It's been a crazy week!

A few things:

1.  I haven't been able to get this running since I can't figure out the permissions.  Do you know what perms to give this API key to enable it to use workers?

2.  I've noticed that my API key doesn't have access to Workers at all (maybe because its a free account?). I tried to update the API's permissions but I get:
```
    integration_test.go:193: runTests: failed fetching worker route list from cloudflare: bad status code from cloudflare: 403 not 200
```

The code either needs to automagically detect if Workers is disabled and be silent or there needs to be a feature-flag to enable Workers.  Otherwise this will be a breaking change... people that don't use workers can't break and can't be asked to modify permissions on the API key just because of a new feature.

I don't know how to detect if workers is enabled, but I'm making a PR that adds the feature flag in case you go that route: https://github.com/fdcastel/dnscontrol/pull/1/files

3. `preprocess_test.go` is a unit test and isn't expected to run as part of the integration tests.  Instead, it runs when you do "go test -count=1 ./..." from `providers/cloudflare`.

The right place for `TestPreprocess_CreateTestWorkers` is in `integrationTest/`.  Yes, that's a layering violation but... that's life.  I would move it to `integrationTest/cloudflare.go` and call it cloudflareSetupWorkers().  You can call it from  `integration_test.go` around line 53 where we do some cloudflare-specific things already.

At least I think that will work. I'm not entirely sure as I can't test it myself until I fix my own permission problem.

I'm in the US/Eastern timezone if you would like to screenshare and work on this in real time. 

Tom

fdcastel:
Thanks, Tom, for the insights and the pragmatism :wink: - I will return to this in a few hours.

About your points:

1. I'm using Cloudflare **API Keys** (which, as I understand, gives me _sudo-like_ access to the account) instead of **API Tokens** (which allows you more granular control about what can be accessed). That said, after a quick glimpse at the options available in Cloudflare dashboard, I believe these are what you will need:

image

2. You definitely can use Workers on a Free account (my own test domain is on this very same plan). So, the problem is probably permission-related. I will investigate a bit more later to know what specific permissions are needed for token-related access (but I would guess that the above image is very close).
 
3. Perfect! Got it!

fdcastel:
Rebase to latest release.

fdcastel:
Added feature flag for CF Workers per @tlimoncelli suggestion.

Porting tests back to `integration_tests/` now -- https://www.youtube.com/watch?v=TnlPtaPxXfc

fdcastel:
All done! Integration tests are creating test workers now.

Two considerations:

- I added 'testing' module in `cloudflare` module. Surely I'm not proud of this and would welcome any suggestions to improve (if deemed necessary).

- Tom: I believe we now have a better understanding about [that comment](https://github.com/StackExchange/dnscontrol/blob/master/integrationTest/integration_test.go#L49) of `integration_test.go` :wink: (needed to avoid breaking changes?). We could improve it explaining the rationale behind the feature flags.

fdcastel:
I ran all tests twice (first without test workers and then with them) in a test domain and everything looks good to me.

Save for further coding improvements (see last comment) I would consider the work now done.

Also, if agreed, I would like to rewrite the commit history to put the commits in a better shape before merge.

tlimoncelli:
Looks very good so far!

I do want to find a way to make legacy domains "just work".  Here's the one I use for tests:

```
$ go test -v -verbose -provider CLOUDFLAREAPI
=== RUN   TestDNSProviders
    cloudflareProvider.go:699: [{"code":7003,"message":"Could not route to /accounts/workers/scripts/dnscontrol_integrationtest_cnn, perhaps your object identifier is invalid?"},{"code":7000,"message":"No route for that URI"}]
--- FAIL: TestDNSProviders (0.92s)
=== RUN   TestDualProviders
    cloudflareProvider.go:699: [{"code":7003,"message":"Could not route to /accounts/workers/scripts/dnscontrol_integrationtest_cnn, perhaps your object identifier is invalid?"},{"code":7000,"message":"No route for that URI"}]
--- FAIL: TestDualProviders (0.87s)
FAIL
exit status 1
FAIL	github.com/StackExchange/dnscontrol/v3/integrationTest	2.041s

```

fdcastel:
Sorry, Tom. I should have stressed this :

**To make the tests work you must now set `CLOUDFLAREAPI_ACCOUNTID` and `CLOUDFLAREAPI_ACCOUNTNAME`.** These are required to the API calls which create the workers.

That `/accounts/workers/scripts/dnscontrol_integrationtest_cnn` path on your error message should be `/accounts/<accoundid>/workers/scripts/dnscontrol_integrationtest_cnn`.

Maybe I should add a test and emit a `Fail()` in this case (missing account id and account name) to warn the user?

P.S: AccountName is not used at all. It is only needed because of another check. Please see [this comment](https://github.com/StackExchange/dnscontrol/pull/1243#issuecomment-917439566).

tlimoncelli:
Hi there!

I'm sorry but I'm having trouble getting this to run on a legacy domain.

It is very important that we make this change in an upwards compatible manner.  Currently we get errors like:

```
$ go test -v -verbose -provider CLOUDFLAREAPI
=== RUN   TestDNSProviders
    cloudflareProvider.go:699: [{"code":7003,"message":"Could not route to /accounts/workers/scripts/dnscontrol_integrationtest_cnn, perhaps your object identifier is invalid?"},{"code":7000,"message":"No route for that URI"}]
--- FAIL: TestDNSProviders (0.96s)
=== RUN   TestDualProviders
    cloudflareProvider.go:699: [{"code":7003,"message":"Could not route to /accounts/workers/scripts/dnscontrol_integrationtest_cnn, perhaps your object identifier is invalid?"},{"code":7000,"message":"No route for that URI"}]
--- FAIL: TestDualProviders (0.80s)
FAIL
exit status 1
FAIL	github.com/StackExchange/dnscontrol/v3/integrationTest	2.087s
```

How about we add a feature-flag?  For example, with "CLOUDFLARE_WORKERS=1" the new code runs. If it isn't set, the new code silently disables itself.

Tom

fdcastel:
Sorry, Tom. I don't understand. Did you see my [last message](https://github.com/StackExchange/dnscontrol/pull/1243#issuecomment-923488723)? 

According to your error message:

```
 [{"code":7003,"message":"Could not route to /accounts/workers/scripts/dnscontrol_integrationtest_cnn, perhaps your object identifier is invalid?"},{"code":7000,"message":"No route for that URI"}]
```

It seems that the `CLOUDFLAREAPI_ACCOUNTID` wasn't set during your test. I can only presume this since the path above should be `/accounts/<ACCOUNT-ID>/workers/...`.

To make the things clearer, this is how I ran the tests:

```powershell
go build 

$env:CLOUDFLAREAPI_DOMAIN='my-test-domain.com'
$env:CLOUDFLAREAPI_KEY='my-api-key'
$env:CLOUDFLAREAPI_USER='my@email.com'

$env:CLOUDFLAREAPI_ACCOUNTID='my-account-id'
$env:CLOUDFLAREAPI_ACCOUNTNAME='MyAccountName'

go test -v ./integrationTest -verbose -provider CLOUDFLAREAPI  
```

The `AccountId` can be seen on the browser url after the first logon to https://dash.cloudflare.com. You could also see it on "Workers" section from Cloudflare administration panel:

image

Please tell me how this has worked for you. I don't see the need for an _additional_ feature flag given that we already added one (`manage_workers`) exactly in the name of backward compatibility.

tresni:
#1267 and this will conflict. Happy to help build this on top of #1267 if that makes sense.

fdcastel:
Whoa! Fantastic work, @tresni ! 😄 

Working on it right now.

fdcastel:
All done! Lots of code slashed 😄. Commit history rewrite. All tests passing.

@tresni could you kindly review https://github.com/StackExchange/dnscontrol/pull/1243/commits/b55278140f63e0f95a3418bcd61b810c731ae08f ? I need `cfClient.AccountID` to be defined and I believe that's the right place (but please correct me if I'm wrong).

@tlimoncelli Did you manage to make [the tests pass](https://github.com/StackExchange/dnscontrol/pull/1243#issuecomment-924072372) with your domain? Do we need anything else to merge this? (Please? Pretty please? 😉)

tlimoncelli:
Sadly I'm still getting an error:

```
$ go test -v -verbose -provider CLOUDFLAREAPI
=== RUN   TestDNSProviders
=== RUN   TestDNSProviders/dnscontroltest-cf.com
=== RUN   TestDNSProviders/dnscontroltest-cf.com/Clean_Slate:Empty
    integration_test.go:197: runTests: failed fetching worker route list cloudflare: HTTP status 403: Authentication error (10000)
=== RUN   TestDNSProviders/dnscontroltest-cf.com/00:GeneralACD:Create_an_A_record
    integration_test.go:197: runTests: failed fetching worker route list cloudflare: HTTP status 403: Authentication error (10000)
=== RUN   TestDNSProviders/dnscontroltest-cf.com/Post_cleanup:Empty
    integration_test.go:197: runTests: failed fetching worker route list cloudflare: HTTP status 403: Authentication error (10000)
=== RUN   TestDNSProviders/dnscontroltest-cf.com/01:WildcardACD:Create_wildcard
    integration_test.go:197: runTests: failed fetching worker route list cloudflare: HTTP status 403: Authentication error (10000)
=== RUN   TestDNSProviders/dnscontroltest-cf.com/Post_cleanup:Empty#01
    integration_test.go:197: runTests: failed fetching worker route list cloudflare: HTTP status 403: Authentication error (10000)
```

tresni:
@fdcastel thats probably a better way to handle it then I did later (in the one place we used it previously.) I’ll check this locally and see if I can help you clean up the tests/get them working.

tlimoncelli:
We're getting closer!

Getting this to work in a way that is upwards compatible with existing users is... difficult.  I'm experimenting with some ideas.  The first is to simply have a creds.json variable that will be a feature-flag, enabling all this stuff.  It's kind of a big hammer for a small problem, but it works.

My work in progress is https://github.com/StackExchange/dnscontrol/pull/1269

providers/cloudflare/cloudflareProvider.go
```diff
@@ -638,3 +680,19 @@ func (c *cloudflareProvider) EnsureDomainExists(domain string) error {
 	fmt.Printf("Added zone for %s to Cloudflare account: %s\n", domain, id)
 	return err
 }
+
+// PrepareCloudflareWorkers creates Cloudflare Workers required for CF_WORKER_ROUTE tests.
```
tresni, line 684: I feel like this should move to the integrationTests source to prevent bringing `testing` into the production build.

providers/cloudflare/rest.go
```diff
@@ -282,6 +282,76 @@ func (c *cloudflareProvider) createPageRule(domainID string, target string) erro
...
+	wr := cloudflare.WorkerRoute{
+		Pattern: parts[0],
+		Script:  parts[1],
+	}
+
+	_, err := c.cfClient.CreateWorkerRoute(context.Background(), domainID, wr)
+	return err
+}
+
+func (c *cloudflareProvider) createTestWorker(workerName string) error {
```
tresni, line 338: If we moved this into a `something_test.go` file would the build still work.  It's not something we are actually using in production code, but I totally understand why it's in the cloudflare provider space.

tresni:
@tlimoncelli this PR seems to work fine for me.  I don't use cloudflare workers normally and everything runs as expected (I don't set accountId and `"manage_workers"` defaults to `false`.)  This is fully backwards compatible with existing code/deployments as is.

@fdcastel  Here are the changes I made to work with latest master and get `testing` out of the main code. Still don't like that there is test setup code in here, but I tried everything I could think of, no easy way to move it out completely unless we make `cloudflareProvider` public (i.e. `CloudflareProvider`) as that would allow us to move all the setup code into intergration_test.go.

Also, this doesn't cleanup the workers after-the-fact which may not be great, but I don't think it's really hurting anything.


```patch
diff --git a/integrationTest/integration_test.go b/integrationTest/integration_test.go
index a8250c51..28e749e4 100644
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -70,8 +70,12 @@ func getProvider(t *testing.T) (providers.DNSServiceProvider, string, map[int]bo
 			}
 		}
 
-		// Cloudflare only. Will do nothing if provider != *cloudflareProvider.
-		cloudflare.PrepareCloudflareTestWorkers(t, provider)
+		if name == "CLOUDFLAREAPI" {
+			// Cloudflare only. Will do nothing if provider != *cloudflareProvider.
+			if err := cloudflare.PrepareCloudflareTestWorkers(provider); err != nil {
+				t.Fatal(err)
+			}
+		}
 
 		return provider, cfg["domain"], fails, cfg
 	}
diff --git a/providers/cloudflare/cloudflareProvider.go b/providers/cloudflare/cloudflareProvider.go
index 9ceb4418..e1e4851e 100644
--- a/providers/cloudflare/cloudflareProvider.go
+++ b/providers/cloudflare/cloudflareProvider.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"net"
 	"strings"
-	"testing"
 
 	"github.com/cloudflare/cloudflare-go"
 	"github.com/miekg/dns/dnsutil"
@@ -506,8 +505,6 @@ func newCloudflare(m map[string]string, metadata json.RawMessage) (providers.DNS
 		api.cfClient.AccountID = m["accountid"]
 	}
 
-	api.cfClient.AccountID = api.AccountID
-
 	if len(metadata) > 0 {
 		parsedMeta := &struct {
 			IPConversions   string   `json:"ip_conversions"`
@@ -674,17 +671,18 @@ func (c *cloudflareProvider) EnsureDomainExists(domain string) error {
 }
 
 // PrepareCloudflareWorkers creates Cloudflare Workers required for CF_WORKER_ROUTE tests.
-func PrepareCloudflareTestWorkers(t *testing.T, prv providers.DNSServiceProvider) {
+func PrepareCloudflareTestWorkers(prv providers.DNSServiceProvider) error {
 	cf, ok := prv.(*cloudflareProvider)
 	if ok {
 		err := cf.createTestWorker("dnscontrol_integrationtest_cnn")
 		if err != nil {
-			t.Fatal(err)
+			return err
 		}
 
 		err = cf.createTestWorker("dnscontrol_integrationtest_msnbc")
 		if err != nil {
-			t.Fatal(err)
+			return err
 		}
 	}
+	return nil
 }
```

tresni:
@tlimoncelli to be clear, I ran the integration tests as follows.  I've sanitized the domain I use as it's not really a testing domain :-P

```shell
$ cd integrationTest
$ export CLOUDFLAREAPI_ACCOUNTID=[ACCOUNT_ID]
$ export CLOUDFLAREAPI_DOMAIN=[DOMAIN]
$ export CLOUDFLAREAPI_TOKEN=[TOKEN]
$ go test -v -verbose -provider CLOUDFLAREAPI --start 41
=== RUN   TestDNSProviders
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/Clean_Slate:Empty
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:simple
    integration_test.go:208: CREATE WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN} cnn.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_cnn ttl=1
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:changeScript
    integration_test.go:208: MODIFY WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN}: (cnn.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_cnn ttl=1) -> (cnn.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_msnbc ttl=1)
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:changePattern
    integration_test.go:208: MODIFY WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN}: (cnn.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_msnbc ttl=1) -> (cable.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_msnbc ttl=1)
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:Empty
    integration_test.go:208: DELETE WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN} cable.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_msnbc ttl=1
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:createMultiple
    integration_test.go:208: CREATE WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN} cnn.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_cnn ttl=1
    integration_test.go:208: CREATE WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN} msnbc.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_msnbc ttl=1
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:addOne
    integration_test.go:208: CREATE WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN} api.${CLOUDFLAREAPI_DOMAIN}/cnn/*,dnscontrol_integrationtest_cnn ttl=1
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:changeOne
    integration_test.go:208: MODIFY WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN}: (msnbc.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_msnbc ttl=1) -> (msn.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_msnbc ttl=1)
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:deleteOne
    integration_test.go:208: DELETE WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN} cnn.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_cnn ttl=1
=== RUN   TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/Post_cleanup:Empty
    integration_test.go:208: DELETE WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN} api.${CLOUDFLAREAPI_DOMAIN}/cnn/*,dnscontrol_integrationtest_cnn ttl=1
    integration_test.go:208: DELETE WORKER_ROUTE ${CLOUDFLAREAPI_DOMAIN} msn.${CLOUDFLAREAPI_DOMAIN}/*,dnscontrol_integrationtest_msnbc ttl=1
--- PASS: TestDNSProviders (31.58s)
    --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN} (24.06s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/Clean_Slate:Empty (1.49s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:simple (2.36s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:changeScript (2.69s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:changePattern (2.78s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:Empty (1.06s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:createMultiple (3.17s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:addOne (2.31s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:changeOne (3.45s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/41:CF_WORKER_ROUTE:deleteOne (2.65s)
        --- PASS: TestDNSProviders/${CLOUDFLAREAPI_DOMAIN}/Post_cleanup:Empty (1.99s)
=== RUN   TestDualProviders
    integration_test.go:313: Clearing everything
    integration_test.go:319: Adding nameservers from another provider
WARNING: cloudflare does not support modifying NS records on base domain. ns1.example.com. will not be added.
WARNING: cloudflare does not support modifying NS records on base domain. ns2.example.com. will not be added.
    integration_test.go:322: Running again to ensure stability
WARNING: cloudflare does not support modifying NS records on base domain. ns1.example.com. will not be added.
WARNING: cloudflare does not support modifying NS records on base domain. ns2.example.com. will not be added.
--- PASS: TestDualProviders (9.31s)
PASS
ok  	github.com/StackExchange/dnscontrol/v3/integrationTest	41.469s
$
```

tlimoncelli:
I'm getting a build error:

```
$ go install
# github.com/StackExchange/dnscontrol/v3/providers/cloudflare
providers/cloudflare/cloudflareProvider.go:509:30: api.AccountID undefined (type *cloudflareProvider has no field or method AccountID)
```

tresni:
@tlimoncelli Yeah, need the patch I included above to make it work with latest master.

fdcastel:
Sorry for the delay. Working on it now!

fdcastel:
Rebase done with latest master. @tresni patch applied. All tests passing. 🎉

fdcastel:
@tresni 

> Still don't like that there is test setup code in here, but I tried everything I could think of, no easy way to move it out completely unless we make `cloudflareProvider` public (i.e. `CloudflareProvider`) as that would allow us to move all the setup code into intergration_test.go.

Fully agreed. I don't like this either. We just took the pragmatic approach, here. 

> Also, this doesn't cleanup the workers after-the-fact which may not be great, but I don't think it's really hurting anything.

Yep. Not ideal. But, also agreed, not a big problem.

Thanks for the patch for the latest master. 👍

fdcastel:
@tlimoncelli please let me know if you have any problem with the integration tests.

.github/workflows/build.yml
```diff
@@ -94,6 +94,8 @@ jobs:
       CLOUDFLAREAPI_KEY: ${{ secrets.CLOUDFLAREAPI_KEY }}
       CLOUDFLAREAPI_TOKEN: ${{ secrets.CLOUDFLAREAPI_TOKEN }}
       CLOUDFLAREAPI_USER: ${{ secrets.CLOUDFLAREAPI_USER }}
+      CLOUDFLAREAPI_ACCOUNTID: ${{ secrets.CLOUDFLAREAPI_ACCOUNTID }}
+      CLOUDFLAREAPI_ACCOUNTNAME: ${{ secrets.CLOUDFLAREAPI_ACCOUNTNAME }}
```
tresni, line 98: You can remove this, it's not needed anymore :)

tlimoncelli:
Here's what I get:

```
tlimwork5:~/git/dnscontrol/integrationTest$ export CLOUDFLAREAPI_DOMAIN='dnscontroltest-cf.com'
tlimwork5:~/git/dnscontrol/integrationTest$ export CLOUDFLAREAPI_TOKEN='redacted'
tlimwork5:~/git/dnscontrol/integrationTest$ echo go test -v -verbose -provider CLOUDFLAREAPI
tlimwork5:~/git/dnscontrol/integrationTest$ go test -v -verbose -provider CLOUDFLAREAPI
=== RUN   TestDNSProviders
    integration_test.go:76: account ID required
--- FAIL: TestDNSProviders (0.00s)
=== RUN   TestDualProviders
    integration_test.go:76: account ID required
--- FAIL: TestDualProviders (0.00s)
FAIL
exit status 1
FAIL	github.com/StackExchange/dnscontrol/v3/integrationTest	0.250s

Since that failed, I tried setting the account id:

tlimwork5:~/git/dnscontrol/integrationTest$ export CLOUDFLAREAPI_ACCOUNTID=a5d8f5aa76a5b5518741f059b5189f9d
tlimwork5:~/git/dnscontrol/integrationTest$ go test -v -verbose -provider CLOUDFLAREAPI
=== RUN   TestDNSProviders
=== RUN   TestDNSProviders/dnscontroltest-cf.com
=== RUN   TestDNSProviders/dnscontroltest-cf.com/Clean_Slate:Empty
    integration_test.go:201: runTests: failed fetching worker route list cloudflare: HTTP status 403: Authentication error (10000)
=== RUN   TestDNSProviders/dnscontroltest-cf.com/00:GeneralACD:Create_an_A_record
    integration_test.go:201: runTests: failed fetching worker route list cloudflare: HTTP status 403: Authentication error (10000)
=== RUN   TestDNSProviders/dnscontroltest-cf.com/Post_cleanup:Empty
    integration_test.go:201: runTests: failed fetching worker route list cloudflare: HTTP status 403: Authentication error (10000)
=== RUN   TestDNSProviders/dnscontroltest-cf.com/01:WildcardACD:Create_wildcard
```

However if I run "dnscontrol preview" on Stackoverflow's dnsconfig.js or my personal dnsconfig.js it works just fine.  So, that tells me that it is only the integration tests that are still poking at workers when they shouldn't.

fdcastel:
All done! Clean merge. Thanks @tlimoncelli !! 🙌

PR #1259, opened 2021-09-16
Modify workflows for multiarch builds

m-ueberall:
This is a "proof-of-concept" which demonstrates how to modify the existing `release.yml` action in order to build both packages and binaries for multiple architectures–namely, amd64/x86_64 and arm64/aarch64–given that matching runners are registered.  Primary objective was to keep individual changes _as simple as possible_, so a number of steps have simply been duplicated.

The result of test runs can be found [here](https://github.com/m-ueberall/dnscontrol/releases/tag/v3.12.0.1) (using two runners) and [here](https://github.com/m-ueberall/dnscontrol/releases/tag/v3.12.0.2) (using a single runner). All produced arm64/aarch64 binaries/packages have been tested successfully using both current Ubuntu Focal and Fedora 33 containers.

PR #1260, opened 2021-09-17
Modify workflows for multiarch builds using a single runner

m-ueberall:
This is a "proof-of-concept" which demonstrates how to modify the existing `release.yml` action in order to build both packages and binaries for multiple architectures using a single runner.  Primary objective was to keep individual changes _as simple as possible_, therefore most of the required changes have been moved to two additional shell scripts.

The result of a test run can be found [here](https://github.com/m-ueberall/dnscontrol/releases/tag/v3.12.0.3). All produced arm64/aarch64 Linux binaries/packages have been tested successfully using both current Ubuntu Focal and Fedora 33 containers.


cdhunt:
@m-ueberall  Why are you discontinuing the use of `build.go`? That already provides logic for multi-platform builds including Windows.

m-ueberall:
> @m-ueberall Why are you discontinuing the use of `build.go`? That already provides logic for multi-platform builds including Windows.

I don't necessarily see `.ci/build.sh` as a **replacemant** for "_personal builds_", because the former also creates different OS-/distribution-specific packages.  "Merging" this functionality into a GitHub-specific action w.r.t. to all required mappings (i.e. a modification of the existing `.github/workflows/release.yml`) looks like a pain, though (I wouldn't know how to easily write/maintain it) and also has the disadvantage that you cannot simply execute this on your local machine _if need be_.  Extending `build.go` would be more straightforward, but I still see a lot of overhead w.r.t. required parameters/switches (i.e. code) and documentation – the latter currently tells you everything about `build.go` you need to know and is easy to read. So basically my pull request is primarily about the replacement of the existing `release.yml` action.

m-ueberall:
For the record, in PR #1260, @jauderho mentioned [`goreleaser`](https://github.com/goreleaser/goreleaser/) as a potential alternative approach/replacement for the existing DNSControl `release.yml` action. It currently covers all but the FreeBSD `.txz` archives created by `build.sh` as part of this PR (see [my comment over there](https://github.com/StackExchange/dnscontrol/pull/1259#issuecomment-930543473)).

tlimoncelli:
FYI: In 3-6 months we plan on moving to CircleCI.  That might affect how you do this.  For example, do a quick-fix now and save anything elegant for when we move to CCI.

cdhunt:
+1 for goreleaser.

https://goreleaser.com/ci/circle/

jauderho:
@tlimoncelli are you able to share any insight into the thought process of picking CircleCI over say GHA?

Having this discussion at work right now and it'd be helpful to get some different viewpoints.

tlimoncelli:
> @tlimoncelli are you able to share any insight into the thought process of picking CircleCI over say GHA?
> 
> Having this discussion at work right now and it'd be helpful to get some different viewpoints.

Short version: it's faster ;-)

Long version: talk to @cdhunt

cdhunt:
Like @tlimoncelli said, mostly speed. This job runs on every commit to the stackoverflow.com codebase.

CircleCI: 
image

GitHub Actions:
image

There is also better tooling for what I'd call Enterprise-Scale CI/CD like really important things such as test history reporting.

tlimoncelli:
We've decided to move back to Github Actions.  No more CircleCI for us.

The release process was already in GHA.  As far as I can tell, it looks like the current GHA process generates all the correct architectures.

I'm closing this PR.   Please re-open it if you feel strongly about implementing these features or open an issue to report any problems with the current system.

PR #1262, opened 2021-09-23
docs: add clarity around named profiles

### LINKED ISSUE #1261: Route 53 Named Profiles Fails on >3.8.0 without creds.json
If running on version >3.8.0, any operation that uses creds seems to _require_ a creds.json file, whereas 3.8.0 works fine without one. This isn't made clear in the Route 53 documentation around [named profiles](https://stackexchange.github.io/dnscontrol/providers/route53), it seems like just exporting the relevant environment variables is enough.

Will submit a PR shortly with proposed clarification.

biodrone:
Specifically that creds.json is now required for versions >3.8.0 when using named profiles, as well as the environment variables.

fixes #1261

### DIFF
--- a/docs/_providers/route53.md
+++ b/docs/_providers/route53.md
@@ -43,6 +43,20 @@ $ export AWS_SDK_LOAD_CONFIG=1
 $ export AWS_PROFILE=ZZZZZZZZ
 ```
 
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Thanks for the tip!

tlimoncelli:
P.S. I love this kind of doc update.  Someone had a problem, figured it out, and updated the docs so that the next person won't have that problem.  You've  saved time for all the people that follow in your footsteps!  Thanks!

PR #1265, opened 2021-09-28
Remove SplitSingleLongTxt call for Cloudflare

tresni:
This isn't necessary for Cloudflare, just pass the string you want and they'll take care of the rest.  This was breaking DKIM records as it would pass the split string with quotes, which Cloudflare then interpreted the quotes as being necessary for the records.

Integration tests all pass with this.


tlimoncelli (approved):

tlimoncelli (approved):

tlimoncelli:
Thanks for finding that and providing the fix!

PR #1267, opened 2021-09-29
Use cloudflare-go

tresni:
Previously this was using a homemade implementation of the Cloudflare API.  I really wanted to implement registrar support for Cloudflare, but I didn't want to have to write all the necessary structs by hand.  This also means we should automatically gain benefits of any improvements in the [cloudflare-go](https://github.com/cloudflare/cloudflare-go) library.

Integration tests look good, but go-staticcheck says that `pageRuleConstraint` is unused which is a dirty lie. Soon I should have another PR adding support for Cloudflare as a registrar.

### DIFF
--- a/providers/cloudflare/cloudflareProvider.go
+++ b/providers/cloudflare/cloudflareProvider.go
@@ -73,7 +73,7 @@ type cloudflareProvider struct {
 	ipConversions   []transform.IPConversion
 	ignoredLabels   []string
 	manageRedirects bool
-	apiProvider     *cloudflare.API
+	cfClient        *cloudflare.API
 }
 
 func labelMatches(label string, matches []string) bool {
@@ -224,7 +224,7 @@ func (c *cloudflareProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 		if ex.Type == "PAGE_RULE" {
 			corrections = append(corrections, &models.Correction{
 				Msg: d.String(),
-				F:   func() error { return c.deletePageRule(ex.Original.(*pageRule).ID, id) },
+				F:   func() error { return c.deletePageRule(ex.Original.(cloudflare.PageRule).ID, id) },
 			})
 		} else {
 			corr := c.deleteRec(ex.Original.(cloudflare.DNSRecord), id)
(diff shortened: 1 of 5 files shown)


tresni:
Turns out that activating the proxy is not covered by the test suite.  When I actually try to run this with a domain using proxied records I get:

```
#2: ACTIVATE PROXY for new record @ CNAME 1 $TARGET.
FAILURE! cannot modify record if domain or record id are empty
```

Looking into this now

tresni:
@tlimoncelli I don't know why this isn't doing integration tests for Cloudflare as I have the necessary secrets on my side.  https://github.com/tresni/dnscontrol/pull/6 should show the results here soon as I couldn't think of a better way to prove that it worked.

tlimoncelli:
Holy cow!  This is quite a PR!

“One of my most productive days was throwing away 1,000 lines of code.”
— Ken Thompson, co-inventor of Unix and C

Ever since the official cloudflare-go module was released I'd been thinking about incorporating it in dnscontrol.  Thanks for working through all the changes.  This is going to make cloudflare a lot more reliable.

creds: As far as why aren't the integration tests working: Github takes the secrets from the owner's repo.  I see that's what it is doing in tresni#6.  I also ran the tests from my system and everything is fine.

go-staticcheck: I wouldn't worry about it.  Maybe move the struct's definition to be above its first use?

Tom

CC @captncraig because his excellent, clean, code made this rewrite so easy.  Look at how often our home-grown structs and funcs were basically the same as what the official module had.

providers/cloudflare/cloudflareProvider.go
```diff
@@ -458,6 +459,17 @@ func newCloudflare(m map[string]string, metadata json.RawMessage) (providers.DNS
...
 
+	var err error
+	if api.APIToken != "" {
+		api.cfClient, err = cloudflare.NewWithAPIToken(api.APIToken)
+	} else {
+		api.cfClient, err = cloudflare.New(api.APIKey, api.APIUser)
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("cloudflare did not like whatever you did %s", err)
```
tlimoncelli, line 470: How about:

```
return nil, fmt.Errorf("cloudflare credentials: %w", err)
```

providers/cloudflare/rest.go
```diff
@@ -1,131 +1,63 @@
...
-		for _, zone := range zr.Result {
-			c.domainIndex[zone.Name] = zone.ID
-			c.nameservers[zone.Name] = append(c.nameservers[zone.Name], zone.Nameservers...)
-		}
-		ri := zr.ResultInfo
-		if len(zr.Result) == 0 || ri.Page*ri.PerPage >= ri.TotalCount {
-			break
-		}
-		page++
+	zones, err := c.cfClient.ListZones(context.Background())
```
tlimoncelli, line 17: Does the library handle pagination?

providers/cloudflare/cloudflareProvider.go
tresni: OMG, I thought I changed that. That’s embarrassing…

providers/cloudflare/rest.go
tresni: It does! One less thing to worry about!

tresni:
@tlimoncelli let me know if there’s anything else you’d like to see.

PR #1269, opened 2021-10-01
Get worker-routes working with older domains

tlimoncelli:


fdcastel:
Tom. NOW I believe I'm starting to understand your problem. 

You need to keep the tests working even without an `AccountId` provided? (and, in this case, simply not run the Workers tests?)

If yes... Why?

Sorry. I'm really didn't got it in full. I'm surely missing something. Why don't you simply set the Cloudflare `AccountID` on your tests?

fdcastel:
BTW: Why do we need that `AccountName`? It is not being used anywhere in `cloudflare` provider.

tlimoncelli:
> Tom. NOW I believe I'm starting to understand your problem.
> 
> You need to keep the tests working even without an `AccountId` provided? (and, in this case, simply not run the Workers tests?)
> 
> If yes... Why?

Excellent question!

It is all about backwards compatibility.   If people's old configs need to be updated, we'd have to increment the major version number.

> 
> Sorry. I'm really didn't get it entirely. I'm surely missing something. Why don't you simply set the Cloudflare `AccountID` on your tests?

I confess I actually don't understand the purpose of AccountID so that was just easier.  Are you suggesting that enable_workers isn't needed because we can just use `AccountID != ""` instead?

Tom

tlimoncelli:
> BTW: Why do we need that `AccountName`? It is not being used anywhere in `cloudflare` provider.
> 
> It seems that `AccountName` is only really used in `transipProvider.go`.

I couldn't figure that out either.  If we can delete it, let's delete it.

fdcastel:
Thanks for the clarification. 😄 

I surely can understand the need of keeping backward compatibility for our USERS. And that's exactly why we added `manageWorkers` in cloudflare config. Old users which will not use this feature would not be affected without setting it to `true`. In the same way that was done with `manageRedirects` in the past.

Just to be perfectly clear: **The users WILL NOT need to change anything after updating to a version with this PR**.

But here we are talking about the DEVELOPERS infrastructure. As I understand, you are having problems to run THE TESTS in YOUR CLOUDFLARE ACCOUNT. But that's because, as I believe, you are not setting the `AccountID` of your cloudflare account in your tests environment.

The usage of `AccountID` is needed because Workers are set in the "Account" level and not in "Zone" level (in which case we need a ZoneID):

fdcastel:
> Are you suggesting that enable_workers isn't needed because we can just use AccountID != "" instead?

Actually I don't understand WHY you are trying to put another flag, when we already have `manageWorkers` for that 😅.

But answering your question: Yes. The test for `AccountID != ""` should be enough.

PR #1277, opened 2021-10-02
Initial support for easyname as registrar

tresni:
Started working on a Cloudflare registrar, but official API docs indicate the necessary information isn't available via their API.  So I did a registrar provider for [easyname](https://easyname.com) for the time being.

No official golang API library for them, so wrote this all from scratch based on the [easyname API documentation](https://api-docs.easyname.com/) and the [easyname php-sdk](https://github.com/easyname/php-sdk/).  Tested against my personal account


tlimoncelli (commented):
Looks good so far. I'll have more time to review this later this weekend.

My one global suggestion is to move the protocol stuff to a separate file.

tresni:
@tlimoncelli I thought about it and then didn't.  I've gone ahead and separated the provider logic from the api logic.

providers/easyname/provider.go
```diff
@@ -0,0 +1,70 @@
...
+
+func init() {
+	providers.RegisterRegistrarType("EASYNAME", newEasyname)
+}
+
+func newEasyname(m map[string]string) (providers.Registrar, error) {
+	api := &easynameProvider{}
+
+	if m["email"] == "" || m["userid"] == "" || m["apikey"] == "" || m["authsalt"] == "" || m["signsalt"] == "" {
+		return nil, fmt.Errorf("missing easyname email, userid, apikey, and/or salt")
```
tlimoncelli, line 27: error message should list authsalt, not salt.

providers/easyname/provider.go
```diff
@@ -0,0 +1,70 @@
...
+			nss = append(nss, ns)
+		}
+	}
+	foundNameservers := strings.Join(nss, ",")
+
+	expected := []string{}
+	for _, ns := range dc.Nameservers {
+		expected = append(expected, ns.Name)
+	}
+	sort.Strings(expected)
```
tlimoncelli, line 56: Does found need to be sorted too? (add a comment if not)

providers/easyname/api.go
```diff
@@ -0,0 +1,140 @@
...
+
+	er := easynameResponseData{}
+	if err = c.request("POST", fmt.Sprintf("https://api.easyname.com/domain/%d/nameserverchange", domain), bytes.NewBuffer(body), &er); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *easynameProvider) getDomain(domain string) (easynameDomain, error) {
```
tlimoncelli, line 116: getDomain should be getDomains or getDomainList

providers/easyname/provider.go
```diff
@@ -0,0 +1,70 @@
...
+
+func newEasyname(m map[string]string) (providers.Registrar, error) {
+	api := &easynameProvider{}
+
+	if m["email"] == "" || m["userid"] == "" || m["apikey"] == "" || m["authsalt"] == "" || m["signsalt"] == "" {
+		return nil, fmt.Errorf("missing easyname email, userid, apikey, and/or salt")
+	}
+
+	api.apikey, api.signSalt = m["apikey"], m["signsalt"]
+	composed := fmt.Sprintf(m["authsalt"], m["userid"], m["email"])
```
tlimoncelli, line 31: Missing format string.

providers/easyname/provider.go
tresni: Good catch, that was from before I added `signsalt` so `authsalt` was just `salt`.  Needs to include both of them.

providers/easyname/api.go
tresni: This returns a single domain's nameservers so getDomains would not be correct (and that's the function fetchDomainList below.)  Might be better to name it getNameservers (to match updateNameservers above) or getDomainNameservers?

providers/easyname/provider.go
tresni: No, `authsalt` is the format string ;-)  It's a really weird system.

providers/easyname/api.go
tresni: Error message needs updated though

tresni:
@tlimoncelli I've addressed all your comments and merged in latest master.  Please let me know if you want to see any other changes.

tlimoncelli:
Two easy things to fix:  Update README.md and OWNERS.  See https://stackexchange.github.io/dnscontrol/writing-providers for details.

tresni:
@tlimoncelli I didn't like the idea of calling it a DNS Provider, so I moved it to a new section called "Domain Registrars" and populated that with all providers that act as domain registrar providers.

tlimoncelli:
> @tlimoncelli I didn't like the idea of calling it a DNS Provider, so I moved it to a new section called "Domain Registrars" and populated that with all providers that act as domain registrar providers.

Excellent idea.  We've actually had confusion about that in the past.

I'll merge it shortly.

PR #1279, opened 2021-10-03
Update deps -- more

tlimoncelli:


tresni:
@tlimoncelli might want to look at [workflow concurrency](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) to prevent the errors from happening.  Will slow down tests but prevent runs from interacting with eachother (basically you need a global concurrency lock unless you move each provider to it's own job, then you could do a per-provider lock.)

PR #1280, opened 2021-10-03
CLOUDFLARE: Stop requiring accountname

### LINKED ISSUE #1275: CLOUDFLAREAPI: Remove AccountName
I'm in favor of removing `AccountName` from Cloudflare configuration.

To the best of my understanding, this is not being used anywhere in this provider. Except in (unnecessary) validation checks.

@tresni: In light of your (excellent) recent changes, could you give us your opinion about this matter? Do you see any need to keep it?

CC @tlimoncelli

tresni:
There is no API call I've found that requires it, only the accountID.  Also, we now set the cfClient.AccountID similar to b55278140f63e0f95a3418bcd61b810c731ae08f (h/t @fdcastel) and no longer store duplicate information in the cfClient and api objects.

Fixes #1275

### DIFF
--- a/docs/_providers/cloudflare.md
+++ b/docs/_providers/cloudflare.md
@@ -43,7 +43,6 @@ If your Cloudflare account has access to multiple Cloudflare accounts, you can s
   "cloudflare": {
     "apitoken": "...",
     "accountid": "your-cloudflare-account-id",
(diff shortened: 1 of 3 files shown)


tlimoncelli:
This is so much cleaner!  Thanks!

PR #1281, opened 2021-10-03
Implement job concurrency on build

tresni:
This should prevent builds from interacting with each other thereby keeping builds from failing when 2 provider tests happen at the same time.

@tlimoncelli I think this should do the trick, but having some problems getting it to work on my fork.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,7 @@ jobs:
   integration-tests:
     runs-on: ubuntu-latest
     needs: build
+    concurrency: ${{ matrix.provider }}
     strategy:
       fail-fast: false
       matrix:


tlimoncelli:
Ah, I was wondering how to do that!

tresni:
Good news is it is working as intended :) I verified by watching some builds collide and actually defering certain provider checks until the previous build completed them.  Azure and Hexonet look like 2 that tend to take a bit.

tresni:
@tlimoncelli Looks like there's a slight issue with this.  I mis-interpreted part of what the documentation was saying

> When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending. Any previously pending job or workflow in the concurrency group will be canceled. To also cancel any currently running job or workflow in the same concurrency group, specify cancel-in-progress: true.

If we have 3 builds happen roughly at the same time, the first build will grab the lock, the second job will sit as pending, when the third job tries to lock, it will cancel the second job and then sit as pending.  They don't stack so you can never have more then 1 job pending for each provider... That's not quite what I think you are looking for...  It will stop the builds from stomping on eachother, but it will lead to a bunch of cancelled builds in some cases that have to be requeued

tlimoncelli:
It seems to be working. I haven't seen anything being cancelled.

PR #1296, opened 2021-10-31
GANDI_V5 & DESEC: correctly handle multiple RR in a RRset

vincentbernat:
When having multiple RR in a RRset, only a few of them may be applied.
In my case, when I have two A records, only one of them makes its way
to Gandi.

In `convert.go`, we had:

```go
var zrs []livedns.DomainRecord
// [...]
zrs = append(zrs, zr)
keys[key] = &zrs[len(zrs)-1]
```

If the slice needs to be extended when appending, the reference we got
in `keys[key]` may be outdated because the new slice contains a copy
of the old one. We either need to store references to domain records
in the slice or we need to stop keeping reference of items in the
slice.

I have fixed this with the second solution as I think the order of the
RRsets is not important.

### DIFF
--- a/providers/desec/convert.go
+++ b/providers/desec/convert.go
@@ -57,10 +57,7 @@ func recordsToNative(rcs []*models.RecordConfig, origin string) []resourceRecord
 				Subname: label,
 				Records: []string{r.GetTargetCombined()},
 			}
-			zrs = append(zrs, zr)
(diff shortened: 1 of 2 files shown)


tlimoncelli:
CC @D3luxee

tlimoncelli:
Thank you for finding and fixing this bug!  As you can see in the code, the current implementation was iffy.

PR #1302, opened 2021-11-02
Add test cases for IDN domains

juliusrickert:
Added test cases for every permutation of IDN domains I could think of.

Tests pass with current master (866aa798e2a3acbb3b595afdbbe20ee74b0604c5) but fail with `v3.12.0`.

P.S. Recreated IR JSON file, had to set the `CNAME` tests (760ee8dc46f74f48e216bcaeb82f87f133ca4e16 `029-extendsub.js` lines 138 and 139) to be relative manually to get the tests to pass.

### DIFF
--- a/pkg/js/parse_tests/029-dextendsub/düsseldorf.example.net.zone
+++ b/pkg/js/parse_tests/029-dextendsub/düsseldorf.example.net.zone
@@ -1,0 +1,9 @@
+$TTL 300
+@                IN A     10.0.0.7
+düsseltal       IN A     10.0.0.11
+www.düsseltal   IN A     10.0.0.12
+subdomain        IN A     10.0.0.9
+www.subdomain    IN A     10.0.0.10
+www              IN A     10.0.0.8
+ü               IN A     10.0.0.13
+www.ü           IN A     10.0.0.14
--- a/pkg/js/parse_tests/029-dextendsub.js
+++ b/pkg/js/parse_tests/029-dextendsub.js
@@ -3,6 +3,7 @@ var CF = NewDnsProvider("Cloudflare", "CLOUDFLAREAPI");
 
 // Zone that gets extended by subdomain
 D("foo.net", REG, DnsProvider(CF),
+  DefaultTTL(300),
   A("@", "10.1.1.1"),
   A("www", "10.2.2.2")
 );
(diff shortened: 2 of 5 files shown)


pkg/js/parse_tests/029-dextendsub/düsseldorf.example.net.zone.ACTUAL
```diff
@@ -0,0 +1,9 @@
+$TTL 300
```
tlimoncelli, line 1: The ".ACTUAL" files shouldn't be checked into git. They're artifacts left around to help people debug.

pkg/js/parse_tests/029-dextendsub/düsseldorf.example.net.zone
```diff
@@ -0,0 +1,9 @@
+$TTL 300
+@                IN A     10.0.0.7
+düsseltal       IN A     10.0.0.11
+www.düsseltal   IN A     10.0.0.12
+subdomain        IN A     10.0.0.9
+www.subdomain    IN A     10.0.0.10
+www              IN A     10.0.0.8
+ü               IN A     10.0.0.13
```
tlimoncelli, line 8: Should a Unicode character like that appear in the zone file? I would think it should be punycode instead (but I confess I'm no expert in this area)

pkg/js/parse_tests/029-dextendsub/düsseldorf.example.net.zone
```diff
@@ -0,0 +1,9 @@
+$TTL 300
```
tlimoncelli, line 1: Can this filename be the punycode version instead?

pkg/js/parse_tests/029-dextendsub/düsseldorf.example.net.zone.ACTUAL
juliusrickert: Oh, yeah, sorry about that. Should we add them to `.gitignore` maybe?

pkg/js/parse_tests/029-dextendsub/düsseldorf.example.net.zone
juliusrickert: Actually, I really don't know. Luckily, I haven't dealt with zone files in years :D

Would the test code load the correct zone file then? I'd expect a change to it to be necessary then, but it's certainly possible.

tlimoncelli:
I'm going to merge this. I admit I'm concerned that unicode in the filenames will break something (Windows maybe?) but let's risk it.  We might have to figure out a workaround or revert this but that's life.

tlimoncelli:
Thanks for creating these tests!

juliusrickert:
> I'm going to merge this. I admit I'm concerned that unicode in the filenames will break something (Windows maybe?) but let's risk it. We might have to figure out a workaround or revert this but that's life.

I'd guess that that isn't a problem anymore. I have used umlauts in filenames on Windows, macOS, and Linux without problems.

> Thanks for creating these tests!

De rien :)

Issue #1263, opened 2021-09-25
CloudNS API "The NS record needs to be pointed to a hostname that has an A or an AAAA record" Error


df-a:
I keep getting this error whenever I try to update cloudns nameservers

> ----- DNS Provider: cloudns...
> 2 corrections
> #1: MODIFY NS domain.tld: (ns1.domain.tld. ttl=3600) -> (c0.domain.tld. ttl=3600), ClouDNS ID: 242876057:
> FAILURE! failed update (ClouDNS): ClouDNS API error: The NS record needs to be pointed to a hostname that has an A or an AAAA record. URL:api.cloudns.net/dns/mod-record.json?auth-id=3965&auth-password=pass-phrase&domain-name=domain.tld&host=&record=c0.domain.tld.&record-id=242876057&record-type=NS&sub-auth-id=&ttl=3600
> #2: MODIFY NS domain.tld: (ns2.domain.tld. ttl=3600) -> (c1.domain.tld. ttl=3600), ClouDNS ID: 242876060:
> FAILURE! failed update (ClouDNS): ClouDNS API error: The NS record needs to be pointed to a hostname that has an A or an AAAA record. URL:api.cloudns.net/dns/mod-record.json?auth-id=3965&auth-password=pass-phrase&domain-name=domain.tld&host=&record=c1.domain.tld.&record-id=242876060&record-type=NS&sub-auth-id=&ttl=3600

As part of the push, I also update and add whichever A and AAAA records to the appropriate IPs and URLs as required. But DNSControl keeps telling me that "The NS record needs to be pointed to a hostname that has an A or an AAAA record" despite these also being pushed in the same operation (as needed). This workflow functioned seamlessly before, but lately I keep getting the aforementioned error, and I can only correct it by manually inputting the NS into cloudns, defeating the purpose.

What is going on? Could this somehow be an error on my end? TIA.

tlimoncelli:
CC @pragmaton (owner of the CLOUDNS provider)

P.S. I haven't used that provider but it sounds like there may be an ordering issue.  Often the A/AAAA records need to exist before they can be mentioned in an NS.  Can you try doing this in two steps? (1) add the A/AAA records, [wait for (1) to propagate],  (2) change the NS record.

riku22:
I have confirmed that the same problem occurs here.
It seems that ClouDNS does not require a trailing period only for editing NS records.
Therefore, you need to modify the part that changes the record.
[...]

PR #1303, opened 2021-11-03
CLOUDNS: Fix name server updates #1263

riku22:
Fix issue [#1263].
ClouDNS does not require the trailing period to be specified when updating an NS record where the A or AAAA record exists in the zone.
So, modify it to remove the trailing period.
I used the `strings` module to implement this.


tlimoncelli:
CC: @pragmaton

tlimoncelli:
Friendly ping?

tlimoncelli:
Friendly ping?

Issue #1139, opened 2021-04-29
Long TXT not working with NS1


kurthockenbury:
Similar to #996 and #1003, I would like to see long TXT support for NS1. Testing the API with curl, and later confirmed when talking with NS1 over Zoom today, NS1 falls into the group of providers that accept TXT records >255 characters and autosplit them on their end.

tlimoncelli:
Sadly the maintainer of the NS1 provider is no longer with the project.

I'd be glad to walk you through fixing this.  The code for TXT record handling was recently rewritten and is much cleaner than before.

kurthockenbury:
I may be stepping in over my head, but I can try. I have never coded in golang before. I have followed the instructions, used "go get" and was able to build 3.8.1-dev, so I suppose I am ready to take my first bite of the elephant.

kurthockenbury:
I was going to take a crack at this, but work has now decided on Cloudflare DNS (since we were paying for Cloudflare anyway), so this has dropped off my priority list.

norman-zon:
@tlimoncelli, I was looking into making the NS1 provider work with long TXT records, but found the `TxtNoMultipleStrings` function not to be suitable, as it [also checks for string length](https://github.com/StackExchange/dnscontrol/blob/3405757271b66c7e679b78af74810bedd3e7b50b/pkg/recordaudit/txt.go#L100).
NS1 does not support multiple strings, but supports any length of string, as long as it is only a single string.

How would you want to move forward from here? Implement a `TxtNoMultipleStrings` directly in the `auditrecords.go` file of the provider, omitting the length check? Or would you prefer to make changes to the general implementation in `pkg/recordaudit/txt.go`, so other provider with the same capabilities could use it too?

tlimoncelli:
Hi there!  Thanks for looking into this!

I would love to see this added to `pkg/recordaudit/txt.go`.  I think a few providers will benefit. (In fact, I think some have it in their own `auditrecords.go` file).  `providers/cloudns/auditrecords.go`

In hindsight, each function in txt.go should test exactly one thing. I think `TxtNoMultipleStrings` is the one that tests 2 attributes.

tlimoncelli:
Fixed and shipped!

PR #1308, opened 2021-11-23
CLOUDNS: More accurate TXT record checking (remove length check)

norman-zon:
Fix issue [#1139].
All functions in should test for only one condition. There already is a
function that tests for long TXT records: `TxtNoLongStrings`.
Add calls to `TxtNoLongStrings` in all providers that use
`TxtNoMultipleStrings`, to keep functionality, except for NS1 and ClouDNS,
which allow for any TXT record length, but not for multiple strings per
TXT.


tlimoncelli:
Thanks for fixing this!

norman-zon:
You're very welcome. Thank you for maintaining this project!
Can you estimate when you will tag the next release? It would be quite critical for us to have this fix available in a release.

tlimoncelli:
Would next week be ok?

norman-zon:
Absolutely. Thanks again for your effort.

PR #1373, opened 2022-01-19
NS1: Enable get-zones

norman-zon:
Refactor code to get zones from NS1 API into GetZonesRecords function to
enable get-zones

### DIFF
--- a/docs/_includes/matrix.html
+++ b/docs/_includes/matrix.html
@@ -1409,47 +1409,6 @@
 		<td><i class="fa fa-minus dim"></i></td>
 		</tr>
 	<tr>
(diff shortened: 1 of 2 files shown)


tlimoncelli:
Thanks for fixing that!

tlimoncelli:
By the way... we're currently without a maintainer for the NS1 provider. I don't have an account there, so I can't even validate other changes haven't broken NS1.  Would you be interested in being the maintainer for this provider?

norman-zon:
I can fix things in the provider that I stumble upon while using it, but I don't think I feel fit to completely maintain it. This is my first contact with Go, so bug-hunting feels OK, but writing new features would be too much at the moment.
I asked NS1 if they would step up and maintain it themselves, but they declined.

PR #1377, opened 2022-01-24
CAA records may include quoted spaces (#1374)

### LINKED ISSUE #1374: Fails on Cloudflare CAA record
I can push a CAA record to Cloudflare, but then when I preview again I get an error:

```
----- DNS Provider: cloudflare...
ERROR
Error getting corrections: unparsable record received from cloudflare: CAA value does not contain 3 fields: ("0 issue \"letsencrypt.org; validationmethods=dns-01; accounturi=https://acme-v02.api.letsencrypt.org/acme/acct/1234\"")
```

Record in `dnsconfig.js`:
`CAA('@', 'issue', 'letsencrypt.org; validationmethods=dns-01; accounturi=https://acme-v02.api.letsencrypt.org/acme/acct/1234'),`

tlimoncelli:
Fixes #1374

Providers such as Cloudflare return quoted data in the API protocol.  The code had assumed simple space-separated data.

For example, a CAA record might return this data:

    0 issue "letsencrypt.org; validationmethods=dns-01; accounturi=https://acme-v02.api.letsencrypt.org/acme/acct/1234"

Doing a simple strings.Fields() was not sufficient for parsing this.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -1085,6 +1085,11 @@ func makeTests(t *testing.T) []*TestGroup {
 			// Test support of ";" as a value
 			tc("CAA many records", caa("@", "issuewild", 0, ";")),
 		),
+		testgroup("Issue 1374",
+			requires(providers.CanUseCAA), not("DIGITALOCEAN"),
+			// Test support of ";" as a value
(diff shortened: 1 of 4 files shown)


tlimoncelli:
Wow!  I can't believe the fix was so easy.  Of course, I owe it all to [https://stackoverflow.com/a/47489846/71978](https://stackoverflow.com/a/47489846/71978).  (and, yes, I really should stop being surprised when the website I work for has the answers I need)

PR #1396, opened 2022-02-04
Documentation: Customizing nameservers for hosting.de provider

juliusrickert:
The hosting.de provider can be used with http.net's offerings too, but http.net uses different nameservers.

This PR adds documentation on how to change the nameservers used and on how to simplify using them, including an example.

Note: Furthermore, it switches to  `fenced_code_blocks` in an effort to improve the legibility when using Markdown preview. If this poses problems, feel free to remove the commit.

### DIFF
--- a/docs/_providers/hostingde.md
+++ b/docs/_providers/hostingde.md
@@ -7,6 +7,7 @@ jsId: hostingde
 # hosting.de Provider
 
 ## Configuration
+
 In your credentials file, you must provide your [`authToken` and optionally an `ownerAccountId`](https://www.hosting.de/api/#requests-and-authentication).
 
 **If you want to use this provider with http.net or a demo system you need to provide a custom `baseURL`.**
@@ -28,6 +29,7 @@ In your credentials file, you must provide your [`authToken` and optionally an `
 {% endhighlight %}
 
 ## Usage
+
 Example JavaScript:
 
 {% highlight js %}
--- a/docs/_providers/hostingde.md
+++ b/docs/_providers/hostingde.md
@@ -4,33 +4,26 @@ title: hosting.de Provider
 layout: default
 jsId: hostingde
 ---
+
 # hosting.de Provider
 
 ## Configuration
 
 In your credentials file, you must provide your [`authToken` and optionally an `ownerAccountId`](https://www.hosting.de/api/#requests-and-authentication).
 
-**If you want to use this provider with http.net or a demo system you need to provide a custom `baseURL`.**
-
-* hosting.de (default): `https://secure.hosting.de`
-* http.net: `https://partner.http.net`
-* Demo: `https://demo.routing.net`
+### Example `creds.json`
 
 ```json
 {
   "hosting.de": {
     "authToken": "YOUR_API_KEY"
-  },
-  "http.net": {
-    "authToken": "YOUR_API_KEY",
-    "baseURL": "https://partner.http.net"
   }
 }
 ```
 
 ## Usage
 
-Example JavaScript:
+### Example `dnsconfig.js`
 
 ```js
 var REG_HOSTINGDE = NewRegistrar('hosting.de', 'HOSTINGDE')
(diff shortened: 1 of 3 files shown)


tlimoncelli:
Doc changes: Looks good to me.

Changing to `fenced_code_blocks`:  I'm open to this. I want to see how it looks first.  Sadly the docker build for the documentation isn't working for me right now and I don't see an obvious fix.

```
$ docker run --rm -it --volume="$PWD:/srv/jekyll" --volume="$PWD/vendor/bundle:/usr/local/bundle" --env JEKYLL_ENV=production jekyll/jekyll:3.8 jekyll build -V
...
...
...
  Rendering Markup: docs/getting-started.md
  Rendering Layout: docs/getting-started.md
     Build Warning: Layout 'default' requested in docs/getting-started.md does not exist.
         Rendering: docs/index.md
  Pre-Render Hooks: docs/index.md
  Rendering Liquid: docs/index.md
  Liquid Exception: Could not locate the included file 'feature.html' in any of ["/srv/jekyll/_includes"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source. in docs/index.md
jekyll 3.8.6 | Error:  Could not locate the included file 'feature.html' in any of ["/srv/jekyll/_includes"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source.
```

Are you seeing the same issue?  It might be just me.

juliusrickert:
You have to mount the `$PWD/docs` directory to `/src/jekyll`, not `$PWD`, or switch into the `docs` folder before running the command :)

tlimoncelli:
Ah!  Thanks for figuring that out.  I'll update the docs!  https://github.com/StackExchange/dnscontrol/pull/1397/files

juliusrickert:
Just figured that using the `providermeta` for this could be a more elegant way to achieve this (See [bind provider](https://stackexchange.github.io/dnscontrol/providers/bind)).

What do you think?

tlimoncelli:
Yes, `default_ns` is the way to go.  It isn't supported by that provider but I'd be glad to accept a PR.  (if you want to merge this in the meanwhile, let me know when its ready)

tlimoncelli:
P.S.  Using `fenced_code_blocks` is fine.  I just did a simple A-B test and they look the same to me.

juliusrickert:
> Yes, default_ns is the way to go.

Alright, I'll add the functionality on Monday and wouldn't merge this in the meantime.

> P.S. Using fenced_code_blocks is fine.

Nice. Maybe I'll find time to change this globally in a scripted manner.

juliusrickert:
@tlimoncelli Please take a look :)

Ready from my point of view.

docs/_providers/hostingde.md
```diff
@@ -38,3 +40,21 @@ D('example.tld', REG_HOSTINGDE, DnsProvider(DNS_HOSTINGDE),
     A('test', '1.2.3.4')
 );
 {% endhighlight %}
+
+## Customize nameservers
```
tlimoncelli, line 44:
```suggestion
## Using this provider with HTTP.NET and others

HTTP.NET and other DNS service providers use an API that is compatible with hosting.de.  Using them requires overriding the default nameservers.

### Example HTTP.NET configuration:

``js
var DNS_HTTPNET = NewDnsProvider('http.net', 'HOSTINGDE', {
  default_ns: [
    'ns1.routing.net.',
    'ns2.routing.net.',
    'ns3.routing.net.',
  ],
});
``

Why this works:  hosting.de has the concept of *nameserver sets* but this provider does not implement it.  The `HOSTINGDE` provider **ignores the default nameserver set** defined in your account!  Instead, it uses hosting.de's nameservers (`ns1.hosting.de.`, `ns2.hosting.de.`, and `ns3.hosting.de.`) by default, regardless of your account settings.  The above examples use the `default_ns` meta data to tell the provider what to use instead.
```

As we find other compatible services it will be easy to add them as examples.

NOTE: The 3x backticks are replaced by 2x backticks in the above suggestion due to quoting rules.

tlimoncelli (commented):
Would you also add `HTTP.NET` to the `README.md` and `docs/provider-list.md` files?

docs/_providers/hostingde.md
juliusrickert: > Would you also add HTTP.NET to the README.md and docs/provider-list.md files?

Sounds good! I'll happily do so.

> NOTE: The 3x backticks are replaced by 2x backticks in the above suggestion due to quoting rules.

Ah, so a GitHub limitation? Got it :D

tlimoncelli:
What do you think about the rewrite that I proposed?  (https://github.com/StackExchange/dnscontrol/pull/1396/files#r807347995)

tlimoncelli (approved):

tlimoncelli:
Thanks for your flexibility on how to structure the doc.

PR #1404, opened 2022-02-08
ns1: fix multivalue CAA handling

costasd:
Introducing better multivalue support for the CAA entry broke CAA support for ns1, failing the relevant test.

Failing case:
```
=== RUN   TestDNSProviders/example.com/25:Issue_1374:CAA_spaces
```

This is was happening because provider's CAA handling was splitting fields on spaces, which trashed answers containing  spaces in the 3rd field (ie. splitting the 3rd field into multiple fields)

While dnscontrol has support to handle such cases of fields containing spaces via `ParseQuotedFields`, ns1's client isn't enclosing such fields in quotes and `ParseQuotedFields()` can't work its magic.

So to fix this issue, rework ns1's CAA handling a bit to not use `string.Split(str, " ")` anymore.

While there, clear up panics in the provider and surface errors to be picked and handled as needed.

### DIFF
--- a/providers/ns1/ns1Provider.go
+++ b/providers/ns1/ns1Provider.go
@@ -170,7 +170,12 @@ func buildRecord(recs models.Records, domain string, id string) *dns.Record {
 		} else if r.Type == "TXT" {
 			rec.AddAnswer(&dns.Answer{Rdata: r.TxtStrings})
 		} else if r.Type == "CAA" {
-			rec.AddAnswer(&dns.Answer{Rdata: strings.Split(fmt.Sprintf("%v %s %s", r.CaaFlag, r.CaaTag, r.GetTargetField()), " ")})
+			rec.AddAnswer(&dns.Answer{
+				Rdata: []string{
+					fmt.Sprintf("%v", r.CaaFlag),
+					r.CaaTag,
+					fmt.Sprintf("%s", r.GetTargetField()),
+			}})
 		} else if r.Type == "SRV" {
 			rec.AddAnswer(&dns.Answer{Rdata: strings.Split(fmt.Sprintf("%d %d %d %v", r.SrvPriority, r.SrvWeight, r.SrvPort, r.GetTargetField()), " ")})
 		} else {
@@ -198,6 +203,13 @@ func convert(zr *dns.ZoneRecord, domain string) ([]*models.RecordConfig, error) 
 			rec.Type = rtype
 			if err := rec.SetTarget(ans); err != nil {
 				panic(fmt.Errorf("unparsable %s record received from ns1: %w", rtype, err))
+			}
+		case "CAA":
+			//dnscontrol expects quotes around multivalue CAA entries, API doesn't add them, so let's force them
+			x_ans := strings.SplitN(ans, " ", 3)
+			answer := fmt.Sprintf(`%s %s "%s"`, x_ans[0],x_ans[1],x_ans[2])
+			if err := rec.PopulateFromString(rtype, answer, domain); err != nil {
+				panic(fmt.Errorf("unparsable %s record received from ns1: %w", err))
 			}
 		default:
 			if err := rec.PopulateFromString(rtype, ans, domain); err != nil {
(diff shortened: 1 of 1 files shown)


providers/ns1/ns1Provider.go
```diff
@@ -199,6 +204,13 @@ func convert(zr *dns.ZoneRecord, domain string) ([]*models.RecordConfig, error)
 			if err := rec.SetTarget(ans); err != nil {
 				panic(fmt.Errorf("unparsable %s record received from ns1: %w", rtype, err))
 			}
+		case "CAA":
+			//dnscontrol expects quotes around multivalue CAA entries, API doesn't add them, so let's force them
+			x_ans := strings.SplitN(ans, " ", 3)
+			answer := fmt.Sprintf(`%s %s "%s"`, x_ans[0],x_ans[1],x_ans[2])
+			if err := rec.PopulateFromString(rtype, answer, domain); err != nil {
+				panic(fmt.Errorf("unparsable %s record received from ns1: %w", err))
```
tlimoncelli, line 212: Don't panic.  Return an error.

providers/ns1/ns1Provider.go
```diff
@@ -199,6 +204,13 @@ func convert(zr *dns.ZoneRecord, domain string) ([]*models.RecordConfig, error)
 			if err := rec.SetTarget(ans); err != nil {
 				panic(fmt.Errorf("unparsable %s record received from ns1: %w", rtype, err))
 			}
+		case "CAA":
+			//dnscontrol expects quotes around multivalue CAA entries, API doesn't add them, so let's force them
+			x_ans := strings.SplitN(ans, " ", 3)
+			answer := fmt.Sprintf(`%s %s "%s"`, x_ans[0],x_ans[1],x_ans[2])
+			if err := rec.PopulateFromString(rtype, answer, domain); err != nil {
```
tlimoncelli, line 211: Since you know it is a CAA record, call `rec.SetTargetCAAStrings()` instead.

providers/ns1/ns1Provider.go
```diff
@@ -199,6 +204,13 @@ func convert(zr *dns.ZoneRecord, domain string) ([]*models.RecordConfig, error)
 			if err := rec.SetTarget(ans); err != nil {
 				panic(fmt.Errorf("unparsable %s record received from ns1: %w", rtype, err))
 			}
+		case "CAA":
+			//dnscontrol expects quotes around multivalue CAA entries, API doesn't add them, so let's force them
```
tlimoncelli, line 208: Please add an integration test that demonstrates this bug.  `integrationTest/integration_test.go` around line 2082/

tlimoncelli (commented):
Interesting bug!

providers/ns1/ns1Provider.go
costasd: thanks!

providers/ns1/ns1Provider.go
costasd: This was actually uncovered with the existing test, `Issue_1374:CAA_spaces` (the testsuite was failing), so I think it's covered.

providers/ns1/ns1Provider.go
costasd: You're right, cleared all panics from the provider :)

providers/ns1/ns1Provider.go
```diff
@@ -192,16 +197,22 @@ func convert(zr *dns.ZoneRecord, domain string) ([]*models.RecordConfig, error)
 		case "ALIAS":
 			rec.Type = rtype
 			if err := rec.SetTarget(ans); err != nil {
-				panic(fmt.Errorf("unparsable %s record received from ns1: %w", rtype, err))
+				return found, fmt.Errorf("unparsable %s record received from ns1: %w", rtype, err)
```
tlimoncelli, line 200: When returning an error, the other parameters should be nil or zero.

```suggestion
				return nil, fmt.Errorf("unparsable %s record received from ns1: %w", rtype, err)
```

(Same for the other cases)

PR #1405, opened 2022-02-10
ns1: add support for NAPTR

costasd:
Enable support for NAPTR support in NS1.

Closes: #1407

### DIFF
--- a/providers/ns1/ns1Provider.go
+++ b/providers/ns1/ns1Provider.go
@@ -19,6 +19,8 @@ var docNotes = providers.DocumentationNotes{
 	providers.CanUseAlias:            providers.Can(),
 	providers.CanUseCAA:              providers.Can(),
 	providers.CanUsePTR:              providers.Can(),
+	providers.CanUseNAPTR:            providers.Can(),
+	providers.CanUseTLSA:             providers.Can(),
 	providers.DocCreateDomains:       providers.Can(),
 	providers.DocDualHost:            providers.Can(),
 	providers.CanGetZones:            providers.Can(),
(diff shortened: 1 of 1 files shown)


providers/ns1/ns1Provider.go
```diff
@@ -173,6 +175,20 @@ func buildRecord(recs models.Records, domain string, id string) *dns.Record {
 			rec.AddAnswer(&dns.Answer{Rdata: strings.Split(fmt.Sprintf("%v %s %s", r.CaaFlag, r.CaaTag, r.GetTargetField()), " ")})
 		} else if r.Type == "SRV" {
 			rec.AddAnswer(&dns.Answer{Rdata: strings.Split(fmt.Sprintf("%d %d %d %v", r.SrvPriority, r.SrvWeight, r.SrvPort, r.GetTargetField()), " ")})
+		} else if r.Type == "NAPTR" {
+			rec.AddAnswer(&dns.Answer{Rdata: []string{
+				fmt.Sprintf("%d", r.NaptrOrder),
```
tlimoncelli, line 180: Consider using `strconv.Itoa()` for these.

costasd: Done, thanks!

costasd:
Removed TLSA support, as I opened a new ns1 account for testing and the record isn't supported in that account.

Will readd TLSA in a new PR as soon as I clarify TLSA support in NS1

tlimoncelli:
Two more:

```
$ staticcheck ./... | grep -i ns1
providers/ns1/ns1Provider.go:179:6: the argument is already a string, there's no need to use fmt.Sprintf (S1025)
$ golint ./...
providers/ns1/ns1Provider.go:219:4: don't use underscores in Go names; var x_ans should be xAns
```

costasd:
> Two more:
> 
> ```
> $ staticcheck ./... | grep -i ns1
> providers/ns1/ns1Provider.go:179:6: the argument is already a string, there's no need to use fmt.Sprintf (S1025)
> $ golint ./...
> providers/ns1/ns1Provider.go:219:4: don't use underscores in Go names; var x_ans should be xAns
> ```

Fixed, thanks! I have to configure an ns1 pipeline at some point :)

tlimoncelli:
Thanks for the PR!

PR #1409, opened 2022-02-11
GCLOUD: Stop using deprecated New() call

tlimoncelli:
Remediates this warning:

```
providers/gcloud/gcloudProvider.go:80:15: gdns.New is deprecated: please use NewService instead. To provide a custom HTTP client, use option.WithHTTPClient. If you are using google.golang.org/api/googleapis/transport.APIKey, use option.WithAPIKey with NewService instead.  (SA1019)
```

### DIFF
--- a/models/t_parse.go
+++ b/models/t_parse.go
@@ -12,43 +12,43 @@ import (
 //
 // If this doesn't work for all rtypes, process the special cases then
 // call this for the remainder.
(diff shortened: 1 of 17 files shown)


tlimoncelli:
Fixed in https://github.com/StackExchange/dnscontrol/pull/2482

Issue #1410, opened 2022-02-15
BIND provider assumes trailing dot, blindly removes last character


juliusrickert:
When setting custom nameservers without a trailing dot with the BIND provider, the last character gets removed, regardless of its value.

This causes—in my opinion—unexpected behaviour.
For example, entering `ns1.example.com` leads to `ns1.example.co` being set.

Code causing this behaviour:

https://github.com/StackExchange/dnscontrol/blob/886dbf6d0070ea7efc37e8235bb9bae50a1f0b61/providers/bind/bindProvider.go#L71-L74

tlimoncelli:
Thanks for reporting this!

Would you please you give a (redacted) example from a dnsconfig.js file?

juliusrickert:
Sure!

Example dnsconfig.js:

```js
var DOH = NewRegistrar('doh', 'DNSOVERHTTPS');

var BIND = NewDnsProvider('bind', 'BIND', {
  default_ns: [
    'ns1.example.com',
  ],
});

D('example.org', DOH, DnsProvider(BIND));
```

Example output:

```
******************** Domain: example.org
----- Getting nameservers from: bind
----- DNS Provider: bind...
File does not yet exist: "zones/example.org.zone"
1 correction
#1: GENERATE_ZONEFILE: 'example.org' (new file with 2 records)

----- Registrar: doh...
1 correction
#1: Update nameservers a.iana-servers.net,b.iana-servers.net -> ns1.example.co
Done. 2 corrections.
```

PR #1411, opened 2022-02-15
BIND: Bug: DefaultNS strips last char

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/1410


tlimoncelli:
I don't think this is a breaking change because it only breaks broken configurations.  (check my logic on that)

juliusrickert:
> I don't think this is a breaking change because it only breaks broken configurations. (check my logic on that)

I'd say so too! Looking good.

Thank you for the quick fix :)

Issue #1402, opened 2022-02-07
DNS over HTTPS (DoH) provider not checking delegation


juliusrickert:
The DNS over HTTPS (DoH) provider is solely a _registrar_ provider, intended to check whether the nameservers at the registrar are matching those of the zone.

But instead, it checks the _zone's_ `NS` records, so it does not check whether the _delegation_ is correct.

#### Example (from my list of weird DNS setups :D)

```bash
$ dig @a.gtld-servers.net vallourec.com ns
…
;; AUTHORITY SECTION:
vallourec.com.		172800	IN	NS	dns1.vallourec.fr.
vallourec.com.		172800	IN	NS	dns2.vallourec.fr.
vallourec.com.		172800	IN	NS	dns3.vallourec.fr.
…

$ dig @dns1.vallourec.fr vallourec.com ns
…
;; ANSWER SECTION:
vallourec.com.		28800	IN	NS	dns2.vallourec.fr.
vallourec.com.		28800	IN	NS	dns1.vallourec.fr.
…
```

##### dnsconfig.js

```js
var DOH = NewRegistrar('doh', 'DNSOVERHTTPS');

var BIND = NewDnsProvider('bind', 'BIND');

D('vallourec.com', DOH, DnsProvider(BIND), [
  NAMESERVER('dns1.vallourec.fr.'),
  NAMESERVER('dns2.vallourec.fr.'),
]);
```

##### Result

```bash
$ dnscontrol push
******************** Domain: vallourec.com
----- Getting nameservers from: bind
----- DNS Provider: bind...
0 corrections
----- Registrar: doh...
0 corrections
Done. 0 corrections.
```

##### Expected

```bash
$ dnscontrol push
******************** Domain: vallourec.com
----- Getting nameservers from: bind
----- DNS Provider: bind...
0 corrections
----- Registrar: doh...
1 correction
#1: Update nameservers dns1.vallourec.fr,dns2.vallourec.fr,dns3.vallourec.fr -> dns1.vallourec.fr,dns2.vallourec.fr
FAILURE! DNS-over-HTTPS 'Registrar' is read only, changes must be applied to vallourec.com manually
Done. 1 corrections.
completed with errors
```

tlimoncelli:
CC @mikenz

tlimoncelli:
Here's my take on this:

This is a documentation issue, not a code bug.  The documentation should be clear about that.

[...]

PR #1413, opened 2022-02-17
DOH: Clarify DOH does not reference whois data

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/1402


juliusrickert:
It may not be the case that WHOIS offers this information. It (usually) does when talking about "proper" domains but doesn't for subdomain delegations like `sub.example.com`.

It's rather the `NS` record in the parent zone.

tlimoncelli:
True. I don't literally mean whois. I just couldn't think of how to explain it simply.

PR #1414, opened 2022-02-17
add DS support on PowerDNS

aba-rechsteiner:


tlimoncelli (approved):
Looks like ya got it all right on the first try!  Congrats!

tlimoncelli:
There's no well-defined release schedule but I plan on doing a release after https://github.com/StackExchange/dnscontrol/issues/1420 is fixed.

PR #1421, opened 2022-03-01
ClouDNS:  Add CLOUDNS_WR (Web Redirects)

tpdn:
I defined a ClouDNS-specific rtype `CLOUDNS_WR` similar to `NS1_URLFWD`.
It is a minimal implementation, supports only simple 302 redirects with default settings; 301 and frame redirects are not supported.

## Motivation
I run `dnscontrol preview` with some web redirects already configured, and get the following results.
```
#1: DELETE WR example.tld http://example.org ttl=3600, ClouDNS ID: 1234567
```
In this situation, If I run `dnscontrol push` without setting `NO_PURGE`, all existing ClouDNS WR records will be deleted.
This behavior is inconvenient for me.

## Another solution
Another solution is to use the generic redirects e.g. `URL`, `URL301`, `FRAME`. 

There are discussions below, but it seems to be inconclusive:
- https://github.com/StackExchange/dnscontrol/pull/237
- https://github.com/StackExchange/dnscontrol/issues/455
- https://github.com/StackExchange/dnscontrol/pull/549


tlimoncelli:
Looks good! 

Your description mentioned NS1_URLFWD but I don't see any related code changes.  When you wrote `as well as NS1_URLFWD` did you mean "similar to NS1_URLFWD"?

PR #1435, opened 2022-03-03
Fix examples in documentation

### LINKED ISSUE #1420: Documentation Examples not being formatted correctly
Function example are showing raw text instead of parsing the Markdown tags

````
```js
````

image

### PR DESCRIPTION
**Pull request by juliusrickert**
Fixes #1420.

Replaces Bootstrap's JS-based "details" with native HTML `<details>`.

### CODE REVIEW

**Comment by tlimoncelli** [comment]
When I view this using Docker I don't get working examples. Is it just me?

**Comment by tlimoncelli** [comment]
When I run this through:

```
cd docs
docker run --rm -it --volume="$PWD:/srv/jekyll" --volume="$PWD/vendor/bundle:/usr/local/bundle" --env JEKYLL_ENV=production jekyll/jekyll:3.8 jekyll build -V
```

I see just the word "Example":

screencapture_2022-03-03_16 56 28

juliusrickert:
> I see just the word "Example"

Which browser are you using?

By default, the browser would display it like in [this example from MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#a_simple_disclosure_example).
Due to normalize.css being used, the browser's defaults are removed to even out differences between them.

I tried re-enabling this behaviour but apparently that didn't go as planned.

juliusrickert:
MDN lists `list-style-type` of `disclosure-open` and `disclosure-closed` to have a bad [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type#browser_compatibility). But it lists Chrome to not be supported even thought I tested in Chrome.

I'm thinking about creating the marker using [this trick](https://css-tricks.com/snippets/css/css-triangle/).

juliusrickert:
Removed the `<details>` for now to get this fixed quickly.

Consolidated the two includes into one.

tlimoncelli:
Chrome displays it correctly now.  Thanks!

tlimoncelli (approved):

PR #1450, opened 2022-03-08
ns1: enable autoDNSSEC capability

### LINKED ISSUE #1444: NS1: Enable AUTODNSSEC capability
NS1 provides a way to enable/disable managed DNSSEC through their [DNSSEC API](https://ns1.com/api/#post-enable-dnssec-for-a-zone), with the only caveat being that DNSSEC capability has to be enabled across the account first via support.

The API part of the above process seems to map directly to the AUTODNSSEC capability.

Investigate whether that's actually the case for the ns1 provider and if yes, whether we can handle DNSSEC being disabled on the account level.

costasd:
NS1's API has support for enabling or disabling DNSSEC on zones under
DNSSEC-enabled accounts.

This commit enables support for detecting a zone's DNSSEC state and,
given the autoDNSSEC setting, pushing the right domain corrections
to our corrections stack.

Fixes #1444

### DIFF
--- a/providers/ns1/ns1Provider.go
+++ b/providers/ns1/ns1Provider.go
@@ -19,6 +19,7 @@ import (
 var docNotes = providers.DocumentationNotes{
 	providers.CanGetZones:            providers.Can(),
 	providers.CanUseAlias:            providers.Can(),
+	providers.CanAutoDNSSEC:	  providers.Can(),
 	providers.CanUseCAA:              providers.Can(),
 	providers.CanUseDS:		  providers.Can(),
 	providers.CanUseDSForChildren:	  providers.Can(),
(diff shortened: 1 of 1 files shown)


providers/ns1/ns1Provider.go
```diff
@@ -88,6 +89,52 @@ func (n *nsone) GetZoneRecords(domain string) (models.Records, error) {
 	}
 	return found, nil
 }
+// Get DNSSEC status for zone. Returns true for enabled, false for disabled
```
tlimoncelli, line 92: Go style: Start the comment with the name of the function.
```suggestion
// GetZoneDNSSEC gets the DNSSEC status for zone. Returns true for enabled, false for disabled
```

providers/ns1/ns1Provider.go
```diff
@@ -88,6 +89,52 @@ func (n *nsone) GetZoneRecords(domain string) (models.Records, error) {
...
+
+	// any other errors not expected, let's surface them
+	if err != nil {
+		return false, err
+	}
+
+	// no errors returned, we assume DNSSEC is enabled
+	return true, nil
+}
+// Create DNSSEC zone corrections based on current state and preference
```
tlimoncelli, line 113:
```suggestion
// getDomainCorrectionsDNSSEC creates DNSSEC zone corrections based on current state and preference
```

providers/ns1/ns1Provider.go
```diff
@@ -88,6 +89,52 @@ func (n *nsone) GetZoneRecords(domain string) (models.Records, error) {
...
+	}
+
+	// any other errors not expected, let's surface them
+	if err != nil {
+		return false, err
+	}
+
+	// no errors returned, we assume DNSSEC is enabled
+	return true, nil
+}
```
tlimoncelli, line 112: Add a blank line between each function unless there is a specific reason not to.

providers/ns1/ns1Provider.go
```diff
@@ -158,6 +211,21 @@ func (n *nsone) modify(recs models.Records, domain string) error {
 	return err
 }
 
+// Configures DNSSEC for a zone. Set 'enabled' to true to enable, false to disable.
```
tlimoncelli, line 214:
```suggestion
// configureDNSSEC configures DNSSEC for a zone. Set 'enabled' to true to enable, false to disable.
```

providers/ns1/ns1Provider.go
```diff
@@ -158,6 +211,21 @@ func (n *nsone) modify(recs models.Records, domain string) error {
...
+// There's a cornercase, in which DNSSEC is globally disabled for the account.
+// In that situation, enabling DNSSEC will always fail with:
+//   #1: ENABLE DNSSEC
+//   FAILURE! POST https://api.nsone.net/v1/zones/example.com: 400 DNSSEC support is not enabled for this account. Please contact support@ns1.com to enable it
+//
+// Unfortunately this is not detectable otherwise, so given that we have a nice error message, we just let this through.
+//
+func (n *nsone) configureDNSSEC(domain string, enabled bool) error {
+	z, _, err := n.Zones.Get(domain)
```
tlimoncelli, line 223: Check `err`.  (otherwise, it is overwritten by line 225)

tlimoncelli (commented):
Excellent work so far!  Just a few minor suggestions.

providers/ns1/ns1Provider.go
costasd: ah yeah, always forget these. Thanks!

providers/ns1/ns1Provider.go
costasd: Ditto.

providers/ns1/ns1Provider.go
costasd: good catch!

tlimoncelli (approved):

Issue #1461, opened 2022-03-24
HEXONET: Bad printf-style format string


tlimoncelli:
CC @papakai 

I saw this error:

```
    integration_test.go:116: Failed getting nameservers Could not get status for domain yodream.com. [%!s(int=423) Service is not active!]
```

The code that generates it is providers/hexonet/nameservers.go:50

```
    return nil, n.GetHXApiError("Could not get status for domain", domain, r)
```

I think the intention is to include the HTTP status code 423, which is an int, not a string.

If we change the code as follows, the error message will be:

```
    integration_test.go:116: Failed getting nameservers Could not get status for domain "yodream.com". [423 Service is not active!]
```

I like using `%q` for domain names because it puts quotes around the value. If someone accidentally specified the domain with a space, dot, or special character, `%q` will highlight that.  (Ever spend a day debugging something only to find that the input had an extra space in it?  Nah, me neither :-) :-) :-) )

Suggested fix:

```
diff --git a/providers/hexonet/error.go b/providers/hexonet/error.go
index e4197ac1..6713110b 100644
--- a/providers/hexonet/error.go
+++ b/providers/hexonet/error.go
@@ -8,5 +8,5 @@ import (
 
 // GetHXApiError returns an error including API error code and error description.
 func (n *HXClient) GetHXApiError(format string, objectid string, r *response.Response) error {
-       return fmt.Errorf(format+" %s. [%s %s]", objectid, r.GetCode(), r.GetDescription())
+       return fmt.Errorf(format+" %q. [%v %s]", objectid, r.GetCode(), r.GetDescription())
 }
```

PR #1462, opened 2022-03-24
HEXONET: Fix error format string

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/1461


tlimoncelli:
CC: @papakai

papakai:
@tlimoncelli thanks! Shall I support with something, or just fyi?
btw. CircleCI looks really promising. We are actually using github actions.

tlimoncelli:
Just an FYI (or a chance for you to chime in if you have a different direction)

PR #1467, opened 2022-03-27
Warn if _PROVIDER not set

tlimoncelli:


tlimoncelli:
Replaced by https://github.com/StackExchange/dnscontrol/pull/1485

PR #1468, opened 2022-03-28
DNSIMPLE: bug-fix, allow non-apex NS records

philpennock:
The DNSimple provider is rejecting all NS records, on the basis that they can't be changed.  This is bogus: it's only the apex records which can not be changed.  Child domains can have delegation glue NS records.

There's no internal tests for the provider, so I tested with the domain where I encountered the bug.  I got the diffs to disappear with `dnscontrol preview`; then I removed one of the NS records in an rrset and re-ran, and saw the diff, and then `dnscontrol push` successfully reconciled.  `go test -v ./...` passes still.

@aeden FYI

### DIFF
--- a/providers/dnsimple/dnsimpleProvider.go
+++ b/providers/dnsimple/dnsimpleProvider.go
@@ -78,7 +78,7 @@ func (c *dnsimpleProvider) GetZoneRecords(domain string) (models.Records, error)
 		if r.Name == "" {
 			r.Name = "@"
 		}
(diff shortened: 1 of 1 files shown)


philpennock:
Looks to me like the build is failing for my not being a trusted person so CircleCI is not exposing secrets for my branch, so there's nothing for me to remediate here.

tlimoncelli (approved):

tlimoncelli:
Thanks for finding and fixing this!

PR #1470, opened 2022-03-31
Fix autodns documentation

riku22:
Fix autodns documentation example javascript.
The provider name was "HETZNER" instead of "AUTODNS".


tlimoncelli (approved):
Good catch!

PR #1485, opened 2022-04-21
Provider type ID should be stored in creds.json (with backwards compatibility)

tlimoncelli:


philpennock:
Not too fond of the magic `"-"` which is easy to overlook and will need lookups by users not familiar with the history, each time.  `"autoProvider"` might be better and will convey semantic meaning when seen.

Is the structure such that this can't just be dropped as mandatory and reflection used to decide if the next parameter is a provider type or an option?

tlimoncelli:
Interesting point!

Looking over the work I've done so far, the reflection is working in all cases I've tried.  I guess I can just get rid of the whole "-" concept.  What I haven't tried so far is updating the command-line tools.  They have positional arguments but I might find some way to handle the old and new format.

Thanks for the feedback!  I'll look into this later this week.

tlimoncelli:
> Not too fond of the magic `"-"` which is easy to overlook and will need lookups by users not familiar with the history, each time. `"autoProvider"` might be better and will convey semantic meaning when seen.

Updated as suggested.  The only place `-` is needed now is on the command line tools that use it as a positional argument.  Next step is to eliminate that last case when the positional argument is the last one on the command line. If that's all cases, then we can eliminate all the `-` shenanigans.

commands/previewPush.go
```diff
@@ -232,6 +243,154 @@ func InitializeProviders(credsFile string, cfg *models.DNSConfig, notifyFlag boo
 	return
 }
 
+// providerTypeFieldName is the name of the field in creds.json that species the provider type id.
```
philpennock, line 246: specifies

pkg/js/helpers.js
```diff
@@ -36,7 +36,27 @@ function getConfiguredDomains() {
...
+// NewRegistrar returns an registrar object.
+// For backwards compatibility, it accepts (name), (name, meta),
+// (name, type), (name, type, meta).
+function NewRegistrar() {
+  // For backwards compatibility, this is a wrapper around the legacy
+  // version of this function.
+  switch (arguments.length) {
+    case 1:
+      return oldNewRegistrar(arguments[0], "-")
+      break;
```
philpennock, line 48: Is `break` after `return` a stylistic choice?  (My JS is poor so I'm a little confused)

philpennock (approved):

pkg/js/helpers.js
tlimoncelli: Nope.  Just a typo.

tlimoncelli:
Good news, @philpennock !  While working through all the remaining code I was able to find ways to make everything forwards and backwards compatible.  From v3.16 on the new format will work.  In v4.0 we can remove the backwards compatibility.  All files and command line syntax works both ways, as long as people update their creds.json file to include a TYPE field.

philpennock:
(double-check: you will merge when you're happy, since I've already approved and you're not using a workflow where the approver merges, right?)

tlimoncelli:
Sadly I found a bug.  I'll work on this more later in the week. I have a big deadline Wed.

tlimoncelli:
Bug fixed.

@philpennock and others: I'd appreciate some testing before I do this merge.

* Try the new binary with zero changes to your config. (dnscontrol preview is sufficient).
* Update your configuration based on the warnings and informational messages that are output.  Try "dnscontrol preview" and push.

Let me know if there are bugs, if anything is confusing, etc.

tlimoncelli:
Closing.  All changes for this were moved to other PRs.

PR #1496, opened 2022-05-03
Update cloudflare with PTR support

adrian-hoasted:


tlimoncelli:
Tested locally.  All passed!

Thanks for adding this!

Issue #1372, opened 2022-01-13
DNSControl permits different TTLs within a resource recordset.


tlimoncelli:
For a given DNS label, all the records of the same type need to have the same TTL. (i.e. all the TXT records on foo.bar.com need to have the same TTL.

This isn't enforced by all providers, and DNScontrol doesn't check for out of alignment TTLs.  This confuses providers.

For example:  With the following TXT records, the provider will set the TTL for `TXT'bar.example.com` to be 300, then 9999, then fail when the records are re-checked and the TTL is found to be 9999 instead of 300.

```
D("example.com", FOO,
    // ...
    TXT("bar", "one", TTL(300)),
    TXT("bar", "two", TTL(9999)),
    // ...
)
```

Proposed solution:

1. Add a function call to `auditrecords.go` which checks for misaligned TTLs.
2. Initially make this a warning.  Then make it an error a few releases later.

Tom

juliusrickert:
Clarified in [RFC 2181 Section 5.2](https://datatracker.ietf.org/doc/html/rfc2181#section-5.2).

tlimoncelli:
Alternatively.... we include this globally, not just on a per-provider basis.  Initially it could be a warning but then would be an error.

juliusrickert:
Sounds good. Have to do justice to being an opinionated project somehow :D

But this would be a breaking change too.

tlimoncelli:
Yes, that would be a breaking change.  I'd make it a warning now so that we can make it an error the next time we bump the major revision number.  No need to bump the major rev just for this.

costasd:
Hey, had some spare time so I took a stab at this :)

PR #1489, opened 2022-04-27
Emit warning in case of label having multiple TTLs

costasd:
An RRSet (=label) consisting of multiple records with different TTLs is
something not supported by most providers, and should be avoided.
Furthermore it is deprecated in rfc2181#section-5.2

Emit a warning for now during validation, eventually turning it into a full-blown error.

Fixes #1372

PR #1498, opened 2022-05-03
pkg/js/parse_tests: fix a broken test

costasd:
012_duration creates multiple records under the same label, with different TTLs
We plan to deprecate this going forward, in #1372.

Modify the test in a way that creates one label per record, each having a
different TTL format as before, to be validated.


costasd:
#1489 requires this

tlimoncelli (approved):

PR #1501, opened 2022-05-08
Update maintainers for Route53 and Cloudflare

tresni:
I'm happy to take these two over in addition to EasyName.


tlimoncelli:
Thank you so much for stepping up!  I really appreciate it, as do the users of Route 53 and Cloudflare!

PR #1512, opened 2022-05-28
Add calling an executable for credentials

plttn:
Certain password managers include a CLI that allows for injecting credentials into a template file. Using `!` works for a single shell script, but doesn't work so well for calling a program, considering it explicitly ignores items in the path. This allows for tools like 1Password CLI to be used to inject the secrets into the json, without ever needing to write the credentials to disk.

As a note, I didn't write any tests for this (simply because I don't really know Go 😅), so while I've tested it with intentional correct input on both macOS and Windows (successfully on both), it may fail spectacularly if bad input is entered.

`dnscontrol preview --creds="@op inject -i creds.json.tpl"` will result in `op` injecting the credentials into the templated file, outputting it via stdout, and `dnscontrol` working as expected.

### DIFF
--- a/pkg/credsfile/providerConfig.go
+++ b/pkg/credsfile/providerConfig.go
@@ -57,6 +57,11 @@ func LoadProviderConfigs(fname string) (map[string]map[string]string, error) {
 		if err != nil {
 			return nil, err
 		}
+	} else if strings.HasPrefix(fname, "@") {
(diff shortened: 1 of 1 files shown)


plttn:
Realized I didn't do this "properly" from a branch, so I'm going to go ahead and close this and re-PR.

PR #1513, opened 2022-05-28
Add calling an executable for credentials.

plttn:
Certain password managers include a CLI that allows for injecting credentials into a template file. Using `!` works for a single shell script, but doesn't work so well for calling a program, considering it explicitly ignores items in the path. This allows for tools like 1Password CLI to be used to inject the secrets into the json, without ever needing to write the credentials to disk.

As a note, I didn't write any tests for this (simply because I don't really know Go 😅), so while I've tested it with intentional correct input on both macOS and Windows (successfully on both), it may fail spectacularly if bad input is entered.

`dnscontrol preview --creds="@op inject -i creds.json.tpl"` will result in op injecting the credentials into the templated file, outputting it via stdout, and dnscontrol working as expected.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,6 +60,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
+
 require (
 	cloud.google.com/go/compute v1.6.1 // indirect
 	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
--- a/go.sum
+++ b/go.sum
@@ -320,6 +320,8 @@ github.com/google/pprof v0.0.0-20210601050228-01bbb1931b22/go.mod h1:kpwsk12EmLe
 github.com/google/pprof v0.0.0-20210609004039-a478d1d731e9/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/pprof v0.0.0-20210720184732-4bb14d4b1be1/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
+github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510 h1:El6M4kTTCOh6aBiKaUGG7oYTSPP8MxqL4YI3kZKwcP4=
+github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510/go.mod h1:pupxD2MaaD3pAXIBCelhxNneeOaAeabZDe5s4K6zSpQ=
 github.com/google/uuid v1.1.1/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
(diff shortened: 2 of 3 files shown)


tlimoncelli:
Hi!

Thanks for the PR.  It's a good idea.  I do have some questions/suggestions:

1.  Why introduce `@` as a prefix?  Why not add this functionality to `!`?  (i.e. would it break any backwards compatibility?)

2. The reason that `!` only takes one item is that I assumed most people would call a wrapper that would do the right thing. The wrapper doesn't need to take any parameters.   The benefit of this PR is that it would eliminate the need for a wrapper, and therefore eliminate the complexity of transporting the wrapper.  However your example includes a template file which would need to be copied around.  I don't see how this is any less complex than using a wrapper.

3. Rather than strings.Fields(), I think shlex.Split() might do a better job.  https://pkg.go.dev/github.com/google/shlex#Split


5.

plttn:
Great questions:

1. As I was taking a look at implementing it, I saw that there's a specific [callout](https://github.com/StackExchange/dnscontrol/blob/39b9dfe3d16827f49cbb8f85b612d271bf008c5a/pkg/credsfile/providerConfig.go#L116) to ensure that `!` isn't relying on `$PATH`. Given that `executeCredsFile` is effectively enforcing `./` on bare calls, while I can't immediately think of a situation where it would occur, it feels questionable at best to change that behavior after the fact to have that prefix resolve the path. Given that, I thought it was best to use a new prefix for saying "yes I want to use my path". 
2. I haven't found any clean ways to write a script / wrapper that would function on multiple operating systems with minimal dependencies. While yes it's true that the template file still needs to be included in the repository, this means that I can run this on both Windows and Mac, with the only two required dependencies being DNSControl, and the external program itself. 
3. This actually makes sense (given that the string may not be split properly by strings.Fields()), so I've modified my branch to use that.

tlimoncelli:
Ah, yes, I remember now.  We add `./` to the front because `--creds filename` assumes a non-executable file is is in the current directory (or relative to it), therefore we wanted `--creds !filename` would find the same file.

The proposal is that `--creds @string` is treated as a shell one-liner, which would search the PATH as needed.

I'm struggling to think of a way to do this PR without introducing a new symbol (`@`).  But maybe that's not too important. 

CC: @jpbede for input/feedback.


@plttn Either way, you'll also want to update `docs/creds-json.md`

jpbede:
Thanks @plttn for this PR from my side too 🙂. I'm with @tlimoncelli and would like to not introduce a new symbol. I refactored the code a bit and used some code from you @plttn to create the following: 

https://github.com/StackExchange/dnscontrol/compare/master...jpbede:feature/creds-command

This should allow for the following commands:
`dnscontrol preview --creds !./creds.sh`
`dnscontrol preview --creds ./creds.sh`
`dnscontrol preview --creds creds.sh`
`dnscontrol preview --creds !creds.sh`
`dnscontrol preview --creds !/some/absolute/path/creds.sh`
`dnscontrol preview --creds /some/absolute/path/creds.sh`

And of course shell commands:
`dnscontrol preview --creds "!op inject -i creds.json.tpl"`

What is your opinion on this?

tlimoncelli:
> What is your opinion on this?

Looks good to me!  @plttn ?

plttn:
Seems like a cleaner solution to avoid adding a second prefix character! Looks good to me as well! 🙂

Since you've got a branch going on your end, I'll go ahead and close this PR out.

tlimoncelli:
@jpbede One comment about your code.  Use https://pkg.go.dev/path/filepath#Join to join the "." to the filename.

PR #1518, opened 2022-05-31
VULTR: Enable long TXT

androw:
Vultr seems to support long TXT record as of now.


tlimoncelli (approved):
Thanks for identifying this!

PR #1523, opened 2022-06-03
docker: Add CA certificates to Docker image

cr1xu5:
This change makes sure dnscontrol can connect to the providers over https when the Docker image is used.


tlimoncelli:
Thanks for the update!

Issue #1369, opened 2022-01-06
Eliminate the need for create-zones: Create the zones dynamically


tlimoncelli:
"dnscontrol push" should automatically add the zone to your account.  There should be no need for a "create-zones" command.

The `dnscontrol create-zones` subcommand was created when DNSControl was new and we weren't sure how we wanted it to work.

This is a breaking change and should be hidden behind a flag that is initially default-off, but will be default-on at the next major release.

juliusrickert:
… and `dnscontrol preview` should show to-be-created zones.

I think this should be considered a breaking change :D

tlimoncelli:
@juliusrickert Why would this be a breaking change?  I think that if "dnscontrol preview" outputs a sufficient warning, it shouldn't be a surprise to see the domain created during a push.

That is...

`preview` should output something like:

```
Domain "example.com" does not exist in the "myroute53" profile and will be added automatically.
```

Remember: This is adding it to the DSP, not registering it or doing anything that costs money.

PR #1528, opened 2022-06-06
Create non-existing zones while push or print warning while preview

jpbede:
Proposal which fixes https://github.com/StackExchange/dnscontrol/issues/1369.

It checks with `ListZones()`during preview if the zone already exists, if not, a warning is issued. During push it is ensured that the domain exists. But I'm not quite happy with the flag name, do you have a suggestion?

### DIFF
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -469,9 +469,6 @@ func printOrRunCorrections(domain string, provider string, corrections []*models
 	if len(corrections) == 0 {
 		return false
 	}
-	for i := 0; i < len(corrections); i += 900 {
-
-	}
 	for i, correction := range corrections {
 		out.PrintCorrection(i, correction)
 		var err error
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -36,9 +36,9 @@ type PreviewArgs struct {
 	GetDNSConfigArgs
 	GetCredentialsArgs
 	FilterArgs
-	Notify          bool
-	WarnChanges     bool
-	CreateWhilePush bool
+	Notify             bool
+	WarnChanges        bool
+	PopulateAtProvider bool
 }
 
 func (args *PreviewArgs) flags() []cli.Flag {
@@ -56,9 +56,9 @@ func (args *PreviewArgs) flags() []cli.Flag {
 		Usage:       `set to true for non-zero return code if there are changes`,
 	})
 	flags = append(flags, &cli.BoolFlag{
-		Name:        "create-while-push",
-		Destination: &args.CreateWhilePush,
-		Usage:       `set to true to create non-existing zones on provider`,
+		Name:        "populate-at-provider",
+		Destination: &args.PopulateAtProvider,
+		Usage:       `set to true to create non-existing zones at provider`,
 	})
 	return flags
 }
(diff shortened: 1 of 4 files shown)


commands/previewPush.go
```diff
@@ -52,6 +55,11 @@ func (args *PreviewArgs) flags() []cli.Flag {
 		Destination: &args.WarnChanges,
 		Usage:       `set to true for non-zero return code if there are changes`,
 	})
+	flags = append(flags, &cli.BoolFlag{
+		Name:        "create-while-push",
```
tlimoncelli, line 59: Hmm... I'm terrible at names too.  Just some ideas:

`--populate-at-provider` ?

`--pre-populate` ?

`--prepare-domain-at-provider` ?

jpbede: I think `populate-at-provider` would fit.

commands/previewPush.go
```diff
@@ -52,6 +55,11 @@ func (args *PreviewArgs) flags() []cli.Flag {
 		Destination: &args.WarnChanges,
 		Usage:       `set to true for non-zero return code if there are changes`,
 	})
+	flags = append(flags, &cli.BoolFlag{
+		Name:        "populate-at-provider",
+		Destination: &args.PopulateAtProvider,
+		Usage:       `set to true to create non-existing zones at provider`,
```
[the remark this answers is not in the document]
jpbede: Changed the description as it is just required to use the flag

tlimoncelli (approved):
Question: Do we want to make this the default and have the flag be reversed?

jpbede:
Indeed a good question, I did it for now as it described in the issue. But the more I think about it, I think it would be better to reverse the logic, since dnscontrol will not be as useful as with a zone ;). Moreover, this is not destructive and does not cause any damage.

tlimoncelli:
I love it when a plan comes together!

I'm ready to merge this. I just want to double-check with you that it is ready.

tlimoncelli:
Before I merge this I want to give a special shout-out to @jpbede for contributing this code and why it's so awesome.

A little DNSControl history:

When DNSControl was new, none of the providers required populating zones in accounts (BIND, for example).  We hit our first provider that required zones to be populated and we weren't sure what to do.  Should we just automatically create/populate the zone?  What if a user doesn't want that?  What if there are typos in the dnsconfig.js file and it creates the wrong zone name?  What if adding a zone to an account has terrible, terrible, terrible side-effects and our auto-create code causes the world to end? (Actually I don't remember all the fears... maybe it was all fear of the unknown).

As a result, we took the cautious route and created the "create-domains" subcommand.

Years later it's pretty clear that our fears were unwarranted.   It isn't the end of the world if a typo leads to the wrong name being created, or if a lot of zones are populated, or anything else.  It caused a lot of user-confusion: Both because people would be confused that dnscontrol didn't do this already (and had to be told about the create-domains subcommand), or people misunderstood the create-domains command to be something about purchasing domains.

Now dnscontrol will automatically populate the zones.  Best of all, thanks to a last minute change, the flag name became obvious!  `--no-populate`!

I'm super happy to see this added.  It corrects a decision made years ago and will make dnscontrol easier to use by new people.

Thanks!

jpbede:
You're welcome. Thanks for all your work on dnscontrol, it saved me and my colleagues many hours of work :)

PR #1531, opened 2022-06-08
Preventing segfault when Cloudflare API limit is reached

tlimoncelli:
Signed-off-by: Jan-Philipp Benecke <jan-philipp@bnck.me>


tlimoncelli:
CC @tresni

PR #1543, opened 2022-06-17
CLOUDFLAREAPI: Fix TXT quoting bug

tlimoncelli:


tlimoncelli:
Second try is much better!

The problem was even deeper than I thought!  

Cloudflare's API only supports one TXT string of any non-zero length. No multiple strings (TXTMulti).  When serving the DNS record, it splits strings >255 octets into individual segments of 255 each. However that is hidden from the API.
  Therefore, whether the string is 1 octet or thousands, just store it as one string in the first element of .TxtStrings.

The code was assuming that multiple strings could be sent to the API using `"foo" "bar"` format (RFC 1035). Surprisingly this actually didn't fail any of the integration tests.  In fact, I'm not sure if there is any way to add an integration test that would detect this bug.

After many iterations of trying to devise a parser for something that wasn't meant to be parsed, I realized I was making everything too difficult for myself and the solution was: don't parse anything at all!

Along the way I found some related issues including a function that should be deprecated. See https://github.com/StackExchange/dnscontrol/issues/1545

https://github.com/StackExchange/dnscontrol/pull/1543 now passes all test.

PR #1544, opened 2022-06-18
MAINT: Enforcing formatted code

jpbede:
I would like to suggest that the code must be formatted with `go fmt ./...`, which is checked and enforced by CircleCI.
This results in consistent formatting and has proven to work well in our company.


tlimoncelli:
Brilliant!

tlimoncelli:
Note: We do something similar with "go generate" in the build process.

PR #1548, opened 2022-06-18
MAINT: Update TXT docs, suggest not using TxtNoLen255

tlimoncelli:
I'm renaming 2 confusing functions:

The confusing names are:

* `TxtNoLen255()` means "no strings that are exactly 255 octets long".  
* `TxtNoLongStrings()` means "no strings longer than 255 octets".

Why is this confusing?  I realized I had used `TxtNoLen255()` in a provider when I had intended to use `TxtNoLongStrings()`.  You can see how they might be confusing.

Therefore, I'm renaming them to be very explicit: `TxtNoStringsExactlyLen255()` and `TxtNoStringsLen256orLonger()`

This change only affects 3 providers: MSDNS, CSCGLOBAL, and INWX.

CC @svenpeter42


tlimoncelli:
@svenpeter42  I apologize for the confusion. I'm going to move the INWX-related part of this to a separate issue.

Issue #1545, opened 2022-06-18
Replace SetTargetTXTString with alternatives


tlimoncelli:
The quote parsing in SetTargetTXTString() doesn't handle escaped quotes. It is also badly named, because it doesn't imply that the string is parsed, which may lead to providers using it accidentally on strings that should not be parsed.

SetTargetTXTString should be depreciated.  Users are better off with the new function SetTargetTXTfromRFC1025Quoted() or simply SetTargetTXT() (single string) or SetTargetTXTs() (multiple strings).

digitalocean, msdns and packetframe are the only providers that use the function.  Please eliminate the use of SetTargetTXTString() when you can.

- [x] Fix DIGITALOCEAN.  @Deraen
- [x] Fix MSDNS.  @tlimoncelli 
- [x] Fix PACKETFRAME.  @hamptonmoore
- [ ] When the above are done, remove the old function from the repo  @tlimoncelli 

To clarify: Most providers should use SetTargetTXT() or SetTargetTXTs().   I've written a longer explanation in `models/t_txt.go`.

tlimoncelli:
@Deraen @hamptonmoore Friendly ping?

Deraen:
It seems like Digitalocean is already done:
https://github.com/StackExchange/dnscontrol/commit/eb4af05d319c1ee6127306cb5b450e8499ded259
https://github.com/StackExchange/dnscontrol/commit/2586e2b611c0054abbc25c15fe64a25d0ee7b51f#diff-b71933063db51c94fac5af976d4589ff5b578d283d1f857cd37676cc5fcd20d2

The tests are passing for me.

(Seems like Packetframe is also using `SetTargetTXT` already.)

tlimoncelli:
Thanks for checking!

Yup!  Looks like we're ready to close this.

PR #1556, opened 2022-06-20
DIGITALOCEAN: Replace SetTargetTXTString use with a new function

Deraen:
Related to #1545

I've run the tests with the change.

### DIFF
--- a/providers/digitalocean/digitaloceanProvider.go
+++ b/providers/digitalocean/digitaloceanProvider.go
@@ -276,7 +276,7 @@ func toRc(domain string, r *godo.DomainRecord) *models.RecordConfig {
 	t.SetTarget(target)
 	switch rtype := r.Type; rtype {
 	case "TXT":
-		t.SetTargetTXTString(target)
+		t.SetTargetTXTfromRFC1035Quoted(target)
 	default:
 		// nothing additional required
 	}


Deraen:
Looks like go fmt check for PRs is broken on CI, this doesn't change formating, and go fmt doesn't output anything for me locally.

tlimoncelli:
@jpbede fixed the CI issue. Thanks for pointing that out!

tlimoncelli:
Before we merge this, I'd like you to try something.

I recently realized that Cloudflare uses `SetTargetTXTfromRFC1035Quoted()` when it should be using SetTargetTXT().  This makes me wonder if other providers made the same mistake.

Sadly I can't think of a way for the integration tests to determine if the wrong function was used.

Does the DO API really return a single string that needs to be parsed?  I looked at the API documentation and it didn't say anything about how they store TXT records.

Could you try your PR using SetTargetTXT() instead?  It might require you to update auditrecords.go and add recordaudit.TxtNoStringsLen256orLonger and recordaudit.TxtNoMultipleStrings.

Thanks!

tlimoncelli:
The problem is that I don't have a good way to explain to people under what situations its best to use (a) SetTargetTXTfromRFC1035Quoted or (b) SetTargetTXT + fix the bugs.  This is my problem to solve.  It's not about you.

The problem is that SetTargetTXTfromRFC1035Quoted will pass all integration tests... yet it is still the wrong choice.

My gut tells me that the right thing to do is:

Go to the web control panel. If it permits you to enter multiple strings, use SetTargetTXTfromRFC1035Quoted.  If it only permits a single string, switch to SetTargetTXT and fix the bugs. (Possibly adding to auditrecord.go to eliminate troublesome edge cases)
 
Sorry for being confusing.  This is kind of a mess and I'm struggling to find a way to put into words the entire situation.  I appreciate your patience.

tlimoncelli:
The changes I made in https://github.com/StackExchange/dnscontrol/pull/1567 might fix this.

tlimoncelli:
Fixed in https://github.com/StackExchange/dnscontrol/pull/1567

PR #1567, opened 2022-06-20
BUGFIX: PopulateFromString should not parse TXT strings

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/1568


tlimoncelli:
I debugged GANDI_V5 and ROUTE53.  Basically I just special-cased TXT and called the parser compatible with SetTargetTXTString()

tlimoncelli:
Well folks.... this is a bit difficult to say but... I think I'm going to cancel this PR.

I've always said "if you have 2 solutions, pick the most simple one".    This PR is getting longer and longer, with more changes and more complexity.  A few days ago I was digging into the diff.go code because changes I made in one place was breaking algorithms in the diff engine.  I was chasing problems left and right. Every time I fixed one bug, I created a bug some place else.

Oh a whim, I decided to revisit the reason this PR was started.  Could we fix the same problem some other way?

The original bug was fixed in https://github.com/StackExchange/dnscontrol/pull/1543 and fundamentally I wanted to make sure that other providers didn't have the same bug. I wanted to add code to make it difficult for future providers to make the same mistake.  This PR is a complex way of doing all that.

The new solution I came up with was a simple document (`docs/testing-txt-records.md`) that tells people how to check to see if the bug exists in their provider.

That is a much more simple solution than this PR.

I realize that people have put a lot of work into this PR already.  Many people have run integration tests, rewritten code, offered new parsers and decoders.   The new code has a lot of good ideas in it, but it is a big change when a small change will do.  I'm grateful to everyone that put in a lot of excellent work on this.  I'm sorry I sent everyone on a while goose chase.

What's next:

* Merge the new solution https://github.com/StackExchange/dnscontrol/pull/1622
* Salvage any useful bits from this PR and merge them in.  I'll file bugs and PRs.
* Everyone should take a look at `docs/testing-txt-records.md` to see if they are affected by this bug.

Again I appreciate all the help with this PR.  I apologize for creating extra work for people.  The bottom line is that this is a better, and less error-prone, solution.

tlimoncelli:
I've salvaged anything useful from this PR and merged them via other PRs.

Closing.

PR #1569, opened 2022-06-20
POWERDNS: Cleanup and fix parsing of TXT records

jpbede:
The PowerDNS API is quite strange. It accepts long TXTs without requiring to split them and returning them as they came in.
e.g. `"averylooooooo[...]oooooongstring"` or `"string" "string"`

Splitting on DNS level will be done on the fly.


tlimoncelli (approved):

tlimoncelli:
Thanks!

PR #1572, opened 2022-06-22
Use HTTPS for git endpoint for cloning

pnorman:
Github no longer supports the unsecured git protocol, so use HTTPS instead.


tlimoncelli:
Thanks!  Fixing these little paper-cuts makes it easier for people to join the dnscontrol family.

Issue #1568, opened 2022-06-20
MAINT: PopulateFromString should use SetTargetTXT, not SetTargetTXTString


tlimoncelli:
[NOTE: This is not urgent.  I'm asking that everyone check their code in the next 4-5 weeks: I'd like to merge this on Aug 1, 2022]

The good news: PopulateFromString is very useful and a lot of people use it.

The bad news: PopulateFromString has a bug, and I need to check with everyone to make sure they aren't affected.

WHAT IS THE PROBLEM?

PopulateFromString calls SetTargetTXTString, which is almost never the right thing to do.  It should use SetTargetTXT() instead.

SetTargetTXTString parses the string for quotes... incorrectly.  Generally provider code receives a big TXT string that doesn't need to be parsed.

WHY CHANGE?

The integration tests do not detect if the incorrect `SetTargetTXT*` function is called.  I can't figure out a test that would do that.  Therefore, we have to make this change carefully.

Here is a checklist of who uses this function:

- [ ] akamaiedgedns @svernick
- [ ]  autodns @arnoschoon
- [X]  cloudflare @tresni
- [ ]  desec @D3luxee
- [ ]  dnsimple @aeden
- [x]  dnsmadeeasy @vojtad
- [ ]  exoscale pierre-emmanuelJ
- [x]  gandiv5 @TomOnTime
- [x]  gcloud @riyadhalnur
- [ ]  hedns @rblenkinsopp
- [x]  hetzner @das7pad
- [X]  hexonet @papakai
- [x]  hostingde @juliusrickert
- [ ]  inwx @svenpeter42
- [ ]  linode @koesie10
- [x]  namecheap @captncraig
- [X]  namedotcom
- [x]  ns1 @costasd
- [ ]  oracle @kallsyms
- [x]  ovh @masterzen
- [x]  powerdns @jpbede
- [x] route53 @tresni
- [x]  transip @blackshadev
- [X]  vultr @pgaskin

WHAT SHOULD I DO?

Git clone [this](https://github.com/StackExchange/dnscontrol/pull/1567) and run your integration tests.

If your integration tests work, let me know.

if the integration tests fail, I'd be glad to help you fix them.

I've written a longer explanation of how TXT records are handled in [models/t_txt.go](https://github.com/StackExchange/dnscontrol/blob/master/models/t_txt.go)

tlimoncelli:
If this change breaks your integration tests, I have a few suggestions:

[...]

PR #1579, opened 2022-06-24
DNSMADEEASY: fix parsing TXT records

vojtad:
This PR fixes parsing TXT records for #1568.

Integration tests are now passing for the DNSMADEEASY provider.


tlimoncelli:
Thanks for helping out with this effort!

PR #1581, opened 2022-06-24
Fix population of txt records from TransIP

blackshadev:
As requested by tlim, the PopulateFromString has changed and thus
some providers need to handle TXT records differently. See issue
linked below for more information

https://github.com/StackExchange/dnscontrol/issues/1568


tlimoncelli (approved):

tlimoncelli:
Thanks!

PR #1582, opened 2022-06-27
Fix handling of TXT for the OVH provider

masterzen:
OVH only supports arbitrary long string even though it is possible to put more than
one quoted string in a TXT target with the API.

Consequently, the best solution for OVH is to:
* quote the string before emitting to OVH
* unquote if needed when getting the string from OVH
* let dnscontrol know OVH doesn't support multiple TXT strings


tlimoncelli:
Thanks!

PR #1583, opened 2022-06-29
Enable ADC for Gcloud provider

miono:
This is a draft-PR where we want to have some feedback on whether you
think this is a good idea, if so, we will make the required
documentation changes etc.

Fallback to using Application Default Credentials if no
private_key is available in the creds.json-file.

This allows usage with various short lived credentials, for example
GCE-instances, gcloud auth application-default login, OIDC Workload
Identity etc


miono:
CC: @riyadhalnur

riyadhalnur:
@miono Thanks for this. Afaik though,  GCP services will have the `GOOGLE_APPLICATION_CREDENTIALS` environment variables set on their services and the client libraries will look for the credentials automatically. Do we need to set it explicitly like this? (ref https://cloud.google.com/docs/authentication/production#automatically)

miono:
> @miono Thanks for this. Afaik though, GCP services will have the `GOOGLE_APPLICATION_CREDENTIALS` environment variables set on their services and the client libraries will look for the credentials automatically. Do we need to set it explicitly like this? (ref https://cloud.google.com/docs/authentication/production#automatically)

The way I read the documentation and the code, when using the `JWTConfigFromJSON` function we will never touch the library built in auto-discovery of ADC.

So our reasoning was to use JWTConfigFromJSON if we actually have a JSON private_key (current behavior) but use `DefaultClient` if there isn't one, getting the library to do the standard automatic discovery of credentials outlined here: https://cs.opensource.google/go/x/oauth2/+/02e64fa5:google/default.go;l=114

tlimoncelli:
FYI: IIRC, the ROUTE53 and AWS_DNS providers do something similar.  I haven't used the feature, but if I understand it correctly, if you are configured properly, the provider leverages the existing authentication.  That is, dnscontrol leverages the same authentication as the CLI tools `aws` and `az`.  If you've authenticated with them, and now no longer need to re-authenticate for each command, dnscontrol should work the same way.

miono:
@tlimoncelli @riyadhalnur Yes, that's our understanding aswell. That's the same behaviour we are adding to the gcloud provider with this change.

riyadhalnur:
@miono I understand what you mean. Makes sense to have this feature. We should try and do a fallback with the default environment variables. We can move ahead with this change.

miono:
@riyadhalnur I have now changed the provider documentation and removed the Draft-marker from the PR.

tlimoncelli:
The code LGTM.  @riyadhalnur: please give this a :+1: and I'll merge it.

riyadhalnur:
@tlimoncelli LGTM. We can go ahead and merge this

@miono Thanks for this change. It would be nice to have unit and integration tests for this as a follow-up.

riyadhalnur (approved):

PR #1584, opened 2022-07-01
Simplify digitalocean using new helper functions

tlimoncelli:
@Deraen I added a helper function that makes your code a lot smaller.


tlimoncelli:
Fixed elsewhere. Sorry for the bother.

PR #1608, opened 2022-07-08
DNSIMPLE: do not support DS records

onlyhavecans:
DNSimple does not support adding your own DS records at this time.


tlimoncelli:
Thanks for spotting this!

PR #1609, opened 2022-07-08
DNSIMPLE: Change code owner/maintainer

onlyhavecans:
Anthony delegates all tasks relating to our DNSControl DNSimple integration to me. It would be more efficient to set me as the code owner.

I am an employee of DNSimple and use DNSControl's DNSimple integration in my other projects as well.

CC @aeden in case he needs to sign off.


tlimoncelli:
Sounds good!  Welcome to the club!

PR #1596, opened 2022-07-05
Update TXT handling for DNSimple

tlimoncelli:
@aeden please run the integration tests and verify I didn't break anything.

PR #1624, opened 2022-07-15
DNSIMPLE: Fix TXT Handling, Second Edition

onlyhavecans:
After reviewing the comments and steps listed in https://github.com/StackExchange/dnscontrol/pull/1596 and [Testing TXT Records](https://stackexchange.github.io/dnscontrol/testing-txt-records.html) I discovered the way we handle TXT is still incorrect and buggy.

With this PR all manual and integration tests pass or have the appropriate audit. 

I added a new audit and test to support DNSimple's use case where we support double quotes, but only when paired in a record. ie `t"hi"s` and `"th" "is"` are valid but `th"is` is not.


onlyhavecans:
All right! I found the offending issue with long lines being broken up and added a bit more to get our txt parsing working correctly.

tlimoncelli (approved):
LGTM (and thank you for updating the docs, and contributing to the recordaudit/txt.go library of functions!)

PR #1629, opened 2022-07-18
Add RWTH provider

MisterErwin:
The RWTH (a German university) has a custom system for subdomains (called RWTH DNS Admin).

A blog post about this can be found [here](https://blog.rwth-aachen.de/itc/2022/07/13/api-im-dns-admin/) and the API is documented [here](https://noc-portal.rz.rwth-aachen.de/dns-admin/en/api_tokens) (although it requires users to be logged in).

An interesting quirk of this provider is that changes are only deployed four times an hour (and deleted records can still be recovered). (Which is why many integration tests fail.)

While I am not certain if it makes sense to add this non-public-available provider to the list, this PR at least will start a discussion about it.


tlimoncelli:
How interesting!

This is the first time someone has suggested supporting a non-public API!

I'm very excited, because it means people value DNSControl.  However I'm not sure how we can support this kind of thing.   I guess the worse-case scenario would be we'd drop the provider from the `_all.go` file.

Let's move forward with this provider and see how it goes!  

Tom

tlimoncelli:
Before we merge this I'd like to request that you reorganize the filenames (and move functions if needed).  `providers/cscglobal` is the best example of how things should be done in the future.  I'm in the process of updating the docs to reflect this.  See https://github.com/StackExchange/dnscontrol/pull/1636 for what I've written so far.

I know this is a big change, but doing it now is better than rewriting it all later.    I'd really appreciate it!

Thanks!
Tom

MisterErwin:
I reorganized it to (hopefully) fit in the new layout.

Although I am a bit unsure if I managed to find the right balance between putting functions into the api and convert files (As I am not wrapping an existing API library).

Feel free to correct me (both yourself in the code or via text in a comment) if I did not hit the spot :) 

Cheers
~Alex

tlimoncelli:
Staticcheck reports:

```
providers/rwth/api.go:51:6: func checkIsLockedSystemApiRecord should be checkIsLockedSystemAPIRecord (ST1003)
providers/rwth/convert.go:41:1: comment on exported function NewRR should be of the form "NewRR ..." (ST1020)
providers/rwth/dns.go:19:9: unnecessary assignment to the blank identifier (S1005)
providers/rwth/dns.go:30:26: methods on the same type should have the same receiver name (seen 12x "api", 1x "apo") (ST1016)
```

tlimoncelli:
Thanks for contributing this!    I hope it helps everyone at RWTH have success with DNS!

PR #1636, opened 2022-07-20
Better explain how to write providers

tlimoncelli:


docs/writing-providers.md
```diff
@@ -147,6 +147,20 @@ The function `GetRegistrarCorrections()` returns
...
+
+Here are the filenames and their purpose.  (This is new and most providers have not been updated to use this layout, but
+`providers/cscglobal` is a good example.)
+
+- `cscglobalProvider.go` - Initializers, capabilities, and struct/type definitions
+- `dns.go` - Get/fix DNS records: GetZoneRecords, GetDomainCorrections, GenerateDomainCorrections and related functions.
+- `registrar.go` - Get/fix registrar (nameserver) info: GetRegistrarCorrections and related functions
+- `listzones.go` - ListZones and related functions
+- `convert.go` - Converting between models.RecordConfig and the provider's native format
+- `api.go` - API primatives, often an ediface in front of a vendor-provided library
```
SimenBai, line 160: Might be my non-native English knowledge, but AFAIK ediface is not a word. At least not one that is commonly used.

tlimoncelli: Very valuable feedback!    The correct word is "facade" as in  https://en.wikipedia.org/wiki/Facade_pattern

SimenBai:
I also noticed that some of the internal references on the page (for instance Step 14.2) is pointing to the wrong step (8 instead of 9).

tlimoncelli:
Yes, this is an early draft. I'll be sure to fix the internal references.  Thanks

docs/writing-providers.md
tlimoncelli: FYI: The correct spelling is edifice.  "edifice" is a word for "building".  The "facade of an edifice" is the "front of a building".

None of that matters. I'm closing this PR.

Issue #1641, opened 2022-07-28
preview and push commands exit when they encounter a Hetzner zone that doesn't exist


gregdotca:
Hello! It's my understanding that the _create-domains_ command is now deprecated, and going forward the _push_ command will create DNS zones for you, however this doesn't seem to work with Hetzner. Also, running the _preview_ command exits as soon as it gets to a Hetzner zone that doesn't exist.

**preview**
```
./dnscontrol preview
******************** Domain: example.com
----- Getting nameservers from: hetzner
"example.com" is not a zone in this HETZNER account
(There are more domains after example.com, however DNSControl exits right here)
```
**push**
```
./dnscontrol push
******************** Domain: example.com
----- Getting nameservers from: hetzner
"example.com" is not a zone in this HETZNER account
(There are more domains after example.com, however DNSControl exits right here)
```

Have I misunderstood how this is supposed to work? If so, is there a way to do a preview without having it exit when it encounters a zone that doesn't exist, or a way to create DNS zones at Hetzner if they don't already exist?

DNSControl version v3.17.0
OS: Debian 11  (running in an LXC container on Proxmox 7.2-7)

tlimoncelli:
CC @das7pad

Not all providers in DNSControl supports populating domains in the account.  If it doesn't, it errors out as you described.  The user must either create the domain in the account (usually via a web portal) or remove the domain from dnsconfig.js.

However it looks like HETZER supports populating domains in the account.   In Go terms, it "completes the DomainCreator interface" (i.e. has a function called `EnsureDomainExists`)

Jakob, could you investigate?

Thanks!
Tom

das7pad:
Looking at the preview process in dnscontrol core, it fetches nameservers ahead of checking whether a given zone exists.

https://github.com/StackExchange/dnscontrol/blob/c0450223c2e3eb0a67a42092cc5620bb05ebab8d/commands/previewPush.go#L132-L158

In other words, the HETZNER provider has not had a chance to create the zone before the listing of nameservers started.

We need to flip the order here:
old:
1. list nameserver
2. for each provider
    1. create zone if needed
    2. get corrections

new:
1. for each provider
    1. create zone if needed
2. list nameserver
3. for each provider
    1. get corrections

I'll open a PR for this fix.

tlimoncelli:
Dude, you're awesome!

das7pad:
[...]

PR #1642, opened 2022-07-28
Create zone before use

das7pad:
Closes https://github.com/StackExchange/dnscontrol/issues/1641

This PR is doubling down on the zone creation before using them for diffing.

The PR contains three fixes:
- a core fix as detailed in https://github.com/StackExchange/dnscontrol/issues/1641#issuecomment-1198426896
- two provider specific fixes for clearing a zone cache when creating a new zone (HETZNER and ROUTE53)

I've reviewed all the other providers with zone caching and they either clear it (dnsmadeeasy, gcloud) or do not support domain creation (ovh).

cc @tresni on the ROUTE53 changes


tlimoncelli (approved):

tresni:
Ran this last night on my test R53 domain, no issues that I saw.

tlimoncelli:
Thanks for fixing this!  Quite an interesting bug!

CC: @jpbede

jpbede (approved):
Interesting, but that definitely makes a lot more sense. Thanks for that!

Issue #1570, opened 2022-06-20
MAINT: Refactor auditrecords.go to do everything in one loop


tlimoncelli:
Each provider's auditrecords.go file loops over rc.Records once for each test.  That is, if there are 5 tests, it loops over the records 5 times.  This is a waste, since most of the records are not TXT records.  Also, there is a lot of repetition in the code.

Instead, auditrecords.go should take a list of checks to be performed. Common functions come from recordaudit, while tests that are specific to the provider can be in the auditrecord.go file itself:

```
var recordaudits = providers.RecordAudits{
	{"TXT", {
		recordaudit.TxtNoDoubleQuotes,
		recordaudit.TxtNoMultipleStrings,
		recordaudit.TxtNoTrailingSpace,
		localTXTCheck1,
	}},
	{"A", {
		localACheck1,
	}},
}

func localTXTCheck1(rc models.RecordConfig) error {
   // do a check that is specific to this provider
}

func localACheck1(rc models.RecordConfig) error {
   // do a check that is specific to this provider
}
```

The system would call each function on the matching record types in one big loop, thus reducing the overhead.

PR #1658, opened 2022-08-04
MAINT: Refactor auditrecords.go to loop only once #1570

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/1570

The old AuditRecord does a lot of unneeded work.  It loops through every record once for each test.  This change loops through all the records once, calling checking functions appropriate to that record type.  It also makes the format of the initializers readable.

I timed it on a small dataset and got a 1-3% improvement in total run time, I suspect that if I timed the function itself it would be more for providers that have many tests.


tlimoncelli:
@tresni @pragmaton @mikenz @Deraen @onlyhavecans @vojtad @papakai @svenpeter42 @tlimoncelli @kordianbruck  @costasd @mistererwin  @pgaskin

Hey folks!  I rewrote the record audit stuff to be significantly faster and less verbose.  I believe I've converted the old code properly.  Please take a look.

costasd (approved):
LGTM for NS1

MisterErwin:
LGTM for the rwth provider (although it is a blatant copy of the other providers)

tlimoncelli:
Flakey tests.  They passed other times.

PR #1661, opened 2022-08-07
Update TransIP CanUseAlias

cafferata:
You can use the ALIAS-record with TransIP since September 2021.
- https://www.transip.nl/nieuws/alias-records/


cafferata:
cc: @blackshadev because I see you have been designated as maintainer of TransIP ☺️

tlimoncelli (approved):

blackshadev:
Sorry for the late reply. But thanks  @cafferata !

cafferata:
There is no need to say sorry, everyone has their own agenda. ☺️

Issue #1667, opened 2022-08-09
NS1 cannot push since v3.17.0


norman-zon:
Starting with version 3.17 I cannot successfully push records via the NS1 provider:

```
WARNING: Error creating domain: PUT https://api.nsone.net/v1/zones/XXXX.YY: 400 invalid: FQDN already exists in the view
```

Running the same on 3.16.2 it works as excpected.

I skimmed through the[ commits between 3.16.2 and. 3.17.0](https://github.com/StackExchange/dnscontrol/compare/v3.16.2...v3.17.0) but nothing there struck me as an obvious reason.

It also persists with v3.18.0. Maybe @costasd could take a look here?

costasd:
Hi Norman, interesting. 

Didn't occur to me during testing. I'll see into it today. My guess is some bug with the create zone/get zone code that had some updates recently.

Just so that I can try and reproduce it, what you're doing is that you're trying to just push changes in an existing zone, right?

Thanks,
Costas

norman-zon:
Thanks for taking a look!



> Just so that I can try and reproduce it, what you're doing is that you're trying to just push changes in an existing zone, right?

Exactly. The zone exists and I add a CNAME record.

tlimoncelli:
Tip: This might be related to #1528.  It caused some caching issues for other providers, such as  a00572af4a2d5b12f3a8bf387dcce7a35cfea3ee.  HTH

costasd:
I managed to reproduce it easily and debug it.

So, the root cause is a bug with the ns1-go module we're using for all the low-level API communication. See https://github.com/ns1/ns1-go/issues/163 that I just opened.

The reason behind this showing up only now is that as of #1369/#1528 we're attempting to create zones every time dnscontrol runs - whereas earlier, this could only be triggered through issuing a `create-zone` operation in a zone that already exists, which is a more rare case:
```
$ ~/go/bin/dnscontrol create-domains 
***  example.com
  - ns1
Error creating domain: PUT https://api.nsone.net/v1/zones/example.com: 400 invalid: FQDN already exists in the view
$ ~/go/bin/dnscontrol version
dnscontrol v3.15.0-dev
```

So, I'll go ahead and prepare an ugly workaround for now to keep dnscontrol working, which I'll remove as soon as ns1-go approves my PR there.

tresni:
With #1671 merged should this be closed?

tlimoncelli:
[...]

PR #1671, opened 2022-08-09
ns1: workaround an issue with ns1 API

costasd:
NS1 api returns a different message these days, which breaks error handling.

Until a fix is merged upstream, work around the issue by attempting to match the
stray error ourselves.

relates to #1667.

integration & manual tests seem to pass.


tlimoncelli:
Thanks for the quick work-around!

Issue #1698, opened 2022-08-14
AZUREDNS: Migrate to newer library


tlimoncelli:
CC: @vatsalyagoel 

staticcheck reports:

providers/azuredns/azureDnsProvider.go:12:7: "github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2018-05-01/dns" is deprecated: Please note, this package has been deprecated. A replacement package is available [github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/dns/armdns](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/dns/armdns). We strongly encourage you to upgrade to continue receiving updates. See [Migration Guide](https://aka.ms/azsdk/golang/t2/migration) for guidance on upgrading. Refer to our [deprecation policy](https://azure.github.io/azure-sdk/policies_support.html) for more details. (SA1019)

vatsalyagoel:
Will have a PR soon.

vatsalyagoel:
This issue can be closed @tlimoncelli

PR #1701, opened 2022-08-14
Switch to the new Azure sdk for go

vatsalyagoel:
Fixing #1698 

Reverting back to using getExistingZones() properly fixing #792


tlimoncelli:
Wow!  That was fast!

Thanks for taking care of that!  The code looks a lot cleaner with the new SDK.

PR #1703, opened 2022-08-14
Revert jquery update to version supported by bootstrap

Firefishy:
In https://github.com/StackExchange/dnscontrol/pull/1637 jquery was updated to a version not supported by bootstrap version used.

This causes the docs to throw an exception and example expand elements not to work.


Firefishy:
Screenshot of the exception from https://stackexchange.github.io/dnscontrol/:

Screenshot 2022-08-15 at 02 30 12

tlimoncelli:
Thanks for figuring that out!

Issue #1720, opened 2022-08-16
New brief format. Should it be the default?


tlimoncelli:
There is a new output format that is much more brief.  Should it be the default (in which case a `--verbose` flag will get the old format), or should it require a flag (such as `--brief`)?

I currently achieve the "brief" format by piping the output to a filter.

Current format:

```
$ dnscontrol preview
******************** Domain: sstatic.net
----- Getting nameservers from: gcloud
----- Getting nameservers from: r53
----- DNS Provider: gcloud...
0 corrections
----- DNS Provider: r53...
0 corrections
----- Registrar: namedotcom_main...
0 corrections
******************** Domain: blogoverflow.com
----- Getting nameservers from: gcloud
----- Getting nameservers from: r53
----- DNS Provider: gcloud...
1 correction
#1: CREATE A testrecord.blogoverflow.com 1.2.3.4 ttl=300
----- Registrar: cscglobal_ltom215...
0 corrections
******************** Domain: example.com
----- Getting nameservers from: gcloud
----- Getting nameservers from: r53
----- DNS Provider: gcloud...
0 corrections
----- DNS Provider: r53...
0 corrections
----- Registrar: namedotcom_main...
0 corrections
******************** Domain: example2.com
----- Getting nameservers from: gcloud
----- Getting nameservers from: r53
----- DNS Provider: gcloud...
0 corrections
----- DNS Provider: r53...
0 corrections
----- Registrar: namedotcom_main...
0 corrections
Done. 0 Corrections.
```

Filtered format:

```
$ dnscontrol preview | bin/filter-preview.sh
******************** Domain: sstatic.net
******************** Domain: blogoverflow.com
1 correction
#1: CREATE A testrecord.blogoverflow.com 1.2.3.4 ttl=300
******************** Domain: example.com
******************** Domain: example2.com
Done. 0 Corrections.
```

In case you were wondering, here is the `filter-preview.sh` shell script:

```shell
#!/bin/bash

exec grep -v -e '^\.\.\.0 corrections$' \
   -e '^ *$' \
   -e '^0 corrections' \
   -e '\.\.\. \(skipping\)' \
   -e '^----- DNS Provider: ' \
   -e '^----- Registrar: ' \
   -e '^----- Getting nameservers from:'
```

When you have many, many domains, this makes it easier to visually
spot a change.  This is particularly useful in a GitOps-style or CI
system where a second person must review a change.

The question is:

[...]

PR #1721, opened 2022-08-18
Add --verbose flag, default to less verbose output

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/1720


tlimoncelli:
This is a lazy hack to fix https://github.com/StackExchange/dnscontrol/issues/1720. It will probably be replaced when https://github.com/StackExchange/dnscontrol/pull/1574 arrives.

Hey, @jpbede!  This overlaps some of the work you're doing.  I hope you don't mind.

jpbede:
No problem, I'm fine with that 🙂

PR #1723, opened 2022-08-19
DESEC: Check authentication via domains API

glueckself:
Hi all!

I want to use restricted tokens to manage the RRSets of a single domain from a specific server (i.e., this server shouldn't be able to modify other domains in my account). I've noted that while I can use `curl` to modify the domain, dnscontrol failed with `authentication failure`.

According to the [official documentation](https://desec.readthedocs.io/en/latest/auth/tokens.html#token-scoping-domain-policies): "Tokens with at least one policy are considered restricted, with their scope explicitly limited to DNS record management. They can perform neither [Retrieve Account Information](https://desec.readthedocs.io/en/latest/auth/account.html#retrieve-account-information) nor [Domain Management](https://desec.readthedocs.io/en/latest/dns/domains.html#domain-management) (such as domain creation or deletion)."

Thus my change to use `/domains`, as this enables to check the validity of restricted tokens too.

I do not have a Go development environment set up, so I've only tested that with `curl --header "Authorization: Token my-restricted-token" https://desec.io/api/v1/domains` I get return code 0 and the domain list. A invalid token returns HTTP 401 and `{"detail":"Invalid token."}`

Please let me know if should setup the development environment and test dnscontrol itself.

Thanks!

### DIFF
--- a/providers/desec/protocol.go
+++ b/providers/desec/protocol.go
@@ -67,7 +67,7 @@ type nonFieldError struct {
 }
 
 func (c *desecProvider) authenticate() error {
-	endpoint := "/auth/account/"
+	endpoint := "/domains/"
 	var _, _, err = c.get(endpoint, "GET")
 	if err != nil {
 		return err
--- a/providers/desec/protocol.go
+++ b/providers/desec/protocol.go
@@ -67,8 +67,13 @@ type nonFieldError struct {
 }
 
 func (c *desecProvider) authenticate() error {
-	endpoint := "/domains/"
-	var _, _, err = c.get(endpoint, "GET")
+	endpoint := "/auth/account/"
+	var _, resp, err = c.get(endpoint, "GET")
+	//restricted tokens are valid, but get 403 on /auth/account
+	//invalid tokens get 401
+	if resp.StatusCode == 403 {
+		return nil
+	}
 	if err != nil {
 		return err
 	}


tlimoncelli:
@D3luxee please review

D3luxee:
I've used auth/account as suggested in https://github.com/StackExchange/dnscontrol/issues/1177 
@peterthomassen do you mind giving some advise here as a member of the desec team?

peterthomassen:
Great to see scoped API tokens in use! :tada: @nils-wisiol

That functionality was only deployed a few months ago. Only for tokens which are explicitly configured with restrictions by the user, the `auth/account/` endpoint is not accessible. In other words, for tokens that don't use the new optional settings, everything still works fine.

If you do `GET auth/account/`, the response will be 401 if the token is invalid, and 403 if the token is valid but restricted. I think the solution is thus to treat 403 as "success" for the purpose of validating a restricted token.

I would advise against using `GET domains/` because that can lead to a 4xx reponse for other reasons (Pagination required, see #1177, as pointed out bey @D3luxee).

glueckself:
Thanks for the feedback!

I'll try to implement checking the response code this weekend.

glueckself:
One more Go first timer starting with your project :)

The fix looks simple. I've looked at some other code where response code is checked and used that as a template.

I've tested it manually with my domain:

* non-restricted tokens:
  - authentication works
  - can access any domain
* restricted tokens:
  - authentication works
  - can access domains with permissions
  - fails on restricted domains with `Error getting corrections: failed fetching records for domain sap.sml.name (deSEC): You do not have permission to perform this action.`
* invalid token:
  - authentication fails

I had a short look at the integration testing, but it looks like you need to add a second domain at deSEC* and (restricted) token to the secrets first.

Please let me know if I missed anything or if you have any feedback. Thanks!

\* You can use a sub-domain there, no need to register a second domain. I.e. both `example.com` and `dev.example.com` can be managed at deSEC, with NS and DS records in `example.com` pointing to/having keys of `dev.example.com`.

D3luxee:
looks good to me

@tlimoncelli feel free to merge

tlimoncelli:
Thanks for working together on this!  I'm sure a lot of DESEC users will appreciate being able to use this!

PR #1749, opened 2022-09-14
chore(DigitalOcean): Support semicolon as CAA target

tresni:
This appears to be a change as it is supported in the interface and in the API. Verified with my own personal domains (e.g. brianandjenny.net)


tlimoncelli:
Thanks for the update!

PR #1753, opened 2022-09-20
Output check "No errors" message to stdout

Firefishy:
Output the check "No errors" information message to stdout.

Check errors are output to Stderr, instead outputting the information message to stdout allows scripted redirect of stdout to avoid any logging "errors".


tomhughes:
Basically we added a `dnscontrol check` to our deployment scripts, to run a check before making any changes, and as result we started getting email every five minutes that just said `No errors.` as a result because the script runs from cron on a regular basis.

That script already sends stdout to `/dev/null` but for obvious reasons we want to get actual errors so we don't do that for stderr.

Firefishy:
Having `dnscontrol check` output to stdout (instead of stderr) allows the `check` to be run via cron or a CI job which throws away stdout if no error.

Without this patch, the following script would cause cron to send an email on every run. With this patch no email is triggered.
```
#!/bin/sh
dnscontrol check >/dev/null
```

An alternative to this flag would be a `--quiet` option which would silence information message like `No errors`

tlimoncelli:
Ah, yes, we've all been in that "email every 5 minutes" situation.

Why not send stderr to /dev/null too?  i.e.  Instead of `>/dev/null` use `>/dev/null 2>&1`  (Cite https://unix.stackexchange.com/questions/100722/how-do-i-completely-silence-a-cronjob-to-dev-null)

Tom

P.S.  By the way... I strongly recommend using `dnscontrol check` in something that triggers once per change (like a CI system or git hooks), not every 5 minutes (like cron).

tomhughes:
As I said we don't redirect stderr because we want actual errors - the cron thing is sort of accidental as the code in question runs both on CI pushes and also from cron to deal with automatically updating things in reponse to statuscake monitoring but it's true that side of things doesn't really need to run the check it will just need some refectoring.

tlimoncelli:
Ah, yes. Sorry for asking you to repeat yourself.

My concern is that with `check`, errors and the "No errors." message will go to different places.  It would make more sense if "check" sent both to stdout.  (At the same time, we should remove the `--out` or `--pretty` flags (which are there just because `check` is really `print-ir`))

tlimoncelli:
If we're going to do this, we should do it all the way.  Everything should go to stdout when "check" is in use, and stderr when "print-ir" is in use.

Please take a look at https://github.com/StackExchange/dnscontrol/pull/1755

tlimoncelli:
Fixed in #1755

PR #1755, opened 2022-09-21
FEATURE: "check" subcommand should send to stdout

tlimoncelli:
As suggested in  https://github.com/StackExchange/dnscontrol/pull/1753,  the `check` subcommand should send output to stdout.

* Output any warnings or errors to stdout
* Remove flags that aren't relevant (`--pretty` and `--out`)

To verify that this works, I made some intentional typos and verified that the output is going to stdout by prefixing stdout with "STDOUT:".  Anything to stderr should NOT have have the prefix.

No errors: 

```
$ dnscontrol check | awk '{ print "STDOUT: " $0 }'
STDOUT: No errors.
```

Errors in javascript:

```
$ dnscontrol check | awk '{ print "STDOUT: " $0 }'
STDOUT: executing dnsconfig.js: (anonymous): Line 91:1 Unexpected token , (and 1 more errors)
```

Errors in validation:

```
$ dnscontrol check | awk '{ print "STDOUT: " $0 }'
STDOUT: 2022/09/21 18:13:05 printIR.go:107: 1 Validation errors:
STDOUT: 2022/09/21 18:13:05 printIR.go:113: ERROR: in A @.stackoverflow.co.il: WARNING: target (sfludlfj) is not an IPv4 address
STDOUT: exiting due to validation errors

```


tlimoncelli:
CC @Firefishy  @tomhughes

Firefishy:
My preference would be for any errors to go to stderr. ;-) But all to stdout is workable.

tlimoncelli:
Maybe I'm over-thinking this but I think this is the right thing to do.

PR #1758, opened 2022-09-25
DOCS: Fix dns.he.net link

Firefishy:
Fix the markdown for the dns.he.net link.


tlimoncelli:
Thanks for catching that!

PR #1759, opened 2022-09-28
DOCS: Add a dynamic macro example (Fastmail)

tlimoncelli:


docs/examples.md
```diff
@@ -143,3 +143,42 @@ DEFAULTS(
...
+  ]
+}
+```
+
+We can then use the macros as such:
+
+```
+D("example.com", REG_NONE, DnsProvider(DSP_R53_MAIN),
+    FASTMAIL_MX,
+    FASTMAIL_ONLY_SPF,
```
cafferata, line 181: Perhaps it is better to remove the `FASTMAIL_SPF_ONLY` variable in the example, or include the contents of the `FASTMAIL_SPF_ONLY` variable as an example so that the developer experience (copy/paste) is better.

tlimoncelli: good point. what do you ise for FASTMAIL_SPF_ONLY?

cafferata: I don't use Fastmail. I also have no idea what Jay's ideas are. If you ask me I would remove the full SPF variable (as you already did). 👍

PR #1768, opened 2022-10-07
VULTR: use API v2

androw:
Vultr v1 API has been deprecated and govultr has been upgraded to v2.
I've adapted VULTR code to use v2 API instead of v1.

I've tried to test on my side but if someone can confirm changes works that be awesome.

### DIFF
--- a/providers/vultr/vultrProvider.go
+++ b/providers/vultr/vultrProvider.go
@@ -5,13 +5,14 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strconv"
 	"strings"
+
+	"golang.org/x/oauth2"
 
 	"github.com/StackExchange/dnscontrol/v3/models"
 	"github.com/StackExchange/dnscontrol/v3/pkg/diff"
 	"github.com/StackExchange/dnscontrol/v3/providers"
-	"github.com/vultr/govultr"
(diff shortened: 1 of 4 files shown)


providers/vultr/vultrProvider.go
```diff
@@ -62,27 +63,44 @@ func NewProvider(m map[string]string, metadata json.RawMessage) (providers.DNSSe
...
 // GetZoneRecords gets the records of a zone and returns them in RecordConfig format.
 func (api *vultrProvider) GetZoneRecords(domain string) (models.Records, error) {
-	records, err := api.client.DNSRecord.List(context.Background(), domain)
-	if err != nil {
-		return nil, err
-	}
+	listOptions := &govultr.ListOptions{}
+	records, meta, err := api.client.DomainRecord.List(context.Background(), domain, listOptions)
+	curRecords := make(models.Records, meta.Total)
+	next_i := 0
```
tlimoncelli, line 80: staticcheck reports:

```
providers/vultr/vultrProvider.go:80:2: should not use underscores in Go names; var next_i should be nextI (ST1003)
providers/vultr/vultrProvider.go:86:3: should not use underscores in Go names; var current_i should be currentI (ST1003)
```

androw: Fixed

PR #1784, opened 2022-10-29
DOCS: Inline creds.json.tpl-example.txt example

cafferata:
I have deleted the sample txt file. The old method results in unnecessary extra actions. The file needs to be downloaded, moved to the developer's project/IDE, the filename needs to be renamed. All in all, too many actions. I believe that our developer colleagues are perfectly capable of copying/pasting a code sample.

**Before**
before

**After**
after


tlimoncelli:
How about including both?  I prefer to have a clean, downloadable, version that I can link to.  Maybe we can avoid the rename somehow?

cafferata:
Where do you draw the line for downloadable code samples? The documentation is full of code examples that cannot be downloaded and must be copied/pasted manually.

tlimoncelli:
Good question!  I make an exception here because I want new users to have a smooth experience getting started.  In my mind, having to cut-and-paste leads to a sloppy start (for lack of a better word).  If people start with files that are sloppy, its difficult to fix.  Start out with a clean file, there is hope they'll maintain it that way.

Maybe a tar file with a "starter config" would be better?

cafferata:
Personally I don't like downloading (archive) files. I would rather think of an example (git) project with all the files in it. The advantage is that it is under version control. It is also possible to provide separate README/Getting started steps in this project.

For now I would suggest narrowing the scope of the PR to pulling the JSON example [creds.json.tpl-example.txt](https://github.com/StackExchange/dnscontrol/blob/master/docs/assets/creds.json.tpl-example.txt) inline as you suggested. 👍

tlimoncelli:
Ah, the root of the problem is that when I called the file `creds.json` browsers made downloading the file difficult, so I renamed it with a `.txt` extension to fool github into making it easy to download.  However now the user had to rename the file.

Instead, we should use a url to the blob (for example https://github.com/StackExchange/dnscontrol/blob/master/docs/assets/creds.json-example.txt) and then the person gets the github experience where the file is displayed, but they can click "raw" for the raw file.

Therefore... my suggestion is:
1. Include the file inline.
2. ALSO include a link to `https://github.com/StackExchange/dnscontrol/blob/master/docs/assets/creds.json` AND rename the file to creds.json
3. Same for dnsconfig.js

Tom

cafferata:
I renamed the sample files and applied deep links to the Github assets for:

1. https://github.com/cafferata/dnscontrol/blob/docs-inline-dnsconfigjs-example/docs/assets/1password/creds.json
2. https://github.com/cafferata/dnscontrol/blob/docs-inline-dnsconfigjs-example/docs/assets/getting-started/creds.json
3. https://github.com/cafferata/dnscontrol/blob/docs-inline-dnsconfigjs-example/docs/assets/getting-started/dnsconfig.js

tlimoncelli:
Thanks! it's great that we found a solution that satisfies both your criteria and mine.

PR #1793, opened 2022-10-30
DOCS: Missing record modifiers

cafferata:
I noticed that the information is fragmented, there is not a good overview of all Javascript DSL. With this PR I want to add the two missing record modifiers `CAA_BUILDER()` and `SPF_BUILDER()` to the Javascript DSL overview.


tlimoncelli:
Ah!  Good catch!  Thanks for cleaning those up!

PR #1806, opened 2022-11-04
[DNSimple] Update name server handling

onlyhavecans:
DNSimple has added new name server suffixes and changed the default name server set for new domains.

In order to support this this adds new logic for the new domains.

It also changes the default set by the provider, this will be a change for any domains still using the legacy ns4.


tlimoncelli:
Thanks!

PR #1808, opened 2022-11-06
Add rTypes restrictions to IGNORE_NAME

### LINKED ISSUE #1757: Ignore A record only for dynamic DNS while updating others
A sub-domain `foo.example.com` has an A record (IPv4 address) set by a _dynamic DNS_ client, static IPv6 (AAAA) managed by dnscontrol and TXT/SPF record also managed by dnscontrol.

Is there a way to tell dnscontrol to not touch the A record but update only the AAAA and TXT records? IGNORE_NAME and IGNORE_TARGET do not seem to do the trick.

bits01:
Adds record type restrictions to IGNORE_NAME. Useful when only specific records are controlled externally, e.g. a dynamic DNS client manages the `A` record but the rest of the records are managed by `dnscontrol`.

```js
D("example.com",
  IGNORE_NAME("foo"), // ignore all record types for name foo
  IGNORE_NAME("baz", "*"), // ignore all record types for name baz
  IGNORE_NAME("bar", "A,MX"), // ignore only A and MX records for name bar
  CNAME("bar", "www"), // CNAME is not ignored
  A("baz", "1.2.3.4")
);
```

Fixes #1757

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -2,6 +2,7 @@ package diff
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 
 	"github.com/StackExchange/dnscontrol/v3/models"
@@ -42,11 +43,18 @@ func New(dc *models.DomainConfig, extraValues ...func(*models.RecordConfig) map[
 	}
 }
 
+// An ignoredName must match both the name glob and one of the recordTypes in rTypes. If rTypes is empty, any
+// record type will match.
+type ignoredName struct {
+	nameGlob glob.Glob
+	rTypes   []string
+}
+
 type differ struct {
 	dc          *models.DomainConfig
 	extraValues []func(*models.RecordConfig) map[string]string
 
-	compiledIgnoredNames   []glob.Glob
+	compiledIgnoredNames   []ignoredName
 	compiledIgnoredTargets []glob.Glob
 }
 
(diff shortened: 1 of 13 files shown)


bits01:
@tlimoncelli 🙏 when you have a chance, could you please take a look at the build, it fails with some `Unauthorized` errors

tlimoncelli:
Hi!

Wow!  Thanks for working on this issue!

Github is very paranoid about letting non-admins have access to the API keys. Therefore, you'll always get your jobs marked as "Unauthorized".  I have to push it to a branch that I own to get all the tests to run. (this is super annoying but something I hope we can fix some day)

Here's my branch: https://github.com/StackExchange/dnscontrol/pull/1811

pkg/diff/diff.go
```diff
@@ -345,16 +353,22 @@ func sortedKeys(m map[string]*models.RecordConfig) []string {
 	return s
 }
 
-func compileIgnoredNames(ignoredNames []string) []glob.Glob {
-	result := make([]glob.Glob, 0, len(ignoredNames))
+func compileIgnoredNames(ignoredNames []*models.IgnoreName) []ignoredName {
+	rTypesRegexp := regexp.MustCompile(`\s*,\s*`)
```
tlimoncelli, line 357: Is this `regexp.MustCompile()` going to run multiple times? It should probably be a global variable so that it runs once. That's what we did in `pkg/transform/ptr.go` and should probably do elsewhere too.

tlimoncelli (commented):
That's some excellent coding!  I have one minor suggestion otherwise this looks ready to merge.  (anything else you'd like to do before the merge?)

pkg/diff/diff.go
bits01: I will do that no problem, though in practice I doubt it has any real perf implications, most of the time will be spent on the network interacting with the provider APIs.

bits01:
@tlimoncelli Done. If you could please do the build magic one more time. Thank you.

tlimoncelli:
See https://github.com/StackExchange/dnscontrol/pull/1813

tlimoncelli:
Merging this.  

@timkgh please give this a try and let us know if it meets your needs

timkgh:
> Merging this.
> 
> @timkgh please give this a try and let us know if it meets your needs

Thank you!! Waiting for the next release.

PR #1811, opened 2022-11-07
Bits01 ignore name rtypes

tlimoncelli:


bits01:
🎉 Looks like the build passed. Thanks for your quick help.

PR #1821, opened 2022-11-25
RWTH: Fix expected key in documentation

nemunaire:
Hi,

While testing the RWTH provider, I found a mistake in the documentation.

The provider takes `api_token` and not an `api_key`: https://github.com/StackExchange/dnscontrol/blob/master/providers/rwth/rwthProvider.go#L44


tlimoncelli:
Good catch!  Thanks for the pR!

PR #1822, opened 2022-11-26
VULTR: Fix setting RecordConfig Original

androw:
RecordConfig.Original was set with a pointer that was being reused.
This caused the wrong record to be modified/removed in some cases.


tlimoncelli:
Thanks for finding and fixing this!

PR #1824, opened 2022-11-29
Improve layout and display of the feature matrix table

j-f1:
- Expands the table to take up the full horizontal width of the screen
- Allows the table to scroll horizontally independently from the main text on the page
- Pins the header column to the left edge of the screen so it remains visible even as you scroll to the right, making the table easier to parse
- For a bit of visual polish, extends the white background behind the fixed header columns to the first column and blocks the provider names as the table column corresponding to each one scrolls out of view:
Screenshot_2022-11-28 22 09 50


tlimoncelli (approved):

tlimoncelli:
Thanks!

PR #1826, opened 2022-11-29
Fix the title of the Namecheap docs, remove false statement that there is no DNS provider implementation

j-f1:


tlimoncelli:
Good catch!  Nobody has noticed this since it was written in 2017!

PR #1827, opened 2022-11-29
Remove Cannot() records for Azure and R53 ALIAS records

j-f1:
It was confusing to me that most services had a “-” in the table for `R53_ALIAS` and `AZURE_ALIAS` while some had an “X.” I’ve removed the `Cannot()` calls for the corresponding capabilities from the providers that had them since it will never make sense for another provider to support these record types.


tlimoncelli:
I agree that R53_ALIAS and AZURE_ALIAS make the chart look ugly.  However they should be listed somewhere.  Maybe a bullet list of provider-specific types?

As far as `URL`... different providers implement that kind of thing differently.  We probably should have called that `NAMECHEAP_URL`.  My preference is to have `GENERIC_REDIRECT()` that supports a lowest-common-denominator redirect that "does the right thing" for each provider. (maybe flags for 301 vs 301 redirects but nothing else).

PR #1809, opened 2022-11-06
DOCS: Migrated documentation website to Gitbook

cafferata:
## Summary

Hi @tlimoncelli! After many documentation pull requests, and setting up more and more implementations at @jcid customers, I still have the desire to set up a better documentation website.

I miss features like:
- being able to search through the available content
- a good navigation structure
- a good mobile responsive website

I'm having trouble with the magic of the static site generator Jekyll. Raw Markdown is easier to read from the repository files. That's why I did some research and quickly came across https://www.gitbook.com. I wanted to set up a proof of concept and present it to you.

I am proud to present a new documentation website!
https://docs.dnscontrol.org 🚀 

docs dnscontrol zone

_Disclaimer, it's a proof-of-concept and there's still some work to be done (_see TODO_). Nevertheless, I would like to hear whether we as a community want to go in this direction._

## The big advantages of gitbook.io
- A out of the box a [search engine](https://docs.gitbook.com/tour/quick-find).
- On every page an overview 'on this page' with deep links to the correct sections.
- A preview website is set up for every Github pull request. This gives you better insight into which changes are made and a developer does not have to work with local tools.
- All content can be rated, which provides better insight into the quality of the documentation.
- You will see the link 'Edit on GitHub' on every page so that you can make even faster/simple errors and/or improvements.

## Cost overview

- We can be brief about that, gitbook.io is free for open source.
- ~~The domain 'dnscontrol.zone' is sponsored by @jcid.~~
    - _Is no longer applies due to the domain migration from `dnscontrol.zone` to `dnscontrol.org`._

By introducing the custom domain, examples can be applied to this domain. For example, the DNS configuration is included in this git repository. The perfect example of 'eating your own dog food'.

## TODO

- [x] Prepare a Gitbook account for @tlimoncelli.
- [x] Better first impression of the homepage. [See issue comment 1325623247](https://github.com/StackExchange/dnscontrol/pull/1809#issuecomment-1325623247).
- [x] Convert the latest Jekyll content to raw markdown/alternate renders.
    - [x] Inline includes.
    - [x] The [Service provider matrix](https://stackexchange.github.io/dnscontrol/provider-list#features).
        - [POC table 1](https://docs.dnscontrol.org/service-providers/provider-list#poc-table-1)
        - [POC table 2](https://docs.dnscontrol.org/service-providers/provider-list#poc-table-2)
    - [ ] The [SPF flattener](https://stackexchange.github.io/dnscontrol/flattener/index.html).
[...]

PR #1830, opened 2022-11-29
Generate TypeScript typings for functions supplied to dnsconfig.js

j-f1:
A `.d.ts` file with full documentation is now generated from existing docs files (with a few manual additions). If placed next to a `dnsconfig.js` file, supporting editors (VS Code but also others as well) will have a significantly improved DX.

I’m marking this as WIP because I need to do a bunch more, but I’m very interested in feedback!
- [x] find a better front matter parsing library (the current one behaves strangely when `---` appears later on in the document)
- [x] manually review the generated typings for inaccuracies when compared with the prose docs
- [x] Fix typings for `FETCH` (currently it uses `fetch`’s typings which are only available to TS code that enables the DOM library)
- [x] come up with a way to ship the generated file
  - one option would be to bundle it with the program (ensuring that it always provides the correct typings that correspond to the available functionality) via a new command that spits out a `dnscontrol.d.ts` file into the current directory
  - another would be to host the files somewhere (maybe check the generated `types.d.ts` file into Git so that it can be fetched from the current release’s tag?)
  - maybe add a way to generate a suitable `jsconfig.json` to configure TypeScript in editors? 
- [x] Figure out if there’s a way to make `CLI_DEFAULTS` cause the names it’s passed to show up as variables **(Result: Probably not)**
- [x] Docs: How to add new record types and get correct TS typings
- [x] Docs: How to use TS with DNSControl

Documentation is shown inline:
Screenshot_2022-11-29 15 37 36

Autocomplete shows all available record types:
Screenshot_2022-11-29 15 39 40


tlimoncelli (commented):
Holy cow!  I didn't even know that kind of thing was possible!

This is awesome!  Yes, please and thank you!

j-f1:
Some questions (actually 1 question) based on my reading through the docs:

- For [`SSHFP`](https://dnscontrol.org/js#SSHFP), arę the reserved “0” algorithm/type supported as values? Or should it just be 1/2/3 allowed as parameters? (also, it would be a neat addition to support the string values like `"DSA"` too)

tlimoncelli:
> Some questions (actually 1 question) based on my reading through the docs:
> 
> * For [`SSHFP`](https://dnscontrol.org/js#SSHFP), arę the reserved “0” algorithm/type supported as values? Or should it just be 1/2/3 allowed as parameters? (also, it would be a neat addition to support the string values like `"DSA"` too)

According to [the source of all knowledge](https://en.wikipedia.org/wiki/SSHFP_record#) 0 is reserved. I guess we could play it safe and permit it.  Though, the world won't end if we guess wrong as I don't think many people use SSHFP ;-)

Tom

j-f1:
Ok, I think this is ready for review! I’ve added a `dnscontrol types` command that downloads the version of the `.d.ts` file corresponding to the current build of `dnscontrol` from GitHub. I definitely want to add some more documentation but before I do that I want to make sure this is a good approach. (I’d also love any feedback on the code since this is my first time writing Go!)

commands/generateTypes.go
```diff
@@ -0,0 +1,73 @@
...
+	"strings"
+
+	versionInfo "github.com/StackExchange/dnscontrol/v3/pkg/version"
+	"github.com/urfave/cli/v2"
+)
+
+var _ = cmd(catUtils, func() *cli.Command {
+	var args TypesArgs
+	return &cli.Command{
+		Name:  "types",
```
tlimoncelli, line 16: I'm not in love with having the subcommand called "types". However if that's what other similar systems use, I'd rather be consistent with them.  Could the name refer to typescript (or ts) to be clear what it is for?

tlimoncelli (commented):
I only did a shallow review of the Go code. I'll do a deeper look soon.

Two more notes:
1) Documentation: Add docs about how to use, etc.
2) Release engineering: Document what needs to be done to generate the file and any other release engineering steps (i.e. is there a "go generate" or equivalent that needs to run?)

build/build.go
```diff
@@ -16,7 +16,7 @@ var goos = flag.String("os", "", "OS to build (linux, windows, or darwin) Defaul
 
 func main() {
 	flag.Parse()
-	flags := fmt.Sprintf(`-s -w -X main.SHA="%s" -X main.BuildTime=%d`, getVersion(), time.Now().Unix())
+	flags := fmt.Sprintf(`-s -w -X "main.SHA=%s" -X main.BuildTime=%d`, getVersion(), time.Now().Unix())
```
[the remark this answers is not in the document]
j-f1: Yes. Previously, the SHA ended up being like `"abc123"` instead of `abc123`. (not sure why, it’s weird that the quotes aren’t properly handled by the shell)

commands/generateTypes.go
j-f1: The key in the JS `package.json` is `types` (which I kind of based it on). I kind of shy away from using `typescript` since I want it to be clear that this doesn’t actually *run* TypeScript, it just emits the file. Maybe `dnscontrol export-ts-types` or `emit-ts-types` or something?

j-f1:
- docs: added to TODO above
- release engineering: added! you just need to do `go generate` (which _should_ be run periodically anyway to update the matrix)

docs/release-engineering.md
```diff
@@ -12,10 +12,23 @@ CircleCI will do most of the work for you. You will need to edit the draft relea
 Please change the version number as appropriate.  Substitute (for example)
 `3.20.0` any place you see `VERSION` in this doc.
 
-## Step 1. Tag the commit in master that you want to release
+## Step 1. Rebuild generated files
 
 ```shell
 git checkout master
+go generate
```
cafferata, line 19: I make a note to myself to do one last `git pull`. Maybe you also think this is a good idea and can you include it as a necessary step?

```suggestion
git pull
go generate
```

build/generate/functionTypes.go
```diff
@@ -0,0 +1,232 @@
...
+			frontMatter, body, err := parseFrontMatter(string(content))
+			if err != nil {
+				return "", err
+			}
+			if frontMatter["ts_ignore"] == true {
+				continue
+			}
+
+			body = body + "\n"
+			body = strings.ReplaceAll(body, "{{site.github.url}}", "https://www.dnscontrol.org/")
```
cafferata, line 101: After @tlimoncelli has undone the Github custom domain configuration https://github.com/StackExchange/dnscontrol/issues/1843#issuecomment-1337828223, these (links) no longer work.

docs/_functions/domain/SRV.md
```diff
@@ -7,6 +7,9 @@ parameters:
   - port
   - target
   - modifiers...
+parameter_types:
+  weight: number
+  port: number
```
cafferata, line 12: Let's face it, I'm not familiar with TypeScript typings (configurations). So I was wondering, when do you include which `parameter_types`? I feel like I'm missing `priority` here?

```suggestion
parameter_types:
  priority: number
  weight: number
  port: number
```

j-f1: I set a bunch of defaults here: https://github.com/StackExchange/dnscontrol/pull/1830/files#diff-76e760c4509933f6592ab7a1e50b4934736913b3cdfd91c00e1232ff33a81e82R52-R62. That said, it might make sense to get rid of that and explicitly specify all the parameter types for every function. Thoughts?

cafferata: Aha! I think I personally would opt for the duplication;
1. Everything in one place
2. No discussion when something is or isn't 'defaults'.
3. No need to look back on a PR if the missing param happens to be defined as a default.
4. Defaults could be left behind during refactoring activities.

cafferata:
In response to commit 'Remove default parameter types' https://github.com/StackExchange/dnscontrol/pull/1830/commits/7d6de8b9d5d408252d3f90735eb75bebfa40a8c4, I now see the knowledge of a function being recorded in three places. Perhaps it is easier to talk in an example situation. 

Let's take [SPF_BUILDER](https://stackexchange.github.io/dnscontrol/js#SPF_BUILDER) as an example. The knowledge about the `parameters` is currently recorded in:
1. `parameters` which are shown in the header.

    dnscontrol-function-header

2. `parameter_types` which is not shown, is only used for `TypeScript typings`.

    dnscontrol-text-documentation

3. `The parameters` section, which is described textually in the documentation page.

I don't want to make your pull request unnecessarily bigger. I did want to express my concerns about this error-proneness that this can create.

I'm leaning towards a solution where only the new `parameter_types` is used. My documentation pull request https://github.com/StackExchange/dnscontrol/pull/1809 can then use this. In which the header display (_screenshot 1_) is removed and more work is done towards an elaboration as can be seen in screenshot 2. With possibly a textual addition as is now done within `SPF_BUILDER`.

What do you guys think @j-f1, @tlimoncelli?

tlimoncelli:
Some feedback:

1. This is awesome. I'm no VS Code expert but I typed "dnscontrol types" and then started VS Code and suddenly typing `D(` autocompleted! Wow!
2. We need to document this really well.  Assume someone with no VS Code (or other) experience wants to use this. How do they get started?  What do they need to do to update this (if anything)?
3. I don't like the fact that having `dnsconfig.js` and `dnscontrol.d.ts` in the same directory means bash autocompletion stops at `dnscon` when it used to go all the way to `dnsconfig.js`.  (is there an emoji for "crying like a baby"?)

Looking forward to seeing this in production!

tlimoncelli:
Cool cool.

The points in (2) should be put in the docs somewhere.  Please add a new page in `$git/dnscontrol/docs` and link to it on the first page.

Point (3): I like `types-dnscontrol.d.ts`.  Does this file need to be in the same directory as `dnsconfig.js`?

j-f1:
> Does this file need to be in the same directory as `dnsconfig.js`?

Yes and no. It can’t be outside of the directory `dnsconfig.js` is in, but it can be in a subdirectory.

commands/generateTypes.go
tlimoncelli: How about `write-types`?  Its short and easy(ish) to remember.

docs/typescript.md
```diff
@@ -0,0 +1,27 @@
+---
+layout: default
+title: Using TypeScript with DNSControl
+---
+
```
tlimoncelli, line 5: Please add a note that: This is is an experimental feature and may change.

tlimoncelli (commented):
Just a few small comments and I think we're ready to go.  I'm asking that we mark this feature as "experimental" so that we can be empowered to iterate on it as we get feedback.

tlimoncelli:
Great!  Thanks!

tlimoncelli:
I'm merging this as-is.   In the release notes I'll point out that it is experimental and ask people to submit issues if they have suggestions and/or improvements.  Once the comments die down we'll make it non-experimental.

I'm super excited about this change. I can't wait to show it to my coworkers!

docs/_functions/global/FETCH.md
```diff
@@ -3,6 +3,8 @@ name: FETCH
 parameters:
   - url
   - args
+ts_ignore: true
+# Make sure to update fetch.d.ts if changing the docs below!
```
cafferata, line 7: @j-f1 I've made some changes from https://github.com/StackExchange/dnscontrol/pull/1809, how do I generate the new `fetch.d.ts? `go generate` only generates `types/dnscontrol.d.ts`.

j-f1: Copy+paste, unfortunately.

cafferata: Perhaps it is a good idea to ensure this necessary step in the docs? https://docs.dnscontrol.org/release/release-engineering#step-1.-rebuild-generated-files

j-f1: What do you mean? Unless someone changes the `FETCH` function’s interface without updating the docs (which is unlikely), they’ll see this message.

cafferata: Thanks for your comment. I'll further explain my comment with a code example. From https://github.com/StackExchange/dnscontrol/commit/55889582aeddc92d424a50f4577be030dfd04f49#diff-bda8825028c14bd7feb8eb5463e708023019b3ff7f710af883e4b63eba0b9ec3R12 I had to change the URLs. By copy/paste `global/FETCH.md` it's generates this diff which (*tested with JetBrains IDE*) is not valid.

```diff
- Compared to `fetch` from Fetch API, `FETCH` will call [PANIC](https://dnscontrol.org/js#PANIC) to terminate the execution of the script, and therefore DNSControl, if a network error occurs.
+ Compared to `fetch` from Fetch API, `FETCH` will call [PANIC](PANIC.md) to terminate the execution of the script, and therefore DNSControl, if a network error occurs.
```

This should be (*I guess*) a full URL reference?

```diff
- Compared to `fetch` from Fetch API, `FETCH` will call [PANIC](https://dnscontrol.org/js#PANIC) to terminate the execution of the script, and therefore DNSControl, if a network error occurs.
+ Compared to `fetch` from Fetch API, `FETCH` will call [PANIC](https://docs.dnscontrol.org/language-reference/top-level-functions/panic) to terminate the execution of the script, and therefore DNSControl, if a network error occurs.
```

I suspect you corrected this URL once? So the previous answer 'copy/paste' doesn't seem to be entirely valid. Or maybe I'm overlooking something?

j-f1: Ah yeah, great point. I must have changed those URLs, then. I’m not convinced it’s worth the effort to thoroughly document this since you’re likely to be the last person to touch the file (and especially the links) for a long time.

PR #1838, opened 2022-12-02
BUGFIX: PrettySort helper doesn't sort

tlimoncelli:
PrettySort doesn't call sort!  Luckily the only code that calls this is the "get-zones" subcommand, and nobody noticed that the output isn't sorted.

By making this change, people using "get-zones" to make their initial dnsconfig.js file will find that their draft D() code is a little prettier.

### DIFF
--- a/pkg/prettyzone/prettyzone.go
+++ b/pkg/prettyzone/prettyzone.go
@@ -90,6 +90,7 @@ func PrettySort(records models.Records, origin string, defaultTTL uint32, commen
 	}
 	z.Records = nil
 	z.Records = append(z.Records, records...)
+	sort.Sort(z)
 	return z
 }
 


gotjoshua:
> nobody noticed that the output isn't sorted.

actually, i am rerunning get-zones and noticing that the results are sorted differently each time

```
$ dnscontrol version
DNSControl version dev
```
installed via [nix latest](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=+dnscontrol).

is this expected?

tlimoncelli:
It isn't guaranteed to be the same every time but I'm surprised it isn't.  Would you please open a new bugid with a repro example?

PR #1850, opened 2022-12-09
REFACTOR: Rewrite diff'ing mechanism

tlimoncelli:
The DNSControl diffing module is being rewritten to be cleaner, more functional, and easier to use.

The old module is `pkg/diff`, the new module is `pkg/diff2`.  Both will exist for a while.

The code is hidden behind a flag.  By default the old code is used.  If you add the `--diff2` flag (or the `-diff2` flag when running integration tests), you get the new code.

The following modules have been updated: BIND, GANDI_V5, AZURE_DNS (still buggy).

Goals:

1. Less brittle INGORE() (and similar) functionality. Currently the IGNORE*() functions are difficult to test and therefore I can't really trust them fully.
2. Easier to use.  Authors of DNS Providers have to do a lot of work; a lot of this is repeated between providers.

What's new:

1. Easy-to-use API: diff2.IncrementalDiff() are replaced by diff2.ByRecordSet(), diff2.ByLabel(), diff2.ByRecord(), diff2.ByZone() ... these return the data in the format needed for providers that deal with updates by RSet, Label, Record, or entire Zone at a time.
2. Easy-to-use results: Rather than return a list of records to be added, changed, and deleted, instead the system returns a list of instructions to be executed in order.  The provider simply needs to translate these into API calls. This is a much more simple paradigm.
3. 
4. The code in the providers is a LOT shorter and easier to understand. Writing new providers is going to be significantly easier now.
5. The new system is easier to test.
6. The diff algorithm is cleaner, more simple, and has fewer edge cases.
7. UNMANAGED() is like IGNORE_NAME() and IGNORE_TARGET() rolled into one.  The old functions call UNMANAGED().

What's next:

1. Convert more DNS Providers!  Top priority is HEXONET, CLOUDFLAREAPI, GCLOUD, NAMEDOTCOM, ROUTE53, DIGITALOCEAN, POWERDNS.
2. Port over any relevant unit tests from `pkg/diff/diff_test.go`
3. Update the "writing providers" documentation.
4. Feedback!  Feedback!  Feedback!  What do you like? What do you dislike?

On a personal note...

I'm super excited about this change!  We know so much more now than we did when we wrote the original diff system. A few weeks ago I came up with 3 novel ideas:

(1) IGNORE() shouldn't ignore records (which has a zillion edge cases), it should pull records from "existing" to "desired". Now there's no edge cases!  The "ignore" happens "for free".

(2) The results of diff.Incremental() are difficult to work with. Everyone ends up regrouping or rewriting the data. Odd edge-cases such as "CNAMEs must be deleted first, and added last" shouldn't have to be implemented in each provider; instead we should hand the data to the providers pre-sorted.

(3) The diff algorithm has a LOT of difficult edge cases and complexity because the models.Records structure requires a lot of cross-checking, etc.  Instead, we copy the records into a hierarchy of Label -> Type -> Record and all the diff'ing algorithm is easier to write.  In other words, rather than solve the difficult problem, we restructure the data and transform the problem into an easy one.  Then we solve the easy problem.

Take a look!  If you'd like to convert your own provider, I'd be glad to help.  Take a look at GANDI_V5 and BIND. They're the best examples.

CC @D3luxee @Deraen @SimenBai @SphericalKat @TomOnTime @arnoschoon @blackshadev @captncraig @costasd @das7pad @hamptonmoore @hnrgrgr @imlonghao @jpbede @juliusrickert @kallsyms @koesie10 @kordianbruck @masterzen @mikenz @mistererwin @onlyhavecans @papakai @patschi @pgaskin @philhug @pierre-emmanuelJ @pragmaton @rblenkinsopp @riyadhalnur @svernick @tlimoncelli @tresni @vatsalyagoel @vojtad @xddxdd


tresni:
@tlimoncelli I'm working to convert the Cloudflare provider to support the new setup, but I'm running into an issue.  The definition for `GetDomainCorrections` calls `PrepFoundRecords` and `PrepDesiredRecords` without the context of the client. I'm not sure how to support certain behaviors in the provider. (e.g. if `manage_redirects` is unset or set to false, that in `PrepDesiredRecords` I should throw an error in if someone attempts to give me a `CF_REDIRECT` or `CF_TEMP_REDIRECT` record, but I can't do that as the knowledge of `manage_redirects` is on the client object.) Am I thinking about this all wrong?

blackshadev:
@tlimoncelli I am not sure if I will be able to provide some changes in the upcoming 2 weeks, but after that I should be able to find the time and make some changes. Will that be fast enough?

tlimoncelli:
> @tlimoncelli I am not sure if I will be able to provide some changes in the upcoming 2 weeks, but after that I should be able to find the time and make some changes. Will that be fast enough?

Oh!  I forgot to mention the timeline.  I have no plan for how we'll roll this out.  Because both the old and new code co-exist, we can take our time (but I do eventually want to remove the old code).  Because the new code is hidden behind a "secret" flag, we can do A-B comparisons, let people test it on their own, etc.

So... we're talking months, not days or weeks.

tlimoncelli:
> @tlimoncelli I'm working to convert the Cloudflare provider to support the new setup, but I'm running into an issue. The standard implementation for `GetDomainCorrections` that we are supposed to copy from the gandi provider calls `PrepFoundRecords` and `PrepDesiredRecords` without the context of the client. I'm not sure how to support certain behaviors in the provider. (e.g. if `manage_redirects` is unset or set to false, that in `PrepDesiredRecords` I should throw an error in if someone attempts to give me a `CF_REDIRECT` or `CF_TEMP_REDIRECT` record, but I can't do that as the knowledge of `manage_redirects` is on the client object.) Am I thinking about this all wrong?

Ah, yes.  The cloudflare provider is probably going to be the most difficult provider to adapt. It has special cases like `CF_PROXY_ON`, `CF_REDIRECT` and `CF_WORKER_ROUTE`.

I have plans on how to accommodate those.  The `compFunc ComparableFunc` that gets passed to functions in diff2.go are for Cloudflare's `CF_PROXY_ON` feature, but in a generic way that should also handle AWS and other systems that "add on" to an existing record.  Handling `CF_REDIRECT` and `CF_WORKER_ROUTE` might require  some API changes, but I'll take a look.

Would you like to branch this PR and work together on this?

tresni:
I've got a [a branch](https://github.com/StackExchange/dnscontrol/compare/tlim_newdiff...tresni:dnscontrol:cloudflare/tlim_newdiff?expand=1#diff-c64fff00beea167240b702077c9d1b0367db2fbe4b95aa254c161a9bd0aa1bd6R170) that I started on.  I've only ensured that the old diff method does not appear to do anything unexpected with the changes to the flow.  Right now I put the `PrepDesiredRecords` cloudflareProvider object, I'll look into the `compFun` and see how far I can get, but an example would be helpful :) 

For CF_PROXY_ON and similars, I'm using [checking/ignore records rules](https://github.com/StackExchange/dnscontrol/compare/tlim_newdiff...tresni:dnscontrol:cloudflare/tlim_newdiff?expand=1#diff-c64fff00beea167240b702077c9d1b0367db2fbe4b95aa254c161a9bd0aa1bd6R438-R505) similar to what you had in the Gandi provider for TTL and NS records at apex.

tlimoncelli:
For logistical reasons* I'm merging this under https://github.com/StackExchange/dnscontrol/pull/1852 instead.

I'll close this PR and create a new one for future work.

Tom

tlimoncelli:
Let's continue the discussion on https://github.com/StackExchange/dnscontrol/issues/1854

PR #1855, opened 2022-12-12
PORKBUN: support diff2 and two small changes

imlonghao:
All tests passed using `go test -v -verbose -provider PORKBUN -diff2`


tlimoncelli:
Looks great! Thanks for doing this so quickly!

Issue #1854, opened 2022-12-12
DIFF2: Ongoing discussion


tlimoncelli:
- [x] akamaiedgedns @svernick https://github.com/StackExchange/dnscontrol/pull/1872/files?diff=split&w=1
- [ ] autodns @arnoschoon https://github.com/StackExchange/dnscontrol/issues/1904 
- [x] axfrddns @hnrgrgr https://github.com/StackExchange/dnscontrol/pull/1905/files?diff=split&w=1 
- [x] azuredns @vatsalyagoel (https://github.com/StackExchange/dnscontrol/issues/1907) 
- [X] bind @tlimoncelli 
- [x] cloudflare @tresni https://github.com/StackExchange/dnscontrol/issues/1906 
- [x] cloudns @pragmaton https://github.com/StackExchange/dnscontrol/pull/1874/files?diff=split&w=1 
- [x] cscglobal @mikenz https://github.com/StackExchange/dnscontrol/pull/1875/files?diff=split&w=1
- [x] desec @D3luxee https://github.com/StackExchange/dnscontrol/pull/1876/files?diff=split&w=1 
- [x] digitalocean @Deraen https://github.com/StackExchange/dnscontrol/pull/1877/files?diff=split&w=1 
- [x] dnsimple @onlyhavecans https://github.com/StackExchange/dnscontrol/pull/1878/files?diff=split&w=1 
- [x] dnsmadeeasy @vojtad https://github.com/StackExchange/dnscontrol/pull/1879/files?diff=split&w=1 
- [x] domainnameshop @SimenBai https://github.com/StackExchange/dnscontrol/pull/1880/files?diff=split&w=1 
- [x] exoscale @pierre-emmanuelJ https://github.com/StackExchange/dnscontrol/pull/1881/files?diff=split&w=1 
- [X] gandi_v5 @TomOnTime 
- [x] gcloud @riyadhalnur https://github.com/StackExchange/dnscontrol/pull/1856
- [X] gcore @xddxdd #1867 
- [x] hedns @rblenkinsopp https://github.com/StackExchange/dnscontrol/pull/1887 and https://github.com/StackExchange/dnscontrol/pull/1915 
- [x] hetzner @das7pad https://github.com/StackExchange/dnscontrol/pull/1888/files?diff=split&w=1 
- [x] hexonet @papakai https://github.com/StackExchange/dnscontrol/pull/1889/files?diff=split&w=1 
[...]

PR #1858, opened 2022-12-15
ns1: Add support for diff2

costasd:
This changeset adds proper support for conditionally enabling diff2 in the NS1 provider.

While there, some code was shuffled around:
* diff-specific parts were moved inside the diff section
* DNSSEC corrections were moved out of the conditional, they're independent of diff/diff2 being used.

Relates to #1854


costasd:
Hey @tlimoncelli, this should be all that's needed to add diff2 support to the ns1 provider :)

tlimoncelli (commented):
Looks good so far.  One request: When we delete the "if diff2.EnabledDiff2" and legacy code, it would  be nice if the remaining code didn't need to change indenting.  That way "git blame" will show you as the author of these lines.

providers/ns1/ns1Provider.go
```diff
@@ -198,11 +199,44 @@ func (n *nsone) GetDomainCorrections(dc *models.DomainConfig) ([]*models.Correct
 				})
 			}
 		}
-		return corrections, nil
+
 	}
 
-	// Insert Future diff2 version here.
+	if diff2.EnableDiff2 {
```
tlimoncelli, line 205: Please add the `	return corrections, nil` to the old diff section, then you can remove this line and the ending `}`.  The result will be the "git blame"-friendly version I mentioned earlier (in case it wasn't clear).

providers/ns1/ns1Provider.go
```diff
@@ -198,11 +199,44 @@ func (n *nsone) GetDomainCorrections(dc *models.DomainConfig) ([]*models.Correct
...
+
+		changes, err := diff2.ByRecordSet(existingRecords, dc, nil)
+		if err != nil {
+			return nil, err
+		}
+		fmt.Printf("DEBUG: NS1 changes=%v\n", changes)
```
tlimoncelli, line 211: Please remove this debug line.

providers/ns1/ns1Provider.go
costasd: heh, was just pushing exactly this :)

providers/ns1/ns1Provider.go
costasd: done!

tlimoncelli:
Looks great!  Thanks!

PR #1860, opened 2022-12-17
TransIP

blackshadev:
Hi Tom,

I believe this will be required changes for the new diff2 differ.

I tested it using `go test -v -verbose -provider TRANSIP -diff2 -end 1` and I ran a preview against my own DNS config.

I notices same annoyances with the new differ package which I like to state as well. Do note, I get that sometimes you have to choose a solution rather than maintain many differences ones to please everybody. So don't see this as a show stopper. But rather suggestions which can maybe help you obtain a easier to integrate package.

- First, the `Change` fields. The rename from `Desired` and `Existing` to `Old` and `New` really took me by surprise. I liked the old field names allot more because it ties more into what we are doing. It is really weird for me te 'remove' old DNS record, instead of 'Existing' DNS records. It seems also weird to 'Change' using the 'New' field instead of the 'Desired'. 
- Second, In every example I see people doing `strings.Join(change.Msgs, "\n")` to pass to `models.Correction`'s `Msg` field. Why not make it an public function on `Change` or `Change.Msgs`?
- The rest API of TransIP does everything record for record, but your new differ seems to favor batching in some form or another. This required me to introduce allot of helper functions and boilerplate functions to 'get around' it. A solution is just assuming the `New` and `Old` fields are always a slice of length 1 or 0, but this to me is just bad design and will lead to bugs down the road. I rather see a separate return structure for the `diff2.ByRecord` version. In the defence of your design, it came to rescue as I needed to switch to `diff2.ByRecordSet` (and even tried diff2.ByLabel) because TransIP's API really can't handle updating a record from a multi record set. If this design intentionally favors batching updates, good job, I am just a terrible person trying my best to go against your better judgement. If your new design stays I am thinking about rewriting my provider to always replace the whole zone, it is allot more destructive, but it seems easier from my perspective. If this is unintentional, you should try and mold the `ChangeList` to support these `ByRecord` changes more fluently.

### DIFF
--- a/providers/transip/transipProvider.go
+++ b/providers/transip/transipProvider.go
@@ -98,12 +98,8 @@ func (n *transipProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*mode
 		return corrections, err
 	}
 
-	if diff2.EnableDiff2 {
-		corrections, err := n.getCorrectionsUsingDiff2(dc, curRecords)
-		return corrections, err
-	}
-
-	return nil, fmt.Errorf("unexpected error: unreachable code. should either use old Diff or Diff2, but neither used")
+	corrections, err := n.getCorrectionsUsingDiff2(dc, curRecords)
+	return corrections, err
 }
 
 func (n *transipProvider) getCorrectionsUsingDiff2(dc *models.DomainConfig, records models.Records) ([]*models.Correction, error) {
--- a/providers/transip/transipProvider.go
+++ b/providers/transip/transipProvider.go
@@ -136,7 +136,7 @@ func (n *transipProvider) getChangeFunction(changeType diff2.Verb, dc *models.Do
 		return wrapChangeFunction(change.Old, func(rec domain.DNSEntry) error { return n.domains.UpdateDNSEntry(dc.Name, rec) })
 	}
 
-	panic(fmt.Sprintf("Unsupported change type %s", changeType))
+	return nil
 }
 
 func wrapChangeFunction(records models.Records, executer func(rec domain.DNSEntry) error) func() error {
--- a/providers/transip/transipProvider.go
+++ b/providers/transip/transipProvider.go
@@ -114,42 +114,33 @@ func (n *transipProvider) getCorrectionsUsingDiff2(dc *models.DomainConfig, reco
 	}
 
 	for _, change := range instructions {
-		msg := strings.Join(change.Msgs, "\n")
 
-		switch change.Type {
-
-		case diff2.CREATE:
-			corrections = append(corrections, &models.Correction{
-				Msg: msg,
-				F:   wrapChangeFunction(change.New, func(rec domain.DNSEntry) error { return n.domains.AddDNSEntry(dc.Name, rec) }),
-			})
-
-		case diff2.CHANGE:
-			if canUpdateDNSEntries(change.New, change.Old) {
-				corrections = append(corrections, &models.Correction{
-					Msg: msg,
-					F:   wrapChangeFunction(change.New, func(rec domain.DNSEntry) error { return n.domains.UpdateDNSEntry(dc.Name, rec) }),
-				})
-			} else {
-				corrections = append(corrections, &models.Correction{
-					Msg: fmt.Sprintf("[1/2] delete of %s", msg),
-					F:   wrapChangeFunction(change.Old, func(rec domain.DNSEntry) error { return n.domains.RemoveDNSEntry(dc.Name, rec) }),
-				})
-				corrections = append(corrections, &models.Correction{
-					Msg: fmt.Sprintf("[2/2] create of %s", msg),
-					F:   wrapChangeFunction(change.New, func(rec domain.DNSEntry) error { return n.domains.AddDNSEntry(dc.Name, rec) }),
-				})
-			}
-
-		case diff2.DELETE:
-			corrections = append(corrections, &models.Correction{
-				Msg: msg,
-				F:   wrapChangeFunction(change.Old, func(rec domain.DNSEntry) error { return n.domains.RemoveDNSEntry(dc.Name, rec) }),
-			})
+		if canDirectApplyDNSEntries(change) {
+			changeFunction := n.getChangeFunction(change.Type, dc, change)
+			corrections = append(corrections, change.CreateCorrection(changeFunction))
+		} else {
+			deleteFunction := n.getChangeFunction(diff2.DELETE, dc, change)
+			createFunction := n.getChangeFunction(diff2.CREATE, dc, change)
+			corrections = append(corrections, change.CreateCorrectionWithMessage("[1/2] delete", deleteFunction), change.CreateCorrectionWithMessage("[2/2] create", createFunction))
 		}
 	}
 
 	return corrections, nil
+}
+
+func (n *transipProvider) getChangeFunction(changeType diff2.Verb, dc *models.DomainConfig, change diff2.Change) func() error {
+	switch changeType {
+	case diff2.DELETE:
+		return wrapChangeFunction(change.Old, func(rec domain.DNSEntry) error { return n.domains.RemoveDNSEntry(dc.Name, rec) })
+
+	case diff2.CREATE:
+		return wrapChangeFunction(change.New, func(rec domain.DNSEntry) error { return n.domains.AddDNSEntry(dc.Name, rec) })
+
+	case diff2.CHANGE:
+		return wrapChangeFunction(change.Old, func(rec domain.DNSEntry) error { return n.domains.UpdateDNSEntry(dc.Name, rec) })
+	}
+
+	panic(fmt.Sprintf("Unsupported change type %s", changeType))
 }
 
 func wrapChangeFunction(records models.Records, executer func(rec domain.DNSEntry) error) func() error {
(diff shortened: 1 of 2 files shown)


tlimoncelli:
Hi!

Thank you for the thoughtful feedback!

1. Agreed.  Old/New vs Desired/Existing is a bit confusing.  In an OO world we would have been easier.  If gorename can fix this, I'll consider renaming them.
2. Agreed. I was considering providing both []string and the joined result but didn't get around to it because I think I'd rather go the other direction: Keep everything as []string upstream also so that the output functions can indent the output or transform it any way they want.  

3. This is the kind of feedback I find the most valuable!  Here's my response...

Your point "your new differ seems to favor batching in some form or another" is interesting.  My goal was to provide enough `By*()` functions that every kind of batching was provided, including no batching (ByRecord).  The problem I wanted to solve was the fact that every provider was re-batching the data. Many providers repeated the same code, but often one had a bug and other didn't.

However I also wanted all `By*()` functions to return the same struct. So, yes, New and Old fields are always a slice of length 1 or 0 for ByRecord, but may be any length for ByLabel and ByRecordSet.  I guess if GO was more OO there would be a cleaner way to do that.   Anyway... My suggestion is to use go's idiomatic `s = append(s, y.Old...)` no matter what. That has the benefit of handling 0 (no-op), 1, and more. 

Looking at your code, it looks like `diff2.ByRecordSet` worked well for you.  Yes, you could switch to `ByZone()` but that might not work as well for large zones.  Either way what the user will see is the same msgs.

Tom

blackshadev:
Ok, so I am not having anything against the different By*()  functions, I think they are a wild improvement. So lets focus the discussion on the main reason why I think it favors batching: The return types. 

If your main goal is to limit the amount of boilerplate in the providers modules, you will need to return the data as they expect them to.  I don't think providing a uniform return type really suits this need. For example, every `ByRecord` API wil have the same functions / code to "assume" 1 or 0 records.
The interesting thing is, you are already doing this a bit: The return type of the ByZone() is wildly different than any of the other By*() functions. 

But there can be something said about limiting the exposed types by having a smaller and more uniform internal API, which you did very nicely. (apart from the ByZone one, but yeah, uniforming that one would be really nasty)

tlimoncelli:
Yes, I see your point. However I don't think it is that much of a burden.  When using ByRecord() Old[0] always exists for delete and change; New[0] always exists for change and create.  There's no need to check len(Old) or len(New) as a precaution.  Maybe I should document that?

blackshadev:
Sure, If you do not embed it into you typing model, documenting it is the next best thing.

tlimoncelli:
Here's a thought...

What if the `By*()` functions returned []*models.Correction with the .Msg field filled out for you.  The provider would just have to update the .F field.

There would be other fields that would provide anything needed to generate F.

The only problem I see with this is that sometimes a single Correction needs to be split into many individual actions. At that point the iteration over the corrections becomes difficult.

I mocked up what this would look like here: https://github.com/StackExchange/dnscontrol/pull/1873

It looks very good... a lot less boilerplate.  However, things get more complex in other ways as you'll see.

Thoughts?

blackshadev:
I think it is a lot better towards having a smaller and more uniform API.  For allot of providers which align with the diff method perfectly this will be way less boilerplate and easier to integrate. For providers like TransIP, which have some some weird exceptions: It will likely not matter all that much. All in all this is an improvement towards your goals. 

Personally I don't like how you mix the responsibilities of the diff-ing method with the API integration (function F). And how you actively ask to mutate state, instead of returning an new instance. These things make it a bit harder to change later on. But in our line of work you always need to make some form of trade-off and these changes are a great step forwards towards your states goals.

Why will these changes not matter al that much for the TransIP provider, and maybe even make it a bit harder: I still need to iterate all corrections because the API actually need to changed ByRecord, but the API cannot handle an edit on an multi record recordset. So I diff byRecordSet, to find these multi record recordsets and splice in a remove and a create API call per record. For all single record record sets I can just do it in one single call.

What I do think will help providers like TransIP allot are an utility function to create a new correction given a previous one. That way I can create two corrections (one for removal and one for creation) when the API doesn't allow me to mutate an existing recordset. 

One small note on your changes: You added a comment behind the `Old` and `New` fields. I still think the naming of `Existing` and `Desired` is a bit better, but I think you did it to minimize changes, which makes sense. What doesn't make sense, is `nil` description, you swapped those. `Old` is not null for delete, it is `nil` for creations because there weren't any existing records. And the same for `New` which is `nil` for delete and not for `create`. 

All in all, great work! Thanks for taking the time. I hope my rambling helps you in one way or another.

tlimoncelli:
> I think it is a lot better towards having a smaller and more uniform API. For allot of providers which align with the diff method perfectly this will be way less boilerplate and easier to integrate. For providers like TransIP, which have some some weird exceptions: It will likely not matter all that much. All in all this is an improvement towards your goals.

Thanks!

> What I do think will help providers like TransIP allot are an utility function to create a new correction given a previous one. That way I can create two corrections (one for removal and one for creation) when the API doesn't allow me to mutate an existing recordset.

What would that design look like?  Could you sketch that out for me?

> One small note on your changes: You added a comment behind the `Old` and `New` fields. I still think the naming of `Existing` and `Desired` is a bit better, but I think you did it to minimize changes, which makes sense. What doesn't make sense, is `nil` description, you swapped those. `Old` is not null for delete, it is `nil` for creations because there weren't any existing records. And the same for `New` which is `nil` for delete and not for `create`.

I do plan on renaming those. I've found VS Code's "rename variable" works brilliantly for this (I assume it leverages gorename).

Naming is hard.  Before we standardized on "existing" and "desired" there was some code that used old/new, before/after, and I think a few other conventions.  My OCD-ness wishes that alphabetically "existing" came before "desired". (Isn't it odd that o/n, b/a, e/d all have that problem?)

But I digress.... I can't find the comment you are referring to.  Pointer?  I may have had too much eggnog while writing those comments and would like to correct them.

> All in all, great work! Thanks for taking the time. I hope my rambling helps you in one way or another.

I appreciate your "rambling" :-).  It's good feedback!  I don't have a local peer that can do PR reviews and appreciate all the feedback I get from the community.

Best,
Tom

blackshadev:
> What would that design look like? Could you sketch that out for me?

I hope I can create a PR based on yours in the upcoming week. The idea I had with an utility function is maybe not really a good way to do it, because you want the key field to be unique and with a function which just makes a copy for you you'd still end up changing all those fields manually. So that it kind of pointless. 

> But I digress.... I can't find the comment you are referring to. Pointer? I may have had too much eggnog while writing those comments and would like to correct them.

models/dns.go:114-115
```
	Old Records // The existing records related to this change (nil for a DELETE)
	New Records // The desired records related to this change (nil for CREATE)
```
The part between brackets is the wrong way around I believe.

blackshadev:
@tlimoncelli 
I made some change on: https://github.com/blackshadev/dnscontrol/pull/1/files

And I have good and bad news.  I went back on your original design and tried to create something how it suited me the most. While still reducing the boilerplate required for providers with a more one to one mapping to the diff method.
And I actually didn't change all that much to be honest. 
I only needed to:
- Restructure my own provider. 
- Add a helper to the Change type.

While using it I think "original" diff2 changes are better / easier to use because:
 - It gives the provider maintainers easier control over the corrections they push
 - It is easier to create a new Correction and append it to the change list, than it is to create a new Change
 - Not all Corrections map one to one to the Change (see my whole Remove and Add shenanigans), if I push this into one function which does both, there is a change it will fail on one of the two functions and it will not help debugging for the end user.

I hope this makes sense and feel free to copy my changes to the extend you like (or not).

tlimoncelli:
Just a small suggestion then it's ready to merge!

providers/transip/transipProvider.go
```diff
@@ -114,42 +114,33 @@ func (n *transipProvider) getCorrectionsUsingDiff2(dc *models.DomainConfig, reco
...
-		}
+	case diff2.CREATE:
+		return wrapChangeFunction(change.New, func(rec domain.DNSEntry) error { return n.domains.AddDNSEntry(dc.Name, rec) })
+
+	case diff2.CHANGE:
+		return wrapChangeFunction(change.Old, func(rec domain.DNSEntry) error { return n.domains.UpdateDNSEntry(dc.Name, rec) })
 	}
 
-	return corrections, nil
+	panic(fmt.Sprintf("Unsupported change type %s", changeType))
```
tlimoncelli, line 143: Suggestion:

    return nil, fmt.Errorf("Unsupported change type %s", changeType)

blackshadev: I tried doing it this way, but I opt-ed out of it. Optionally returning an error in this case will make the usage statements on line 119, 121 and 122 really awkward full with conditional statements. And to be honest, the sole reason to have this is, because I need to return (or panic). But this statement should and will not execute ever, not in the "the user made an error" kind of way, no that statement should be statically unreachable at all times. 

From what I can see I can do one of the following:

1. Return an error and make the code  using it full with conditionals to check the errors (which should not happen)
1. Return an error and ignore the error...?
1. Leave it as is and panic (this is my suggested option)
1. Return an empty function
1. Return an change function (which is the least destructive) 
1. or write function without a switch and default to the change function (same as option above)

pick your poison, I will follow

providers/transip/transipProvider.go
```diff
@@ -94,76 +93,167 @@ func (n *transipProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*mode
 
 	models.PostProcessRecords(curRecords)
 
+	if !diff2.EnableDiff2 {
+		corrections, err := n.getCorrectionsUsingOldDiff(dc, curRecords)
+		return corrections, err
+	}
+
```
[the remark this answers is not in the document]
blackshadev: Nice one, done.

providers/transip/transipProvider.go
blackshadev: Sure changed it to return `nil`.

tlimoncelli:
Thanks for the PR and the lively discussion!

PR #1861, opened 2022-12-18
Adapt OVH provider to Diff2

masterzen:
This adds support for Diff2 (cf #1854) for the OVH provider.
OVH API deals with records so the new Diff2 based implementations is not very different than the actual Diff1 one.
Integration tests pass in both diff1 and diff2 modes.


tlimoncelli:
Excellent!  I like how you split out the diff1 vs diff2 code.  Thanks!

PR #1867, opened 2022-12-29
GCORE: Implement diff2 and greatly improve performance for getting record sets

xddxdd:
- Implement diff2 method, tested with `go test -v -verbose -provider GCORE -diff2`
- Use a hidden API only documented in Gcore's OctoDNS provider (`https://api.gcorelabs.com/dns/v2/zones/example.com/rrsets?all=true`, documented at <https://github.com/octodns/octodns-gcore/blob/main/octodns_gcore/__init__.py#L105>), to get all record sets in one HTTP request instead of many requests. This greatly improves performance handling large zones.


tlimoncelli:
Looks great! Thanks for the move to diff2 and the awesome speedup!

PR #1873, opened 2022-12-31
Maybe this should be the interface to diff2.By*() ?

tlimoncelli:


tlimoncelli:
Closing.  It was a useful discussion but we resolved it other ways.

PR #1890, opened 2023-01-01
HOSTINGDE: Adopt diff2 in compatibility mode

tlimoncelli:
Hi @juliusrickert:

Please run:
```
    go test -v -verbose -provider HOSTINGDE
```
and
```
    go test -v -verbose -provider HOSTINGDE -diff2
```

Thanks and happy new years!!!

Tom
@tlimoncelli


juliusrickert:
Hey @tlimoncelli,

Happy new year to you too!

I'm quite busy at the moment but I'll see whether I can test it within the next week.

juliusrickert:
Finally found some time to run the integration tests.

Both, old and new, fail.
This is, from what I can tell, the result of Hosting.de running zone updates asynchronously. I didn't bother to implement it because it would have made the implementation way more complex. It only occurs when applying multiple changes within a few seconds, an operation which usually isn't encountered, as my implementation applies all changes in a single change request per zone.

In more detail: Whenever applying a change, a job is created, whose status can be observed at another endpoint. We could change the implementation to wait for that to finish, but it'd prolong the runtime per zone by an order of magnitude, I'd guess.

tlimoncelli:
Interesting situation!

How long do these jobs take to complete?  For other providers it is less than a second. This slows things down a bit but it is worth checking.

For CSCGLOBAL the job takes 7+ minutes to complete.  Yes, over 400 seconds and typically a lot more!  For that provider, we do something weird.  We do a pre-check (are there outstanding jobs?) instead of a post-check (did this job complete?).  That way the last change doesn't involve any waiting.  Since the typical "dnscontrol push" is only making one change, this means we almost never wait.

The way CSCGLOBAL implements this is that before any change, a query is done to see if there are any incomplete jobs related to that domain. If there are, we wait for each of them to complete.

The integration tests takes hours (7-10 hours!) but typical changes take almost no time at all.

Hope that helps!

juliusrickert:
> How long do these jobs take to complete?

From experience, a second to ten seconds, max. But @membero may be able to give more insight.

> This slows things down a bit but it is worth checking.

Agreed! But I certainly won't have time to implement this within the next month.

tlimoncelli:
Cool cool. Thanks for the info.  Let's loop back when you're less busy.

tlimoncelli:
Hi @membero! Thanks for the info!

What happens if one sends an update and the zone is not active yet?  Does the API return an error?

Most provider's API reports that the update is "done" when the change is committed to the provider's database; actual propagation to DNS will happen later.  DNSControl assumes that's sufficient.

membero:
Hello @tlimoncelli,

yes, the API will return an error if the zone is not "updateable". The JSON output looks like this:

`{
    "errors": [
        {
            "code": 10205,
            "contextObject": "",
            "contextPath": "",
            "details": [
            ],
            "text": "The current status of an object you are trying to use does not allow that operation.",
            "value": "blocked"
        }
    ],
    "metadata": {
        "clientTransactionId": "",
        "serverTransactionId": "stid"
    },
    "status": "error",
    "warnings": [
    ]
}`

tlimoncelli:
Perfect!  That will allow us to do it like CSCGLOBAL.    i.e. Send update. If you get a "blocked" error, sleep then resend the exact same request.

tlimoncelli:
hi hi!

Would you please re-run the tests? I want to verify I didn't break anything while resolving a conflict.

Thanks!

juliusrickert:
Hi @tlimoncelli,

I've run `go test -v -verbose -provider HOSTINGDE` and `go test -v -verbose -provider HOSTINGDE -diff2`.

Both yield the same result. Everything but one test is passing:
`FAIL: TestDNSProviders/testing-hostingde-dnscontrol.de/21:SRV:Null_Target`
As far as I can tell, hosting.de doesn't support null targets for SRV records.

So LGTM.

tlimoncelli:
Cool cool.  I've added rejectif.SrvHasNullTarget to auditrecords.go

juliusrickert:
> I've added rejectif.SrvHasNullTarget to auditrecords.go

Thanks, Tom. Haven't been keeping track of the developments regarding audit records.

tlimoncelli:
No worries!  I added audit records last year and its become a great timesaver!

PR #1895, opened 2023-01-01
NETCUP: Adopt diff2 in compatibility mode

tlimoncelli:
Hi @kordianbruck:

Please run:
```
    go test -v -verbose -provider NETCUP
```
and
```
    go test -v -verbose -provider NETCUP -diff2
```

Thanks and happy new years!!!

Tom
@tlimoncelli


tlimoncelli:
Friendly ping? @kordianbruck

kordianbruck:
A lot here doesn't work because apparently there is now an assumption baked in that Netcup supports TTLs (which it does not). For example the fifth or so test case fails on a missing diff:

```
=== RUN   TestDNSProviders/mflang.de/00:Protocol-Plain:Change_a_ttl
    integration_test.go:219: Expected changes, but got none
```

In reality `GetDomainCorrections` in the `netcupProvider` just overwrites any records that try to set a specific TTL for a record. The provider however doesn't support TTLs for records. I don't see a feature flag that would be gating this or a way in auditrecords, or am I missing something?

kordianbruck:
Yeah, those problems seemingly also exist on master, so I don't think this is a new failing behavior of this change. You have a suggestion on how to address this inability here?

tlimoncelli:
That's an interesting situation!  I think this is the first provider that completely ignores TTLs.  Some providers have a minimum TTL or a list of permitted TTLs.  Cloudflare has a magic value (TTL=0 and TTL=300 are magic).  Some providers don't enforce the rule that all records at the same level must have the same TTL.

Here's my recommendation:  Set all TTLs to the same value.  That will trick diff2 into doing the right thing.   `toRecordConfig()` already sets the TTL to 0.  You just need to iterate over `dc.Records` and set `.TTL` to zero in `GetDomainCorrections()`. Since `GetDomainCorrections()` works on a copy of the records, that shouldn't affect other providers.

Tom

kordianbruck:
> Here's my recommendation: Set all TTLs to the same value.

Yes, that is what is currently being done: https://github.com/StackExchange/dnscontrol/blob/master/providers/netcup/netcupProvider.go#L79

Which however leads to tests like `Change a ttl` to not pass: https://github.com/StackExchange/dnscontrol/blob/master/integrationTest/integration_test.go#L702

I don't know if we particularly care.

tlimoncelli:
Agreed.  You can skip those integration tests with `not()`.

tlimoncelli:
I'm merging this.  If NETCUP should opt-out of other tests we'll handle that separately.  Please re-run the tests and either file an issue (if you'd like me to make the change) or a PR (if you'd like to make the change).

kordianbruck:
Cool, thanks! Checked both runs of integration tests and results look ok to me. I'll send a separate PR to fix/exclude some of the other failing integration test.

tlimoncelli:
> Cool, thanks! Checked both runs (diff and diff2) of integration tests and results look ok to me. I'll send a separate PR to fix/exclude some of the other failing integration test.

Thank you very much!

PR #1902, opened 2023-01-01
VULTR: Adopt diff2 in compatibility mode

tlimoncelli:
Hi @pgaskin:

Please run:
```
    go test -v -verbose -provider VULTR
```
and
```
    go test -v -verbose -provider VULTR -diff2
```

Thanks and happy new years!!!

Tom
@tlimoncelli


tlimoncelli:
Friendly ping? @pgaskin

androw:
I've been testing this change on VULTR. Both commands hangs up on the "IDNAs in CNAME targets" test.

After disabling this test, both commands failed on "add_to_label_and_change_orig_ttl:Add_at_same_label,_new_ttl" and "SRV_record":
```
=== RUN   TestDNSProviders/androw.cf/02:add_to_label_and_change_orig_ttl:Add_at_same_label,_new_ttl
    integration_test.go:223: CREATE A www.androw.cf 1.2.3.4 ttl=700
    integration_test.go:227: {"error":"Unable to add record: New records TTL (700) must match the TTL of existing records (400)","status":400}

=== RUN   TestDNSProviders/androw.cf/19:SRV:SRV_record
    integration_test.go:223: CREATE SRV _sip._tcp.androw.cf 5 6 7 foo.com. ttl=300
    integration_test.go:242: Expected 0 corrections on second run, but found 1.
    integration_test.go:244: UNEXPECTED #0: MODIFY SRV _sip._tcp.androw.cf: (5 6 7 foo.com ttl=300) -> (5 6 7 foo.com. ttl=300); Vultr RecordID: 217bd3e7-9ece-4e6a-be9b-73711df1b183
```

Same behaviour was observed without this patch.

androw:
I've rebased the diff2 compat here: [androw:vultr](https://github.com/androw/dnscontrol/tree/vultr)

I've been looking at the IDNA problem. It looks like the we should punycode the name but not the target.

If I remove dc.punycode() and do the test with a "classic" name I'm getting this:
```
=== RUN   TestDNSProviders/androw.cf/13:IDNAs_in_CNAME_targets:IDN_CNAME_AND_Target
    integration_test.go:223: CREATE CNAME androw.androw.cf \195\182\195\182\195\182.\228\188\129\228\184\154. ttl=300
    integration_test.go:242: Expected 0 corrections on second run, but found 1.
    integration_test.go:244: UNEXPECTED #0: MODIFY CNAME androw.androw.cf: (xn--ndaaa.xn--vhquv. ttl=300) -> (\195\182\195\182\195\182.\228\188\129\228\184\154. ttl=300); Vultr RecordID: 8a6e22a4-f180-4d1b-81fc-4d2607618260
```

It seems the record is correctly created at Vultr but dnscontrol shows it in another form.

@tlimoncelli can you help me with how I should handle this?

tlimoncelli:
Excellent question!  I'm not sure if I have an excellent answer! :-(

IDNA can be difficult to handle.  In fact, lately I've been thinking that we should handle Punycode differently in the system. However that's not going to help this particular problem.

The good news is that if this exists in the old and new system, then it isn't a problem with diff2.  :-)

My suggestions:

* Keep`dc.punycode()`.
* `IDNAs in CNAME targets`: My guess is that we have to punycode (or un-code?) the target.
* `19:SRV:SRV_record`: Looks like a trailing dot is missing.  Either remove it from both or add it to both.
* `02:add_to_label_and_change_orig_ttl:Add_at_same_label `: This means that the provider requires the same TTL on all records at a certain label.  This is actually required by the RFCs but most providers ignore this. My suggestion is to set the TTL to the lowest of the existing records, or just skip this test.

Tom

androw:
Thanks for helping out. Everything thing should be okay in my PR.

PR #1905, opened 2023-01-01
AXFRDDNS: Adopt diff2 in compatibility mode

tlimoncelli:
Hi @hnrgrgr:

Please run:
```
    go test -v -verbose -provider AXFRDDNS
```
and
```
    go test -v -verbose -provider AXFRDDNS -diff2
```

Thanks and happy new years!!!

Tom
@tlimoncelli


hnrgrgr:
Hi @tlimoncelli, happy new year.

I didn't reviewed the proposed patch yet, but I quickly ran the tests without and with `-diff2`. There is only one failing test, in both cases:
```
=== RUN   TestDNSProviders/example.com/04:TypeChange:Change_back_to_CNAME
    integration_test.go:223: DDNS UPDATES to 'example.com' (primary master: '172.27.82.2:53'). Changes:
        CREATE CNAME foo.example.com google2.com. ttl=300
        DELETE A foo.example.com 1.2.3.4 ttl=300
        
    integration_test.go:242: Expected 0 corrections on second run, but found 1.
    integration_test.go:244: UNEXPECTED #0: DDNS UPDATES to 'example.com' (primary master: '172.27.82.2:53'). Changes:
        CREATE CNAME foo.example.com google2.com. ttl=300
```

I will investigate further this week.

Grégoire

tlimoncelli:
Interesting!   I didn't look deeply into it but my guess is that CNAMEs need to be created after the existing items at that label are deleted.  It's a common API requirement. (the new pkg/diff2/By*() functions handle that automatically but the backwards compatible function doesn't)

PR #1926, opened 2023-01-09
DOCS: Akamai: Updated the Akamai URLs

cafferata:
Updated the Akamai URLs.

Found from the work https://github.com/StackExchange/dnscontrol/pull/1809.


tlimoncelli:
Thanks for the update!

PR #1937, opened 2023-01-12
HOSTINGDE: Add simple exponential back-off retry

juliusrickert:
While a zone is propagating, the zone cannot be edited. This commit adds an exponential back-off retry mechanism for writes.

### DIFF
--- a/providers/hostingde/hostingdeProvider.go
+++ b/providers/hostingde/hostingdeProvider.go
@@ -3,8 +3,10 @@ package hostingde
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/StackExchange/dnscontrol/v3/models"
 	"github.com/StackExchange/dnscontrol/v3/pkg/diff"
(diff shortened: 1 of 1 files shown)


providers/hostingde/hostingdeProvider.go
```diff
@@ -145,16 +147,25 @@ func (hp *hostingdeProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
...
+					// Code:10205 indicates the zone is currently blocked due to a running zone update.
+					if !strings.Contains(err.Error(), "Code:10205") {
+						return err
+					}
+
+					// Exponential back-off retry.
+					// Base of 1.8 seemed like a good trade-off, retrying for approximately 45 seconds.
+					time.Sleep(time.Duration(math.Pow(1.8, float64(i))) * 100 * time.Millisecond)
+				}
+				return fmt.Errorf("timeout: zone blocked for over 10 seconds")
```
tlimoncelli, line 168: "10 seconds" or "10 tries"?

juliusrickert: Good catch! Was initially a linear retry with 10 attempts. So that's about 9 seconds to be precise :D

I'll fix that.

PR #1941, opened 2023-01-13
Fix GitHub Pages build failure

j-f1:
See: https://github.com/StackExchange/dnscontrol/actions/runs/3913042186/jobs/6688447274#step:4:561


tlimoncelli:
Thanks!

PR #1942, opened 2023-01-13
Embed types-dnscontrol.d.ts into the binary

j-f1:


j-f1:
Hm, not sure what the best way forward is here. I intentionally added `types-dnscontrol.d.ts` to the gitignore list so that it would never be out of date (and so that people wouldn’t be tempted to hotlink it from GitHub, but it seems `go generate` isn’t being run before building so the file isn’t being generated.

tlimoncelli:
> Hm, not sure what the best way forward is here. I intentionally added `types-dnscontrol.d.ts` to the gitignore list so that it would never be out of date (and so that people wouldn’t be tempted to hotlink it from GitHub, but it seems `go generate` isn’t being run before building so the file isn’t being generated.

Ah yes, go generate isn't run automatically by design. (I think Pike has a talk about this at gophercon... but he hints around it in  https://go.dev/blog/generate and the design doc https://go.googlesource.com/proposal/+/refs/heads/master/design/go-generate.md).

Here's how it should work:

0. The file should not be in .gitignore. It is part of the repo.
1. Anyone that modifies helpers.js or related files should run "go generate". If the file changes, commit the change as part of the change to helpers.js, etc.
2. "go build" will embed the file in the binary at build time.

That's it!

j-f1:
Ok, I’ve added it back. 

> If the file changes, commit the change as part of the change to helpers.js, etc.

To be clear, `helpers.js` has no impact on the TS definitions. Most definitions are automatically generated from the doc pages, but some were manually written in `types/*.d.ts`. Those files are concatenated together with the generated files before being written to `commands/types-dnscontrol.d.ts`, and then an additional header is generated at runtime by `dnscontrol write-types` with the version info (and warning not to edit the file)

tlimoncelli:
> Ok, I’ve added it back.
> 
> > If the file changes, commit the change as part of the change to helpers.js, etc.
> 
> To be clear, `helpers.js` has no impact on the TS definitions. Most definitions are automatically generated from the doc pages, but some were manually written in `types/*.d.ts`. Those files are concatenated together with the generated files before being written to `commands/types-dnscontrol.d.ts`, and then an additional header is generated at runtime by `dnscontrol write-types` with the version info (and warning not to edit the file)

Ah. Good to know!

j-f1:
Another good reason to integrate the types into the binary: the version info didn’t get picked up correctly for the v3.24.0 version, so the types file ends up looking like this:

```ts
// This file was automatically generated by DNSControl. Do not edit it directly.
// To update it, run `dnscontrol write-types`.

// DNSControl version: -dev
// Source: https://raw.githubusercontent.com/StackExchange/dnscontrol//types/dnscontrol.d.ts

404: Not Found
```

j-f1:
Updated!

PR #1947, opened 2023-01-15
NETCUP: Handle CNAMEs with a single @ as destination correctly

### LINKED ISSUE #1362: NETCUP: Parsed `CNAME`, `TLSA` entries cannot be used w/ other providers
When trying to push `CNAME`, `TLSA` entries in `js` format to another provider (tested w/ `DESEC`), the following cause problems:

```
CNAME('www', '\@.')
-->
2022/01/04 17:42:50 printIR.go:94: ERROR: in CNAME www.example.org: target (@.) includes invalid char
```
_Here, unless there is a unified placeholder for the domain name, the latter should be included in verbatim._

```
TLSA('_tcp', 0, 0, 0, '2 1 2 77[…]')
-->
CREATE TLSA _tcp.example.org 0 0 0 2 1 2 77[…] ttl=3600

FAILURE! failed create RRset (deSEC): HTTP status 400 Bad Request Body: [{},{},{},{},{"non_field_errors":["Record content for type TLSA malformed: Odd-length string"]},{}], the API does not provide more information
```
_Here, the correct values for parameters 2–4 should be stripped from parameter 5._

kordianbruck:
This should fix #1362 and address the concerns there

### DIFF
--- a/providers/netcup/types.go
+++ b/providers/netcup/types.go
@@ -71,6 +71,17 @@ type responseLogin struct {
 	SessionID string `json:"apisessionid"`
 }
 
+// addTailingDot adds a dot if it's missing from what the netcup api has returned to us.
+func addTailingDot(destination string) string {
+	if destination == "@" || len(destination) == 0 {
+		return destination
+	}
+	if destination[len(destination)-1:] != "." {
+		return destination + "."
+	}
+	return destination
+}
+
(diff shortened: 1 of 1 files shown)


tlimoncelli (approved):

tlimoncelli:
Thanks!

PR #1981, opened 2023-01-23
VULTR: Adopt diff2 in compatibility mode and fix handling of some integrations tests

androw:
Overtake #1902

Tried both
```
    go test -v -verbose -provider VULTR
```
and
```
    go test -v -verbose -provider VULTR -diff2
```
and both succeed


tlimoncelli:
Thanks!  Looks great!

androw:
Can check vultr on #1854

tlimoncelli:
> Can check vultr on #1854

Checked!  Thanks for the reminder!

PR #1983, opened 2023-01-25
Domains flag should accept simple wildcards

tlimoncelli:
The `preview` and `push` commands accept a flag `--domains` which restricts which domains are processed.  Currently they are exact matches.  This PR permits a wildcard at the start of the string.  In other words `--domains foo.com,*.10.in-addr.arpa` will match `foo.com`, `0.0.0.10.in-arpa.com`, but not `0.0.0.6.in-addr.arpa.com`.

The goal here isn't to add a complete glob match. We can add that later.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -294,8 +294,18 @@ func (args *FilterArgs) shouldRunDomain(d string) bool {
 	if args.Domains == "" {
 		return true
 	}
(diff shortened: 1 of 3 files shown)


tlimoncelli:
NOTE: I haven't updated the docs.  This is a hidden feature.  I want to document it when it supports full globs.

PR #1996, opened 2023-01-27
Linting

tlimoncelli:
Fixed these golint and staticcheck issues:

```
build/generate/featureMatrix.go:72:3: const ProviderDnsProvider should be ProviderDNSProvider
build/generate/functionTypes.go:158:6: exported type Function should have comment or be unexported
build/generate/functionTypes.go:167:6: exported type Param should have comment or be unexported
build/generate/functionTypes.go:196:9: if block ends with a return statement, so drop this else and outdent its block
commands/writeTypes.go:4:2: a blank import should be only in a main or test package, or have a comment justifying it
pkg/prettyzone/sorting.go:108:1: exported function LabelLess should have comment or be unexported
providers/hedns/hednsProvider.go:211:9: if block ends with a return statement, so drop this else and outdent its block
build/generate/featureMatrix.go:27:2: should replace loop with tableHeaders = append(tableHeaders, matrix.Features...) (S1011)
build/generate/featureMatrix.go:61:5: should omit comparison to bool constant, can be simplified to featureMap[featureName].HasFeature (S1002)
build/generate/featureMatrix.go:63:12: should omit comparison to bool constant, can be simplified to featureMap[featureName].Unimplemented (S1002)
pkg/diff2/analyze.go:255:10: empty branch (SA9003)
```

CC @cafferata

### DIFF
--- a/build/generate/featureMatrix.go
+++ b/build/generate/featureMatrix.go
@@ -24,9 +24,7 @@ func generateFeatureMatrix() error {
 func markdownTable(matrix *FeatureMatrix) (string, error) {
 	var tableHeaders []string
 	tableHeaders = append(tableHeaders, "Provider name")
-	for _, featureName := range matrix.Features {
-		tableHeaders = append(tableHeaders, featureName)
-	}
+	tableHeaders = append(tableHeaders, matrix.Features...)
 
 	var tableData [][]string
 	for _, providerName := range allProviderNames() {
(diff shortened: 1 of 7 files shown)


cafferata:
Thank you! How is the linting tool used? Why is this linting not guaranteed within GitHub actions?

cafferata:
Found it! `go fmt ./...` 😅

cafferata:
Perhaps a necessary step to include in the online documentation/pull request submission?

tlimoncelli:
Internally we're discussing moving to a new build pipeline. Until that decision is made, I don't want to invest too much in the pipeline.  However adding it to the docs is fine.

cafferata:
Thanks for your response, see https://github.com/StackExchange/dnscontrol/pull/2006 for a possible solution.

PR #2010, opened 2023-01-28
ROUTE53: (CHORE) Rename variables for clarity

tlimoncelli:
CC @tresni


tlimoncelli:
Just making these more clear before I re-use that code to make diff2 work.

PR #2012, opened 2023-01-29
ROUTE53: Rewrite to use diff2

tlimoncelli:
CC @tresni 

It is roughly working!  Now I just have to clean up the little stuff.


tlimoncelli:
This is ready for review, @tresni !

It doesn't do batching (i.e. it sends 1 RPC per change) but it passes all tests.  I'd like to merge this (pending any feedback) then come back later and add batching. (Or, not add batching?  I dunno how important it is.)

tresni:
R53's [API limits](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests) are very  low (5 requests per second). For mass operations, batching is basically a requirement.

Retries may alleviate this for some cases, but I'm not sure it would be a good strategy overall.  (e.g. if you have 1000 records changing, without batching you are looking at ~5 minutes to process all changes, vs a potential single request with batching.)

Issue #2020, opened 2023-01-31
Pull request template asks to run `go tidy`, which does not exist


Yannik:
The pull request template (https://github.com/StackExchange/dnscontrol/blob/master/.github/pull_request_template.md)  asks to run `go tidy`. However, that seems to be an invalid command:

```
$ go tidy            
go tidy: unknown command
Run 'go help' for usage.
```

cafferata:
Thanks for the feedback. Fixed by https://github.com/StackExchange/dnscontrol/pull/2021

PR #2006, opened 2023-01-28
DOCS: GitHub pull request template

cafferata:
@tlimoncelli https://github.com/StackExchange/dnscontrol/pull/1996#issuecomment-1407388782
>Internally we're discussing moving to a new build pipeline. Until that decision is made, I don't want to invest too much in the pipeline. However adding it to the docs is fine.

The Go format and Go mod tidy commands listed within the GitHub pull request template.

See [GitHub docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository).

PR #2021, opened 2023-01-31
DOCS: Pull request template: Fixed `go mod tidy` command

cafferata:
Fixes https://github.com/StackExchange/dnscontrol/issues/2020

The origin is copy/paste of comment https://github.com/StackExchange/dnscontrol/pull/2006#discussion_r1090024971

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,6 +3,6 @@ Before you submit a pull request, please make sure you've run the following comm
 
 go fmt ./...
 go generate
-go tidy
+go mod tidy
 !-->
 


Yannik:
For me, `go mod tidy` results in the following error:

```
github.com/juju/testing@v0.0.0-20210324180055-18c50b0c2098 requires
	github.com/juju/utils/v2@v2.0.0-20200923005554-4646bfea2ef1 requires
	launchpad.net/gocheck@v0.0.0-20140225173054-000000000087: unrecognized import path "launchpad.net/gocheck": GOVCS disallows using bzr for public launchpad.net/gocheck; see 'go help vcs'
```

I was able to fix that by setting the env variable `GOVCS='private:all,public:git|hg|bzr'`, however, it would perhaps be nice to have this documented.

tlimoncelli:
thanks for the fix!

Yannik:
> For me, `go mod tidy` results in the following error:
> 
> ```
> github.com/juju/testing@v0.0.0-20210324180055-18c50b0c2098 requires
> 	github.com/juju/utils/v2@v2.0.0-20200923005554-4646bfea2ef1 requires
> 	launchpad.net/gocheck@v0.0.0-20140225173054-000000000087: unrecognized import path "launchpad.net/gocheck": GOVCS disallows using bzr for public launchpad.net/gocheck; see 'go help vcs'
> ```
> 
> I was able to fix that by setting the env variable `GOVCS='private:all,public:git|hg|bzr'`. However, it would perhaps be nice to have this documented.


@tlimoncelli @cafferata Any thoughts about this?

cafferata:
Yes, I do have thoughts about this, because I have not experienced this problem myself before, I wonder to what extent this is a machine-specific problem. I wonder how many other DNSControl users are encountering this.

In my case the entire environment variable `GOVCS` (default installation) is empty.

```shell
go env | grep GOVCS
```

```text
GOVCS=""
```

Apart from that, I am very much in favor of describing these kinds of problems, but GoLang also helps with that. See https://github.com/golang/go/issues/41730 https://github.com/golang/go/issues/43596.

```shell
go help vcs
```
```text
The 'go get' command can run version control commands like git
to download imported code. This functionality is critical to the decentralized
Go package ecosystem, in which code can be imported from any server,
but it is also a potential security problem, if a malicious server finds a
way to cause the invoked version control command to run unintended code.

To balance the functionality and security concerns, the 'go get' command
by default will only use git and hg to download code from public servers.
But it will use any known version control system (bzr, fossil, git, hg, svn)
to download code from private servers, defined as those hosting packages
matching the GOPRIVATE variable (see 'go help private'). The rationale behind
allowing only Git and Mercurial is that these two systems have had the most
attention to issues of being run as clients of untrusted servers. In contrast,
Bazaar, Fossil, and Subversion have primarily been used in trusted,
authenticated environments and are not as well scrutinized as attack surfaces.

The version control command restrictions only apply when using direct version
control access to download code. When downloading modules from a proxy,
'go get' uses the proxy protocol instead, which is always permitted.
By default, the 'go get' command uses the Go module mirror (proxy.golang.org)
for public packages and only falls back to version control for private
packages or when the mirror refuses to serve a public package (typically for
legal reasons). Therefore, clients can still access public code served from
Bazaar, Fossil, or Subversion repositories by default, because those downloads
use the Go module mirror, which takes on the security risk of running the
version control commands using a custom sandbox.

The GOVCS variable can be used to change the allowed version control systems
for specific packages (identified by a module or import path).
The GOVCS variable applies when building package in both module-aware mode
and GOPATH mode. When using modules, the patterns match against the module path.
When using GOPATH, the patterns match against the import path corresponding to
the root of the version control repository.

The general form of the GOVCS setting is a comma-separated list of
pattern:vcslist rules. The pattern is a glob pattern that must match
one or more leading elements of the module or import path. The vcslist
is a pipe-separated list of allowed version control commands, or "all"
to allow use of any known command, or "off" to disallow all commands.
Note that if a module matches a pattern with vcslist "off", it may still be
downloaded if the origin server uses the "mod" scheme, which instructs the
go command to download the module using the GOPROXY protocol.
The earliest matching pattern in the list applies, even if later patterns
might also match.

For example, consider:

	GOVCS=github.com:git,evil.com:off,*:git|hg

With this setting, code with a module or import path beginning with
github.com/ can only use git; paths on evil.com cannot use any version
control command, and all other paths (* matches everything) can use
only git or hg.

The special patterns "public" and "private" match public and private
module or import paths. A path is private if it matches the GOPRIVATE
variable; otherwise it is public.

If no rules in the GOVCS variable match a particular module or import path,
the 'go get' command applies its default rule, which can now be summarized
in GOVCS notation as 'public:git|hg,private:all'.

To allow unfettered use of any version control system for any package, use:

	GOVCS=*:all

To disable all use of version control, use:

	GOVCS=*:off

The 'go env -w' command (see 'go help env') can be used to set the GOVCS
variable for future go command invocations.
```

As far as I am concerned there is no reason to include this in DNSControl documentation. Of course you are free to make a proposal.

tlimoncelli:
I'm in favor of this.

Most open source projects would say, "Hey! This is a generic Go issue. The user should find the answer on Stackoverflow.com or whatever".

However it is a high priority to me for this project to be friendly to people that are new to Go, new to open source, or just new in general.  This is the kind of blocker that can prevent a person from getting involved in the project.

My suggestion is that we create a page in the Developer Info section that is used to collect random tips and shorts bits of advice that don't deserve their own page.  If this particular issue goes away (suppose Go 2.0 eliminates this problem from ever happening) we can remove it from the documentation.

I do ask that the documentation we provide includes a link to the general solution or the Go documentation where appropriate.

Tom

PR #2023, opened 2023-01-31
HOSTINGDE: Implement SOA record (fixes #1972)

### LINKED ISSUE #1972: HOSTINGDE: SOA Support
The `HOSTINGDE` API does support setting custom SOA records for zones, however this needs to be done in the `ZoneConfig` and not as a regular record.

Currently, this is not implemented within dnscontrol. I'm creating this issue to track this feature.

@juliusrickert @membero Would you be interested in adding this?

Yannik:
This is not 100% done, there is one open question:

Currently, the SOA record expects the zone mailaddress as `hostmaster.example.org.`.

This is the standard syntax bind expects, which is the only other provider currently supporting SOA records.
However, hosting.de expects a rfc5322-formatted emailadress (`hostmaster@example.org`), as do probably most api-based providers do.

Additionally, converting from rfc5322 to bind-format is trivial, the other way around is not. (I have not implemented such a conversion yet due to this)

Therefore, I'd suggest expecting a rfc5322 formatted emailaddress for `SOA(..)` and only converting it within bind.

### DIFF
--- a/providers/hostingde/api.go
+++ b/providers/hostingde/api.go
@@ -18,6 +18,7 @@ type hostingdeProvider struct {
 	ownerAccountID string
 	baseURL        string
 	nameservers    []string
+	defaultSoa     soaValues
 }
 
 func (hp *hostingdeProvider) getDomainConfig(domain string) (*domainConfig, error) {
--- a/providers/hostingde/api.go
+++ b/providers/hostingde/api.go
@@ -125,12 +125,7 @@ func (hp *hostingdeProvider) updateNameservers(nss []string, domain string) func
 	}
 }
 
-func (hp *hostingdeProvider) updateRecords(domain string, create, del, mod diff.Changeset) error {
-	zc, err := hp.getZoneConfig(domain)
-	if err != nil {
-		return err
-	}
-
+func (hp *hostingdeProvider) updateZone(zc *zoneConfig, create, del, mod diff.Changeset) error {
 	toAdd := []*record{}
 	for _, c := range create {
 		r := recordToNative(c.Desired)
@@ -158,7 +153,7 @@ func (hp *hostingdeProvider) updateRecords(domain string, create, del, mod diff.
 		RecordsToModify: toModify,
 	}
 
-	_, err = hp.get("dns", "zoneUpdate", params)
+	_, err := hp.get("dns", "zoneUpdate", params)
 	if err != nil {
 		return err
 	}
--- a/providers/hostingde/api.go
+++ b/providers/hostingde/api.go
@@ -125,44 +125,6 @@ func (hp *hostingdeProvider) updateNameservers(nss []string, domain string) func
 	}
 }
 
-func (hp *hostingdeProvider) getRecords(domain string) ([]*record, error) {
-	zc, err := hp.getZoneConfig(domain)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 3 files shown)


tlimoncelli:
> Therefore, I'd suggest expecting a rfc5322 formatted emailaddress for SOA(..) and only converting it within bind.

I see the dilemma.  However we have to do this in a backwards-compatible way.

While is it non-trivial to turn bind-format (`foo.example.com`) to RFC5322 (`foo@example.com`)?  Why not just convert the first `.` to `@`?  Yes, an email address could have `.` in it, but that's pretty rare.

Yannik:
> > Therefore, I'd suggest expecting a rfc5322 formatted emailaddress for SOA(..) and only converting it within bind.
> 
> I see the dilemma. However we have to do this in a backwards-compatible way.
> 
> While is it non-trivial to turn bind-format (`foo.example.com`) to RFC5322 (`foo@example.com`)? Why not just convert the first `.` to `@`? Yes, an email address could have `.` in it, but that's pretty rare.

Both the user-part and the domain-part of an email address can contain dots. Look at this [soa parser](https://github.com/wesinator/soaparse) for example. It's offering converting either the first dot or the last dot, since it is just impossible to know which is correct, it's simply a bad format.

I think it would really be better to store rfc5322-email-addresses. For the bind provider, one might implement a check to see whether the email-address includes `@` or not, and convert based on that. This could then be dropped at the next major release (breaking change).

tlimoncelli:
> I think it would really be better to store rfc5322-email-addresses. For the bind provider, one might implement a check to see whether the email-address includes `@` or not, and convert based on that. This could then be dropped at the next major release (breaking change).

Sounds good!  It should also print a warning that the change is coming.

Tom

dkim1970:
Just remember reading about this not to long ago in RFC-1035 [Section-8](https://datatracker.ietf.org/doc/html/rfc1035#section-8 which specifics the zonefile format

> If the <local-part> contains dots or other special characters, its
> representation in a master file will require the use of backslash
> quoting to ensure that the domain name is properly encoded.  For
> example, the mailbox Action.domains@ISI.EDU would be represented as
> Action\.domains.ISI.EDU. 

So converting from RFC-1035 to RFC-5322 format wouldn't it just be looking for a backslash otherwise just sub the first . with a @

Yannik:
> Just remember reading about this not to long ago in RFC-1035 [Section-8](https://datatracker.ietf.org/doc/html/rfc1035#section-8 which specifics the zonefile format
> 
> > If the  contains dots or other special characters, its
> > representation in a master file will require the use of backslash
> > quoting to ensure that the domain name is properly encoded.  For
> > example, the mailbox [Action.domains@ISI.EDU](mailto:Action.domains@ISI.EDU) would be represented as
> > Action.domains.ISI.EDU.
> 
> So converting from RFC-1035 to RFC-5322 format wouldn't it just be looking for a backslash otherwise just sub the first . with a @

Thank you, I have taken the RFC you mentioned into account regarding the conversion. However, it still feels more correct to store the actual email-address and convert in the few places it may be necessary, than the other way around.

PR in #2037

Yannik:
@tlimoncelli With #2037, I have now rebased this on master and it should be ready to merge.

tlimoncelli (approved):
Looks good!

PR #2030, opened 2023-01-31
Remove all template tag lines from the .d.ts file

j-f1:
While looking at https://github.com/StackExchange/dnscontrol/pull/2029, I noticed that some template tags had made it into the generated doc comments.


cafferata:
Thanks! 🙏 I had the same idea, you are ahead of me with the technical implementation. 👏

PR #2036, opened 2023-02-01
HOSTINGDE: Fix dnssec error resulting from non-go-conformant comparison of three values

Yannik:
This fixes an dnssec error resulting from non-go-conformant comparison of three values (introduced due to my limited go experience :-))

### DIFF
--- a/providers/hostingde/hostingdeProvider.go
+++ b/providers/hostingde/hostingdeProvider.go
@@ -159,7 +159,7 @@ func (hp *hostingdeProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 	var DnsSecOptions *dnsSecOptions = nil
 
 	// ensure that publishKsk is set for domains with AutoDNSSec
-	if existingAutoDNSSecEnabled == desiredAutoDNSSecEnabled == true {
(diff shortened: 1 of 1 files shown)


providers/hostingde/hostingdeProvider.go
```diff
@@ -159,7 +159,7 @@ func (hp *hostingdeProvider) GetDomainCorrections(dc *models.DomainConfig) ([]*m
 	var DnsSecOptions *dnsSecOptions = nil
 
 	// ensure that publishKsk is set for domains with AutoDNSSec
-	if existingAutoDNSSecEnabled == desiredAutoDNSSecEnabled == true {
+	if existingAutoDNSSecEnabled == desiredAutoDNSSecEnabled && desiredAutoDNSSecEnabled == true {
```
tlimoncelli, line 162: If desiredAutoDNSSecEnabled is a bool, you don't need to compare it to true.

Isn't this equivalent?

```
if desiredAutoDNSSecEnabled && existingAutoDNSSecEnabled {
```

Yannik: Absolutely right. That's better.

PR #2037, opened 2023-02-01
Expect SOA mailbox in hostmaster@example.org format instead of hostmaster.example.org

Yannik:
As discussed in https://github.com/StackExchange/dnscontrol/pull/2023#issuecomment-1410645574

### DIFF
--- a/documentation/functions/domain/SOA.md
+++ b/documentation/functions/domain/SOA.md
@@ -28,6 +28,8 @@ D("example.com", REG_THIRDPARTY, DnsProvider("DNS_BIND"),
 );
 ```
 
+The email address should be specified like a normal RFC822/RFC5322 address (user@hostname.com). It will be converted into the required format (e.g. BIND format: `user.hostname.com`) by the provider as required. This has the benefit of being more human-readable plus DNSControl can properly handle escaping and other issues.
+
 ## Notes
 * Previously, the accepted format for the SOA mailbox field was `hostmaster.example.org`. This has been changed to `hostmaster@example.org`
 * The serial number is managed automatically.  It isn't even a field in `SOA()`.
--- a/documentation/functions/domain/SOA.md
+++ b/documentation/functions/domain/SOA.md
@@ -24,12 +24,12 @@ parameter_types:
 
 ```javascript
 D("example.com", REG_THIRDPARTY, DnsProvider("DNS_BIND"),
-  SOA("@", "ns3.example.org.", "hostmaster.example.org.", 3600, 600, 604800, 1440),
+  SOA("@", "ns3.example.org.", "hostmaster@example.org", 3600, 600, 604800, 1440),
 );
 ```
 
 ## Notes
-
+* Previously, the accepted format for the SOA mailbox field was `hostmaster.example.org`. This has been changed to `hostmaster@example.org`
 * The serial number is managed automatically.  It isn't even a field in `SOA()`.
 * Most providers automatically generate SOA records.  They will ignore any `SOA()` statements.
 
--- a/providers/bind/soa.go
+++ b/providers/bind/soa.go
@@ -31,8 +31,8 @@ func makeSoa(origin string, defSoa *SoaDefaults, existing, desired *models.Recor
 		user_part = strings.ReplaceAll(user_part, ".", "\\.")
 		soaMail = user_part + "." + domain_part
 	} else {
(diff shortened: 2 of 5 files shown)


documentation/functions/domain/SOA.md
```diff
@@ -24,12 +24,12 @@ parameter_types:
...
 ```javascript
 D("example.com", REG_THIRDPARTY, DnsProvider("DNS_BIND"),
-  SOA("@", "ns3.example.org.", "hostmaster.example.org.", 3600, 600, 604800, 1440),
+  SOA("@", "ns3.example.org.", "hostmaster@example.org", 3600, 600, 604800, 1440),
 );
 ```
 
 ## Notes
-
+* Previously, the accepted format for the SOA mailbox field was `hostmaster.example.org`. This has been changed to `hostmaster@example.org`
```
tlimoncelli, line 32: Suggestion:

* Previously, the accepted format for the SOA mailbox field was `hostmaster.example.org`. This has been changed to `hostmaster@example.org`. DNSControl will change the `@` to `.` when generating the SOA record.

providers/bind/soa.go
```diff
@@ -19,10 +23,22 @@ func makeSoa(origin string, defSoa *SoaDefaults, existing, desired *models.Recor
 		desired = &models.RecordConfig{}
 	}
 
+	soaMail := firstNonNull(desired.SoaMbox, existing.SoaMbox, defSoa.Mbox, "DEFAULT_NOT_SET.")
```
tlimoncelli, line 26: This should be a separate function, with unit tests.

providers/bind/soa.go
```diff
@@ -19,10 +23,22 @@ func makeSoa(origin string, defSoa *SoaDefaults, existing, desired *models.Recor
...
+	if strings.Contains(soaMail, "@") {
+		res := strings.SplitN(soaMail, "@", 2)
+		user_part, domain_part := res[0], res[1]
+		// RFC-1035 [Section-8]
+		user_part = strings.ReplaceAll(user_part, ".", "\\.")
+		soaMail = user_part + "." + domain_part
+	} else {
```
tlimoncelli, line 33: change `Note:` to `WARNING:` (in both places)

documentation/functions/domain/SOA.md
Yannik: The latter sentence is only accurate for the bind provider. I'd leave that out.

providers/bind/soa.go
Yannik: Is there a good place to globally declare this function, so all providers can use it? (Additionally, which function name would you think would be good here?)

providers/bind/soa.go
Yannik: Done

providers/bind/soa.go
tlimoncelli: I can't find an existing place that is good. However there is `pkg/txtutil` so why not create `pkg/soautil`?

Suggestion: create `pkg/soautil/soautil.go` and put the function in there.

Yannik: Done

Yannik:
@tlimoncelli I have extracted the function and added tests as you suggested.

providers/bind/soa.go
```diff
@@ -19,10 +24,18 @@ func makeSoa(origin string, defSoa *SoaDefaults, existing, desired *models.Recor
 		desired = &models.RecordConfig{}
 	}
 
+	soaMail := firstNonNull(desired.SoaMbox, existing.SoaMbox, defSoa.Mbox, "DEFAULT_NOT_SET.")
+	if strings.Contains(soaMail, "@") {
+		soaMail = soautil.RFC5322MailToBind(soaMail)
+	} else {
+		fmt.Println("WARNING: SOA hostmaster address must be in the format hostmaster@example.com")
+		fmt.Println("WARNING: hostmaster.example.com is deprecated and will be dropped in a future version")
```
tlimoncelli, line 32: Please document this in documentation/functions/domain/SOA.md

Something like: NOTE: The email address should be specified like a normal RFC822/RFC5322 address (`user@hostname.com`). It will be converted into the BIND format (`user.hostname.com`) internally.  This has the benefit of being more human-readable plus DNSControl can properly handle escaping and other issues.

Yannik: @tlimoncelli done

tlimoncelli:
Looks good!

PR #2038, opened 2023-02-01
Rename DomainCreator to the more accurate ZoneCreator

Yannik:
`DomainCreator` and `EnsureDomainExists` are misleading, because it actually creates zones, not domains.

### DIFF
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -151,12 +151,12 @@ DomainLoop:
 						out.Warnf("DEBUG: zones: %v\n", zones)
 						out.Warnf("DEBUG: Name: %v\n", domain.Name)
 
-						out.Warnf("Domain '%s' does not exist in the '%s' profile and will be added automatically.\n", domain.Name, provider.Name)
+						out.Warnf("Zone '%s' does not exist in the '%s' profile and will be added automatically.\n", domain.Name, provider.Name)
 						continue // continue with next provider, as we can not determine corrections without an existing zone
 					}
-				} else if creator, ok := provider.Driver.(providers.DomainCreator); ok && push {
+				} else if creator, ok := provider.Driver.(providers.ZoneCreator); ok && push {
 					// this is the actual push, ensure domain exists at DSP
-					if err := creator.EnsureDomainExists(domain.Name); err != nil {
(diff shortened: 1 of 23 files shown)


tlimoncelli:
Thanks for fixing this inconsistency.  Please also fix the comments and other strings that surround these changes.

Yannik:
@tlimoncelli done.

tlimoncelli (approved):

commands/previewPush.go
```diff
@@ -154,9 +154,9 @@ DomainLoop:
 						out.Warnf("Domain '%s' does not exist in the '%s' profile and will be added automatically.\n", domain.Name, provider.Name)
```
tlimoncelli, line 154: Change "Domain" to "Zone" in out.Warnf

providers/dnsmadeeasy/dnsMadeEasyProvider.go
```diff
@@ -177,8 +177,8 @@ func (api *dnsMadeEasyProvider) GetDomainCorrections(dc *models.DomainConfig) ([
 	return corrections, nil
 }
 
-// EnsureDomainExists returns an error if domain doesn't exist.
-func (api *dnsMadeEasyProvider) EnsureDomainExists(domainName string) error {
+// EnsureZoneExists returns an error if domain doesn't exist.
+func (api *dnsMadeEasyProvider) EnsureZoneExists(domainName string) error {
```
tlimoncelli, line 181: domainName should be zoneName.

If we're going to make this change, we might as well change the "domain string" parameter in most of these other functions. I wouldn't change much more (like the api.domainExists() call below)

Yannik: done

commands/previewPush.go
Yannik: done

Yannik:
@tlimoncelli I have resolved the points you have raised.

Yannik:
I have reverted the `domain->zoneName` renaming. IMO, "Creating a zone for a domain" is quite accurate. Additionally, this creates major inconsistencies with other variable names.

Issue #1976, opened 2023-01-23
Feature: Exclusive mode (remove obsolete domains/zones)


Yannik:
Currently, it is not possible to delete obsolete Domains/Zones by simply deleting them from the `dnsconfig.js` (I think).

It would be great if there was an `exclusive` Parameter that could be passed to `NewRegistrar` and `NewDnsProvider` that would allow removing all domains/zones not configured within dnscontrol.

tlimoncelli:
Sadly we don't be adding this feature.

The risk of accidentally deleting a domain outweighs the benefit. 

That should be all the reason needed. However I'll also add:

* Most providers have different processes for deleting domains. Implementing this feature in a provider-agnostic way is difficult. The time spent designing this would outweigh the benefit.
* There's no harm in leaving a domain around unused.  If nothing delegates to it, the internet doesn't use it.
* DNSControl doesn't manage any registration or billing aspects of a domain.  If the domain needs to be deleted for billing reasons, this isn't the right tool.

Thank you for the suggestion but I have to reject this.

Tom

Yannik:
Hi Tom,

IaC components should, in my opinion, provide the following benefits:

   -  idempotency,
   - diff support,
   - check mode compatibility,
   - correct change signaling, and
   - lifecycle management.

The last one, lifecycle management, is the one we are talking about here: Being able to delete outdated objects.

We have a customer with hundreds of domains for various projects, regional entities etc. Domains (and their respective zones!) are added and deleted all the time. Of course, managing deletes by hand is possible, but it is, depending on provider, quite a tedious process.

We are currently using a self-built ansible module to deal with all things domains & zones (including deletion ;-)), however we would very much like to switch to dnscontrol to benefit from the possibilities of a interpreted language instead of just static yaml. Unfortunately, it would really be a step back to losing lifecycle management features.

> The risk of accidentally deleting a domain outweighs the benefit.

For deleting domains I do agree that there is some risk involved. Which might outweigh the benefit, if you do not have a good change management process. 

For zones I cannot see a risk increase. Currently, you could accidentally delete all records by deleting them from the config, then you can delete the zone by accidentally deleting it from the config (and ignoring the preview!).  
To fix it: Just run `git reset --hard` and re-deploy.

To reduce risks further: Of course this would only be enabled with explicit configuration, and not by default. (Making it a default would be quite risky, I agree!)

I would like to add some more benefits to this discussion as well:  
[...]

PR #2050, opened 2023-02-02
 FEATURE: New flag --delete-unmanaged: Remove unmanaged domains & zones (implements #1976)

Yannik:
@tlimoncelli 

This is an implementation of the feature we discussed in #1976. 

The feature can be used like this:

```
dnscontrol preview --delete-unmanaged-domains
``` 

```
dnscontrol preview --delete-unmanaged-zones
``` 

```
dnscontrol preview --delete-unmanaged # same as --delete-unmanaged-domains --delete-unmanaged-zones
``` 

The output looks like this:

```
Checking Domain Removal:
Removing domain mydomain.com from provider hosting.de
Checking Zone Removal:
Removing zone mydomain.com from provider hosting.de
```

### DIFF
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -38,6 +38,7 @@ type PreviewArgs struct {
 	WarnChanges            bool
 	NoPopulate             bool
 	Full                   bool
+	DeleteUnmanaged        bool
 	DeleteUnmanagedDomains bool
 	DeleteUnmanagedZones   bool
 }
@@ -66,6 +67,12 @@ func (args *PreviewArgs) flags() []cli.Flag {
 		Destination: &args.DeleteUnmanagedDomains,
 		Usage:       `Use this flag to delete unmanaged domains`,
 	})
+	flags = append(flags, &cli.BoolFlag{
+		Name:        "delete-unmanaged",
+		Destination: &args.DeleteUnmanaged,
+		Usage:       `Use this flag to delete unmanaged domains and zones`,
+	})
+
 	flags = append(flags, &cli.BoolFlag{
 		Name:        "delete-unmanaged-zones",
 		Destination: &args.DeleteUnmanagedZones,
@@ -229,7 +236,7 @@ DomainLoop:
 		anyErrors = printOrRunCorrections(domain.Name, domain.RegistrarName, corrections, out, push, interactive, notifier) || anyErrors
 	}
 
-	if args.DeleteUnmanagedDomains {
+	if args.DeleteUnmanaged || args.DeleteUnmanagedDomains {
 		fmt.Printf("Checking Domain Removal:\n")
 		for registrarName, registrar := range createdRegistrars {
 			domainLister, ok := registrar.(providers.DomainLister)
@@ -254,7 +261,7 @@ DomainLoop:
 		}
 	}
 
-	if args.DeleteUnmanagedZones {
+	if args.DeleteUnmanaged || args.DeleteUnmanagedZones {
 		fmt.Printf("Checking Zone Removal:\n")
 
 		for providerName, provider := range createdProviders {
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -281,7 +281,7 @@ func DeleteUnmanagedDomains(cfg *models.DNSConfig, createdRegistrars map[string]
 		for _, domain := range deployedDomains {
 			if !IsDomainManagedByRegistrar(cfg, domain, registrarName) {
 
-				fmt.Printf("Removing from provider %s: domain %s\n", registrarName, domain)
+				fmt.Printf("Removing domain from provider %s: %s\n", registrarName, domain)
 				*totalCorrections += 1
 				if domainRemover, ok := registrar.(providers.DomainRemover); ok && push {
 					err := domainRemover.EnsureDomainAbsent(domain)
@@ -311,7 +311,7 @@ func DeleteUnmanagedZones(cfg *models.DNSConfig, createdProviders map[string]pro
 		}
 		for _, zone := range deployedZones {
 			if !IsZoneManagedByProvider(cfg, zone, providerName) {
-				fmt.Printf("Removing from provider %s: zone %s\n", providerName, zone)
+				fmt.Printf("Removing zone from provider %s: %s\n", providerName, zone)
 				*totalCorrections += 1
 				if zoneRemover, ok := provider.(providers.ZoneRemover); ok && push {
 					err := zoneRemover.EnsureZoneAbsent(zone)
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -237,57 +237,19 @@ DomainLoop:
 	}
 
 	if args.DeleteUnmanaged || args.DeleteUnmanagedDomains {
-		fmt.Printf("Checking Domain Removal:\n")
-		for registrarName, registrar := range createdRegistrars {
-			domainLister, ok := registrar.(providers.DomainLister)
-			if !ok {
-				out.Warnf("provider type of %s cannot list domains\n", registrarName)
-				continue
-			}
-			deployedDomains, err := domainLister.ListDomains()
-			if err != nil {
-				return fmt.Errorf("failed ListZones: %w\n", err)
-			}
-			for _, domain := range deployedDomains {
-				if !IsDomainManagedByRegistrar(cfg, domain, registrarName) {
-					fmt.Printf("Removing domain %s from provider %s\n", domain, registrarName)
-					totalCorrections += 1
-					if remover, ok := registrar.(providers.DomainRemover); ok && push {
-						if err := remover.EnsureDomainAbsent(domain); err != nil {
-							out.Errorf("Error deleting domain: %s\n", err)
-						}
-					}
-				}
-			}
+		err := DeleteUnmanagedDomains(cfg, createdRegistrars, push, out, &totalCorrections)
+		if err != nil {
+			return err
 		}
 	}
 
 	if args.DeleteUnmanaged || args.DeleteUnmanagedZones {
-		fmt.Printf("Checking Zone Removal:\n")
-
-		for providerName, provider := range createdProviders {
-			zoneLister, ok := provider.(providers.ZoneLister)
-			if !ok {
-				out.Warnf("provider type of %s cannot list zones\n", providerName)
-				continue
-			}
-			deployedZones, err := zoneLister.ListZones()
-			if err != nil {
-				return fmt.Errorf("failed ListZones: %w\n", err)
-			}
-			for _, zone := range deployedZones {
-				if !IsZoneManagedByProvider(cfg, zone, providerName) {
-					fmt.Printf("Removing zone %s from provider %s\n", zone, providerName)
-					totalCorrections += 1
-					if remover, ok := provider.(providers.ZoneRemover); ok && push {
-						if err := remover.EnsureZoneAbsent(zone); err != nil {
-							out.Errorf("Error deleting zone: %s\n", err)
-						}
-					}
-				}
-			}
+		err := DeleteUnmanagedZones(cfg, createdProviders, push, out, &totalCorrections)
+		if err != nil {
+			return err
 		}
 	}
+
 	if os.Getenv("TEAMCITY_VERSION") != "" {
 		fmt.Fprintf(os.Stderr, "##teamcity[buildStatus status='SUCCESS' text='%d corrections']", totalCorrections)
 	}
@@ -298,6 +260,62 @@ DomainLoop:
 	}
 	if totalCorrections != 0 && args.WarnChanges {
 		return fmt.Errorf("there are pending changes")
+	}
+	return nil
+}
+
+func DeleteUnmanagedDomains(cfg *models.DNSConfig, createdRegistrars map[string]providers.Registrar, push bool, out printer.CLI, totalCorrections *int) error {
+	fmt.Printf("Checking Domain Removal:\n")
+	for registrarName, registrar := range createdRegistrars {
+		domainLister, ok := registrar.(providers.DomainLister)
+		if !ok {
+			out.Warnf("--purge-unmanaged-domains not implemented: provider %s\n", registrarName)
+			continue
+		}
+		deployedDomains, err := domainLister.ListDomains()
+		if err != nil {
+			return fmt.Errorf("failed ListDomains for provider %s: %w\n", registrarName, err)
+		}
+		for _, domain := range deployedDomains {
+			if !IsDomainManagedByRegistrar(cfg, domain, registrarName) {
+				fmt.Printf("Removing from provider %s: domain %s\n", registrarName, domain)
+				*totalCorrections += 1
+				if domainRemover, ok := registrar.(providers.DomainRemover); ok && push {
+					err := domainRemover.EnsureDomainAbsent(domain)
+					if err != nil {
+						out.Errorf("Error deleting domain: %s\n", err)
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func DeleteUnmanagedZones(cfg *models.DNSConfig, createdProviders map[string]providers.DNSServiceProvider, push bool, out printer.CLI, totalCorrections *int) error {
+	fmt.Printf("Checking Zone Removal:\n")
+	for providerName, provider := range createdProviders {
+		zoneLister, ok := provider.(providers.ZoneLister)
+		if !ok {
+			out.Warnf("--purge-unmanaged-zones not implemented: provider %s\n", providerName)
+			continue
+		}
+		deployedZones, err := zoneLister.ListZones()
+		if err != nil {
+			return fmt.Errorf("failed ListZones for provider %s: %w\n", providerName, err)
+		}
+		for _, zone := range deployedZones {
+			if !IsZoneManagedByProvider(cfg, zone, providerName) {
+				fmt.Printf("Removing from provider %s: zone %s\n", providerName, zone)
+				*totalCorrections += 1
+				if zoneRemover, ok := provider.(providers.ZoneRemover); ok && push {
+					err := zoneRemover.EnsureZoneAbsent(zone)
+					if err != nil {
+						out.Errorf("Error deleting zone: %s\n", err)
+					}
+				}
+			}
+		}
 	}
 	return nil
 }
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -363,9 +363,6 @@ type ProviderState struct {
 func InitializeProviders(cfg *models.DNSConfig, providerConfigs map[string]map[string]string, notifyFlag bool) ProviderState {
 	var notificationCfg map[string]string
 	result := ProviderState{}
-	defer func() {
-		result.notifier = notifications.Init(notificationCfg)
-	}()
 	if notifyFlag {
 		notificationCfg = providerConfigs["notifications"]
 	}
@@ -418,6 +415,7 @@ func InitializeProviders(cfg *models.DNSConfig, providerConfigs map[string]map[s
 	}
 	result.createdRegistrars = registrars
 	result.createdDnsProviders = dnsProviders
+	result.notifier = notifications.Init(notificationCfg)
 	return result
 }
 
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -229,28 +229,20 @@ DomainLoop:
 		anyErrors = printOrRunCorrections(domain.Name, domain.RegistrarName, corrections, out, push, interactive, notifier) || anyErrors
 	}
 
-	managedDomains := make([]string, 0, len(cfg.Domains))
-	for _, zone := range cfg.Domains {
-		managedDomains = append(managedDomains, zone.Name)
-	}
-
 	if args.DeleteUnmanagedDomains {
 		fmt.Printf("Checking Domain Removal:\n")
-		if len(createdRegistrars) != 1 {
-			return fmt.Errorf("To delete unmanaged domains, exactly one registrar must be used")
-		}
-		for name, registrar := range createdRegistrars {
+		for registrarName, registrar := range createdRegistrars {
 			domainLister, ok := registrar.(providers.DomainLister)
 			if !ok {
-				fmt.Errorf("provider type of %s cannot list zones\n", name)
+				fmt.Errorf("provider type of %s cannot list zones\n", registrarName)
 			}
 			deployedDomains, err := domainLister.ListDomains()
 			if err != nil {
 				return fmt.Errorf("failed ListZones: %w\n", err)
 			}
 			for _, domain := range deployedDomains {
-				if !slices.Contains(managedDomains, domain) {
-					fmt.Printf("Removing domain %s from provider %s\n", domain, name)
+				if !IsDomainManagedByRegistrar(cfg, domain, registrarName) {
+					fmt.Printf("Removing domain %s from provider %s\n", domain, registrarName)
 					totalCorrections += 1
 					if remover, ok := registrar.(providers.DomainRemover); ok && push {
 						if err := remover.EnsureDomainAbsent(domain); err != nil {
@@ -265,18 +257,18 @@ DomainLoop:
 	if args.DeleteUnmanagedZones {
 		fmt.Printf("Checking Zone Removal:\n")
 
-		for name, provider := range createdProviders {
+		for providerName, provider := range createdProviders {
 			zoneLister, ok := provider.(providers.ZoneLister)
 			if !ok {
-				fmt.Errorf("provider type of %s cannot list zones\n", name)
+				fmt.Errorf("provider type of %s cannot list zones\n", providerName)
 			}
 			deployedZones, err := zoneLister.ListZones()
 			if err != nil {
 				return fmt.Errorf("failed ListZones: %w\n", err)
 			}
 			for _, zone := range deployedZones {
-				if !slices.Contains(managedDomains, zone) {
-					fmt.Printf("Removing zone %s from provider %s\n", zone, name)
+				if !IsZoneManagedByProvider(cfg, zone, providerName) {
+					fmt.Printf("Removing zone %s from provider %s\n", zone, providerName)
 					totalCorrections += 1
 					if remover, ok := provider.(providers.ZoneRemover); ok && push {
 						if err := remover.EnsureZoneAbsent(zone); err != nil {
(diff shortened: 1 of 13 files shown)


commands/previewPush.go
```diff
@@ -231,7 +302,7 @@ DomainLoop:
 }
 
 // InitializeProviders takes (fully processed) configuration and instantiates all providers and returns them.
-func InitializeProviders(cfg *models.DNSConfig, providerConfigs map[string]map[string]string, notifyFlag bool) (notify notifications.Notifier, err error) {
+func InitializeProviders(cfg *models.DNSConfig, providerConfigs map[string]map[string]string, notifyFlag bool) (createdRegistrars map[string]providers.Registrar, createdDnsProviders map[string]providers.DNSServiceProvider, notify notifications.Notifier, err error) {
```
tlimoncelli, line 305: This signature has changed a few times in  the last few years.  Each time we add more return values then update all the other code as appropriate.  My suggestion is that we return a pointer to a structure that has fields for all the current return values.  In the future we can add new fields and the signature won't change.

commands/previewPush.go
```diff
@@ -216,6 +228,65 @@ DomainLoop:
...
 		anyErrors = printOrRunCorrections(domain.Name, domain.RegistrarName, corrections, out, push, interactive, notifier) || anyErrors
 	}
+
+	managedDomains := make([]string, 0, len(cfg.Domains))
+	for _, zone := range cfg.Domains {
+		managedDomains = append(managedDomains, zone.Name)
+	}
+
+	if args.DeleteUnmanagedDomains {
+		fmt.Printf("Checking Domain Removal:\n")
```
tlimoncelli, line 238: Suppose a user has a provider that doesn't support `ListZones()` and they provide the` --delete-unmanaged-zones` flag.  They won't see the `Checking Domain Removal:` message.  They might wonder why "the flag isn't working".  Can they get a message that lets them know they've used a flag that isn't supported by their provider?

commands/previewPush.go
```diff
@@ -216,6 +228,65 @@ DomainLoop:
...
+		if len(createdRegistrars) != 1 {
+			return fmt.Errorf("To delete unmanaged domains, exactly one registrar must be used")
```
tlimoncelli, line 240: why is that a requirement?

Yannik:
> 1. Deleting unmanaged domains is only possible if exactly one domain registrar is being used (otherwise the command throws an error)
> 
>  2. Deleting unmanaged zones is possible with any number of dns providers, with the limitation that all dns providers need to be configured to serve all the configured zones.

Actually, after reviewing this again, the code I submitted in this PR was the implementation of a different approach.

I will go over this whole thing again tomorrow. (I'm currently in GMT+8, so it's really late here :-))
Thanks for the code review, I will try to address your comments tomorrow as well.

tlimoncelli:
> I will go over this whole thing again tomorrow. (I'm currently in GMT+8, so it's really late here :-))

Sleep is more important that dnscontrol!  I promise!

Yannik:
Hi @tlimoncelli

I have gone over this again and pushed a new version of my PR.

Some of the things from my OP were incorrect and based on wrong assumptions. I think I got a bit caught up with understanding go and lost sight of the bigger picture.

With this new version any number of registrars/dns providers can be correctly checked for unmanaged domains/zones. (Therefore, the limit of one registrar has been removed)

We can now also use `--delete-unmanaged` which acts as an alias for the rather unwieldy `--delete-unmanaged-domains --delete-unmanaged-zones`.

It would be great to get a new review on this. I have also added documentation for this feature with my new commits.

Regarding the output of the command, I think there is room for improvement. If you have any good ideas, please bring them forward!

commands/previewPush.go
Yannik: My mistake, this should not be required.

commands/previewPush.go
```diff
@@ -216,6 +228,65 @@ DomainLoop:
...
+			if !ok {
+				fmt.Errorf("provider type of %s cannot list zones\n", name)
+			}
+			deployedDomains, err := domainLister.ListDomains()
+			if err != nil {
+				return fmt.Errorf("failed ListZones: %w\n", err)
+			}
+			for _, domain := range deployedDomains {
+				if !slices.Contains(managedDomains, domain) {
+					fmt.Printf("Removing domain %s from provider %s\n", domain, name)
```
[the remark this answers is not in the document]
Yannik: IMO, preview outputting the same messages as push should be expected.

Which message would you suggest for preview?

commands/previewPush.go
Yannik: There is a check for that just two lines down.

tlimoncelli:
Looking good so far!

Some notes:

1. My "preview" takes at least a minute to run through all my domains. Waiting for the zone/domain report at the end was a drag.  Maybe the domain/zone deletion should happen first?  (Or maybe this is whyOr maybe this means we should implement the sub-commands instead of putting this in preview/push.

2. I'm getting panics when I run the code:

I got the panics with all the providers I tested it with: NONE, CSCGLOBAL, NAMEDOTCOM

The panic happens on this line:

`      deployedDomains, err := domainLister.ListDomains()`

Here's the panic:

```
dnscontrol preview --delete-unmanaged
Checking Domain Removal:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x178e7a4]

goroutine 1 [running]:
github.com/StackExchange/dnscontrol/v3/commands.run({{{{0x246267e, 0xc}, {0x0, 0x0}, 0x0, {{...}, {...}, 0x0}}, {0x0, 0x0}}, ...}, ...)
	/Users/tlimoncelli/git/dnscontrol/commands/previewPush.go:246 +0xfe4
github.com/StackExchange/dnscontrol/v3/commands.Preview(...)
	/Users/tlimoncelli/git/dnscontrol/commands/previewPush.go:119
github.com/StackExchange/dnscontrol/v3/commands.glob..func7.1(0xc00038e160?)
	/Users/tlimoncelli/git/dnscontrol/commands/previewPush.go:26 +0xa6
github.com/urfave/cli/v2.(*Command).Run(0xc00038e160, 0xc0006978c0, {0xc0003ba840, 0x2, 0x2})
	/Users/tlimoncelli/go/pkg/mod/github.com/urfave/cli/v2@v2.24.2/command.go:273 +0xa42
github.com/urfave/cli/v2.(*Command).Run(0xc00038edc0, 0xc000697740, {0xc0000401b0, 0x3, 0x3})
	/Users/tlimoncelli/go/pkg/mod/github.com/urfave/cli/v2@v2.24.2/command.go:266 +0xc97
github.com/urfave/cli/v2.(*App).RunContext(0xc0000423c0, {0x26e0b20?, 0xc000046098}, {0xc0000401b0, 0x3, 0x3})
	/Users/tlimoncelli/go/pkg/mod/github.com/urfave/cli/v2@v2.24.2/app.go:332 +0x616
github.com/urfave/cli/v2.(*App).Run(...)
	/Users/tlimoncelli/go/pkg/mod/github.com/urfave/cli/v2@v2.24.2/app.go:309
github.com/StackExchange/dnscontrol/v3/commands.Run({0xc0003dd560?, 0x2460c00?})
	/Users/tlimoncelli/git/dnscontrol/commands/commands.go:70 +0x325
main.main()
	/Users/tlimoncelli/git/dnscontrol/main.go:34 +0x134
```

Yannik:
@tlimoncelli 
Regarding 2: this is surely due to `ListDomains()` not being implemented for these providers. However, it shouldn't crash but throw an error.

```
			domainLister, ok := registrar.(providers.DomainLister)
			if !ok {
				fmt.Errorf("provider type of %s cannot list domains\n", registrarName)
			}
```

Do you have any idea why this is not working? I copied this from the getZone [ZoneLister check](https://github.com/StackExchange/dnscontrol/blob/master/commands/getZones.go#L173). I'm not 100% certain how this check works, I just assumed it would check whether the methods of the interface are implemented.

You could also try `--delete-unmanaged-zones`, and skip checking domains for now.

tlimoncelli:
I don't understand why that's happening.  I read the code and though it should skip the provider properly.  :-(

Smart idea to try `--delete-unmanaged-zones`.  That finds a number of unmanaged zones (they were moved to a different provider, but we never cleaned up the old location).  However it crashes on a different location.

```
Checking Zone Removal:
Removing zone stackpromos.com from provider gcloud
Removing zone miniprofiler.com from provider gcloud
Removing zone stackoverflow.co from provider gcloud
Removing zone stackenterprise.co from provider gcloud
Removing zone asksf.com from provider r53
Removing zone stackoverflow.net from provider r53
Removing zone aws.ds.stackexchange.com from provider r53
Removing zone stackoverflow.co from provider r53
Removing zone bosun.org from provider r53
Removing zone selfserve-stackexchange.com from provider r53
Removing zone apptivate.ms from provider r53
Removing zone stackpromos.com from provider r53
Removing zone stackenterprise.co from provider r53
Removing zone miniprofiler.com from provider r53
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x178e1ee]

goroutine 1 [running]:
github.com/StackExchange/dnscontrol/v3/commands.run({{{{0x24626fe, 0xc}, {0x0, 0x0}, 0x0, {{...}, {...}, 0x0}}, {0x0, 0x0}}, ...}, ...)
	/Users/tlimoncelli/git/dnscontrol/commands/previewPush.go:276 +0xa2e
github.com/StackExchange/dnscontrol/v3/commands.Preview(...)
	/Users/tlimoncelli/git/dnscontrol/commands/previewPush.go:119
github.com/StackExchange/dnscontrol/v3/commands.glob..func7.1(0xc00032ab00?)
	/Users/tlimoncelli/git/dnscontrol/commands/previewPush.go:26 +0xa6
github.com/urfave/cli/v2.(*Command).Run(0xc00032ab00, 0xc0000d0480, {0xc00014d6a0, 0x2, 0x2})
	/Users/tlimoncelli/go/pkg/mod/github.com/urfave/cli/v2@v2.24.2/command.go:273 +0xa42
github.com/urfave/cli/v2.(*Command).Run(0xc00019c2c0, 0xc0000d0300, {0xc000110180, 0x3, 0x3})
	/Users/tlimoncelli/go/pkg/mod/github.com/urfave/cli/v2@v2.24.2/command.go:266 +0xc97
github.com/urfave/cli/v2.(*App).RunContext(0xc0000421e0, {0x26e0be0?, 0xc000122008}, {0xc000110180, 0x3, 0x3})
	/Users/tlimoncelli/go/pkg/mod/github.com/urfave/cli/v2@v2.24.2/app.go:332 +0x616
github.com/urfave/cli/v2.(*App).Run(...)
	/Users/tlimoncelli/go/pkg/mod/github.com/urfave/cli/v2@v2.24.2/app.go:309
github.com/StackExchange/dnscontrol/v3/commands.Run({0xc000047780?, 0x2460c80?})
	/Users/tlimoncelli/git/dnscontrol/commands/commands.go:70 +0x325
main.main()
	/Users/tlimoncelli/git/dnscontrol/main.go:34 +0x134
```

Yannik:
@tlimoncelli 

It doesn't crash for me (otherwise I wouldn't have submitted the PR :D).

Can you confirm the line of code it is crashing on? (Line 276 in my branch is `for _, zone := range deployedZones {`). 

Do you have more providers than gcloud and r53 configured?

Yannik:
@tlimoncelli 
> I don't understand why that's happening. I read the code and though it should skip the provider properly. :-(

I have fixed that with my latest commit. The `if !ok` branch was simply missing a `return`/`continue` to stop processing the current provider.

tlimoncelli:
> @tlimoncelli
> 
> > I don't understand why that's happening. I read the code and though it should skip the provider properly. :-(
> 
> I have fixed that with my latest commit. The `if !ok` branch was simply missing a `return`/`continue` to stop processing the current provider.

Looks like that fixed it!  Thanks!

Yannik:
@tlimoncelli did that also fix the second crash you experienced, or is that still present?

commands/previewPush.go
```diff
@@ -216,6 +235,59 @@ DomainLoop:
...
+				out.Warnf("provider type of %s cannot list zones\n", providerName)
+				continue
+			}
+			deployedZones, err := zoneLister.ListZones()
+			if err != nil {
+				return fmt.Errorf("failed ListZones: %w\n", err)
+			}
+			for _, zone := range deployedZones {
+				if !IsZoneManagedByProvider(cfg, zone, providerName) {
+					fmt.Printf("Removing zone %s from provider %s\n", zone, providerName)
```
tlimoncelli, line 280: Sadly CSCGLOBAL (and others?) are showing every domain as unmanaged.

I think the problem is that the code should compare deployedZone to the list of zones that are in dnsconfig.js (`cfg.Domains`).

Suggestion 1:
Turn this into a function. All the code included by `if DeleteUnmanaged || args.DeleteUnmanagedZones {` should be its own function.  Rather than including complex algorithms in the middle of an already long function, separate it out to make it easier to read and test.

Suggestion 2:
My other suggestion is to use "sets" instead of iteration.

Step 1: Extract the data you need into 3 lists:

```
configList := // zone names in dnsconfig.js. The `.Name` of each `cfg.Domains` item.
providerZoneList := // The zones that ListZones() output
providerDomainList := // The zones that ListDomains() output
```

Step 2: Now generate the list of zones that are unmanaged:

```
unmanaged_zones := providerZoneList - configList
unmanaged_domains := providerDomainList - configList
```

(In these formula `-` means "intersection".... Go Syntax doesn't support that but you can use https://pkg.go.dev/github.com/golang-collections/collections/set )


Tom

commands/previewPush.go
```diff
@@ -216,6 +235,59 @@ DomainLoop:
 		totalCorrections += len(corrections)
 		anyErrors = printOrRunCorrections(domain.Name, domain.RegistrarName, corrections, out, push, interactive, notifier) || anyErrors
 	}
+
+	if args.DeleteUnmanaged || args.DeleteUnmanagedDomains {
+		fmt.Printf("Checking Domain Removal:\n")
+		for registrarName, registrar := range createdRegistrars {
+			domainLister, ok := registrar.(providers.DomainLister)
+			if !ok {
+				out.Warnf("provider type of %s cannot list domains\n", registrarName)
```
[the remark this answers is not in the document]
Yannik: Done, however I kept is as `WARNING`

commands/previewPush.go
```diff
@@ -216,6 +235,59 @@ DomainLoop:
...
+				continue
+			}
+			deployedDomains, err := domainLister.ListDomains()
+			if err != nil {
+				return fmt.Errorf("failed ListZones: %w\n", err)
+			}
+			for _, domain := range deployedDomains {
+				if !IsDomainManagedByRegistrar(cfg, domain, registrarName) {
+					fmt.Printf("Removing domain %s from provider %s\n", domain, registrarName)
```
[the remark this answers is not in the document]
Yannik: I have gone with your Idea 1, as it makes parsing the command output with unix tools a bit easier.

commands/previewPush.go
Yannik: Suggestion 1: Implemented, I have extracted the code into `DeleteUnmanagedDomains` and `DeleteUnmanagedZones`.

Regarding your problem with CSCGLOBAL and your suggestion 2:
If we have the domain `test.com` deployed on dns provider 1 and dns provider 2, but only configured for dns provider 1, it would stay deployed on dns provider 2 if we only compare to `cfg.Domains`

Therefore we need to check whether a domain is actually not managed for provider (instead of your suggestion of checking whether it is managed at all).

The code that is being used for this is:
```
func IsZoneManagedByProvider(cfg *models.DNSConfig, zone string, dnsProviderName string) bool {
	domainCfg := GetDomainCfg(cfg, zone)

	if domainCfg == nil {
		return false
	}
	for managedProviderName, _ := range domainCfg.DNSProviderNames {
		if managedProviderName == dnsProviderName {
			return true
		}
	}
	return false
}
```

For me that is working fine. Can you check on your side why this might not be working for CSCGLOBAL?

Yannik: I have just added a test `TestIsDomainOrZoneManaged`. 

To reproduce your CSCGLOBAL issue, you could try breaking it by modifying `test_data/dnsconfig.js`

commands/previewPush.go
Yannik: Done.

commands/previewPush.go
```diff
@@ -230,12 +264,105 @@ DomainLoop:
...
+	err                 error
+}
+
 // InitializeProviders takes (fully processed) configuration and instantiates all providers and returns them.
-func InitializeProviders(cfg *models.DNSConfig, providerConfigs map[string]map[string]string, notifyFlag bool) (notify notifications.Notifier, err error) {
+func InitializeProviders(cfg *models.DNSConfig, providerConfigs map[string]map[string]string, notifyFlag bool) ProviderState {
 	var notificationCfg map[string]string
-	defer func() {
-		notify = notifications.Init(notificationCfg)
-	}()
```
Yannik, line 238: Using this (with `notify` changed to `result.notifier`)  did result in the returned `providerState.notifier` being nil. 
I don't know why. 
To fix this, I have moved the `notifications.Init` out from the defer to before the (successful) return.

Is this okay?

Yannik, line 238: @tlimoncelli 
Using this (with `notify` changed to `result.notifier`)  did result in the returned `providerState.notifier` being nil. 
I wasn't able to deduce why this is the case. 
To fix this, I have moved the `notifications.Init` out from the defer to before the (successful) return.

Is this okay?

Yannik:
rebased on master

Yannik:
@tlimoncelli 

> Sadly CSCGLOBAL is still listing every zone :(

Unfortunately I cannot solve that without more information. I listed some things to look at in [this comment](https://github.com/StackExchange/dnscontrol/pull/2050#discussion_r1096624605).

Additionally, could you create a fresh creds file with cscglobal only, and a dnsconfig.js with only a single (already deployed!) domain configured, to see if the issue can be reproduced in this minimal testcase? Maybe also post these configs here.

tlimoncelli:
Hi @Yannik 

Good news!  The problem with CSCGLOBAL is fixed. It wasn't your code. It was our dnsconfig.js!

*  A macro we use to specify a parked domain was missing the `DnsProvider(DSP_CSCGLOBAL)` statement in the `D()`.  Therefore it wasn't managed!
* A few domains had been added to our account without adding them to dnsconfig.js.  Maybe we need a `--report-unmanaged-zones` flag!

Not so good news:

It took me quite a long time to trace the code and understand what it was trying to do. The layers of nested loops is elegant but difficult to understand unless you really know the structs well.  In fact, I wasn't sure what CSCGOBAL's problem was until I started reimplementing the algorithm.  Now the logic is more visible to the reader.

The results of my rewrite are here: https://github.com/StackExchange/dnscontrol/pull/2061 

I'd like you to consider rewriting the domains code the same way.

Thanks!

Yannik:
Hi @tlimoncelli 

I'm glad you found the issue!

I have also reviewed #2061:
Personally, I feel like the `setFromStrings` is too much boilerplate (that needs brainpower to be understood as well) just to be able to use the `Difference` function.
What about a compromise: Extract getting the managed/deployed zones from the loop, but still loop over the deployedZones and use the `slices.Contains(managedZones, zone)` function to check if it is managed or not?
I think this would be a good solution.

tlimoncelli:
Using slices.Contains instead of sets seems reasonable!  Show me what that would look like.

Yannik:
@tlimoncelli I have changed some things to make the code easier to understand (i.e. replaced the loop over `DnsProviderNames` with a simple check if the relevant key exists, split up if statements, ordered returns more logically). I've also added comments to explain what each function does. 

In addition, I've improved the output format a bit to match the other output of dnscontrol and make it more readable.

IMO, the code is in pretty good shape to be merged now. :)

tlimoncelli:
Thank you for showing what the compromise would look like.  I was willing to see what you were proposing but sadly the new revision has the same issues as before. Specifically:

1. It is impossible to understand the intended algorithm by looking at the code.
2. It doesn't work for CSCGLOBAL.

These two issues are related: I can't debug the CSCGLOBAL problem because I can't create a mental model of what the intended algorithm is and compare that to the code that I see on my screen.

Another issue is that this code isn't testable. My proposed alternative is broken down into smaller functions, each could have a unit test.  A unit test for DeleteUnmanagedZones/DeleteUnmanagedDomains in https://github.com/StackExchange/dnscontrol/pull/2050 would be "all or nothing" which would be brittle.

I've offered you a PR showing how it should be implemented.  You're going to have to adopt that PR's method (i.e. adapt DeleteUnmanagedDomains and DeleteUnmanagedZones to by like DeleteUnmanagedDomains in https://github.com/StackExchange/dnscontrol/pull/2061) or I will have to reject this PR because it fails to work on CSCGLOBAL.

I have spent too much time trying to debug the CSCGLOBAL issue and can not dedicate more time to it.

My recommended next steps:

* Adopt the design of DeleteUnmanagedDomains in https://github.com/StackExchange/dnscontrol/pull/2061) for DeleteUnmanagedDomains in https://github.com/StackExchange/dnscontrol/pull/2061
* Replace the use of mapset and use slices.Contains() as you previously suggested.

Tom

commands/previewPush.go
```diff
@@ -230,12 +264,119 @@ DomainLoop:
...
+		return false
+	}
+	return true
+}
+
+type ProviderState struct {
+	createdRegistrars   map[string]providers.Registrar
+	createdDnsProviders map[string]providers.DNSServiceProvider
+	notifier            notifications.Notifier
+	err                 error
```
tlimoncelli, line 373: Remove `err` and return it seperately.

commands/previewPush.go
```diff
@@ -230,12 +264,119 @@ DomainLoop:
...
+	_, ok := domainCfg.DNSProviderNames[dnsProviderName]
+	// domain is configured within dnscontrol, but not for this dns provider
+	if !ok {
+		return false
+	}
+	return true
+}
+
+type ProviderState struct {
+	createdRegistrars   map[string]providers.Registrar
```
tlimoncelli, line 370: Use `instantiated` instead of `created`.

commands/previewPush.go
```diff
@@ -230,12 +264,119 @@ DomainLoop:
...
+// are not configured within dnscontrol
+func DeleteUnmanagedZones(cfg *models.DNSConfig, createdProviders map[string]providers.DNSServiceProvider, push bool, out printer.CLI, totalCorrections *int) error {
+	for providerName, provider := range createdProviders {
+		out.StartUnmanagedZonesCheck(providerName)
+		zoneLister, ok := provider.(providers.ZoneLister)
+		if !ok {
+			out.Warnf("--purge-unmanaged-zones not implemented: provider %s\n", providerName)
+			continue
+		}
+		deployedZones, err := zoneLister.ListZones()
```
tlimoncelli, line 308: Change "deployed" to "existing".  The convention in the rest of the code is to refer to things as "existing" and "desired", meaning "what the provider state is currently" and "what the configuration requests" respectively.

"existing" and "desired"

commands/previewPush.go
```diff
@@ -230,12 +264,119 @@ DomainLoop:
...
+		if err != nil {
+			return fmt.Errorf("failed ListZones for provider %s: %w\n", providerName, err)
+		}
+		for _, zone := range deployedZones {
+			if !IsZoneManagedByProvider(cfg, zone, providerName) {
+				fmt.Printf("Removing zone from provider %s: %s\n", providerName, zone)
+				*totalCorrections += 1
```
tlimoncelli, line 315: Don't pass by reference `*totalCorrections`.  Return a count of changes. The caller can add count to totalCorrections.

tlimoncelli:
> @tlimoncelli
> 
> > It doesn't work for CSCGLOBAL.
> 
> I'm a bit confused. I thought the CSCGLOBAL issue was caused by not wrapping the provider in `DnsProvider(..)` in your `dnsconfig.js`?

That was required to get https://github.com/StackExchange/dnscontrol/pull/2061 working.  It still has the same bug in https://github.com/StackExchange/dnscontrol/pull/2050

commands/previewPush.go
```diff
@@ -214,12 +233,27 @@ DomainLoop:
 			continue
 		}
 		totalCorrections += len(corrections)
-		anyErrors = printOrRunCorrections(domain.Name, domain.RegistrarName, corrections, out, push, interactive, notifier) || anyErrors
+		anyErrors = printOrRunCorrections(domain.Name, domain.RegistrarName, corrections, out, push, interactive, providerState.notifier) || anyErrors
+	}
+
+	if args.DeleteUnmanaged || args.DeleteUnmanagedDomains {
```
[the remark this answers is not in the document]
Yannik: Done

commands/previewPush.go
```diff
@@ -230,12 +264,119 @@ DomainLoop:
 	return nil
 }
 
+// DeleteUnmanagedDomains iterates over all registrars and their respective deployed domains, and removes those, which
+// are not configured within dnscontrol
+func DeleteUnmanagedDomains(cfg *models.DNSConfig, createdRegistrars map[string]providers.Registrar, push bool, out printer.CLI, totalCorrections *int) error {
```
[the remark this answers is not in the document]
Yannik: Done

commands/previewPush.go
Yannik: That makes sense! Done.

commands/previewPush.go
Yannik: I also thought about this one before... 
Naming this var `instantiated` would make me think the variable is one of  `RegistrarInstance` or `DNSProviderInstances`. That seemed confusing.

commands/previewPush.go
Yannik: Done

commands/previewPush.go
Yannik: Done

Yannik:
> > @tlimoncelli
> > > It doesn't work for CSCGLOBAL.
> > 
> > 
> > I'm a bit confused. I thought the CSCGLOBAL issue was caused by not wrapping the provider in `DnsProvider(..)` in your `dnsconfig.js`?
> 
> That was required to get #2061 working. It still has the same bug in #2050

Could you kindly check that you really tested this with a corrected `dnsconfig.js`?
Would it be possible for you to post the output of `dnscontrol print-ir` for one of the affected domains?

tlimoncelli:
> Could you kindly check that you really tested this with a corrected `dnsconfig.js`? Would it be possible for you to post a snippet of the `dnscontrol print-ir` output for one of the affected domains?

Sure!  Could you send me your email address and I'll send it.  You can send to me privately via this link: https://transfer.secretoverflow.com/u/tlimoncelli

TomOnTime:
I was thinking about this command and I realized it might be better as a subcommand instead of part of preview/push.  When I wanted to try out this command, a co-worker was making changes to the domain and I wished this wasn't part of "preview" or that there was a flag that would only do domain/zone purges, no DNS record updates.  That's when I realized this might be better as a subcommand:

Maybe something like this?

```
dnscontrol purge [--zones|--domains] [--dry-run]
```

(The default would be to do both zones and domains)

Yannik:
@tlimoncelli I will try to get back onto this in april, currently I am very short on time for working on this.

tlimoncelli:
The bad news is: The new ppreview code obsoletes all of the code in this PR. So, implementing this will have to be done differently. 

The good news is: The new ppreview code makes implementing this much easier.

PR #2055, opened 2023-02-06
HOSTINGDE: Remove dnssec key from domain upon autodnssec disable

Yannik:
I noticed an issue where if you enable autodnssec, and then disable it, the DS key is left at the domain, resulting in the zone not being able to be used anymore. (Because it's not signed anymore)

This PR fixes that.


tlimoncelli:
Looks good!  Please rebase and I'll merge.

Yannik:
@tlimoncelli done

PR #2072, opened 2023-02-13
Fix parameter type for DEFAULTS

### LINKED ISSUE #2070: DEFAULTS should accept DomainModifier in TypeScript typings
I think `DEFAULTS` expects `modifiers` of type `DomainModifier[]`. The TypeScript typings declare `modifiers` to be  `RecordModifier[]` though.

j-f1:
Fixes #2070.


tlimoncelli:
Thanks!

PR #2076, opened 2023-02-14
NEW FEATURE: Colorize preview/push output

fritterhoff:
When using dnscontrol in CI/CD pipelines or changing several records/zones at once the output can be quite confusing. In order to make the output clearer, we would like to highlight the different operations in color.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,6 +59,7 @@ require (
 
 require (
 	github.com/G-Core/gcore-dns-sdk-go v0.2.3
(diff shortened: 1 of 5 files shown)


tlimoncelli:
I love this idea!

Sadly I'd prefer it if the CHANGE items were still on a single line... at least for now. I have some plans for changing that format in the future.

tlimoncelli (approved):

PR #2080, opened 2023-02-18
NAMECHEAP: New maintainer: willpower232

willpower232:
As per emails with @tlimoncelli, I should like to volunteer to maintain the namecheap provider.

<details>
<summary>
I have been able to run the tests against the namecheap sandbox and they almost all pass.
</summary>

```
--- FAIL: TestDNSProviders (197.07s)
    --- FAIL: TestDNSProviders/willpower232testsdnscontrol.com (197.07s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/Clean_Slate:Empty (1.36s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/00:A:Create_A (1.17s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/00:A:Change_A_target (1.02s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/Post_cleanup:Empty (0.75s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/01:Attl:Create_Arc (1.04s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/01:Attl:Change_TTL (1.15s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/Post_cleanup:Empty#01 (0.75s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/02:MX:Create_MX (1.05s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/02:MX:Change_MX_target (1.04s)
        --- FAIL: TestDNSProviders/willpower232testsdnscontrol.com/02:MX:Change_MX_p (1.11s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/Post_cleanup:Empty#02 (0.73s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/03:CNAME:Create_a_CNAME (1.05s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/Post_cleanup:Empty#03 (0.61s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/04:ManyAtOne:CreateManyAtLabel (0.93s)
        --- PASS: TestDNSProviders/willpower232testsdnscontrol.com/Post_cleanup:Empty#04 (0.69s)
[...]

PR #2081, opened 2023-02-18
namecheap fix rate limit detection

willpower232:
As part of running the tests for #2080, I discovered that the rate limit detection is currently broken for the namecheap provider.

Unfortunately after it was implemented in #244 the library the namecheap provider depends on changed how it presented this situation https://github.com/billputer/go-namecheap/pull/18

I have verified this works against sandbox and live however I am also curious that namecheap selected a different HTTP status code when rate limits are triggered.


tlimoncelli:
Nice detective work!

(yes, APIs change without much notice. That's one of the reason the integration tests exist: not just to find bugs in the code, but to detect APi changes!)

PR #2082, opened 2023-02-18
test namecheap provider against their sandbox

willpower232:
Namecheap have a nice sandbox environment for testing things, unfortunately their API relies on an IP address allow list so CI is tricky but switching local tests to the sandbox needs this change.


integrationTest/providers.json
```diff
@@ -142,7 +142,8 @@
   "NAMECHEAP": {
     "apikey": "$NAMECHEAP_KEY",
     "apiuser": "$NAMECHEAP_USER",
-    "domain": "$NAMECHEAP_DOMAIN"
+    "domain": "$NAMECHEAP_DOMAIN",
+    "BaseURL": "https://api.sandbox.namecheap.com/xml.response"
```
tlimoncelli, line 146: Hmmm.... this is a good idea but what if we wanted to run tests outside the sandbox?

My suggestion is to set this using an env and document that (if it isn't already) in `documentation/providers/namecheap.md`

As an env variable it would look like this:

```
    "BaseURL": "NAMECHEAP_BASEURL"
```

PR #2101, opened 2023-02-28
DOC: Fixed a bit of havoc which appears in go docs

systemcrash:
at https://pkg.go.dev/github.com/StackExchange/dnscontrol/


tlimoncelli:
Thanks for noticing this and providing a PR!

PR #2105, opened 2023-02-28
Doc: Fixed a bit of havoc which appears in go docs

systemcrash:

### DIFF
--- a/models/t_parse.go
+++ b/models/t_parse.go
@@ -17,7 +17,7 @@ import (
 //
 //		  var err error
 //		  switch rType {
-//	   case "MX":
+//		  case "MX":
 //	       // MX priority in a separate field.
 //	       if err := rc.SetTargetMX(cr.Priority, target); err != nil {
 //	         return nil, fmt.Errorf("unparsable MX record received from cloudflare: %w", err)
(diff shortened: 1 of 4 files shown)


models/t_parse.go
```diff
@@ -14,24 +14,6 @@ import (
...
-//
-//		  var err error
-//		  switch rType {
-//	   case "MX":
-//	       // MX priority in a separate field.
-//	       if err := rc.SetTargetMX(cr.Priority, target); err != nil {
-//	         return nil, fmt.Errorf("unparsable MX record received from cloudflare: %w", err)
-//	       }
-//		  case "TXT":
-//	       // TXT records are stored verbatim; no quoting/escaping to parse.
```
tlimoncelli, line 26: This example disappeared in the new revision.  Please bring it back.

tlimoncelli (commented):
Thanks for cleaning this up!  I really appreciate it!

models/t_parse.go
systemcrash: If there should be any text which explains its presence, that would be great. Like "For example", or "note to self". But I read the code segment and I am none the wiser.

PR #2108, opened 2023-03-01
Update getting-started.md

systemcrash:
tweaks (via github edit)


tlimoncelli:
thank you for taking the time to fix these!

PR #2139, opened 2023-03-11
Update writing-providers.md

systemcrash:
golint is frozen and deprecated. remove.


tlimoncelli:
This is an important point. However I'm going to fix it this way instead: https://github.com/StackExchange/dnscontrol/pull/2142

PR #2133, opened 2023-03-09
Update DefaultTTL.md

systemcrash:
Describe default TTL for NS cases.

PR #2119, opened 2023-03-02
DOCS: Added Gitbook filenames to JSON code snippets.

cafferata:
Added Gitbook filenames to the JSON code snippets.

PR #2140, opened 2023-03-11
Loopia DNS service provider

systemcrash:
Comments welcome! 😄


documentation/providers/loopia.md
```diff
@@ -0,0 +1,155 @@
...
+username and password from your `creds.json` file. If you want to share these, 
+like for a github issue, be sure to redact those from the XML.
+
+
+### Rate Limiting
+
+Loopia rate limits requests to 60 per minute.
+
+From their [web-site](https://www.loopia.com/api/rate_limiting/):
+```
```
cafferata, line 69:
```suggestion
```text
```

documentation/providers/loopia.md
```diff
@@ -0,0 +1,155 @@
...
+This effectively means that this plugin does not access registrar functions.
+
+## Configuration
+
+To use this provider, add an entry to `creds.json` with `TYPE` set to `LOOPIA`
+along with your Loopia API login credentials.
+
+Example:
+
+```json
```
cafferata, line 46: Can you also apply the GitBook code syntax? See https://github.com/StackExchange/dnscontrol/pull/2119 and https://github.com/StackExchange/dnscontrol/pull/2137

documentation/providers/loopia.md
```diff
@@ -0,0 +1,155 @@
...
+// Providers:
+var REG_LOOPIA = NewRegistrar("loopia");
+var DSP_LOOPIA = NewDnsProvider("loopia");
+
+// Set Default TTL for all RR to reflect our Backend API Default
+// If you use additional DNS Providers, configure a default TTL
+// per domain using the domain modifier DefaultTTL instead.
+// also check this issue for [NAMESERVER TTL](https://github.com/StackExchange/dnscontrol/issues/176).
+DEFAULTS(
+    {"ns_ttl":"3600"},
```
cafferata, line 126: See @tlimoncelli his comments https://github.com/StackExchange/dnscontrol/pull/2133#discussion_r1133086503

cafferata (changes requested):

documentation/providers/loopia.md
systemcrash: Think I understood - tags like these? `{% code title="dnsconfig.js" %}`?

documentation/providers/loopia.md
```diff
@@ -0,0 +1,155 @@
+Loopia is an absolutely 💩 provider of DNS API, if you want to use a decent API.
```
tlimoncelli, line 1:
```
Loopia is an absolutely 💩 provider of DNS. Using DNSControl hides some of the 💩.
If you are stuck with Loopia, hopefully this will reduce the pain.

They provide DNS services, both as a registrar, and a provider. 
They provide support in English and other regional variants (Norwegian, Serbian, Swedish).

This plugin is based on API documents found at 
[https://www.loopia.com/api/](https://www.loopia.com/api/)
and by observing API responses. Hat tip to github @hazzeh whose code for the
LEGO Loopia implementation was helpful.

Sadly the Loopia API has some problems:
* API calls are limited to 60 calls per minute.  If you go above this,
  you will have to wait before you can make changes.
* When rate-limited, you will not receive a single
  error: The errors propagate from the back-end, with no headers, or
  Retry-After or anything useful.
* Some testing has revealed weird behavior with NS records
* There are no guarantees of idempotency from their API.
```

providers/loopia/client.go
```diff
@@ -0,0 +1,523 @@
...
+		BaseURL:     DefaultBaseURL,
+		HTTPClient:  &http.Client{Timeout: 10 * time.Second},
+		Debug:       debug,
+	}
+}
+
+//CRUD: Create, Read, Update, Delete
+//Create
+
+// CreateDomainRecord adds a record.
```
tlimoncelli, line 109: Update comment

providers/loopia/convert.go
```diff
@@ -0,0 +1,70 @@
+package loopia
+
+// Convert the provider's native record description to models.RecordConfig.
+
+import (
+	"fmt"
+	// "net"
```
tlimoncelli, line 7: remove `// "net"` and other commented out imports

providers/loopia/loopiaProvider.go
```diff
@@ -0,0 +1,426 @@
...
+// Section 1: Register this provider in the system.
+
+// init registers the provider to dnscontrol.
+func init() {
+	fns := providers.DspFuncs{
+		Initializer:   newDsp,
+		RecordAuditor: AuditRecords,
+	}
+	providers.RegisterDomainServiceProviderType("LOOPIA_XMLRPCv1", fns, features)
+	providers.RegisterRegistrarType("LOOPIA_XMLRPCv1", newReg)
```
tlimoncelli, line 42: Can we call this something more simple than "LOOPIA_XMLRPCv1"?  How about just "LOOPIA"?  That way it matches the directory name and the fooProvider.go name.

providers/loopia/loopiaProvider.go
```diff
@@ -0,0 +1,426 @@
...
+// post-process them, and generate corrections.
+func (client *LoopiaClient) GetDomainCorrections(dc *models.DomainConfig) ([]*models.Correction, error) {
+	existing, err := client.GetZoneRecords(dc.Name)
+	if err != nil {
+		return nil, err
+	}
+	models.PostProcessRecords(existing)
+	clean := PrepFoundRecords(existing)
+	PrepDesiredRecords(dc)
+	return client.GenerateDomainCorrections(dc, clean)
```
tlimoncelli, line 157: I haven't documented this but... please rename "GenerateDomainCorrections" to "GenerateZoneRecordsCorrections".  Thanks!

providers/loopia/loopiaProvider.go
systemcrash: Yeah, I had a similar thought, but I had the hope that one day they might release a better API. I suppose at that point it's possible rename the existing stuff.

providers/loopia/loopiaProvider.go
systemcrash: OK.

cafferata:
Would you like to add this new DNS provider `Loopia` to `SUMMARY.md`?

```diff
    * [Internet.bs](providers/internetbs.md)
    * [INWX](providers/inwx.md)
    * [Linode](providers/linode.md)
+   * [Loopia](providers/loopia.md)
    * [LuaDNS](providers/luadns.md)
    * [Microsoft DNS Server on Microsoft Windows Server](providers/msdns.md)
```

documentation/providers/loopia.md
```diff
@@ -0,0 +1,158 @@
+Loopia is a 💩 provider of DNS. Using DNSControl hides some of the 💩.
+If you are stuck with Loopia, hopefully this will reduce the pain.
+
+They provide DNS services, both as a registrar, and a provider. 
+They provide support in English and other regional variants (Norwegian, Serbian, Swedish).
+
+This plugin is based on API documents found at 
+[https://www.loopia.com/api/](https://www.loopia.com/api/)
+and by observing API responses. Hat tip to github @hazzeh whose code for the
```
cafferata, line 9:
```suggestion
and by observing API responses. Hat tip to GitHub @hazzeh whose code for the
```

documentation/providers/loopia.md
```diff
@@ -0,0 +1,158 @@
...
+* `username` - string - your @loopiaapi created username
+* `password` - string - your loopia API password
+* `debug` - string - Set to true for extra debug output. Remove or set to false to prevent extra debug output. 
+* `rate_limit_per` - string - See Rate Limiting below.
+
+There is no test endpoint. Fly free, grasshopper.
+
+Turning on debug will show the XML requests and responses, and include the
+username and password from your `creds.json` file. If you want to share these, 
+like for a github issue, be sure to redact those from the XML.
```
cafferata, line 60:
```suggestion
like for a GitHub issue, be sure to redact those from the XML.
```

documentation/providers/loopia.md
systemcrash: Good catch. Fixed.

documentation/providers/loopia.md
systemcrash: Fixed. Thanks.

documentation/providers/loopia.md
```diff
@@ -0,0 +1,158 @@
...
+{% code title="dnsconfig.js" %}
+```javascript
+// Providers:
+var REG_LOOPIA = NewRegistrar("loopia");
+var DSP_LOOPIA = NewDnsProvider("loopia");
+
+// Set Default TTL for all RR to reflect our Backend API Default
+// If you use additional DNS Providers, configure a default TTL
+// per domain using the domain modifier DefaultTTL instead.
+// also check this issue for [NAMESERVER TTL](https://github.com/StackExchange/dnscontrol/issues/176).
```
cafferata: There's no Markdown possible in a code block.

```suggestion
// also check this issue for NAMESERVER TTL: https://github.com/StackExchange/dnscontrol/issues/176.
```

systemcrash: Removed this.

systemcrash:
Question for @tlimoncelli @cafferata 

How should one handle NS entries?

Looking at other providers, some have e.g.

```go
// GetNameservers returns a list of nameservers for domain.
func (c *XClient) GetNameservers(domain string) ([]*models.Nameserver, error) {

	nameservers, err := c.GetDomainNS(domain)
	if err != nil {
		return nil, err
	}

	return models.ToNameserversStripTD(nameservers)
}
```

And this either returns the provider default NS entries, or the NS entries from the zone file (which should be the provider default NS entries...?)

Then the dnsconfig.js magic goes like:

```js
...
	// NAMESERVER('ns1.corp.com.'),
	// NAMESERVER('ns2.corp.com.'),
...
```
And dnscontrol only tries to add extra NS if we uncomment these.

Problem 1: if the zone has (by some mistake) multiple copies of (the same) NS entries, dnscontrol does nothing. 
Problem 2: if the zone has foreign NS entries (as created by the test `TestDualProviders`), and e.g. Loopia declares 

```go
...
	providers.DocDualHost:            providers.Can(),
````

One cannot delete NS entries.

If we now do e.g.

```go
// GetNameservers returns a list of nameservers for domain.
func (c *XClient) GetNameservers(domain string) ([]*models.Nameserver, error) {

	return nil, nil
}
```

One can manage NS entries like any other in `dnsconfig.js`:

```js
...
	NAMESERVER('ns1.corp.com.'),
	NAMESERVER('ns2.corp.com.'),
...
```

Which approach "should" one take? I think manually managing NS entries where possible is flexible. But there is probably something important that I have not realized yet.

tlimoncelli:
That's a lot to unpack. Let me see if I can answer the important questions:

The reason get-zones outputs the NAMESERVER() lines commented out because in most cases NAMESERVER() isn't needed.  DNSControl looks up the parent delegation via the API.  The NAMESERVER() commands are only need when that fails.  By outputting a suggestion (commented out), it permits a human to make the decision whether they are needed or not.

> Problem 1: if the zone has (by some mistake) multiple copies of (the same) NS entries, dnscontrol does nothing.

I think that's a problem for any provider.  Isn't it?  (or maybe I'm misunderstanding the question) 

> Problem 2: if the zone has foreign NS entries (as created by the test TestDualProviders), and e.g. Loopia declares

There are many providers that don't allow you to delete certain NS records.  Those providers output a warning instead.

I think transip/transipProvider.go is one example.  removeOtherNS() ermoves the NS records that can't be deleted so that DNSControl doesn't try to delete them.

Tom

systemcrash:
> Tom

I think what I can do is update the PR so that we don't poll the API specially for NS entries, but instead receive them with all other entries. This way NS and NAMESERVER entries in the `.js` file behave like other entries. Also helpful if you want to `dual-host`. This was basically my question - whether there is something bad in doing this, that I may have missed.

systemcrash:
Or, how about having both options, but hide them behind a flag specific to Loopia?

tlimoncelli:
I'm not sure I understand.  I hope this answers the question:

The `GetNameservers()` method should not return the list of NS records in the domain. It should list the parent domain's delegations for this domain.  Usually these are the same, but not necessarily (the NS records may be missing/wrong and everything still works).

```
$ whois plts.org | grep "^Name Server:"
Name Server: ns-104-b.gandi.net
Name Server: ns-207-c.gandi.net
Name Server: ns-32-a.gandi.net
$ dig +short plts.org ns
ns-32-a.gandi.net.
ns-104-b.gandi.net.
ns-207-c.gandi.net.
```

Both have the same information but they come from different sources.

tlimoncelli:
Yes.  GetNameservers() returns the list of glue records.

systemcrash:
OK  - I've added flags to handle how Loopia behaves with NS records, to reconcile dnscontrol's behaviour expectations. 

After plenty of testing, everything is working nicely. All tested. All linted. 

I'm happy with where things are at this stage, and I'm going to let it sit for a diem and see if anything extra comes up.

documentation/providers/loopia.md
```diff
@@ -0,0 +1,214 @@
...
+
+API calls to loopia can be expensive time-wise. Set this to "false" (off) to
+skip the API call to fetch the apex (`@`) entries, and use Loopia's default NS
+servers.
+
+This setting defaults to "true" (on).
+
+### Modify Name Servers
+
+Enabling this setting to "true" (on) allows you to modify NS entries.
```
tlimoncelli, line 92: It isn't clear that "this setting" refers to `modify_name_servers`.  Change to:

Setting `modify_name_servers` in `creds.json' to `true`  (on) allows...

documentation/providers/loopia.md
```diff
@@ -0,0 +1,214 @@
...
+to modification. Beware the consequences of changing from default NS entries. Likely
+nothing will happen since the glue records provided won't match those in the domain,
+and you will need to manually inform Loopia of this so they can update the glue records.
+
+In short: enable this setting to be able to delete NS entries. No `NS()` in your
+`dnsconfig.js`? Existing ones will be deleted. Have some `NS()` or `NAMESERVER()`
+entries? They'll be added.
+
+This setting defaults to "false" (off).
+
```
tlimoncelli, line 106: NOTE: This setting is global.

(do you want to implement this on a per-domain basis?  I can show you how to move that setting from creds.json to be a meta-parameter for the domain)

systemcrash: AI think that would be nice. It shouldn't take me too long to implement. Maybe I could add it later. Are there any examples already in code or do you have boiler-plate?

documentation/providers/loopia.md
systemcrash: Fixed. Thanks!

systemcrash:
I was taking a wallk and thought to ask: what's the recommended way to log? I see various ways through the plugins. Perhaps we can document it?

`log.Print` and debug variants and so?

tlimoncelli:
Yes, things are pretty inconsistent.  Documentation would help!

The person working on the new logging system left Stack before it was documented and finished.  See There is pkg/printer.

There is also https://github.com/StackExchange/dnscontrol/pull/1574 which, if implemented, would let us run DSPs in parallel threads and capture the output (rather than outputting it immediately).

TBH I'm not sure what the state of things is for either of these.  I'm probably making things worse by using fmt.Printf() since I don't know the new stuff.

Would you be interested in looking into the code and making a recommendation?

Tom

PR #2148, opened 2023-03-12
DOCS: Removed the prompt character

cafferata:
Removed the prompt character. The origin comes from pull request [StackExchange/dnscontrol/pull/2142](https://github.com/StackExchange/dnscontrol/pull/2142/files#diff-9844a181252db64e648f9c9e864423bc8b27727870fb9b4b4dc8dd52181ab00eR273).


systemcrash:
Helping users to get going is the priority here. Although the prevailing wisdom from sysadmins is that the `#` or the `$` is there to prevent pasting dangerous commands.

tlimoncelli:
I usually only include the `$` or `#` if I'm showing output.  The `$`/`#` differentiates between commands and output.

In this case, I was showing some output but I removed it.  Removing the prompt character makes sense.

tlimoncelli:
(continued)

The documentation/ci-cd-gitlab.md example has output but we do some formatting to separate it out.  I'm not sure which way is better, but we should be consistent.

PR #2168, opened 2023-03-15
Add tlsa to hetzner provider

Narquadah:
Hetzner now Supports TLSA via their API.


tlimoncelli:
Thanks for the update!

PR #2175, opened 2023-03-15
LOOPIA: Linting corrections

TomOnTime:
CC @riku22 for visibility and approval.

Hi @riku22!

I don't know how I missed these before. go-staticcheck found them all.

```
providers/loopia/client.go:82:2: exported const DefaultBaseNOURL should have comment (or a comment on this block) or be unexported
providers/loopia/client.go:92:6: type name will be used as loopia.LoopiaClient by other packages, and that stutters; consider calling this Client
providers/loopia/client.go:165:38: exported method GetDomains returns unexported type []loopia.domainObject, which can be annoying to use
providers/loopia/client.go:182:75: exported method GetDomainRecords returns unexported type []loopia.zoneRecord, which can be annoying to use
providers/loopia/client.go:223:10: if block ends with a return statement, so drop this else and outdent its block
providers/loopia/loopiaProvider.go:415:9: if block ends with a return statement, so drop this else and outdent its block
providers/loopia/types.go:119:6: exported type Property should have comment or be unexported
providers/loopia/types.go:124:6: exported type Value should have comment or be unexported
providers/loopia/types.go:131:1: exported method Property.Name should have comment or be unexported
providers/loopia/types.go:133:1: exported method Property.Int should have comment or be unexported
providers/loopia/types.go:134:1: exported method Property.Bool should have comment or be unexported
```


riku22:
Hi, @TomOnTime.

I don't know as I am not a maintainer of a loopia provider.
Maybe it is @systemcrash?

TomOnTime:
oops!  sorry, [riku22](https://github.com/riku22)!  I copied the wrong name from OWNERS.

CC @systemcrash

systemcrash:
Yeah sure - smash those in there.

PR #2179, opened 2023-03-15
HETZNER: Don't use string.Title()

TomOnTime:
Hi @das7pad!

`providers/hetzner/api.go` uses `strings.Title()` which is deprecated.  Here's my suggested fix.

Let me know if you agree.

```
providers/hetzner/api.go:36:63: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (SA1019)
```


das7pad (approved):
LGTM, thanks!

PR #2194, opened 2023-03-17
PowerDNS: Implement diff2 differ and some cleanup

jpbede:
I've implemented the diff2 differ and done some cleanup. Integration test with and without the `-diff2` flag are passing.

### DIFF
--- a/providers/powerdns/diff.go
+++ b/providers/powerdns/diff.go
@@ -74,9 +74,6 @@ func (dsp *powerdnsProvider) getDiff2DomainCorrections(dc *models.DomainConfig, 
 		return nil, err
 	}
 
-	// reorder so that DELETE actions come first
-	changes = reorderChanges(changes)
-
 	var corrections []*models.Correction
 
 	for _, change := range changes {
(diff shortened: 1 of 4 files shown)


providers/powerdns/diff.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	return corrections, nil
+}
+
+// reorderChanges returns changes reordered so that DELETE changes run first
+func reorderChanges(changes diff2.ChangeList) diff2.ChangeList {
+	var main, pre diff2.ChangeList
+	for _, change := range changes {
+		if change.Type == diff2.DELETE {
```
tlimoncelli, line 124: Note that .DELETE Is only used when all the records at the Label:Type are being deleted.  If, for example, a label had 3 A's and now there are only 2, the type will be .CHANGE even though (in human terms) an A record is being deleted.

jpbede:
Indeed, the reordering is no longer necessary. The reason was that the CNAMEs had to be deleted first.

tlimoncelli:
> Indeed, the reordering is no longer necessary. The reason was that the CNAMEs had to be deleted first.

Cool.  (Now that I've said that, I hope the code proves me right!   I should add some specific integration tests.)

Tom

tlimoncelli:
I'm merging this but if it turns out moving the .DELETEs to the front was required, I'll buy you a beer :)

P.S. Will you be at SRECon?

jpbede:
> P.S. Will you be at SRECon?

No, unfortunately not.

PR #2195, opened 2023-03-17
Don't add extra dot after dnsimple SRV record contents

### LINKED ISSUE #2157: Unstable SRV records on Dnsimple
I have this (exported from Gandi):
```
    SRV("_imaps._tcp", 0, 1, 993, "mail.messagingengine.com."),
```

Every time I run dnscontrol it reports this modification:
```
#1: ± MODIFY SRV _imaps._tcp.mydomain: (0 1 993 mail.messagingengine.com.. ttl=10800) -> (0 1 993 mail.messagingengine.com. ttl=10800)
```
(removing duplicate final `.`)

On dnsimple's web ui it shows
```
SRV | _imaps._tcp.zarbosoft.com | 1 993 mail.messagingengine.com.
-- | -- | --
```
This is the only record listed there with a final `.` (i.e. there's no duplicate ., and the modifications are doing nothing).

andrewbaxter:
Fix #2157 

The dnsimple API returns
```
Content:"1 993 mail.messagingengine.com."
```
when fetching the records.

The code here I think was to make up for the lack of a `.` for the final field.  This would work if the contents were `1 993 mail.messagingengine.com. ` which would become `1 993 mail.messagingengine.com. .` but since there's no final space it becomes `1 993 mail.messagingengine.com..` (unlike `Split`, `Fields` ignores trailing delimiters).

I think this was not necessary anyways, since `SetTargetSRVPriorityString` does the same thing, but skips this extra work (and encoding error risk) of re-encode-decode by passing `"."` as a separate parameter to the next function.

This solved my reported issue, but I'm not sure if it's possible to get other SRV record formats from the API and so I haven't tested anything else.

### DIFF
--- a/providers/dnsimple/dnsimpleProvider.go
+++ b/providers/dnsimple/dnsimpleProvider.go
@@ -119,10 +119,6 @@ func (c *dnsimpleProvider) GetZoneRecords(domain string) (models.Records, error)
 		case "MX":
 			err = rec.SetTargetMX(uint16(r.Priority), r.Content)
 		case "SRV":
-			parts := strings.Fields(r.Content)
-			if len(parts) == 3 {
-				r.Content += "."
-			}
 			err = rec.SetTargetSRVPriorityString(uint16(r.Priority), r.Content)
 		case "TXT":
 			err = rec.SetTargetTXT(r.Content)


tlimoncelli:
CC @onlyhavecans  for approval

onlyhavecans:
Sorry for missing the review on this and ticket. I was out last week. Thank you so much for the fix

andrewbaxter:
np, thanks for the provider in the first place, it makes things very easy!  And @tlimoncelli thanks for the review :bow:

PR #2201, opened 2023-03-18
MAINT: Use WaitGroup instead of labeled loop

jpbede:
In preparation for https://github.com/StackExchange/dnscontrol/pull/1574 and later, to make DNSControl concurrent, I'm splitting some of the work from https://github.com/StackExchange/dnscontrol/pull/1574 into several smaller PRs. Smaller PRs are always better to keep track of and easier to review :)

This PR changes the labeled loop to a loop using a `WaitGroup`. The `WaitGroup` is maybe a bit overkill in this step, but it will be there when we need it :)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,8 @@ jobs:
 
     - name: Build binaries
       run: go run build/build.go
+      env:
+        CGO_ENABLED: 0
 
     - name: Get release from tag
       run: echo ::set-output name=RELEASE_VERSION::$(echo ${GITHUB_REF:11})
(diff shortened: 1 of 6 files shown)


tlimoncelli (commented):
Ok, I see what you are doing.  Turning the inside of a forloop into an anonymous function so that later it can be a goroutine.  Totally makes sense.

However the anonymous function is very long, and thus the call ( `}(domain)`) at the end is easy to miss.

Could the anonymous function be shorter?  (i.e. move some or all the code into named function(s)?)

Tom

tlimoncelli:
P.S. Rather than moving code around, more comments (written for an audience that isn't familiar with WaitGroups or anonymous functions) might be sufficient.

jpbede:
> P.S. Rather than moving code around, more comments (written for an audience that isn't familiar with WaitGroups or anonymous functions) might be sufficient.

Added some comments instead of moving code around. However, we should still consider refactoring this code for maintainability and readability, as you already said, the function is quite large.

tlimoncelli (approved):

PR #2215, opened 2023-03-22
LOOPIA: Ban Null MX

systemcrash:
`Record pointing to @` test passes fine, however.


integrationTest/integration_test.go
```diff
@@ -818,6 +818,9 @@ func makeTests(t *testing.T) []*TestGroup {
 		),
 
 		testgroup("MX",
+			not(
```
tlimoncelli, line 821: Rather than using "not()" please add this to `loopia/auditrecords.go` (See `azuredns/auditrecords.go` as an example)

PR #2216, opened 2023-03-23
HEDNS: Fix support for LOC records

rblenkinsopp:
This fixes support for LOC records which until were erroneously indicated as supported, despite not working correctly. With this fix, the changes in https://github.com/StackExchange/dnscontrol/issues/2214 all pass correctly for the HEDNS provider.

### DIFF
--- a/providers/hedns/hednsProvider.go
+++ b/providers/hedns/hednsProvider.go
@@ -361,7 +361,7 @@ func (c *hednsProvider) GetZoneRecords(domain string) (models.Records, error) {
 		}
 
 		// Ignore record types that dnscontrol does not support
-		if rc.Type == "HINFO" || rc.Type == "AFSDB" || rc.Type == "RP" || rc.Type == "LOC" {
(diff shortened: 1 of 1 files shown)


tlimoncelli (commented):
One small change:

Line 52: Change CanUseLOC to providers.Cannot().

rblenkinsopp:
Sorry my original message might have been unclear on re-reading:

- HEDNS always supported LOC records, but dnscontrol did not.
- When I previously enabled `CanUseLOC` there was a bug which prevented it working correctly as it was still filtering LOC out as unsupported, I think the `Single_LOC_record` test must not have existed at this point as integration tests all worked.
- This MR fixes the bug so LOC records are fully supported on HEDNS :)

tlimoncelli:
> Sorry my original message might have been unclear on re-reading:

Thanks for the clarification.  The confusion was on my end (I shouldn't read github messages on my phone!)

PR #2217, opened 2023-03-23
NSONE: strip dots from nameservers

norman-zon:
Apparently NS1 changed their API, since 2023/03/22 we see 'ERROR: provider code leaves trailing dot on nameserver'.


tlimoncelli:
Thanks for fixing that!

costasd:
Thanks!

norman-zon:
Would it be possible to do a hotfix release soon for this?

tlimoncelli:
> Would it be possible to do a hotfix release soon for this?

Will do!

norman-zon:
@tlimoncelli it works fine! Thanks for the quick release 👍

PR #2219, opened 2023-03-23
DOCS: Fix number type in typescript 

### LINKED ISSUE #2213: The latest version writes types that contain non standard numbers
I just tried the v3.29.0 preview in my other issues and discovered that when I write the updated types they now contain native number types that do not exist in standard TypeScript, like `float32` and `int`.

This means that TypeScript can't validate the types properly and I get warnings such as these when I try:
```
types-dnscontrol.d.ts:764:28 - error TS2304: Cannot find name 'uint32'.

764 declare function LOC(deg1: uint32, min1: uint32, sec1: float32, deg2: uint32, min2: uint32, sec2: float32, altitude: uint32, size: float32, horizontal_precision: float32, vertical_precision: float32): DomainModifier;
```

The only number type that should be used is `number`.

systemcrash:
fixes #2213

### DIFF
--- a/documentation/functions/domain/LOC.md
+++ b/documentation/functions/domain/LOC.md
@@ -14,17 +14,35 @@ parameters:
 parameter_types:
   name: string
   target: string
-  deg1: uint32
-  min1: uint32
-  sec1: float32
-  deg2: uint32
-  min2: uint32
-  sec2: float32
-  altitude: uint32
-  size: float32
-  horizontal_precision: float32
-  vertical_precision: float32
+  deg1: number
+  min1: number
+  sec1: number
+  deg2: number
+  min2: number
+  sec2: number
+  altitude: number
+  size: number
+  horizontal_precision: number
+  vertical_precision: number
 ---
+
+The parameter number types are as follows:
+
+```
+name: string
+target: string
+deg1: uint32
+min1: uint32
+sec1: float32
+deg2: uint32
+min2: uint32
+sec2: float32
+altitude: uint32
+size: float32
+horizontal_precision: float32
+vertical_precision: float32
+```
+
 
 ## Description ##
 
--- a/documentation/functions/record/LOC_BUILDER_DD.md
+++ b/documentation/functions/record/LOC_BUILDER_DD.md
@@ -6,20 +6,21 @@ parameters:
   - y
   - alt
   - ttl
+parameters_object: true
 parameter_types:
-  label: string
-  x: float32
-  y: float32
-  alt: float32
-  ttl: int
+  label: string?
+  x: number
+  y: number
+  alt: number?
+  ttl: Duration?
 ---
 
 `LOC_BUILDER_DD({})` actually takes an object with the following properties:
 
   - label (optional, defaults to `@`)
-  - x
-  - y
-  - alt
+  - x (float32)
+  - y (float32)
+  - alt (float32, optional)
   - ttl (optional)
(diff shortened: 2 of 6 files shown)


DavidZidar:
Thanks for looking into this, there are more lines that need to be changed though. Look at line 2277 in `dnscontrol.d.ts` for example, there are several more.

systemcrash:
Yeah, forgot about those helpers. This oughtta do it.

DavidZidar:
Looks good to me, the TypeScript compiler seems happy too.

documentation/functions/domain/LOC.md
```diff
@@ -11,6 +11,24 @@ parameters:
...
+  size: number
+  horizontal_precision: number
+  vertical_precision: number
+---
+
+The parameter number types are as follows:
+
+```
 parameter_types:
   name: string
```
j-f1, line 33: Instead of specifying the types in a YAML format, can you mention them in prose wherever they’re relevant? (Like the other record types do)

documentation/functions/record/LOC_BUILDER_DD.md
```diff
@@ -6,13 +6,25 @@ parameters:
...
   label: string
   x: float32
   y: float32
   alt: float32
   ttl: int
----
+```
+
 
 `LOC_BUILDER_DD({})` actually takes an object with the following properties:
```
j-f1, line 29: Can you add a `parameters_object: true` entry to any of these builders that take an object instead of a list of parameters? That way the TS types will be correct.

j-f1 (changes requested):

documentation/functions/domain/LOC.md
j-f1: Here's an example: https://github.com/StackExchange/dnscontrol/blob/master/documentation/functions/domain/TLSA.md

documentation/functions/record/LOC_BUILDER_STR.md
```diff
@@ -5,12 +5,22 @@ parameters:
   - str
   - alt
   - ttl
+parameter_types:
+  label: string
+  str: string
+  alt: number
+  ttl: number
```
j-f1, line 12: Also if possible this should use the `Duration` type (if the TTL is required) or `Duration?` (if it isn’t) to reflect that the helper calls the function to parse values like `"1h"`. (If the helper doesn’t do that it should, but that could be a separate PR)

documentation/functions/record/LOC_BUILDER_DD.md
systemcrash: How would that look?
```yaml
---
name: LOC_BUILDER_DMS_STR
parameters_object: 
  - true
parameters:
  - label
  - str
  - alt
  - ttl
parameter_types:
  label: string
  str: string
  alt: number
  ttl: number
---
```

documentation/functions/record/LOC_BUILDER_STR.md
systemcrash: Sure, that can go in.

systemcrash: Thanks for the tips - news to me :)

documentation/functions/domain/LOC.md
j-f1: It looks like the yaml format for this builder is still there?

PR #2226, opened 2023-03-24
TRANSIP: Enable CI integration tests

cafferata:
Enable CI integration tests for TransIP. cc: @blackshadev 

_The TransIP credentials have been sent via https://transfer.secretoverflow.com/u/tlimoncelli_

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -283,6 +283,7 @@ workflows:
                 - HEDNS
                 - INWX
                 - POWERDNS
+                - TRANSIP
           filters:
             branches:
               ignore: [master, main]
@@ -309,6 +310,7 @@ workflows:
                 - HEDNS
                 - INWX
                 - POWERDNS
+                - TRANSIP
           filters:
             branches:
               ignore: [master, main]


blackshadev:
Thanks, I did notice that the Integration tests do not succeed. It seems TransIP started to check the validity of the whole zone. So if you refer a local record in an MX record it must exists. This does not bode wel with te integration test which at some point tries to remove (incrementally)  an A record which is later referred in an MX record.

blackshadev:
I can resolve the issue by using the zonal differ in the new diff2 branch. But my diff2 changes never seen to hit master. I created an issue for it: #2227

blackshadev:
I am unsure how to resolve these issues to be honest. 

I need either:
 - be able to sort all records by type such that CNAME, NS, MX or SRV are deleted first. and A , AAAA records are created first. But I still need to diff by recordset. So this seem to be really hard to do
 - Or I need to be able to diff on the whole zone and just replace it, but than we do not have a "diff" per record anymore.

Does anybody have a good suggestion for this one?

cafferata:
@blackshadev thanks for your comment! It was my intention to inform you that the automatic CI integration tests are now guaranteed by transferring TransIP credentials.

.circleci/config.yml
```diff
@@ -90,6 +90,10 @@ jobs:
       NAMEDOTCOM_URL: api.name.com
       NAMEDOTCOM_USER: dnscontroltest
 
+      TRANSIP_DOMAIN: $TRANSIP_DOMAIN
```
tlimoncelli, line 93: Any `X: $X` statement can be eliminated (I need to update the docs to mention this).  In your case, that eliminates all of these.

.circleci/config.yml
```diff
@@ -153,6 +157,10 @@ jobs:
...
```
tlimoncelli, line 160: Any `X: $X` statement can be eliminated (I need to update the docs to mention this).  In your case, that eliminates all of these.

.circleci/config.yml
cafferata: Fixed. 👌

.circleci/config.yml
cafferata: Fixed. 👌

PR #2250, opened 2023-03-30
DNSimple: Remove CAA Whitespace in Target restriction

onlyhavecans:
DNSimple has updated how their validators work and fixed the inability to handle whitespace in CAA target records.

I also verified that all the other audits are still required and updated their comments.


tlimoncelli (approved):
Sweet!

Glad to see the improvements!

Issue #2263, opened 2023-04-07
An additional step to enable TypeScript type-checking in VSCode


BirkhoffLee:
Hello! Today I stumbled upon DNSControl. I read about the TypeScript integration and tried to have it enabled in VSCode. However it seemed that VSCode was not recognising the `.d.ts` file automatically -- I had to add this line to the top of the file to have it working properly. Thought this would be helpful to have in the documentations!

```
/// <reference path="types-dnscontrol.d.ts" />
```

j-f1:
Yep, I must have missed that I added that. Would you be willing to submit a PR to the docs to add an instruction about that? (If not, I’d be happy to)

BirkhoffLee:
I'll do that!

tlimoncelli:
Thanks! Looks good!

PR #2264, opened 2023-04-07
DOCS: add reference tag for TypeScript server to recognise the type definitions

BirkhoffLee:
Per #2263 :) This will help reduce confusion on TypeScript setup.

### DIFF
--- a/documentation/typescript.md
+++ b/documentation/typescript.md
@@ -38,6 +38,7 @@ Add this comment to the top of your `dnsconfig.js` file:
 
 {% code title="dnsconfig.js" %}
 ```javascript
+/// <reference path="types-dnscontrol.d.ts" />
 // @ts-check
 ```
 {% endcode %}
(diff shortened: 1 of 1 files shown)


documentation/typescript.md
```diff
@@ -38,6 +38,7 @@ Add this comment to the top of your `dnsconfig.js` file:
 
 {% code title="dnsconfig.js" %}
 ```javascript
+/// <reference path="types-dnscontrol.d.ts" />
 // @ts-check
```
j-f1, line 42: The line above should probably say “add these comments” since there are two now. Also, I think `@ts-check` is supposed to be first.
```suggestion
// @ts-check
/// <reference path="types-dnscontrol.d.ts" />
```

j-f1 (approved):
LGTM!

tlimoncelli (commented):
Can we add a comment that tells people what command to type to generate that file?  I'm sure people will be grateful months later when they realize they need an update.

j-f1:
That’s included in the comment generated on the top of the `.d.ts` file.

PR #2266, opened 2023-04-08
DOCS: Provider requests

cafferata:
Added two provider requests (Plesk #2261 & BunnyDNS #2265) and sorted [the provider requests list](https://docs.dnscontrol.org/service-providers/providers#requested-providers).


tlimoncelli:
Thanks!

PR #2273, opened 2023-04-12
MSDNS: Document pssession options

shodges-vermeer:
Added optional variables of psusername and pspassword

Found in source code at dnscontrol/providers/msdns/powershell.go

### DIFF
--- a/documentation/providers/msdns.md
+++ b/documentation/providers/msdns.md
@@ -31,6 +31,8 @@ along with other settings:
 
 * `dnsserver`: (optional) the name of the Microsoft DNS Server to communicate with.
 * `pssession`: (optional) the name of the PowerShell PSSession host to run commands on.
+* `psusername`: (optional) the username to connect to the PowerShell PSSession host.
+* `pspassword`: (optional) the password to connect to the PowerShell PSSession host.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Those settings are untested.  Have you used them successfully?  My reading of the code says they should work, but I'm not in a position to test them.

shodges-vermeer:
Ahh okay. I'm running into issues with getting it to work, but it may be unrelated to the credentials portion. Let me do some more digging and I'll get back to ya.

shodges-vermeer:
Authentication appears to be working if my understanding of order of operations is correct. When running `dnscontrol preview` it should (I believe) authenticate before echoing "******************** Domain: $domain:". 

For me, once it echos that it just hangs but I'm thinking it may be because a few of our zones are insanely large.

tlimoncelli:
I'm not sure of the order of operations but, yes, large zones take a LONG time.

Fun story: When I was first writing this I thought there was a terrible bug because the zone processing was taking minutes.   I assumed forking the powershell from Go was super slow... that wasn't it at all.

It turns out the zone dump was a 100 MiB of JSON, 99% of which was data that got thrown away.  That's why there is `Select-Object -Property * -ExcludeProperty Cim*` to filter out the cruft.  It runs a lot faster now. 

If you see any improvements (other properties we could exclude), I'd be glad to optimize this further.

Tom

tlimoncelli:
I'm merging this.  It sounds like it works... the slowness isn't related.  

Thanks for finding this and

shodges-vermeer:
Thanks! Turns out it wasn't slow, just had some firewall rules or something in place that was preventing my laptop from hitting our DC directly. Once we opened that it worked flawlessly.

PR #2274, opened 2023-04-13
POWERDNS: Add option to set the PowerDNS zone kind

jpbede:
This PR adds a option to the PowerDNS provider to set the zone kind when creating a new zone.

If this is not specified in the request, PowerDNS uses `Native`, which results in the zone not being transfered via AXFR/IXFR. `Native` is used to let the storage backend of PDNS handle the replication.

The PR does not break the default behavior, if it is not set, the zone is still created with the `Native` type.

I'm currently waiting for a second tester to test this.

### DIFF
--- a/documentation/providers/powerdns.md
+++ b/documentation/providers/powerdns.md
@@ -29,13 +29,15 @@ Following metadata are available:
         'a.example.com.',
         'b.example.com.'
     ],
-    'dnssec_on_create': false
+    'dnssec_on_create': false,
+    'zone_kind': 'Native',
 }
 ```
 {% endcode %}
 
 - `default_ns` sets the nameserver which are used
 - `dnssec_on_create` specifies if DNSSEC should be enabled when creating zones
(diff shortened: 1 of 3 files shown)


documentation/providers/powerdns.md
```diff
@@ -29,13 +29,15 @@ Following metadata are available:
...
-    'dnssec_on_create': false
+    'dnssec_on_create': false,
+    'zone_kind': 'Native',
 }
 ```
 {% endcode %}
 
 - `default_ns` sets the nameserver which are used
 - `dnssec_on_create` specifies if DNSSEC should be enabled when creating zones
+- `zone_kind` is the type that will be used when creating the zone. Can be one of `Native`, `Master` or `Slave`, when not specified it defaults to `Native`. Please see [[PowerDNS documentation](https://doc.powerdns.com/authoritative/modes-of-operation.html) for explanation of the kinds.
```
tlimoncelli, line 40: Suggestion: Point out that those tokens are case-sensitive.

tlimoncelli (approved):

PR #2279, opened 2023-04-16
HEXONET: Remove duplicate code

tlimoncelli:
```
$ staticcheck ./...
providers/hexonet/records.go:61:71: argument actual is overwritten before first use (SA4009)
	providers/hexonet/records.go:63:2: assignment to actual
```


tlimoncelli:
CC @KaiSchwarz-cnic for visibility

Issue #2283, opened 2023-04-17
BIND Provider: Writing to wrong file


fritterhoff:
We are generating bind files using dnscontrol. Since the last update the files are written to the wrong files:
```
******************** Domain: 253.39.141.in-addr.arpa
1 correction (bind)
#1: GENERATE_ZONEFILE: '253.39.141.in-addr.arpa'. Changes:
...
WRITING ZONEFILE: zones/84.187.129.in-addr.arpa.zone
SUCCESS!
******************** Domain: 254.39.141.in-addr.arpa
```

fritterhoff:
P.S. `84.187.129.in-addr.arpa.zone` is the first defined zone in our case. Seems like some properties are not replaced correctly?

fritterhoff:
From my point of view the removal of these lines causes the error:
https://github.com/StackExchange/dnscontrol/compare/v3.30.0...v3.31.0#diff-97c9fab3d30b65d4202799888dd5ecd1ef0ea62955cac3d6fdb07f37f48ca27dL221-L222

PR #2284, opened 2023-04-17
fix: generate file name on bind zone generation

fritterhoff:
Should fix: #2283


tlimoncelli:
Good catch!  Thanks for finding both the problem and fixing it!

fritterhoff:
Do you plan to release some sort of Bugfix release? :)

tlimoncelli:
Working on that later today!

PR #2298, opened 2023-04-26
NS1: Fix NS1_URLFWD records push

flz:
Previous testing was done with dnscontrol preview only. Lesson learnt.


flz:
This is part 2 of #2298.

PR #2301, opened 2023-05-01
HETZNER: fix nameserver trailing dots

willpower232:
As per https://github.com/StackExchange/dnscontrol/issues/491#issuecomment-1529072560 hetzner started being affected by the trailing dots on nameservers thing.

My config stopped reporting this error with the change, not sure if you wanted a specific test for this situation to be added.

cc @das7pad as currently listed hetzner maintainer

### DIFF
--- a/providers/hetzner/hetznerProvider.go
+++ b/providers/hetzner/hetznerProvider.go
@@ -149,11 +149,8 @@ func (api *hetznerProvider) GetNameservers(domain string) ([]*models.Nameserver,
 	if err != nil {
 		return nil, err
 	}
-	nameserver := make([]*models.Nameserver, len(z.NameServers))
-	for i, s := range z.NameServers {
-		nameserver[i] = &models.Nameserver{Name: s}
-	}
-	return nameserver, nil
+
+	return models.ToNameserversStripTD(z.NameServers)
 }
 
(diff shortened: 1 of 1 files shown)


das7pad (approved):
LGTM and tested. Thanks!

Unfortunately, the integration tests did not pick this up and manual testing with my production config did not ran past the code that generates the warning.
(I'm using `DnsProvider(NewDnsProvider('...'), 0)` and hard-code a subset of hosts from two providers. The `0` tells dnscontrol to pick zero nameserver from this provider and it bails out just before the existing run-time warning [here](https://github.com/StackExchange/dnscontrol/blob/a0d491e6ce236ab1f931bb22cb449689c9f956af/pkg/nameservers/nameservers.go#L24-L27).)

Hi @tlimoncelli, WDYT about adding an integration test for the trailing dot behavior, like this 135672bb1aa71d993301dae1980bd68324c136c4? (The new test fails on main and passes with this PR. I'm happy to open a PR tomorrow.)

tlimoncelli:
@das7pad Yes, an integration test would be most useful!  A PR similar to   https://github.com/StackExchange/dnscontrol/commit/135672bb1aa71d993301dae1980bd68324c136c4 would be excellent.

Issue #2307, opened 2023-05-01
Add integration test for nameserver trailing dots


tlimoncelli:
Issue created to track @das7pad 's suggestion to add an integration test similar to https://github.com/StackExchange/dnscontrol/commit/135672bb1aa71d993301dae1980bd68324c136c4

PR #2311, opened 2023-05-02
TESTING: add integration test for no trailing dot in nameservers

das7pad:
As suggested in https://github.com/StackExchange/dnscontrol/pull/2301#pullrequestreview-1408080979 this PR is adding an integration test for checking that dnscontrol providers remove any trailing dot from nameservers as returned by the respective dns provider (https://github.com/StackExchange/dnscontrol/issues/491).

_The next time all maintainers run the tests we might see a few remaining providers fix their handling of nameservers. The test failure message points them right at the solution._

Closes https://github.com/StackExchange/dnscontrol/issues/2307

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -319,6 +319,14 @@ func runTests(t *testing.T, prv providers.DNSServiceProvider, domainName string,
 
 	}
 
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Super!  Thanks!

Issue #2319, opened 2023-05-04
NS1 NS1_URLFWD still failing tests


tlimoncelli:
```
=== RUN   TestDNSProviders/test-ft.com/48:NS1_URLFWD_tests:Add_a_urlfwd
    integration_test.go:230: 
        + CREATE URLFWD urlfwd1.test-ft.com / http://example.com 302 2 0 ttl=300
    integration_test.go:251: Expected 0 corrections on second run, but found 2.
    integration_test.go:253: UNEXPECTED #0: + CREATE URLFWD urlfwd1.test-ft.com / http://example.com 302 2 0 ttl=300
    integration_test.go:253: UNEXPECTED #1: - DELETE NS1_URLFWD urlfwd1.test-ft.com / http://example.com 302 2 0 ttl=300
```

tlimoncelli:
CC [flz](https://github.com/flz)

flz:
It looks like we have two internal representations for URLFWD (URLFWD actual type, NS1_URLFWD alias type). However URLFWD isn't a real type. NS1 is the only provider supporting it, CloudFlare has a similar type but handled separately (CF_REDIRECT).

Given URLFWD isn't a real type, I'd suggest getting rid of it altogether to avoid confusion and use NS1_URLFWD only. We'll still need some level of translation between dnscontrol representation (NS1_URLFWD) and NS1 API (URLFWD).

flz:
The test above should pass now but until the revert from #2317 lands, the trailing dot integration will keep failing.

tlimoncelli:
Reverting now...

PR #2320, opened 2023-05-04
NS1 Fix URLFWD implementation by being more consistent

flz:
URLFWD isn't a real record type. There's no need to maintain both a "virtual" type and an alias type when there's a single implementation.

Discussed in #2319.


tlimoncelli:
Thank you!  Yes, that's much better!

PR #2342, opened 2023-05-09
Fix doc URLs in .d.ts file

j-f1:
They went to the old doc site, but the new one has different URLs (that don’t seem to correspond to anything directly in the source files?)


tlimoncelli:
Thanks for fixing this!

PR #2353, opened 2023-05-11
Added support for SOA record in PowerDNS

BomBardyGamer:
After carefully examining the API, testing myself with the data provided in the tests and custom data, I could see no reason why the SOA record should not be supported. As such, I've enabled support for it.All the integration tests pass, which is further evidence that it works just fine.

If there is a reason why the SOA record is disabled for PowerDNS, please feel free to let me know, but I do not see such a reason myself.


tlimoncelli:
CC @jpbede for approval

jpbede:
Basically this is fine and PDNS supports it, but this is problematic when you use the `default-soa-edit` option (https://doc.powerdns.com/authoritative/settings.html#setting-default-soa-edit) from PDNS. This option automatically takes care of incrementing the SOA serial.

As a result, you will always have one change pending per domain:
```
******************** Domain: jpbe.de
1 correction (powerdns)
#1: ± MODIFY jpbe.de SOA (a.anydns.io. zonemaster.anydns.io. 2023033005 10800 3600 1209600 3600 ttl=300) -> (a.anydns.io. zonemaster.anydns.io. 0 3600 600 604800 1440 ttl=120)

[applied the SOA change]

******************** Domain: jpbe.de
1 correction (powerdns)
#1: ± MODIFY jpbe.de SOA (a.anydns.io. zonemaster.anydns.io. 2023051201 3600 600 604800 1440 ttl=120) -> (a.anydns.io. zonemaster.anydns.io. 0 3600 600 604800 1440 ttl=120)
```

This was the main reason I left it disabled.

BomBardyGamer:
Yeah that is fair enough. I wonder if there is a way we could exclude the serial from the diffing so that, when an option is enabled, we don't run in to this issue, as the serial won't be considered as modified. I'm not fully sure whether this is too complicated and not really worth it for something as infrequently used as SOA though.

tlimoncelli:
I'm curious about the inspiration for this PR.  Is it for completeness? neatness? Or is there a business process at your site that would be improved or enabled by having the SOA record updated automatically?

BomBardyGamer:
I've got a home server that I'm running PowerDNS on. I remembered this tool from a while ago, and thought I'd try using it to configure PowerDNS, instead of setting up PowerDNS Admin (something I've done multiple times before). I just like configuring the SOA record for completeness, it's nothing to do with business processes or anything like that. I only made the suggestion I did as a way to enable SOA and fix the issue that you mentioned was the reason why it was disabled.

Do I really really need this? Probably not. Do I want it though? Yes I do. That's why I made the PR.

tlimoncelli:
I'm ambivalent.  I'll let @jpbede decide: Merge or reject?

jpbede:
Sry for the late reply, had some problems with my internet.... Fiber cut...

I would reject it, I think this would cause some pain for some users. Maybe I have a way to distinguish if the auto-increment is on or not.
Depending on this we can then activate the SOA record.

BomBardyGamer:
Sounds fair. I'll close it then. Thanks.

Issue #2370, opened 2023-05-17
New record type: OPENPGPKEY


systemcrash:
Would this be interesting?

Some providers support this record type. I mainly did this as code-golf, for fun. 

It is PR ready, just needs a re-base. 

It required a PR for the miekg lib to fix a buffer issue, [which is now merged](https://github.com/miekg/dns/pull/1446). 

It requires at least https://github.com/miekg/dns/commit/730c265d18c3f51b447f6a4c61fe3cde336585d8 or 1.1.54

I know many orgs use (ECC) PGP Keys for security disclosures, and some (wish to) post those via DNS.

tlimoncelli:
Hi Paul!

In general I'm weary of adding DNS types that aren't commonly used.  Can I see the PR?


Tom

systemcrash:
You can [preview one here](https://github.com/StackExchange/dnscontrol/compare/master...systemcrash:openpgp_type?expand=1).

tlimoncelli:
Well, that's quite an impressive bit of coding you've done there. You've even implemented the integration tests.

I'd be glad to include this in the project, but you'll be on the hook for maintaining it since it isn't something I have a lot of expertise in.  Ok?

One code change I'd like to suggest: Shift the logic out of helpers.js into the Go code.

A lot of the processing is done in Javascript in helpers.js.  That makes it difficult to test and debug.   (Not that there aren't testing frameworks for Javascript, but we can't use them because of otto, etc.)  In general the functions in helpers.js (like A, CNAME, etc.) just pass along the human input and do the "heavy lifting" in Go.  Go has the advantage of the libraries and testing frameworks that we know well.   I would suggest putting that code in its own module (for example  dnscontrol/pkg/openpgpkey)

Tom

systemcrash:
I looked around for places to put the heavy lifting in go because it seems like better APIs are available especially for the UTF8 normalization parts, but there does not seem to be an appropriate place to do so. The multiple pathways seem to be set up to take data immediately from js files and into use in records. Do I just need to be creative or is there a recommended point to start massaging the data so it's ready for use everywhere, irrespective of the ingress route?

tlimoncelli:
Good question!

The Javascript exports the result as JSON.  The fields are the names of the object keys.  The import is done by the UnmarshlJSON functions in models/record.go:

    func (rc *RecordConfig) UnmarshalJSON(b []byte) error

[...]

PR #2374, opened 2023-05-18
OPENPGPKEY record type

tlimoncelli:
Step 1: Update RecordConfig in models/record.go ✅

Step 2: Add a capability for the record ✅
+providerCapabilityChecks ✅
+TestCapabilitiesAreFiltered ✅

Step 3: Add a helper function ✅

OPENPGPKEY()

Plus extra internal functions to hash and process

Step 4: Search for #rtype_variations ✅

Step 5: Add a parse_tests test case ✅

Step 6: Add an integrationTest test case ✅

--- PASS: TestDNSProviders (0.01s)
    --- PASS: TestDNSProviders/example.com (0.01s)
        --- PASS: TestDNSProviders/example.com/Clean_Slate:Empty (0.00s)
        --- PASS: TestDNSProviders/example.com/28:OPENPGPKEY:OPENPGPKEY_create_first_record_->_36_bytes (0.00s)
        --- PASS: TestDNSProviders/example.com/28:OPENPGPKEY:OPENPGPKEY_create_second_record_->_304_bytes (0.00s)
        --- PASS: TestDNSProviders/example.com/28:OPENPGPKEY:OPENPGPKEY_create_second_subrecord_->_3004_bytes (0.00s)
        --- PASS: TestDNSProviders/example.com/28:OPENPGPKEY:OPENPGPKEY_delete_record_->_3004_bytes (0.00s)
        --- PASS: TestDNSProviders/example.com/28:OPENPGPKEY:OPENPGPKEY_change_target_->_304_to_3004_bytes (0.00s)
        --- PASS: TestDNSProviders/example.com/Post_cleanup:Empty (0.00s)
Step 7: Support more providers ✅ (bind, axfr)

Step 8: Write documentation ✅


brechsteiner:
Hi @tlimoncelli, i am interested in this record. Is there a reason why the PR was not merged?

tlimoncelli:
He withdrew the PR after I asked him to move some logic to the Go module.  See https://github.com/StackExchange/dnscontrol/issues/2370

@systemcrash if you'd like to resubmit, I'll consider it again.

systemcrash:
I never actually made a PR - I was curious whether my code would pass muster. I was not satisfied with where data processing should occur vs what Tim wanted, since this record type needs a bit doing: utf8 normalisation, hashing and truncating. As long as the input text is basic text where one character fits into 8 bits, everything is fine. If it contains obscure utf8 code points, probably not. Some of the tests I wrote handle Japanese, so most text should work fine. 

The js marshals the data into use immediately, despite Go being better to handle data processing. So my code favors getting most common text working correctly over procedural correctness, ie all work is done in js but has virtually no testing. Because the processing is standardized, it's possible for an external implementation to get a correct result while the js code here does not, producing a differing hash. But only if the openpgp corresponding email address is something weird like half width hiragana and complex emoji. 🔢

One way around this limitation that I can think of is having some internal flag for each record to signal that Go does the utf8 normalization etc. Because there is no trivial way to know that data has been otherwise processed since everything is just text. A better one might be to import records with the right hand part and trim it away when done for its respective (sub)domain. This seems better in case we want to be able to import record format ready data.

TL;DR: what we have today should work but is not standardized.

https://www.rfc-editor.org/rfc/rfc7929.html

brechsteiner:
@systemcrash I understand your requirements for the feature. But I think in a basic version it would be enough if you had to provide the hash as input.

systemcrash:
These aren't my requirements. They're a standard. If you just hash, you
might get something completely different without the preprocessing.

brechsteiner:
okay. No, I actually meant that the user input from the config file should be taken over 1:1 without processing.

`OPENPGPKEY("2360697cee...423ebc8d052acb1aecb04", "mQINBF...OuYhQ=="),`

systemcrash:
That's certainly one way of doing it. I don't recall whether my
implementation does that.

brechsteiner:
@tlimoncelli as long as the rtype OPENPGPKEY is not supported, I wanted to use IGNORE() as a workaround. Unfortunately this does not work because the rtype cannot be parsed. Is there a way to exclude an rtype from parsing?

tlimoncelli:
> @tlimoncelli as long as the rtype OPENPGPKEY is not supported, I wanted to use IGNORE() as a workaround. Unfortunately this does not work because the rtype cannot be parsed. Is there a way to exclude an rtype from parsing?

Sadly not.  DNSControl can only ignore records that it understands.

I know that sounds weird.  "Why can't it just ignore what it doesn't understand?"

Here are some examples:

* A BIND user has a zone that wishes to ignore `foo IN OPENPGPKEY x...`.   Bind requires that we generate  the entire zonefile every time. Therefore, to ignore "foo", we must read it from the zonefile then generate the new zonefile with changes.  If there are any ignored records, we must generate them too.  If DNScontrol doesn't "understand" OPENPGPKEY records, it can't generate them.
* GANDI_V5 does updates one "label" at a time.  There is no way to update a particular record. The only API calls for updates are either "delete all the records at label `foo`" or "replace all the records at label `foo`".  A label (`foo`) can have multiple records of different types: A, MX, and so on.  Suppose `foo IN OPENPGPKEY x...` and `foo IN A 1.2.3.4` both exists. If we want to change the "A" record, but ignore the OPENPGPKEY record, we need to be able to re-generate the OPENPGPKEY record and generate the replacement "A" record.  Again, this requires DNSControl to understand OPENPGPKEY.

We could have IGNORE() work for some providers but not others, but that would create a mess.

tacerus:
Hi all, I made https://github.com/StackExchange/dnscontrol/pull/3718 to add a simple implementation, I think it is better than nothing, and if someone wants to pick up the extended variant, it could be added on top in a later improvement.

PR #2375, opened 2023-05-18
DOCS: Fix quote and a broken link

tlimoncelli:


tlimoncelli:
CC @cafferata 

Thanks for pointing out these typos!

PR #2381, opened 2023-05-20
DESEC: implement ListZones

### LINKED ISSUE #2346: DESEC: FEATURE REQUEST: Enable get-zones (ListZones) for deSEC
If I configure `creds.json` and `dnsconfig.js` to use deSEC [as instructed](https://docs.dnscontrol.org/service-providers/providers/desec) and then run either `dnscontrol check-creds desec` or `dnscontrol get-zones --format=nameonly desec - all`, I get the following error:

```
provider type desec: cannot list zones to use the 'all' feature
```

However, if I [query the deSEC API](https://desec.readthedocs.io/en/latest/dns/domains.html#listing-domains) directly, I do get back a json response with all the domains on my account. Is it possible to get this implemented?

imlonghao:
close #2346

the zones list are already loaded in provider initialize, this PR simply load the local cache to implement ListZones

https://github.com/StackExchange/dnscontrol/blob/0edffa198b0a09acdf086cb6fa7026a7aaef628c/providers/desec/protocol.go#L82-L127

```
# dnscontrol get-zones --format=nameonly desec - all
desec.esd.cc
```

### DIFF
--- a/providers/desec/desecProvider.go
+++ b/providers/desec/desecProvider.go
@@ -248,3 +248,12 @@ func (c *desecProvider) GetZoneRecordsCorrections(dc *models.DomainConfig, exist
 
 	return corrections, nil
 }
+
+// ListZones return all the zones in the account
+func (c *desecProvider) ListZones() ([]string, error) {
+	var domains []string
+	for domain := range c.domainIndex {
+		domains = append(domains, domain)
+	}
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Excellent!  Thanks!

PR #2382, opened 2023-05-20
Update version in go.mod

nemunaire:
The recent major release breaks the usage of dnscontrol as a module:

```
go.mod:6:2: require github.com/StackExchange/dnscontrol/v3: version "v4.0.1" invalid: should be v3, not v4
```

or :

```
github.com/StackExchange/dnscontrol/v3/models: github.com/StackExchange/dnscontrol/v4@v4.0.1: go.mod has non-.../v4 module path "github.com/StackExchange/dnscontrol/v3" (and .../v4/go.mod does not exist) at revision v4.0.1
```

In this commit, I changed the module in `go.mod` to `github.com/StackExchange/dnscontrol/v4`, as stated [in the documentation](https://docs.dnscontrol.org/release/release-engineering#tip-how-to-bump-the-major-version).


tlimoncelli:
oops!  I wrote that documentation and I totally forgot to do this when I bumped to v4!  Thanks for catching this!

PR #2407, opened 2023-05-31
BIND: reset zonefile for reading records

### LINKED ISSUE #2406: Bind: Wrong files are used
Similar to #2283 I noticed, that the wrong files are parsed. 

In case of using two zones with a bind provider the first reads the correct zone, the second reuses the first zone ...

From my point of view even in the "non-test" mode this section is used:

https://github.com/StackExchange/dnscontrol/blob/e6f3fb42f2952429f6fad8a91de8f87fe4b3550f/providers/bind/bindProvider.go#L162-L168

fritterhoff:
Fixes #2406

### DIFF
--- a/providers/bind/bindProvider.go
+++ b/providers/bind/bindProvider.go
@@ -158,13 +158,19 @@ func (c *bindProvider) GetZoneRecords(domain string, meta map[string]string) (mo
 	if _, err := os.Stat(c.directory); os.IsNotExist(err) {
 		printer.Printf("\nWARNING: BIND directory %q does not exist! (will create)\n", c.directory)
 	}
-
-	if c.zonefile == "" {
+	_, okTag := meta[models.DomainTag]
+	_, okUnique := meta[models.DomainUniqueName]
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Thanks for finding and fixing that!

fritterhoff:
Do you (maybe) plan a small bugfix release? :)

TomOnTime:
@fritterhoff Shipped! https://github.com/StackExchange/dnscontrol/releases/tag/v4.1.1

fritterhoff:
Yay! Thank's again for this great tool!

TomOnTime:
I'm just glad other people find it useful!

Thanks for all YOUR help!

fritterhoff:
If you are interested in some details: 

We are currently managing 66 different public second level domains + 34 /24 public reverse zones using git & dnscontrol. These zones contain several sub zones and overall approx. 1600 A-records. 
In the near future we plan to also manage our internal records - far more than 5000 A-records. 

So yes ... it is very useful!

PR #2419, opened 2023-06-02
Add ordering of  changes based on the record dependencies

blackshadev:
## What

- Order changes based on dependency using dnssort
- Add dnssort package which creates a dependency graph  and sorts based on the graph
- Report changes are not sorted and tracked
- Using forward and backward dependencies using the newRecords en oldRecords in a change


models/record.go
```diff
@@ -92,6 +92,7 @@ type RecordConfig struct {
 	TTL       uint32            `json:"ttl,omitempty"`
 	Metadata  map[string]string `json:"meta,omitempty"`
 	Original  interface{}       `json:"-"` // Store pointer to provider-specific record object. Used in diffing.
+	Origin    string            `json:"origin"`
```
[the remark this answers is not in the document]
blackshadev: This got removed in favor of making all targets FQDN

models/record.go
```diff
@@ -455,6 +459,31 @@ func (rc *RecordConfig) ToRR() dns.RR {
 	return rr
 }
 
+func (rc *RecordConfig) GetNameFQDN() string {
```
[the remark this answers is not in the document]
blackshadev: The dnssort package uses an interface named `SortableChange` which has `GetNameFQDN`, `GetType` and `GetFQDNDependencies`. Using the interface instead of a direct dependency on RecordConfig makes it more reusable for other uses if need be (and a bit easier to test)

models/record.go
```diff
@@ -455,6 +459,31 @@ func (rc *RecordConfig) ToRR() dns.RR {
...
+	return rc.NameFQDN
+}
+
+func (rc *RecordConfig) GetFQDNDependencies() []string {
```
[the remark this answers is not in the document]
blackshadev: Do note, this method name comes from the `SortableChange` interface. But GetDependencies sounds good enough, I will add a comment above it the returned strings are the dependent record keys as FQDNs

pkg/dnssort/tree_test.go
```diff
@@ -0,0 +1,60 @@
...
+			[]string{"com", "x.example.com", "x.www.example.com", "example.com"},
+		),
+	)
+
+	t.Run("Wildcard",
+		executeTreeTest(
+			[]string{
+				"*.example.com",
+			},
+			[]string{"example.com", "other.example.com"},
```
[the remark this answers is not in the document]
blackshadev: Missing, I will add it to the test case. the wildcard is only resolved when adding things to the tree, not when looking them up. This is on purpose because the record dependencies never contain wildcards, while record names can.

models/record.go
```diff
@@ -161,6 +162,7 @@ func (rc *RecordConfig) UnmarshalJSON(b []byte) error {
 		TTL       uint32            `json:"ttl,omitempty"`
 		Metadata  map[string]string `json:"meta,omitempty"`
 		Original  interface{}       `json:"-"` // Store pointer to provider-specific record object. Used in diffing.
+		Origin    string            `json:"origin"`
```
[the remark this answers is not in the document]
blackshadev: Removed in latest version in favor of making all Targets and Dependencies FQDN

pkg/js/parse_tests/030-dextenddoc.json
```diff
@@ -1,25 +1,104 @@
...
-        { "name": "bbb.sub", "subdomain": "sub", "target": "127.0.0.4", "type": "A" },
-        { "name": "ccc.sub", "subdomain": "sub", "target": "127.0.0.5", "type": "A" },
-        { "name": "e.sub", "subdomain": "sub", "target": "f", "type": "CNAME" },
-        { "name": "ddd.sub.sub", "subdomain": "sub.sub", "target": "127.0.0.6", "type": "A" },
-        { "name": "g.sub.sub", "subdomain": "sub.sub", "target": "h", "type": "CNAME" },
-        { "name": "sub", "subdomain": "sub", "target": "127.0.0.7", "type": "A" },
-        { "name": "i.sub", "subdomain": "sub", "target": "j", "type": "CNAME" }
-      ],
-      "registrar": "Third-Party"
+        {
```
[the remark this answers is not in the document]
blackshadev: I am going to remove all parse_tests changes from my PR since this shouldn't be needed anymore. It was only changed because of the Origin change

models/record.go
```diff
@@ -591,15 +617,15 @@ func CanonicalizeTargets(recs []*RecordConfig, origin string) {
 		switch r.Type { // #rtype_variations
 		case "AKAMAICDN", "ANAME", "CNAME", "DS", "MX", "NS", "NAPTR", "PTR", "SRV":
 			// Target is a hostname that might be a shortname. Turn it into a FQDN.
-			r.target = dnsutil.AddOrigin(r.target, origin)
+			r.target = dns.Fqdn(dnsutil.AddOrigin(r.target, origin))
```
TomOnTime, line 620:
```suggestion
			r.target =dnsutil.AddOrigin(r.target, origin+"."))
```

TomOnTime: Suggestion: Earlier in this function do: `originDot := origin + "."` then use originDot here and on line 625 and 628.

TomOnTime (commented):
This is a lot cleaner now.  Lots of progress!

I think it would be useful to have a comment that explains the high-level algorithm, in pseudo-code, at the top of one of the main files.  I must confess I don't have a very good high-level view of that the algorithm is.  By that I mean... When I look at code, I try to build a mental model of what's going on by looking at the comments and code.  At this point I'm not clear what the high-level algorithm is.  

There's also a chance that there's a way we can simplify this a bit.  The pkg/dnssort/tree.go  code is rather brilliant and I think we can use it to simplify other parts.  Are you open to discussing a radical change?

models/record.go
```diff
@@ -587,19 +613,20 @@ func Downcase(recs []*RecordConfig) {
 
 // CanonicalizeTargets turns Targets into FQDNs
 func CanonicalizeTargets(recs []*RecordConfig, origin string) {
+	originFQDN := dns.Fqdn(origin)
```
TomOnTime, line 616:
```suggestion
	originFQDN := origin+"."
```

There's never a situation where the "." already exists.  If there is, that would be an API change and we'd actually want this to fail.

TomOnTime (commented):
Reviewing the code more deeply I have some more structural suggestions:

I think the code would be more understandable if the code was split up into smaller packages, each doing one thing (the search tree, the graph, etc)
*  pkg/dnstree/  
    * type DNSGraph and all related functions.
*  pkg/graphsort/
    * type DNSGraph, SortUsingGraph(), and related.
*  pkg/reorder/
    * orderByDependencies() // rename to ByDependencies()
    * CreateGraph() // rename to populateGraph()
    * removeNode()/addNode()/addEdge() 

By putting CreateGraph() in a separate module, it would force graphsort to be more generic, which is a good thing.  In fact, (and this is the radical change I was thinking of) the graph could store indexes into the ChangeList.  The output of sorting would be a list of indexes into the change list.  Let's say there are 4 changes. The output might be: [3, 1 , 0, 2]. Indexes are easy to work with, debug, and write tests for.

The last step would be to assemble the new changelist by copying the old changelist in the order of the sort (`[3, 1 , 0, 2]`), then copy any unmentioned indexes to the end (or front?)

models/record.go
blackshadev: Good one, I added the suggestion (roughly, I like `dns.Fqdn` better to ensure we are not adding a dot after an already , made fqdn.
I also added doc comments to the main function of the dnssort and dnstree. 

Of course I am open to discussing radical changes. If these suggested changes are outside of the re-ordering scope, we can always create a seperate PR for it to not incorporate to much changes in this one. What did you have in mind?

TomOnTime:
PS. I'm on sabbatical for the next 8 weeks.  I won't be using my work account (@tlimoncelli ) but instead will use my non-work account (@TomOnTime ).

models/record.go
blackshadev: Ok I will pick this up when I implement the rest of your suggestions.

blackshadev:
> Reviewing the code more deeply I have some more structural suggestions:
> 
> I think the code would be more understandable if the code was split up into smaller packages, each doing one thing (the search tree, the graph, etc)
> 
> * pkg/dnstree/
>   
>   * type DNSGraph and all related functions.
> * pkg/graphsort/
>   
>   * type DNSGraph, SortUsingGraph(), and related.
> * pkg/reorder/
>   
>   * orderByDependencies() // rename to ByDependencies()
>   * CreateGraph() // rename to populateGraph()
>   * removeNode()/addNode()/addEdge()
> 
> By putting CreateGraph() in a separate module, it would force graphsort to be more generic, which is a good thing. In fact, (and this is the radical change I was thinking of) the graph could store indexes into the ChangeList. The output of sorting would be a list of indexes into the change list. Let's say there are 4 changes. The output might be: [3, 1 , 0, 2]. Indexes are easy to work with, debug, and write tests for.
> 
> The last step would be to assemble the new changelist by copying the old changelist in the order of the sort (`[3, 1 , 0, 2]`), then copy any unmentioned indexes to the end (or front?)

I am unsure about the reordering using indices, but I like you suggestion about splitting the package up into multiple. I will start with that and try to form an opinion on the index list als return type.

TomOnTime:
> I am unsure about the reordering using indices, but I like you suggestion about splitting the package up into multiple. I will start with that and try to form an opinion on the index list als return type.

Very wise and reasonable!

TomOnTime:
Looks great!  Splitting those out into separate pkg directories definitely makes it more readable.  Thanks!

blackshadev:
Regarding the restructuring: I think it is complete. Can you re-review on that part? I did not do it exactly as you said, more so because I think the current splitting lends itself better to future re-usage (with the dnsgraph being stand-alone and the dnssort just using the graph. Likewise with the tree).

Regarding the the index list as a return type. But I am finding it really hard to implement and I don't think it will benefit the code all that much. 

This is how I see it:
- If we want to return an index list we need to actively track the indices as part of the data in the graph, which will make it less generic and more focused onto the dnssort use-case.
- It does make writing assertions in tests easier, but less readable. I will end up with writing a helper to spit out proper error messages with the domain in it, instead of "expected 4 got 5".
- It adds complexity as mentioned above, but also to apply said order on a target slice. While the current code builds the sorted result as it resolve less dependent records. Which is what the algorithm lends itself for. 
- It is less performant both in terms of CPU en  memory usage. All be it this is neglect-able. I need to iterate the  record set at least once more (but to keep things readable and separated I think I will end up iterating it thrice more). 

If you still think it is really worth the effort: please do enlighten me and  I will continue with the effort. But for now I will pause it since I am not seeing how this will be better than what we have now.

pkg/diff2/ordering.go
```diff
@@ -0,0 +1,18 @@
+package diff2
+
+import (
+	"log"
+
+	"github.com/StackExchange/dnscontrol/v4/pkg/dnsgraph"
+	"github.com/StackExchange/dnscontrol/v4/pkg/dnssort"
+)
+
+func orderByDependencies(changes ChangeList) ChangeList {
```
tlimoncelli, line 10: Let's put this feature (in all  behind a flag at least temporarily.  The flag should default to on, but `--disableordering` would make this function `return changes` unmodified.

See `dnscontrol/pkg/bindserial/main.go`  for how to add a global flag that is temporary.

I'm not saying that the code is bad or anything... just that new features like this should be easy to disable until we've had them in production for a while.

Tom

pkg/diff2/compareconfig.go
```diff
@@ -93,6 +93,11 @@ type targetConfig struct {
 
 // NewCompareConfig creates a CompareConfig from a set of records and other data.
 func NewCompareConfig(origin string, existing, desired models.Records, compFn ComparableFunc) *CompareConfig {
+	models.Downcase(existing)
+	models.Downcase(desired)
+	models.CanonicalizeTargets(existing, origin)
+	models.CanonicalizeTargets(desired, origin)
```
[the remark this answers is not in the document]
blackshadev: I removed these and added them on the analyze test, where I actually needed them.

pkg/diff2/ordering.go
```diff
@@ -0,0 +1,18 @@
...
+	a := dnssort.SortUsingGraph(changes)
+
+	if len(a.UnresolvedRecords) > 0 {
+		log.Printf("Found unresolved records %v\n", dnsgraph.GetRecordsNamesForGraphables(a.UnresolvedRecords))
```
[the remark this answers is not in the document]
blackshadev: I added allot more body to the message. Do you think it is improved?

pkg/diff2/ordering.go
blackshadev: I understand, the fact that you want to merge this and have it on by default is trust enough ;) 

I added the flag and added an informational message (just like with diff2) as well.

tlimoncelli: Looks good!  Thanks for understanding!

pkg/diff2/ordering.go
```diff
@@ -0,0 +1,27 @@
...
+		return changes
+	}
+
+	a := dnssort.SortUsingGraph(changes)
+
+	if len(a.UnresolvedRecords) > 0 {
+		log.Printf("Warning: Found unresolved records %v.\n"+
+			"This can indicate a circular dependency, please ensure all targets from given records exist and no circular dependencies exist in the changeset."+
+			"These unresolved records are still added as changes and pushed to the provider, but will cause issues if and when the provider checks the changes\n",
+			dnsgraph.GetRecordsNamesForGraphables(a.UnresolvedRecords),
```
tlimoncelli, line 22: (sorry for the delayed response... I was traveling for a week. I'm on sabbatical until July 30, so dnscontrol is low priority right now)

Thanks for adding this explanation!

Now that I've had more time to think about it, I have some new thoughts! :)

The important thing here is that it is a warning, not an error.  It won't stop someone from getting their work done, though they may get an error from their provider.

I think this message should link to a page with more details.  Maybe `documentation/ordering.md`?  The page should give an example dnsconfig.js change that would product this error, and how to fix it.  This doc could also mention the `--disableordering` flag and request that people file a bug.

With that, I think we'll be ready to merge this!

Tom

blackshadev: (No problem, and no need for apologies. We all are normal humans with lives. I am getting married in about 3 months, so my priorities are also a bit less on the open-source side than they normally would be. )

I did as you asked, added docs and linked in the warning to said docs. I also fixed the tests, appearently analyze_test and compareconfig_test rely on the same source records

documentation/ordering.md
```diff
@@ -0,0 +1,34 @@
...
+D('example.com', REG_NONE, DnsProvider(DNS_BIND),
+    CNAME('foo', 'bar')
+    CNAME('bar', 'foo'),
+);
+```
+
+
+## Disabling ordering
+
+If you encounter any issues with the reordering don't hesitate to [open an issue](https://github.com/StackExchange/dnscontrol/issues). On top of that you can opt-out of the reordering with the `--disableordering` global flag.
```
tlimoncelli, line 34:
```suggestion
The re-ordering feature can be disabled using the `--disableordering` global flag (it goes before `preview` or `push`). While the code has been extensively tested, it is new and you may still find a bug.  This flag leaves the updates unordered and may require multiple `push` runs to complete the update.

If you encounter any issues with the reordering please [open an issue](https://github.com/StackExchange/dnscontrol/issues). 
```

documentation/ordering.md
```diff
@@ -0,0 +1,34 @@
...
+);
+```
+
+`foo` requires `bar` to exist. Thus `bar` needs to exist before `foo`. But when deleting these records, `foo` needs to be deleted before `bar`.
+
+## Unresolved records
+
+DNSControl can produce a warning stating it found `unresolved records` this is most likely because of a cycle in the targets of your records. For instance in the code sample below both `foo` and `bar` depend on each other and thus will produce the warning.
+
+But have no fear, DNSControl will still try to push both `foo` and `bar` to your provider, it just can't determine the order and thus will append all these unresolved records as last records to update.
```
tlimoncelli, line 22:
```suggestion
Such updates will be done after all other updates to that domain.

In this (contrived) example, it is impossible to know which CNAME should be created first. Therefore they will be done in a non-deterministic order after all other updates to that domain:
```

documentation/ordering.md
```diff
@@ -0,0 +1,34 @@
+# Ordering of DNS records
+
+DNSControl tries to automatically reorder the pending changes based on the dependencies of the records.
```
tlimoncelli, line 3:
```suggestion
DNSControl tries to automatically reorder the pending changes based on the dependencies of the records.
For example, if an A record and a CNAME that points to the A record are created at the same time, some providers require the A record to be created before the CNAME.
```

tlimoncelli (approved):

tlimoncelli (approved):

tlimoncelli:
We're SO CLOSE!

Sadly I found some bugs.   None are show-stoppers but they should be fixed before we merge this to master.

1) Any provider with custom DNS types fails.

You can see them here:
https://github.com/StackExchange/dnscontrol/actions/runs/5931050001/job/16083128144

I think the solution is to have GetDependencies() handle them as a no-op instead of a log.Fatalf().  If providers want to handle them better, they can add the code to special-case them. (This is the one place that I wish Go had a full OO paradigm.)   That said, we should at least handle the simple ones:

* ALIAS, R53_ALIAS, AZURE_ALIAS: basically the same as CNAME
* These are no-ops: CF_REDIRECT, CF_TEMP_REDIRECT, CF_WORKER_ROUTE, IMPORT_TRANSFORM, NAMESERVER, NO_PURGE, NS1_URLFWD, PAGE_RULE, PURGE, URL, URL301, WORKER_ROUTE, AKAMAICDN, IMPORT_TRANSFORM, LOC


2) An MX record pointing to itself shouldn't be a circular dependency.

This is a very typical situation:

D(...,
    MX("foo", 10, "foo"),
   ..
)

Test 19 is a good repo:

```
$ go test -v -verbose -provider BIND -diff2 -start 19 -end 19
=== RUN   TestDNSProviders
=== RUN   TestDNSProviders/example.com
=== RUN   TestDNSProviders/example.com/Clean_Slate:Empty
    integration_test.go:256: 
        - DELETE example.com NS ns1.example.com. ttl=300
        - DELETE example.com NS ns2.example.com. ttl=300
WRITING ZONEFILE: zones/example.com.zone
=== RUN   TestDNSProviders/example.com/19:testByRecordSet:initial
2023/08/21 17:47:15 Warning: Found unresolved records [foo.example.com].
This can indicate a circular dependency, please ensure all targets from given records exist and no circular dependencies exist in the changeset.These unresolved records are still added as changes and pushed to the provider, but will cause issues if and when the provider checks the changes
For more information and how to disable the reordering please consolidate our documentation at https://docs.dnscontrol.org/developer-info/ordering
    integration_test.go:256: 
        + CREATE bar.example.com A 1.2.3.4 ttl=300
        + CREATE foo.example.com A 3.4.5.6 ttl=300
        + CREATE foo.example.com MX 10 foo.example.com. ttl=300
        + CREATE foo.example.com MX 20 bar.example.com. ttl=300
        + CREATE foo.example.com A 2.3.4.5 ttl=300
WRITING ZONEFILE: zones/example.com.zone
=== RUN   TestDNSProviders/example.com/19:testByRecordSet:changeOne
    integration_test.go:256: 
        ± MODIFY foo.example.com A (3.4.5.6 ttl=300) -> (8.8.8.8 ttl=300)
WRITING ZONEFILE: zones/example.com.zone
=== RUN   TestDNSProviders/example.com/19:testByRecordSet:deleteOne
    integration_test.go:256: 
        - DELETE foo.example.com A 8.8.8.8 ttl=300
WRITING ZONEFILE: zones/example.com.zone
=== RUN   TestDNSProviders/example.com/19:testByRecordSet:addOne
    integration_test.go:256: 
        + CREATE foo.example.com A 8.8.8.8 ttl=300
WRITING ZONEFILE: zones/example.com.zone
=== RUN   TestDNSProviders/example.com/Post_cleanup:Empty
2023/08/21 17:47:15 Warning: Found unresolved records [bar.example.com foo.example.com].
This can indicate a circular dependency, please ensure all targets from given records exist and no circular dependencies exist in the changeset.These unresolved records are still added as changes and pushed to the provider, but will cause issues if and when the provider checks the changes
For more information and how to disable the reordering please consolidate our documentation at https://docs.dnscontrol.org/developer-info/ordering
    integration_test.go:256: 
        - DELETE foo.example.com A 8.8.8.8 ttl=300
        - DELETE foo.example.com MX 10 foo.example.com. ttl=300
        - DELETE foo.example.com MX 20 bar.example.com. ttl=300
        - DELETE bar.example.com A 1.2.3.4 ttl=300
        - DELETE foo.example.com A 2.3.4.5 ttl=300
WRITING ZONEFILE: zones/example.com.zone
--- PASS: TestDNSProviders (0.01s)
    --- PASS: TestDNSProviders/example.com (0.01s)
        --- PASS: TestDNSProviders/example.com/Clean_Slate:Empty (0.00s)
        --- PASS: TestDNSProviders/example.com/19:testByRecordSet:initial (0.00s)
        --- PASS: TestDNSProviders/example.com/19:testByRecordSet:changeOne (0.00s)
        --- PASS: TestDNSProviders/example.com/19:testByRecordSet:deleteOne (0.00s)
        --- PASS: TestDNSProviders/example.com/19:testByRecordSet:addOne (0.00s)
        --- PASS: TestDNSProviders/example.com/Post_cleanup:Empty (0.00s)
=== RUN   TestDualProviders
    integration_test.go:381: Clearing everything
    integration_test.go:388: Adding nameservers from another provider
    integration_test.go:375: #1:
        + CREATE example.com NS ns1.example.com. ttl=300
        + CREATE example.com NS ns2.example.com. ttl=300
WRITING ZONEFILE: zones/example.com.zone
    integration_test.go:391: Running again to ensure stability
--- PASS: TestDualProviders (0.00s)
=== RUN   TestNameserverDots
=== RUN   TestNameserverDots/No_trailing_dot_in_nameserver
--- PASS: TestNameserverDots (0.00s)
    --- PASS: TestNameserverDots/No_trailing_dot_in_nameserver (0.00s)
PASS
ok  	github.com/StackExchange/dnscontrol/v4/integrationTest	0.408s
```


3) CNAME pointing to itself

Test 61 has a CNAME pointing to itself.  That's not required for the test.  We should probably just have the CNAME point at a different hostname.  

```
$ go test -v -verbose -provider BIND -diff2 -start 61 -end 61
=== RUN   TestDNSProviders
=== RUN   TestDNSProviders/example.com
=== RUN   TestDNSProviders/example.com/Clean_Slate:Empty
    integration_test.go:256: 
        - DELETE example.com NS ns1.example.com. ttl=300
        - DELETE example.com NS ns2.example.com. ttl=300
WRITING ZONEFILE: zones/example.com.zone
=== RUN   TestDNSProviders/example.com/61:IGNORE_TARGET_function_CNAME:Create_some_records
2023/08/21 17:51:42 Warning: Found unresolved records [keep.example.com].
This can indicate a circular dependency, please ensure all targets from given records exist and no circular dependencies exist in the changeset.These unresolved records are still added as changes and pushed to the provider, but will cause issues if and when the provider checks the changes
For more information and how to disable the reordering please consolidate our documentation at https://docs.dnscontrol.org/developer-info/ordering
    integration_test.go:256: 
        + CREATE foo.example.com CNAME test.foo.com. ttl=300
        + CREATE keep.example.com CNAME keep.example.com. ttl=300
```

Here's the fix:

```
$ git diff
diff --git a/integrationTest/integration_test.go b/integrationTest/integration_test.go
index c6cb7a11..e032aa5e 100644
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -1964,15 +1964,15 @@ func makeTests(t *testing.T) []*TestGroup {
                testgroup("IGNORE_TARGET function CNAME",
                        tc("Create some records",
                                cname("foo", "test.foo.com."),
-                               cname("keep", "keep.example.com."),
+                               cname("keep", "keeper.example.com."),
                        ),
                        tc("ignoring CNAME=test.foo.com.",
                                ignoreTarget("test.foo.com.", "CNAME"),
-                               cname("keep", "keep.example.com."),
+                               cname("keep", "keeper.example.com."),
                        ).ExpectNoChanges(),
                        tc("ignoring CNAME=test.foo.com. and add",
                                ignoreTarget("test.foo.com.", "CNAME"),
-                               cname("keep", "keep.example.com."),
+                               cname("keep", "keeper.example.com."),
                                a("adding", "1.2.3.4"),
                                cname("another", "www.example.com."),
                        ),
```

blackshadev:
Welcome back Tom.

1 and 3 are fairly easily and I already fixed them locally (although not tested).

For 2 I am still looking for a solution. The problem is that I do not use the record type in the dependency graph and in the dependency list. So I am just looking for the name "foo.example.com." instead of a A or CNAME for "foo.example.com." . 

I am thinking to only add nodes to the graph we can resolve, if at all possible.

Do you maybe know if: the referred targets (so the target of mx, ns, srv and aliases) are always of the same "type". So they always point to an CNAME, A record or another alias? If so I can shortcut the resolution stage by just adding only CNAME and A records as resolvable records. Otherwise I need to add the record type and referential types to the graph which will be quite the rework.

tlimoncelli:
Sadly targets can point to basically any label. 

What if GetDependencies() returned an empty list if `rc.NameFQDN == rc.target` ?

blackshadev:
> Sadly targets can point to basically any label.
> 
> What if GetDependencies() returned an empty list if `rc.NameFQDN == rc.target` ?

Shame, I don't think your solution would work. In case of a "self-referencing" MX record, the target equals the NameFQDN and thus it wouldn't have any dependencies, which will not ensure the A record it tries to refer is higher.
I also thought to exclude the "same" node from the graph to have a relation on itself. But I didn't yet got that working, although it might work if done correctly.

blackshadev:
Small update, Solving the MX "self-reference" brought a bigger issue to light which needs to be resolved. 
The algorithm which creates edges for the graph "resolves" records only by their DNS name, meaning if we add have two `a.example.com` records, they will be treated the same, even though they are wildly different.

I a way to iterate all added nodes and add their dependencies instead of re-resolving them by name. I see two options forward which I will investigate:

- Not looking up the source nodes since we added the nodes at the first step, it shouldn't need to do this
- Adding unique keys to the graph nodes to tracking them easier. This is be allot harder to do since I need to generate a unique key for each provided entry. On top of that these keys needs to be used to resolve specific entries. 

I will try the first option first, since it is easier, and doesn't need another map structure and field to the already existing loads. The second one will make debugging allot easier though, keeping track of all the nodes in the system and which resolve to what is quite a pain to do.

blackshadev:
Ok I think I solved it. Could you re-check it? I created a separate commit per stated issue to keep it a bit easier to review.

tlimoncelli:
All tests pass!  https://github.com/StackExchange/dnscontrol/actions/runs/5986595523

Congrats!

I'll merge this in the next few days.  (I"m out sick right now)

PR #2418, opened 2023-06-02
CICD: Also push container images to GitHub Container Registry (ghcr.io)

Firefishy:
This adds the support to also push container images to GitHub Container Registry (ghcr.io)

The goreleaser code change have been briefly tested here ([minor changes to disable other builds](https://github.com/Firefishy/dnscontrol/commit/bc0c066aa18d9e1a3b4d643c5e22b22dd6f752ac)): https://github.com/Firefishy/dnscontrol/actions/runs/5153372408/jobs/9280525081 (👍 docker build/push worked, push GH release failed) and the GitHub container images are published here: https://github.com/Firefishy/dnscontrol/pkgs/container/dnscontrol

PR #2424, opened 2023-06-06
CICD: Add required GHA permissions for goreleaser

Firefishy:
Added the missing permissions required by goreleaser for releases.

Link: https://github.com/StackExchange/dnscontrol/pull/2418


Firefishy:
FYI [build here failed](https://github.com/StackExchange/dnscontrol/actions/runs/5187678662) because it doesn't contain the change in this PR.

It might be possible to force re-tag and force push the updated tag again to re-trigger the build.

Issue #2422, opened 2023-06-03
DNSControl behaves differently in Alpine-based Docker container


DvdGiessen:
When run in an Alpine-based Docker container, DNSControl is unable to list all my domain names. This means it doesn't work with [the official Docker image](https://hub.docker.com/r/stackexchange/dnscontrol/), while it works perfectly when run on my Mac, my Arch Linux system at home, or inside a Debian 11 container.

Using binary from `dnscontrol_4.1.0_linux_amd64.tar.gz` from latest release:
```
$ docker run -ti --rm -v $(pwd):/dns:ro -w /dns debian:latest
root@d2c4d8b7e3ec:/dns# cat /etc/issue
Debian GNU/Linux 11 \n \l

root@d2c4d8b7e3ec:/dns# # Note: In the Debian image some CA certificates are missing I need for my provider
root@d2c4d8b7e3ec:/dns# (apt update && apt install -y ca-certificates) >/dev/null 2>&1
root@d2c4d8b7e3ec:/dns# ./dnscontrol version
dnscontrol "4.1.0" ("e97cf017445a1a622ed4a7b2766b1aa83637d185") built 25 May 23 15:48 UTC
root@d2c4d8b7e3ec:/dns# ./dnscontrol check-creds transip | wc -l
56
root@3eba97af219b:/dns# ./dnscontrol check-creds transip | head -n1 | sha256sum 
a9a52a3164777288e7f3c0a13540c47c634743cd6e9aa7c4de5bd441c5d9414c  -
```
```
$ docker run -ti --rm -v $(pwd):/dns:ro -w /dns alpine:latest
/dns # cat /etc/issue
Welcome to Alpine Linux 3.18
Kernel \r on an \m (\l)

/dns # ./dnscontrol version
dnscontrol "4.1.0" ("e97cf017445a1a622ed4a7b2766b1aa83637d185") built 25 May 23 15:48 UTC
/dns # ./dnscontrol check-creds transip | wc -l
1
/dns # ./dnscontrol check-creds transip | sha256sum
a9a52a3164777288e7f3c0a13540c47c634743cd6e9aa7c4de5bd441c5d9414c  -
```

Left out the actual domain names, but as you can see it appears DNSControl is only seeing the very first domain when being run in Alpine.

[...]

PR #2429, opened 2023-06-08
Change Docker to Debian base image

DvdGiessen:
This pull request changes the base image used for Docker from Alpine to Debian to address the issue described in https://github.com/StackExchange/dnscontrol/issues/2422.

The change consists of two lines (and comments):
- The changed base image, now referring to `debian:11` which is the current stable release of Debian.
  - To check the SHA256 I'm proposing in this commit, you can download the latest Debian image yourself using `docker pull debian:11`, and then inspect its digest with `docker inspect --format='{{index .RepoDigests 0}}' debian:11`, and see if it matches with the one mentioned in my commit, assuming they didn't just release a new version which you can check [here](https://hub.docker.com/_/debian/tags?name=latest).

- We explicitly install the [`ca-certificates` package](https://packages.debian.org/bullseye/ca-certificates), which is needed to make providers which use HTTPS work out-of-the-box without throwing CA errors.
  - In the previous Alpine version this was commented out because unlike Debian Alpine actually does already ship with common CA certificates included.
  - The install command and the command before/after it are the way they are to follow the [best practises](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) from the Docker documentation.

Note that this will be a breaking change for some users of the Docker image, if they depended on the image specifically being based on Alpine. The examples outlined in the DNSControl documentation do not specifically depend on Alpine, so those will continue to work, but if people modified them to make use of Alpine-specific parts of the Docker image it will break. This should probably be added to the release notes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 # syntax = docker/dockerfile:1.4
 
-FROM alpine:3.18.0@sha256:02bb6f428431fbc2809c5d1b41eab5a68350194fb508869a33cb1af4444c9b11 as RUN
+# We're using Debian because that is also the base image for the `golang` images
+FROM debian:11@sha256:432f545c6ba13b79e2681f4cc4858788b0ab099fc1cca799cc0fae4687c69070 AS RUN
 
-#RUN --mount=type=cache,target=/var/cache/apk \
-#    apk update \
-#    && apk add ca-certificates \
-#    && update-ca-certificates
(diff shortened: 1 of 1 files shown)


DvdGiessen:
Closing because preference to keep using the small Alpine image and resolve issues with it, see https://github.com/StackExchange/dnscontrol/issues/2422#issuecomment-1583172825.

PR #2434, opened 2023-06-09
CHORE: Add runtime dependencies for container

Firefishy:
Add the required runtime dependencies for running dnscontrol in a container.

* tzdata - Go time required external dependency. See: #2422
* ca-certificates - Required for HTTPS to work out-of-the-box without throwing CA errors

Fixes: #2422
Link: #2429

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,13 @@
 
 FROM alpine:3.18.0@sha256:02bb6f428431fbc2809c5d1b41eab5a68350194fb508869a33cb1af4444c9b11 as RUN
 
-#RUN --mount=type=cache,target=/var/cache/apk \
-#    apk update \
-#    && apk add ca-certificates \
-#    && update-ca-certificates
+# Add runtime dependencies
+# - tzdata: Go time required external dependency eg: TRANSIP and possibly others
+# - ca-certificates: Needed for https to work properly
+RUN --mount=type=cache,target=/var/cache/apk \
+    apk update \
+    && apk add tzdata ca-certificates \
+    && update-ca-certificates
 
 COPY dnscontrol /usr/local/bin/
 


DvdGiessen:
> ca-certificates - Required for HTTPS to work out-of-the-box without throwing CA errors

Note the `ca-certificates` package doesn't do anything since Alpine already ships with a default CA bundle, and HTTPS works without issues out of the box on Alpine.

This can also be verified by inspecting the CA bundle before and after running the commands:
```
$ docker run -ti --rm alpine:3.18.0@sha256:02bb6f428431fbc2809c5d1b41eab5a68350194fb508869a33cb1af4444c9b11
/ # sha256sum /etc/ssl/certs/ca-certificates.crt
824cefcee69de918c76b7b92776f304c3a4b7f6281539118bc1d41a9dd8476d9  /etc/ssl/certs/ca-certificates.crt
/ # apk add --no-cache ca-certificates && update-ca-certificates
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/community/x86_64/APKINDEX.tar.gz
(1/1) Installing ca-certificates (20230506-r0)
Executing busybox-1.36.0-r9.trigger
Executing ca-certificates-20230506-r0.trigger
OK: 8 MiB in 16 packages
/ # sha256sum /etc/ssl/certs/ca-certificates.crt
824cefcee69de918c76b7b92776f304c3a4b7f6281539118bc1d41a9dd8476d9  /etc/ssl/certs/ca-certificates.crt
```

DvdGiessen:
Tested it and I can confirm this does indeed fix the issue of missing timezone data from https://github.com/StackExchange/dnscontrol/issues/2422!

Firefishy:
> > ca-certificates - Required for HTTPS to work out-of-the-box without throwing CA errors
> 
> Note the `ca-certificates` package doesn't do anything to change that since Alpine already ships with a default CA bundle, and thus HTTPS already works without issues out of the box on Alpine.
> 

Yes, the default alpine CA bundle is a static compiled bundle produced from `ca-certificates`. Adding the package makes it easier to understand and simplifies adding/removing certificates from the CA bundle (eg: for  corporate certificates or "Next Generation Firewall" support etc). My preference is to include the full `ca-certificates` package, the size difference is minimal at around 780KB.

TomOnTime:
Sounds good!  Thanks for all the input!

PR #2436, opened 2023-06-13
Add missing var statements at loop initialization

killerbees19:
Just don't ask why I've found this little "bug" 😄

_[My own loop in `dnsconfig.js` (also without proper variable initialization) broke `require_glob()` 😬]_


TomOnTime:
That must have been one difficult bug to fix!

Issue #2387, opened 2023-05-21
DOCS: Code examples in language reference


cafferata:
The code examples within the language reference are:
1. Missing provider constants https://github.com/StackExchange/dnscontrol/pull/2384
2. Missing provider complete
3. Tied to a DNS provider

**Missing provider constants**

```javascript
D("example.com", REGISTRAR, DnsProvider("R53"),
  A("@", "1.2.3.4"),
  A("foo", "2.3.4.5"),
  A("test.foo", IP("1.2.3.4"), TTL(5000)),
  A("*", "1.2.3.4", {foo: 42})
);
```

_<https://docs.dnscontrol.org/language-reference/domain-modifiers/a>_

**Missing provider complete**

```javascript
D("example.com", REGISTRAR, ....,
  TXT('@', '598611146-3338560'),
  TXT('listserve', 'google-site-verification=12345'),
  TXT('multiple', ['one', 'two', 'three']),  // Multiple strings
  TXT('quoted', 'any "quotes" and escapes? ugh; no worries!'),
  TXT('_domainkey', 't=y; o=-;'), // Escapes are done for you automatically.
  TXT('long', 'X'.repeat(300)) // Long strings are split automatically.
);
```
_<https://docs.dnscontrol.org/language-reference/domain-modifiers/txt>_

**Tied to a DNS provider** example below Google Cloud DNS:

```javascript
D("example.com", REGISTRAR, DnsProvider("GCLOUD"),
  // Allow letsencrypt to issue certificate for this domain
  CAA("@", "issue", "letsencrypt.org"),
  // Allow no CA to issue wildcard certificate for this domain
  CAA("@", "issuewild", ";"),
  // Report all violation to test@example.com. If CA does not support
  // this record then refuse to issue any certificate
  CAA("@", "iodef", "mailto:test@example.com", CAA_CRITICAL)
);
```

_<https://docs.dnscontrol.org/language-reference/domain-modifiers/caa>_

Is it an idea to draw a line here? If so, what would you consider a good outcome?

tlimoncelli:
> Is it an idea to draw a line here? If so, what would you consider a good outcome?

Do you mean how should we make the example not tied to a specific provider?  I would just use REG_MYPROVIDER and DSP_MYPROVIDER

cafferata:
> I would just use `REG_MYPROVIDER` and `DSP_MYPROVIDER`

That's fine for me.

> Do you mean how should we make the example ...

[...]

PR #2439, opened 2023-06-15
DOCS: Improve consistency in code examples in language reference

cafferata:
Pull request nr. 2 for https://github.com/StackExchange/dnscontrol/issues/2387.

_Please merge after approval so that no unnecessary conflicts arise._ 🙏


documentation/adding-new-rtypes.md
```diff
@@ -34,14 +34,12 @@ Here are some examples:
 
 ```go
 type RecordConfig struct {
-  ...
```
TomOnTime, line 37: We need a way to indicate that some fields in the struct have been removed for brevity

I agree with the removal of `...` in `D(X, ...` examples.  This is different.

How about:

```
    [...]
    field
    [...]
```

cafferata: > We need a way to indicate that some fields in the struct have been removed for brevity

That's a great suggestion, but I'd like to keep it out of the pull request.

> I agree with the removal of ... in D(X, ... examples.

I've applied a rebase commit that keeps the rest of the `...`.

TomOnTime:
Thanks for cleaning this up!

Issue #2435, opened 2023-06-12
BIND: Long TXT records broken?


killerbees19:
Maybe I'm doing something wrong, but it looks like it's stuck at MODIFY forever? 🤔

```
$ dnscontrol version
dnscontrol "4.1.1" ("9c9ff4d45a970cb851d9fda013dc9da631e54ce5") built 06 Jun 23 17:11 CEST

$ mkdir -p /tmp/test/zones
$ cd /tmp/test
$ echo '{}' > creds.json
$ cat << EOF > dnsconfig.js
  D('example.com', NewRegistrar('none'), DnsProvider(NewDnsProvider('bind')), {no_ns:'true'});
  D_EXTEND('example.com', TXT('test1._domainkey', 'v=DKIM1; k=rsa; p=012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789'));
EOF

$ dnscontrol preview
[INFO: Diff2 algorithm in use. Welcome to the future!]
******************** Domain: example.com
WARNING: Zone 'example.com' does not exist in the 'bind' profile and will be added automatically.
Done. 0 corrections.

$ dnscontrol push
[INFO: Diff2 algorithm in use. Welcome to the future!]
******************** Domain: example.com
File does not yet exist: "zones/example.com.zone" (will create)
1 correction (bind)
#1:
WRITING ZONEFILE: zones/example.com.zone
SUCCESS!
Done. 1 corrections.

$ cat zones/example.com.zone
$TTL 300
; generated with dnscontrol 2023-06-13T00:38:52+02:00
@                IN SOA   DEFAULT_NOT_SET. DEFAULT_NOT_SET. 2023061200 3600 600 604800 1440
test1._domainkey IN TXT   "v=DKIM1; k=rsa; p=012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789"

$ dnscontrol preview
[INFO: Diff2 algorithm in use. Welcome to the future!]
******************** Domain: example.com
1 correction (bind)
[...]

PR #2447, opened 2023-06-19
BUG: BIND not handling long TXT records correctly

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/2435


tlimoncelli:
This time for sure!

PR #2449, opened 2023-06-20
NEW FEATURE: Generate Shell Completion data

whi-tw:
## Why

I would like to enable tab completion for `dnscontrol`. There's currently no good way to do this.

## What

`dnscontrol shell-completion <shell>` will generate a shell completion script for the specified shell (currently `bash` or `zsh`).

If no shell is specified, the script will be generated for the current shell, using `$SHELL`.

The completion scripts I have added to this repo are slightly modified from those in [urfave/cli@v2.25.5](https://github.com/urfave/cli/tree/60247a75d701dd881749eb22fc862f26d604ce72/autocomplete). The modifications just remove the reliance on `$PROG` being set, and instead use the value of `app.Name`.

Additional shells can be 'easily' added by adding a new script:

`commands/completion-scripts/completion.${shell_binary_name}.gotmpl`, where `shell_binary_name` is like `bash`, `zsh`, `fish` etc.

It *should* just be picked up and used automatically, with the tests running against the new script.

### Practical tests

In each implemented shell, do:

1. `export PATH="<directory containing dnscontrol binary for this PR>:${PATH}"`
2. `eval "$(dnscontrol shell-completion <name of shell>)"`
3. do some tab completion things
4. `unfunction _dnscontrol`
5. `eval "$(dnscontrol shell-completion)"`
6. do some tab completion things

### Unit tests

1. Ensure tests are actually testing the right things
2. Ensure tests pass


tlimoncelli:
Thank you!  This is much needed.

I'm not familiar with how other systems do this so excuse this naive question but... should we pre-generate the completion files and include them in the releases or do we expect each user to run `dnsutils shell-completion` on their own?

whi-tw:
> Thank you! This is much needed.
> 
> I'm not familiar with how other systems do this so excuse this naive question but... should we pre-generate the completion files and include them in the releases or do we expect each user to run `dnsutils shell-completion` on their own?

I'd recommend two things: Add some documentation for people to follow, similar to:

```markdown
You can generate shell completions for `bash` and `zsh`:
  - bash: `eval "$(dnscontrol completion bash)"`
  - zsh: `eval "$(dnscontrol completion zsh)"`
```

(I was thinking of doing this, myself, but documentation isn't my strong point!)

Update the homebrew formula similar to how we do it for [`gds-cli`](https://github.com/alphagov/homebrew-gds/blob/a65107c3b72115d109af839106f584fb8faeffbd/Formula/gds-cli.rb#L26-L30):

```ruby
output = Utils.safe_popen_read("#{bin}/dnscontrol", "shell-completion", "bash")
(bash_completion/"dnscontrol").write output
output = Utils.safe_popen_read("#{bin}/dnscontrol", "shell-completion", "zsh")
(zsh_completion/"_dnscontrol").write output
```

Any other packaging systems you're using (ie. Macports) will likely have a similar process for installing the completion scripts.

The benefit of not packaging the files is that they can be updated and people will likely automatically start using them without any messing around with local files.

tlimoncelli:
@whi-tw Agreed on all points.  Thank you for clarifying.

While you may claim that documentation isn't your strong point, the sample doc you wrote look just fine. Please add that to `documentation/getting-started.md` at the end of `## 1. Install the software` ?

Thanks!

tlimoncelli:
(NOTE: The doc should be `shell-completion` not `completion`)

tlimoncelli:
FYI:

I'm having trouble testing this on my mac.

```
host:~/git/dnscontrol$ go install
host:~/git/dnscontrol$ dnscontrol version
dnscontrol -dev
host:~/git/dnscontrol$ echo $BASH_VERSION
3.2.57(1)-release
host:~/git/dnscontrol$ eval $(dnscontrol shell-completion bash)
host:~/git/dnscontrol$ dnscontrol p 
package-lock.json  pkg/               providers/         
host:~/git/dnscontrol$ dnscontrol (TAB gets filenames, not subcommands)
host:~/git/dnscontrol$ ^C
-bash: : command not found
host:~/git/dnscontrol$ dnscontrol shell-completion bash
#! /bin/bash

: dnscontrol

# Macs have bash3 for which the bash-completion package doesn't include
# _init_completion. This is a minimal version of that function.
_dnscontrol_init_completion() {
  COMPREPLY=()
  _get_comp_words_by_ref "$@" cur prev words cword
}

_dnscontrol() {
  if [[ "${COMP_WORDS[0]}" != "source" ]]; then
    local cur opts base words
    COMPREPLY=()
    cur="${COMP_WORDS[COMP_CWORD]}"
    if declare -F _init_completion >/dev/null 2>&1; then
      _init_completion -n "=:" || return
    else
      _dnscontrol_init_completion -n "=:" || return
    fi
    words=("${words[@]:0:$cword}")
    if [[ "$cur" == "-"* ]]; then
      requestComp="${words[*]} ${cur} --generate-bash-completion"
    else
      requestComp="${words[*]} --generate-bash-completion"
    fi
    opts=$(eval "${requestComp}" 2>/dev/null)
    COMPREPLY=($(compgen -W "${opts}" -- ${cur}))
    return 0
  fi
}

complete -o bashdefault -o default -o nospace -F _dnscontrol dnscontrol
```

imlonghao:
It should be `eval "$(dnscontrol shell-completion bash)"`

whi-tw:
@tlimoncelli, @imlonghao is correct - I missed the quotes around the `$(..)`. With those, you should be golden

whi-tw:
@tlimoncelli I've added a bit to the docs, and have rebased against `StackExchange:master`. It'll likely need approval from y'all again.

commands/completion-scripts/completion.bash.gotmpl
```diff
@@ -0,0 +1,34 @@
+#! /bin/bash
```
tlimoncelli, line 1: No space after `#!`

commands/completion-scripts/completion.bash.gotmpl
```diff
@@ -0,0 +1,34 @@
...
+
+_{{.App.Name}}() {
+  if [[ "${COMP_WORDS[0]}" != "source" ]]; then
+    local cur opts base words
+    COMPREPLY=()
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    if declare -F _init_completion >/dev/null 2>&1; then
+      _init_completion -n "=:" || return
+    else
+      _{{.App.Name}}_init_completion -n "=:" || return
```
tlimoncelli, line 20: Is there any possibility that .App.Name would have a space in it?  If so, quote this string.

tlimoncelli:
There are a few conversations that need to be resolved.  Let me know if you have any questions.

tlimoncelli:
Friendly ping?

whi-tw:
Sorry, I did look at this, but then got distracted and then went on vacation!I will take a look ASAP!On Sep 14, 2023, at 18:18, Tom Limoncelli ***@***.***> wrote:﻿
Friendly ping?

commands/completion-scripts/completion.bash.gotmpl
```diff
@@ -0,0 +1,34 @@
+#! /bin/bash
+
+: {{.App.Name}}
```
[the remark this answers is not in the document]
whi-tw: `.App.Name` is the name of the app. For `dnscontrol`, this is hard-coded in [commands/commands.go:53](https://github.com/StackExchange/dnscontrol/blob/e32bdc053f64efc78a9f1fc3ac25cd8af6737425/commands/commands.go#L53), but if that line is removed, it is the name of the running binary.

It's useful not to hardcode the app name in the completion script in case the binary is named differently - that way, completions are generated for the specific name they run the tool with.

This usage of `:` essentially just functions as a comment (literally a noop) [[gnu.org](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Bourne-Shell-Builtins)]. A lot of bash completion scripts I've come across have that at the top.

commands/completion-scripts/completion.bash.gotmpl
whi-tw: As-per https://github.com/StackExchange/dnscontrol/pull/2449#discussion_r1350392872, in `dnscontrol`'s case as-is, it won't. I think it's probably worth quoting though in this case, in case the hard-coding of `app.Name` is removed. will do.

tlimoncelli:
Thanks for the update. Looks good!

Sadly I'm getting an error when I use this.  It might be unique to my situation. That said:

On my mac:

```
$ echo $BASH_VERSION
3.2.57(1)-release
$ . "/opt/homebrew/etc/profile.d/bash_completion.sh"
$ eval "$(dnscontrol shell-completion bash)"
-bash: `"_dnscontrol_init_completion"': not a valid identifier
-bash: `"_dnscontrol"': not a valid identifier
$
```

On a mac with Nix's latest bash:

```
$ echo $BASH_VERSION
5.2.15(1)-release
$ . "/opt/homebrew/etc/profile.d/bash_completion.sh"
bash: shopt: progcomp: invalid shell option name
$ eval "$(dnscontrol shell-completion bash)"
bash: `"_dnscontrol_init_completion"': not a valid identifier
bash: `"_dnscontrol"': not a valid identifier
bash: complete: command not found
$
```

Suggestions on what might be wrong?

Tom


P.S. Sadly while I use bash completion a lot, I've never written a completion file.

whi-tw:
Ah, that'll be the quotes then. I guess zsh is far happier with quotes around function names than bash. Although I'm sure I tested it with bash before!

I'll take another look when this migraine lets me think again though.

tlimoncelli:
Thanks!  I think we're getting closer!

```
$ echo $BASH_VERSION
3.2.57(1)-release
$ eval "$(dnscontrol shell-completion bash)"
$ dnscontrol pu<TAB>-bash: _get_comp_words_by_ref: command not found
```

`<TAB>` is where I pressed the TAB key.

whi-tw:
Now, we're diverging!

## Testing

### `bash`

#### 5.2.15 (from Homebrew)

```
$ echo $BASH_VERSION
5.2.15(1)-release
$ echo $SHELL
/bin/bash
$ eval "$(./dnscontrol shell-completion bash)"
$ dnscontrol <TAB> bash: _get_comp_words_by_ref: command not found
^C
$ source /opt/homebrew/etc/bash_completion
$ dnscontrol <TAB>
check           check-creds     create-domains  fmt             get-certs       get-zone        get-zones       h               help            preview         print-ir        push            version         write-types
```

#### 5.2.57 (Builtin on MacOS 13.6)

```
$ echo $BASH_VERSION
3.2.57(1)-release
$ echo $SHELL
/bin/bash
$ eval "$(./dnscontrol shell-completion bash)"
$ dnscontrol <TAB> bash: _get_comp_words_by_ref: command not found

$ source /opt/homebrew/etc/bash_completion
$ dnscontrol <TAB>
check           check-creds     create-domains  fmt             get-certs       get-zone        get-zones       h               help            preview         print-ir        push            version         write-types
```

### `zsh`

#### 5.9 (Builtin on MacOS 13.6)

```
❯ echo $ZSH_VERSION
5.9
❯ eval "$(./dnscontrol shell-completion zsh)"

❯ dnscontrol <TAB>
Completing values
check                     -- Check and validate dnsconfig.js. Output to stdout.  Do not access providers.
check-creds               -- Do a small operation to verify credentials (stand-alone)
create-domains            -- DEPRECATED: Ensures that all domains in your configuration are activated at their Domain Service Provider (This does not purchase the domain or otherwise interact with Registrars.)
fmt                       -- [BETA] Format and prettify a given file
get-certs                 -- Issue certificates via Let's Encrypt
get-zones       get-zone  -- gets a zone from a provider (stand-alone)
help            h         -- Shows a list of commands or help for one command
preview                   -- read live configuration and identify changes to be made, without applying them
print-ir                  -- Output intermediate representation (IR) after running validation and normalization logic.
push                      -- identify changes to be made, and perform them
version                   -- Print version information
write-types               -- [BETA] Write a TypeScript declaration file in the current directory
```

## Thoughts

It looks like the environments you are testing completion in are not correctly set up for completion:

### `bash`

`_get_comp_words_by_ref` is provided by [scop/bash-completion](https://github.com/scop/bash-completion/). It does look like other projects have come across the same problem though, eg. https://github.com/openshift/origin/issues/14657.

This script is a packaged version of the upstream script from the cli library (https://github.com/urfave/cli/blob/e8abd76fa3ee5f1bacb52eab9b71dedbeab41991/autocomplete/bash_autocomplete), so I'm loath to try to remove the dependency on bash-completion.

```
$ type _get_comp_words_by_ref | wc -l
bash: type: _get_comp_words_by_ref: not found
       0
$ source /opt/homebrew/etc/bash_completion
$ type _get_comp_words_by_ref | wc -l
      65
```

### `zsh`

`compdef` is a shell builtin from zsh:

```
❯ type compdef
compdef is a shell function from /usr/share/zsh/5.9/functions/compinit
```

It appears that the system you are testing on does not have the completion system enabled: [zsh: 20 Completion System](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Initialization). A guide to set it up is [here](https://scriptingosx.com/2019/07/moving-to-zsh-part-5-completions/) ("Turning It On" section), TL;DR: `autoload -Uz compinit && compinit` should be added to your .zshrc.

tlimoncelli:
bash:

Ah, I've fixed my env and now it works.  However I've found a small bug.

```
$ echo $BASH_VERSION
3.2.57(1)-release
$ eval "$(dnscontrol shell-completion bash)"
$ dnscontrol p<TAB>
preview   print-ir  push      
$ dnscontrol --no-colors p<TAB>
package-lock.json  pkg/               providers/         
```

As you can see, `dnscontrol p<TAB>` works but `dnscontrol --no-color p<TAB>` does file-name completion instead of offering the `p*` commands.

cafferata:
> It appears that the system you are testing on does not have the completion system enabled: [zsh: 20 Completion System](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Initialization). A guide to set it up is [here](https://scriptingosx.com/2019/07/moving-to-zsh-part-5-completions/) ("Turning It On" section), TL;DR: autoload -Uz compinit && compinit should be added to your .zshrc.

Is this something that can be detected? I've run into the same problem. 🙈

whi-tw:
@tlimoncelli
> I've found a small bug

I think you're run into one of the limitations of auto-generated completion scripts: [urfave/cli](/urfave/cli) does its best to 'guess' the context you're completing in, but isn't always right. you can [help it out](https://cli.urfave.org/v2/examples/bash-completions/#custom-auto-completion) though, so I'll see what can be done!

@cafferata
> Is this something that can be detected? I've run into the same problem. 🙈

Honestly, I'm not sure if it's something that *needs* to be detected by individual applications. This is a standard requirement for shell completion scripts, so it feels like it should be an expectation if someone is attempting to set up completion for dnscontrol.

However, perhaps there should be a callout to some upstream documentation for enabling completion from the dnscontrol documentation, to try to prevent people getting caught out?

tlimoncelli:
Something is better than nothing. Plus, we can always improve it in future releases.

Therefore... I'm willing to merge this with a warning in the documentation that TAB completion after the subcommand might not work if a global flag is included before the subcommand.

PR #2458, opened 2023-06-22
POWERDNS: Add option to set SOA-EDIT-API when creating zones via the API

dkim1970:
This PR adds a option to the PowerDNS provider to set soa-edit-api when creating a new zone.

This make it possible to choose SOA serial format when creating zones instead of having to set it with CURL or the pdnsutil afterwards and then run DNSControl  again with a new record to update the SOA serial.

This PR doesn't break the default behavior as it always gets created with the value DEFAULT.


tlimoncelli:
CC  @jpbede for approval

jpbede (approved):
LGTM, thanks @dkim1970!

Issue #2444, opened 2023-06-17
DNSControl treats a configured TTL of 0 as "not configured"


str4d:
While discussing the fact that Linode's API explicitly permits and returns a TTL of 0 to represent "Default", I asked in https://github.com/StackExchange/dnscontrol/issues/2440#issuecomment-1595543504:

> The question though is whether `dnscontrol` has a good way to represent "Default" in its own UX. Do any other providers have a similar concept? How do they handle a TTL of 0?

It turns out that DNSControl completely ignores any `DefaultTTL(0)` or `TTL(0)` settings, causing it to instead fall back on the global DNSControl default of 300 seconds. AFAICT the code in `pkg/js/helpers.js` handles things correctly, but when the JSON is parsed into Go, the resulting `RecordConfig` has its TTL replaced:
https://github.com/StackExchange/dnscontrol/blob/5b3bb312ea99853d1741d838d5a789cbc5eec1fb/models/record.go#L377-L379

str4d:
> AFAICT the code in pkg/js/helpers.js handles things correctly,

This is probably actually just by accident. I think the core issue here is that 0 in the TTL field _is_ being treated as "falsey" in `helpers.js` (while not present on the pathways I'm using, there are conditionals of the form `if (r.TTL)`), and the TTL field is fundamentally being treated as an optional non-zero unsigned integer, rather than an optional unsigned integer. The Go code above is just enforcing that at parse time.

Fixing this would require changing both the JavaScript and Go types to separately represent `None` and `Some(0)`. On the JavaScript side that could probably be slipped in via e.g. using `-1` to mean `None`, while on the Go side a proper `Optional` type would help.

str4d:
These are all the spots in the Go code that map `0` to `DefaultTTL`:
https://github.com/StackExchange/dnscontrol/blob/5b3bb312ea99853d1741d838d5a789cbc5eec1fb/models/record.go#L377-L379
https://github.com/StackExchange/dnscontrol/blob/5b3bb312ea99853d1741d838d5a789cbc5eec1fb/pkg/normalize/validate.go#L352-L354
https://github.com/StackExchange/dnscontrol/blob/5b3bb312ea99853d1741d838d5a789cbc5eec1fb/pkg/prettyzone/prettyzone.go#L88-L90
[...]

Issue #2440, opened 2023-06-16
LINODE: TTL does not support "Default" (0)


str4d:
**Describe the bug**
I set up a `dnsconfig.js` matching the records for a domain I manage through Linode. All of the records in Linode have their TTL column set to "Default". I do not configure an explicit TTL in `dnsconfig.js` for any record, leaving it at the `dnscontrol` default.

When I run `dnscontrol preview`, every record is marked as a modification such as:
```
#1: MODIFY A example.com: (1.2.3.4 ttl=0) -> (1.2.3.4 ttl=300), Linode ID: NNN: {"type":"A","name":"","target":"1.2.3.4","ttl_sec":300}
```

**To Reproduce**
1. Configure a domain name to use Linode for DNS records.
2. Add a DNS A record to Linode with some label and IP, and leave the TTL as "Default".
3. Write a `dnsconfig.js` with the same A record label and IP, and no configured TTL.
4. `dnscontrol preview`

**Expected behavior**
No modifications should be proposed.

**Versions**
 - DNSControl version: 4.1.1 (Homebrew)
 - DNS Provider: `LINODE`

**Additional context**
The `LINODE` provider documentation has a [caveat](https://docs.dnscontrol.org/service-providers/providers/linode#caveats) that the TTLs must be from a specific list. [The file that it links to](https://github.com/linode/manager/blob/master/src/domains/components/SelectDNSSeconds.js) does not exist in the Linode source tree anymore. When I searched for an equivalent list, I found this one that explicitly includes the option `{ label: 'Default', value: 0 }`.

https://github.com/linode/manager/blob/07be31a25399c8583a55a3727b8a38e0a6e25cd7/packages/manager/src/features/Domains/DomainRecordDrawer.tsx#L320-L336

str4d:
[Linode's own documentation](https://www.linode.com/docs/products/networking/dns-manager/guides/troubleshooting-dns/#set-the-time-to-live-or-ttl) confirms that the default TTL is still one day.

TomOnTime:
CC @koesie10 

Interesting bug!

That raises an interesting question.  Should allowedTTLValues (https://github.com/StackExchange/dnscontrol/blob/d00804b4f51c5b95137b50b15c747493de1475d5/providers/linode/linodeProvider.go#L29) be changed to include `0` or should the fixTTL function turn 0 into 86400?  (I think the first option)

str4d:
[...]

PR #2475, opened 2023-07-10
Add a TTL model

str4d:
This unifies the logic for handling the dnscontrol-default TTL, and frees up the TTL value 0 for use by providers (e.g. Linode, which uses it as its sentinel for its default TTL).

Closes StackExchange/dnscontrol#2444.


str4d:
With this change, I am able to fully resolve #2440 by adding an explicit `DefaultTTL(0)` to all of my `DSP_LINODE` sections. It would be slightly cleaner if providers could override the global `DefaultTTL` for themselves, but that was a more complex change, whereas the new `models.TTL` is simpler to reason about.

providers/exoscale/exoscaleProvider.go
```diff
@@ -269,8 +269,8 @@ func (c *exoscaleProvider) createRecordFunc(rc *models.RecordConfig, domainID st
 			Priority: prio,
 		}
 
-		if rc.TTL != 0 {
-			ttl := int64(rc.TTL)
+		if rc.TTL.Value() != 0 {
+			ttl := int64(rc.TTL.Value())
```
str4d, line 273: This is an example of where a provider was explicitly checking for 0, in a place where it could never be reached (as `dnscontrol` always replaced `0` with `DefaultTTL` before reaching this point). I presumed that this check was intentional and thus left it in place (as it is now triggerable).

providers/linode/linodeProvider.go
```diff
@@ -277,7 +277,7 @@ func toReq(dc *models.DomainConfig, rc *models.RecordConfig) (*recordEditRequest
 		Type:     rc.Type,
 		Name:     rc.GetLabel(),
 		Target:   rc.GetTargetField(),
-		TTL:      int(rc.TTL),
+		TTL:      int(rc.TTL.Value()),
```
str4d, line 280: This is what fully fixes #2440:
- When no TTL is set, this evaluates to `models.DefaultTTL` (300).
- When `DefaultTTL(0)` is used explicitly, this evaluates to 0.

TomOnTime:
Hi there!  I'm on vacation so please expect delays.

Thank you for taking the time to work on this issue.  The proposed change is pretty fundamental and will require extensive discussion and testing.


I have a number of over-arching concerns:

1. I have a concern about code complexity.  Would it be less complex to declare TTL=0 to be the default?  It is not a valid TTL, thus we should be able to use it. This would eliminate a lot of the complexity especially the use of pointers.
2. Do we want to support a default TTL at all?   Every vendor has a different default TTL.  Therefore depending on the default makes the zone non-portable to other providers.  It also makes the use of dual-hosted domains more difficult.  I could be convinced that zone portability is rare and insignificant.  anyway... maybe we should simply change any "default" to (for example) 1 day; the next "push" would remove the use of default from the zone. Then all our our domain data is explicit.

Code-specific suggestions:
1. Rather than add to dns.go, please move the TTL-related code to models/ttl.go
2. ValueRef(): Is there a better term for this?  
3. A lot of tests are breaking, especially for BIND.
    cd integrationTest && go test -v -verbose -provider BIND

str4d:
> Thank you for taking the time to work on this issue. The proposed change is pretty fundamental and will require extensive discussion and testing.

Yep, as expected! I will say that I'm not claiming this PR is the optimal approach (I am not a Go or JS developer, and have fumbled my way into something that compiles and works with a specific provider), but my hope is that it helps us to coalesce around a viable approach.

> I have a number of over-arching concerns:
> 
> 1. I have a concern about code complexity.  Would it be less complex to declare TTL=0 to be the default?  It is not a valid TTL, thus we should be able to use it. This would eliminate a lot of the complexity especially the use of pointers.

In this PR I didn't do that because I was trying to preserve `dnscontrol`'s global default (to not affect the existing user API) while also enabling the full unsigned integer range to be used for per-provider TTL values (as I have not surveyed what values providers actually use for different settings).

If the global `dnscontrol` default were either removed, or set up to be the fallback if an individual provider doesn't have its own default, and we confirmed that `TTL=0` isn't used as any other kind of sentinel by any provider, then yeah using `TTL=0` for this would simplify the implementation.

As an aside, my first local attempt at this PR did not use pointers, instead opting for an `Option`-based approach (I'm a Rust developer 🦀) via a third party dependency that `dnscontrol` already depended on elsewhere. However, one single provider within `dnscontrol` required that the TTL value could itself be referenced via a pointer, which was not possible for the non-pointer `Option` type. I did not spend the time to figure out if that was a hard requirement for the provider, and instead switched to using a pointer inside an explicit `TTL` model (the latter being a better approach anyway).

> 2. Do we want to support a default TTL at all?   Every vendor has a different default TTL.  Therefore depending on the default makes the zone non-portable to other providers.  It also makes the use of dual-hosted domains more difficult.  I could be convinced that zone portability is rare and insignificant.  anyway... maybe we should simply change any "default" to (for example) 1 day; the next "push" would remove the use of default from the zone. Then all our our domain data is explicit.

I personally don't particularly care whether you want to support a default TTL or not. My own issue is "solved" by this PR, in that I'm wanting to move my DNS settings away from Linode, and all I wanted was to confirm that my `dnsconfig.js` was correct (the existing diff view makes it very hard to verify this). So I don't actually need this PR to be merged.

However, there doesn't appear to be a clear decision one way or the other regarding default TTLs, and the implicit decision to map `0` to `300` instead of having separate "unconfigured" or "default" states causes problematic interactions with individual providers (e.g. #2440). Even if there isn't support for any "default" TTL in the model, I think _having_ a TTL model (and a way to tell when the user is using it) is useful.

> Code-specific suggestions:
> 
> 1. Rather than add to dns.go, please move the TTL-related code to models/ttl.go

Done.

> 2. ValueRef(): Is there a better term for this?

IDK, I can go with whatever is standard Go naming convention here. In Rust we use `*_ref` suffixes to denote APIs that return immutable references (as opposed to owned values).

> 3. A lot of tests are breaking, especially for BIND.
>        cd integrationTest && go test -v -verbose -provider BIND

Thanks, I couldn't figure out how to run the tests. It took me a bit to figure out that I actually had to set an environment variable to get this to run:

```
cd integrationTest && BIND_DOMAIN=example.com go test -v -verbose -provider BIND
```

And the test failure was because `prettyzone` was trying to print the TTL model directly instead of `TTL.Value()`, and so was writing the address of the pointer into the zone file 😅 Fixed.

[Force-pushed](https://github.com/StackExchange/dnscontrol/compare/df6842b06072618d3813cc9560522de1729f77f4..879b16a4149d07ded9121ebfb1235b9bab647679) with changes 1 and 3.

tlimoncelli:
I appreciate all the effort you've put into this.  While this solution will work, I think there are other solutions that are more simple that will also work.  I always favor the more simple solution.

I don't think we can solve this globally until first we decide how we want TTLs to work.  Most of the TTL-handling code was written in 2016-ish when DNSControl had fewer formal design specifications (translation: we were guessing our way through most design decisions and a lot of specifics were just implemented without a formal design).  As a result, each provider handles things slightly differently.  Eventually we should do an RFC and get feedback from the community about how TTLs should work, get consensus, then implement a solution based on that feedback.

In the meanwhile, I think we should fix this just for Linode rather than doing a global change that affects nearly all providers.

Therefore, I'm asking you to start a different PR that fixes this for Linode only.

Thanks for understanding.

str4d:
> I appreciate all the effort you've put into this. While this solution will work, I think there are other solutions that are more simple that will also work. I always favor the more simple solution.

@tlimoncelli If you know of a simpler solution, I am all ears!

The core problem is the aliasing everywhere of `TTL = 0`; this needs to be addressed somehow. The JS model uses it to mean both "nothing configured" and "user explicitly configured `0`:
https://github.com/StackExchange/dnscontrol/blob/af91e37043d49ce1d5f8ecc54c0196d972772bae/pkg/js/helpers.js#L100-L109
https://github.com/StackExchange/dnscontrol/blob/af91e37043d49ce1d5f8ecc54c0196d972772bae/pkg/js/helpers.js#L251-L257
https://github.com/StackExchange/dnscontrol/blob/af91e37043d49ce1d5f8ecc54c0196d972772bae/pkg/js/helpers.js#L1048-L1051
https://github.com/StackExchange/dnscontrol/blob/af91e37043d49ce1d5f8ecc54c0196d972772bae/pkg/js/helpers.js#L222-L228

The Go model `dns.RecordConfig` is parsed directly from the JS model:

https://github.com/StackExchange/dnscontrol/blob/af91e37043d49ce1d5f8ecc54c0196d972772bae/models/record.go#L86-L92

`dns.RecordConfig` is used directly by every single provider, so any solution that alters the type of `TTL` (as this PR does) is by definition a global solution. Your request therefore prevents me from doing this.

Additionally, individual providers never see `RecordConfig.TTL = 0`, because of `normalize.ValidateAndNormalizeConfig`:
https://github.com/StackExchange/dnscontrol/blob/af91e37043d49ce1d5f8ecc54c0196d972772bae/pkg/normalize/validate.go#L348-L354

Delaying or changing this normalization would require changes to every single provider to handle the fact that their assumptions about default TTL handling have been broken. So I can't alter that either if I am to satisfy your request.

As a result of the above two implementation details, there is currently no way for the Linode provider to distinguish these three cases:
- No setting, falling back on `dnscontrol`'s global default TTL of 300 (changing this to instead fall back on the provider's default TTL is something I presume would be part of the RFC you describe).
- The user explicitly configuring `DefaultTTL(0)` (which should always be interpreted as a valid Linode TTL setting).
- The user explicitly configuring `DefaultTTL(300)`.

I'll be honest, I do not see a simpler solution than "change the type of `TTL`". I see (what I believe to be) a _more complex_ solution that allows for an incremental migration, however:
- Add a new TTL field to `dns.RecordConfig` with a different name (say, `TypedTTL`).
- Connect `TypedTTL` to the JS model in one of two ways:
  - Add similar duplicate fields to the JS model, and handle user updates to them in parallel with the existing `defaultTTL` and `ttl` fields.
  - Change the JS model to be nullable (like this PR) and then change the Go parser to parse the `ttl` field into `TypedTTL`, and then copy the parsed value into the existing `TTL` field (aliasing `None` and `Some(0)` to `0`).
- Change every single location in the codebase that modifies `dns.RecordConfig.TTL` to also modify `TypedTTL`, _except for_ `normalize.ValidateAndNormalizeConfig` (which would leverage `TypedTTL`'s `None` case as meaning "use the `dnscontrol` global default").
  - Hopefully no providers are modifying `TTL` so I don't have to modify their code.
- Change the Linode provider to use `TypedTTL` instead of `TTL`.
  - All other providers are left using `TTL` as-is.

If this approach is along the lines of what you were thinking, then I'll go ahead and implement it. And if there is some aspect of `dnscontrol` that I'm completely overlooking (as a non-Go developer or as a new contributor) that would make this way simpler, I'd love to know.

tlimoncelli:
Thank you for this proposal. Sadly I've decided not to move forward with it.

Issue #2476, opened 2023-07-13
preview/push error messages missing new lines


asn-iac:
**Describe the bug**
printer error messages are missing new lines which affects output during certain errors in push/preview.  I'm not sure if this is intentional, as all warn messages have new lines on them already

**To Reproduce**
1. create the following dnsconfig (with valid gcloud credentials)
```
var DNS_GOOGLE = NewDnsProvider('gcloud');
D('atestdomain0.com', REG_NONE, DnsProvider(DNS_GOOGLE),
	NAMESERVER_TTL('21600')
);
D('atestdomain1.com', REG_NONE, DnsProvider(DNS_GOOGLE),
	NAMESERVER_TTL('21600')
);
D('atestdomain2.com', REG_NONE, DnsProvider(DNS_GOOGLE),
	NAMESERVER_TTL('21600')
);
```
2. run 'dnscontrol push'
3. printer output is missing newlines as below:
```
$ dnscontrol push
[INFO: Diff2 algorithm in use. Welcome to the future!]
******************** Domain: atestdomain0.com
Adding zone for atestdomain0.com to gcloud account 
ERROR: ERROR: domain "atestdomain0.com" not found in your GCLOUD account******************** Domain: atestdomain1.com
Adding zone for atestdomain1.com to gcloud account 
ERROR: ERROR: domain "atestdomain1.com" not found in your GCLOUD account******************** Domain: atestdomain2.com
Adding zone for atestdomain2.com to gcloud account 
ERROR: ERROR: domain "atestdomain2.com" not found in your GCLOUD accountDone. 0 corrections.

$ dnscontrol version
dnscontrol "4.1.1" ("9c9ff4d45a970cb851d9fda013dc9da631e54ce5") built 06 Jun 23 08:11 PDT
```

**Expected behavior**
The error messages from push/preview should be terminated with newlines (unless this is intended)

The output for 'dnscontrol push' should appear as follows:
```
$ dnscontrol push
[INFO: Diff2 algorithm in use. Welcome to the future!]
******************** Domain: atestdomain0.com
Adding zone for atestdomain0.com to gcloud account 
ERROR: ERROR: domain "atestdomain0.com" not found in your GCLOUD account
******************** Domain: atestdomain1.com
Adding zone for atestdomain1.com to gcloud account 
ERROR: ERROR: domain "atestdomain1.com" not found in your GCLOUD account
******************** Domain: atestdomain2.com
[...]

PR #2477, opened 2023-07-13
Adjust Error message formatting for preview/push

asn-iac:
Add new line to printer.CLI.Errorf() messages

This should fix output issues as described in #2476


tlimoncelli:
Good catch!  Thanks for fixing that!

PR #2506, opened 2023-08-11
Implementing API retries when receiving 429 (TooManyRequests) HTTP Error

kabenin:
NS1 client has few sleep strategies when receiving error 429 already implemented.
We use RateLimitStrategySleep() as the one allowing for the least possible sleep
amount.
However, client library does not implement retry logic, even when a Strategy is set.
User of the library still gets error. That means, for each API call we need to
implement a retry logic if httpResp has status code of 429.


costasd:
sorry for the delay here; will take a look later today!

PR #2507, opened 2023-08-11
Provide a cmd switch for disabling colors

### LINKED ISSUE #2499: Add no color flag option for output
When running dnscontrol I would like to force it to not output any color

an additional command line flag i.e. `--no-color` 

Currently for a CI process we have an additional job step to strip out ANSI color characters

**Additional context**

Before
image

After
image

costasd:
... called `--no-colors`.

Fixes #2499

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -12,6 +12,8 @@ import (
 	"github.com/StackExchange/dnscontrol/v4/pkg/js"
 	"github.com/StackExchange/dnscontrol/v4/pkg/printer"
 	"github.com/urfave/cli/v2"
+
+	"github.com/fatih/color"
 )
 
 // categories of commands
@@ -67,6 +69,12 @@ func Run(v string) int {
 			Usage:       "Enable replacement diff algorithm",
 			Destination: &diff2.EnableDiff2,
 			Value:       true,
+		},
+		&cli.BoolFlag{
+			Name:        "no-colors",
+			Usage:       "Disable colors",
+			Destination: &color.NoColor,
+			Value:       false,
 		},
 	}
 	sort.Sort(cli.CommandsByName(commands))


cafferata:
Would you mind including this `--no-colors` option in the documentation?

costasd:
> Thanks @costasd for the first version. Maybe also good to discribe the `env-variable mechanism` as you mentioned earlier? [#2499 (comment)](https://github.com/StackExchange/dnscontrol/issues/2499#issuecomment-1674756602)

oh this proved to be a noop and removed in this changeset. I'll add a comment about it in the issue.

documentation/colors.md
```diff
@@ -0,0 +1,27 @@
...
+how to bypass it.
+
+## Disable colors
+
+There are cases that we might need to disable colored output, such as in certain
+CI environments or particular notification mechanisms that don't support ANSI color
+codes.
+
+In order to do so, a global `--no-colors` command option is provided, which when
+set, will disable colors globally.
```
cafferata, line 16: Set to what value? Perhaps we can help a newcomer with a small example, by describing the `--no-colors=true` case here? The two `--no-colors=false` cases are described below (`(Force) Enable colors` & `Force colors in CI environments`).

```suggestion
set `--no-colors=true`, will disable colors globally.
```

tlimoncelli:
This is definitely needed.  I don't think any notification systems accept the ESC codes for colors.

However stripping them out seems less good than, say, generating them correctly in the first place.  Sadly the code isn't really structured for that.  We could make `--notify` imply `--no-color` but that would be... worse.

Any ideas for how we can do this differently?

costasd:
So, if we talked a kinda larger design - what we could do is have dnscontrol generate generic enough color hints of some kind if run without `--no-colors`,  say `<red> - DELETE A lala.test.com</red>`.

...And then have each different functionality either strip these colors if not needed (say, for an sms notification) or convert them into the native representation: ESC codes for console printing, slack markdown colors for slack etc.

Maybe (strong maybe, i haven't paid a lot of attention into that part of the code) diff2 could take care of generating the generic color hints and then dnscontrol console printer and each notification can take care of translating into the native platform's color support (or stripping colors if not supported).

But until we get there, I'm thinking we should as an intermediate step 1) supply a way to switch colors off/on via the cli in #2507 and 2) disable colors completely in notifications to avoid sending ESC codes to platforms that don't support them (what I have in #2508)

Happy to take a stab on the larger design at some point if it makes sense to y'all.

PR #2520, opened 2023-08-17
CHORE: Fix CF and NS1 due to api changes

tlimoncelli:


costasd:
ah just opened a PR for the NS1 part of this, will close. LGTM in any case.

PR #2516, opened 2023-08-17
NS1: update ns1-go to 2.7.8

costasd:

PR #2521, opened 2023-08-17
ns1: fix a leftover change from ns1-go update

costasd:
PR #2506 getting merged while #2516 was still in flight introduced another place we needed to fix to make our code work with NS1's updated go API, which wasn't covered.

This changeset covers this as well.


costasd:
covered in #2520

PR #2524, opened 2023-08-18
ci(gorelease): move to github action

musdav:
## Issue

Installing gorelease in separate step as part of CI

## Resolution

Utilize existing gorelease github action for build and draft release


tlimoncelli:
How to push a release is documented here: documentation/release-engineering.md

ProTip: I usually test the release process on version numbers that won't actually get released.  For example, the current release is 4.1.1 and the next release will be 4.2.0.  Therefore, I've been doing test releases v4.1.12, v4.1.13, .. v4.1.16.  The next test should be on v4.1.17.   When the tests are done, I usually delete the test releases from https://github.com/StackExchange/dnscontrol/tags (both tags and releases)

PR #2531, opened 2023-08-23
Display changes, even if upstream has no records.

tomfitzhenry:
Motivation: It's confusing to see no changes on first creation. This led me to falsely believe that no changes were being made.


tlimoncelli:
Yes, I see your point!

I think my plan was to add a diff2.REPORT saying "New zone uploaded from scratch" but, yes, listing the records is easier and less violates the principle of least surprise.

PR #2543, opened 2023-09-03
NEW FEATURE: Added "push --report" flag which generates JSON report of changes made

fritterhoff:
We want to execute some actions after running dnscontrol. To reduce the amount of actions and improve the speed this PR adds a report option that generates a JSON file containing the Domain, the amount of performed actions and the Provider/Registrar. 

Afterwards we can parse the json and identify the domains that were changed and ignore those that kept unchanged.

### DIFF
--- a/commands/previewPush.go
+++ b/commands/previewPush.go
@@ -47,10 +47,10 @@ type PreviewArgs struct {
 }
 
 type ReportItem struct {
-	Domain      string
-	Corrections int
-	Provider    string
-	Registrar   string
+	Domain      string `json:"domain"`
+	Corrections int    `json:"corrections"`
+	Provider    string `json:"provider"`
+	Registrar   string `json:"registrar"`
 }
 
 func (args *PreviewArgs) flags() []cli.Flag {
@@ -296,7 +296,7 @@ func run(args PreviewArgs, push bool, interactive bool, out printer.CLI, report 
 		return fmt.Errorf("there are pending changes")
 	}
 	if report != nil {
-		f, err := os.OpenFile(*report, os.O_CREATE|os.O_WRONLY, 0644)
+		f, err := os.OpenFile(*report, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


cafferata:
Interesting! How do you use this in the PR context? Maybe you can describe this new option including the possibilities/examples repots in the documentation? I am curious because I see possible improvements for our current working method.

fritterhoff:
In detail we extended dnscontrol a little bit further (https://github.com/hm-edu/dnscontrol-extended). 

We generate reverse zones using dnscontrol and parse them afterwards into GitLab Issues. This allows us to identify free ip addresses and get a quick overview over the utilization of our IP ranges. In the next step we plan/are discussing to push the data into netbox. In a long term view the DNS data shal be the single source of truth for several management operations.

The GitLab issues are currently always created - regardless of whether a change is necessary or not. This is at the expense of the API limitations or, in the case of the large number of IP ranges, at the expense of performance.

fritterhoff:
As an example this is an excerpt of the GitLab Issue for one /24 Zone. 

image

In total we have more than 30 public /24 zones that get managed by dnscontrol and approx. 15 /16 internal zones. Overall we currently manage approx. 1500 public A Records and thousands of private A Records ;-)

cafferata:
Thanks again for sharing! 👏 I like this kind of overview/deeper understanding of the needs. This inspires me, and perhaps others, as well.

PS. Did you also see @tlimoncelli 'endorsements' request on [Google Group](https://groups.google.com/g/dnscontrol-discuss/c/a6CZgJKhvbY)? Sounds like the Munich University of Applied Sciences is a great example. 😄

fritterhoff:
@cafferata @tlimoncelli  feel free to add the Munich University of Applied Sciences to your list of happy users 😉

tlimoncelli:
Great idea!

Could you add documentation (https://docs.dnscontrol.org/developer-info/styleguide-doc#how-to-add-a-new-page) for this feature?  (Please include a sample of what the json output looks like)

Tom

tlimoncelli:
Thanks!  I look forward to using this feature!

Issue #2547, opened 2023-09-05
CLOUDFLAREAPI: NAPTR Record Support


asn-iac:
Add support for NAPTR records to cloudflare provider

asn-iac:
merged in #2548

PR #2548, opened 2023-09-05
CLOUDFLAREAPI: Add NAPTR record support

asn-iac:
Add support for NAPTR records to cloudflare provider.
(#2547)


asn-iac:
I added a new type to handle encoding/decoding of the data object for the NAPTR record to avoid type conflict with the existing "Flags" field in the generic "cfRecData" type.  Please let me know if it's preferred to handle in a different way.  TIA!

tlimoncelli:
> I added a new type to handle encoding/decoding of the data object for the NAPTR record to avoid type conflict with the existing "Flags" field in the generic "cfRecData" type. Please let me know if it's preferred to handle in a different way. TIA!

Looks good to me!

tlimoncelli:
This looks good.  I'll merge it soon.

I'm curious: What do you use NAPTR for?   I'm curious now popular this feature is. (I added the original NAPTR support but I haven't used it in a year or more)

asn-iac:
> This looks good. I'll merge it soon.
> 
> I'm curious: What do you use NAPTR for? I'm curious now popular this feature is. (I added the original NAPTR support but I haven't used it in a year or more)

@tlimoncelli I'd actually not encountered this record type previously and had to go look at the rfc... it came up in a domain portfolio we have in cloudflare that I'm finally pulling into our dnscontrol ci/cd pipeline.  The team using this record develops a telephony-based app, so it wasn't too surprising in the end.  Most of the other domains we've managed over the years are more web app focused

Thanks again for all the hard work!

PR #2567, opened 2023-09-23
Improve output when --full=false

tlimoncelli:


documentation/functions/domain/IGNORE.md
```diff
@@ -64,10 +69,14 @@ following patterns will work:
 * `IGNORE("{bar,[fz]oo}")` will ignore `bar`, `foo` and `zoo`.
 * `IGNORE("\\*.foo")` will ignore the literal record `*.foo`.
 
+NOTE: `.` should not be escaped with a `\`.  These are globs (like filenames), not regular expressions.
```
cafferata, line 72: To make the [GitBook render](https://docs.dnscontrol.org/~/revisions/9uGevTaTgWzzE4wqHcLP/language-reference/domain-modifiers/ignore#globs) look better you can use [hint](https://docs.dnscontrol.org/developer-info/styleguide-doc#hint):

```suggestion
{% hint style="info" %}
**NOTE**: `.` should not be escaped with a `\`.  These are globs (like filenames), not regular expressions.
{% endhint %}
```

tlimoncelli:
never mind.

PR #2577, opened 2023-10-01
inwx: fix fetchNameserverDomains

### LINKED ISSUE #2541: INWX: Does not work with external domains
Currently the `ListZones` enpoint of the INWX provider calls the `fetchNameserverDomains` function, which in turn retrieves the owned domains from the INWX api. I want to argue that this behaviour is wrong and does not match the documentation of `fetchNameserverDomains` (https://github.com/StackExchange/dnscontrol/blob/master/providers/inwx/inwxProvider.go#L402).

**Describe the bug**
The INWX retrieves the owned domains instead of the owned zones to see if zones need to be created.

**To Reproduce**
1. Use the `INWX` provider for any domain you do not own in INWX, you may or may not own the domain through a different domain registrar. The exact content to deploy in the zone does not matter.
2. Executing `dnscontrol preview` will result in the following for that zone:
```
WARNING: Zone 'die-cloud.org' does not exist in the 'inwx' profile and will be added automatically.
```
3. Executing `dnscontrol push` will work, if the zone does not exist yet in INWX
4. Add some change to this zone in the `dnsconfig.js`, like adding a `TXT` record.
5. Executing `dnscontrol push` will fail with the following error, if the zone already exists in INWX and and update would normally be attemted:
```
WARNING: Error creating domain: (2302) Object exists. Reason: (Error_DomainAlreadyInDNS) The domain is already in the DNS.
```

**Expected behavior**
It should be correctly honored, if the zone is already present, so an update of zones for externally owned domains can work.

**DNS Provider**
 - INWX

**Additional context**
The documentation of `fetchNameserverDomains` already reads that it should return the domains configured in the INWX nameserver section (which is the zones configured in their DNS). The api endpoint `nameserver.list` can be used for that. Currently the `domain.list` endpoint is used, which retrieves the owned domains in INWX. This should be changed in my opinon.

It happened to me when switching to INWX as a registrar, and also as an dns provider. Some domains are still paid at a different registrar, so I do not want to migrate them right now, but I could already migrate the dns, if not for this bug.

P.s.: Currently i am in vacation and I do not really want to spend my time fixing this myself right now. Debugging already took more time, that I would have liked.

costasd:
fetchNameserverDomains was erroneously pulling configured domains from the registrar side of INWX, instead of the DNS serving side.

api.client.Domains     -> Registrar
api.client.Nameservers -> DNS Hosting

Fixes #2541

### DIFF
--- a/providers/inwx/inwxProvider.go
+++ b/providers/inwx/inwxProvider.go
@@ -401,9 +401,7 @@ func (api *inwxAPI) GetRegistrarCorrections(dc *models.DomainConfig) ([]*models.
 
 // fetchNameserverDomains returns the domains configured in INWX nameservers
 func (api *inwxAPI) fetchNameserverDomains() error {
-	request := &goinwx.DomainListRequest{}
-	request.PageLimit = 2147483647 // int32 max value, highest number API accepts
-	info, err := api.client.Domains.List(request)
+	info, err := api.client.Nameservers.List("")
 	if err != nil {
 		return err
 	}


patschi:
I think this is breaking the fix for using more than 20 domains. As per INWX's API, there is still a pageLimit of 20:
https://www.inwx.com/en/help/apidoc/f/ch02s15.html#nameserver.list

It looks like there is no `NameserverListRequest` struct in goinwx library as it is with `DomainListRequest`, so this might need to be added ideally to the library first, so we can take use of the same.

Also, I'm not sure if there're any other edge cases where this change might cause anything else to break. Quickly scrolling through it, I don't think so however.

costasd:
Hmm, you are right @patschi! I did check goinwx, but didn't check the upstream API for this one. Will fix goinwx library first and come back to this PR.

tlimoncelli:
@costasd I'm setting this as "draft" so I'm not tempted to merge prematurely.

costasd:
OK, this should be better now. Pulls also the latest goinwx library that is required for enabling the change.

PR #2583, opened 2023-10-08
Fix the return types of `*_BUILDER` functions

### LINKED ISSUE #2582: `dnscontrol write-types` generates incorrect return type for `*_BUILDER`s (except `SPF_BUILDER`)
**Describe the bug**
`*_BUILDER` functions, like `DMARC_BUILDER`, get used like `SPF_BUILDER` and other records, like this:
```javascript
D("example.com", REG_MY_PROVIDER, DnsProvider(DSP_MY_PROVIDER),
    A("@", "1.2.3.4"),
    SPF_BUILDER({
        parts: [
            "v=spf1",
            "ip4:1.2.3.4",
            "-all",
        ]
    }),
    DMARC_BUILDER({  // this is reported as an error in type-checking
        'policy': 'reject',
        'alignmentSPF': 'strict',
    })
)
```
And that works as expected in `dnscontrol <check|preview|push>`, but if you're using the TypeScript type-checking in your IDE/editor generated by `dnscontrol write-types`, it'll report it as an error because in  `types-dnscontrol.d.ts` it says `DMARC_BUILDER` returns type `RecordModifier`, which isn't allowed as an argument in `D` (it expects `DomainModifier`, not `RecordModifier`). But it actually does return type `DomainModifier` like other records do (including `SPF_BUILDER`), so it works and isn't actually an error, but it's wrongly reported as one.

**To Reproduce**
1. Run `dnscontrol write-types` to generate `types-dnscontrol.d.ts`
2. Open your IDE/editor with type-checking enabled and see that using `DMARC_BUILDER` as an argument in `D` returns an error when it shouldn't (like in the above code).

**Expected behavior**
No reported errors when `DMARC_BUILDER` is properly used as an argument in `D`, and a reported error if you try to use it as a record modifier in `A` or something ("reported" because it does actually return an error if you try to run it, but it doesn't report that in type-checking). Like this breaks if you try to run it, but type-checking allows it even though it shouldn't:
```javascript
D("example.com", REG_MY_PROVIDER, DnsProvider(DSP_MY_PROVIDER),
    A('@', '1.2.3.4', DMARC_BUILDER({
        'policy': 'reject',
        'alignmentSPF': 'strict',
    }))
)
```

**Additional context**
This is basically just #2484, but for more `*_BUILDER`s instead of just `SPF_BUILDER`, and only the return type part of that issue.

Originally I noticed this with just `DMARC_BUILDER` because that's the only one I used, but when reading through the code and documentation to figure out how to fix it myself I found there's more `*_BUILDER`s, and this seems to apply to all of them.
* `CAA_BUILDER` I got to work and confirmed is the same as `DMARC_BUILDER` above
* `LOC_BUILDER_*`s I couldn't get to work (my DNS provider doesn't support LOC records), but [the documentation examples](https://docs.dnscontrol.org/language-reference/record-modifiers/loc_builder_dd) show it inside `D` so I'm guessing it's the same.
* `M365_BUILDER` I couldn't figure out (and don't really want to, I don't use it) and its [documentation doesn't show it inside anything](https://docs.dnscontrol.org/language-reference/record-modifiers/m365_builder), but I checked [the code](https://github.com/StackExchange/dnscontrol/blob/60f9e4f87f122fa121c4947164b879ffc853a905/pkg/js/helpers.js#L1703) and it seems like it's [the same](https://github.com/StackExchange/dnscontrol/blob/60f9e4f87f122fa121c4947164b879ffc853a905/pkg/js/helpers.js#L1422).

`SPF_BUILDER` is the only one that's been fixed for some reason, guessing the others were just missed... The fix seems to be moving the documentation pages from `documentation/functions/record/*_BUILDER*.md` -> `documentation/functions/domain/*_BUILDER*.md` (`record` -> `domain`), and then regenerating `dnscontrol.d.ts`. `SPF_BUILDER` has [already been moved](https://github.com/StackExchange/dnscontrol/pull/2487/commits/0d0e1f93f870bb730f4d41e9c77782fbbf19128f) which is why that one's fixed but the others aren't.

Purrrpley:
They actually return `DomainModifier`, not `RecordModifier`. This allows you to use them as arguments in `D` without type-checking reporting an error (because it's not an error, it's correct).

Fixes #2582.

Possibly related note: I noticed the [SPF_BUILDER documentation page](https://docs.dnscontrol.org/language-reference/record-modifiers/spf_builder) is broken. It doesn't display any content, and appears in the wrong section ("Record Modifiers" instead of "Domain Modifiers"), so it seems like the documentation is broken and hasn't updated the moving of the file properly, which is what I've done here but for more things, so they might break too...

I'm not sure how to fix that though, but it's probably a separate issue anyway, but still related so I'm noting it here before it breaks more things. I don't actually have much experience with much of this (I've never used Go or TypeScript before, and rarely use JavaScript), so I'm not 100% sure what I'm doing... but I think I did this properly, it's fixed in my testing.

Thank you. :)


tlimoncelli:
Thanks for noticing and fixing this!

PR #2605, opened 2023-10-24
Change "push -i" prompt text from "Y/n" to "y/N" to match reality

tomfitzhenry:
Per https://askubuntu.com/questions/322600/do-you-want-to-continuey-n-why-the-upper-case , the convention is that the uppercase character is the default.

PromptToRun() defaults to false, so let's prompt with "y/N".


tlimoncelli:
Wow!  How did I not notice that?   Thanks for catching this!

PR #2618, opened 2023-11-01
[DNSimple] Update DNSimple's Default name servers

onlyhavecans:
The primary set of ns2 name servers name is changing.
Per https://support.dnsimple.com/articles/dnsimple-nameservers/


tlimoncelli:
Thanks!

PR #2622, opened 2023-11-14
PORKBUN: implement ListZones()

stefanwascoding:
Small improvement for the Porkbun provider. Enables `all` argument for the `get-zones` subcommand, as well as `check-creds` subcommand.

Side note: My first PR for the project, did not find a "how to contribute" entry on the otherwise really good documentation (I guess I could have asked on the mailing list first 🙂), so I really appreciate the PR template mentioning a few steps (especially as someone who only has occasional contact with `go`).

### DIFF
--- a/providers/porkbun/api.go
+++ b/providers/porkbun/api.go
@@ -41,6 +41,23 @@ type recordResponse struct {
 	Records []domainRecord `json:"records"`
 }
 
+type domainListRecord struct {
+	Domain       string `json:"domain"`
+	Status       string `json:"status"`
+	TLD          string `json:"tld"`
+	CreateDate   string `json:"createDate"`
+	ExpireDate   string `json:"expireDate"`
+	SecurityLock string `json:"securityLock"`
+	WhoisPrivacy string `json:"whoisPrivacy"`
+	AutoRenew    string `json:"autoRenew"`
+	NotLocal     string `json:"notLocal"`
+}
+
+type domainListResponse struct {
+	Status  string             `json:"status"`
+	Domains []domainListRecord `json:"domains"`
+}
+
 type nsResponse struct {
 	Status      string   `json:"status"`
 	Nameservers []string `json:"ns"`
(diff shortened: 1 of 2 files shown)


stefanwascoding:
Do I need to change anything in the [DocumentationNotes,](https://github.com/stefanwascoding/dnscontrol/blob/399170dc4a5bd90e22a6afe30e1710329b80487b/providers/porkbun/porkbunProvider.go#L54) now that this is implemented?

tlimoncelli:
CC @imlonghao (maintainer of the porkbun provider)

tlimoncelli:
> Do I need to change anything in the [DocumentationNotes,](https://github.com/stefanwascoding/dnscontrol/blob/399170dc4a5bd90e22a6afe30e1710329b80487b/providers/porkbun/porkbunProvider.go#L54) now that this is implemented?

Nope!  The code detects if the function exists automatically.

providers/porkbun/api.go
```diff
@@ -150,3 +167,21 @@ func (c *porkbunProvider) updateNameservers(ns []string, domain string) error {
 	}
 	return nil
 }
+
+func (c *porkbunProvider) listAllDomains() ([]string, error) {
+	params := requestParams{}
+	var bodyString, err = c.post("/domain/listAll", params)
```
tlimoncelli, line 173: This looks correct. However, experience tells me I should always ask this question:

Does Porkbun have "zones" and "domains"?  

A domain is something you can register.  Like... example.com or foo.uk.co.  A zone is a DNS zone (list of records).  For example, I registered the domain yesthattom.com but I have a zone called corp.yesthattom.com which is a subdomain (or a sub-zone).

I ask because ListZones() should return zones, not just domains.  Some APIs call them both "domains", others have different API calls for zones but refer to `/domains` only when dealing with registrations.



Thanks!

stefanwascoding: The [Porkbun API documentation](https://porkbun.com/api/json/v3/documentation)  has 0 matches when searching for "zone".

stefanwascoding: The [Porkbun API documentation](https://porkbun.com/api/json/v3/documentation)  has 0 matches when searching for "zone".

imlonghao: Porkbun has both "zones" and "domains".

The worst thing is Porkbun combine them in this one `/domain/listAll` endpoint, and it is tricky to tell which one is zone, and which one is domain.

tlimoncelli: Ok!  Sounds like you've already taken care of this.  Thanks for verifying!

PR #2638, opened 2023-11-21
DOCS: Add commands for pre-generated shell completions

x70b1:
I have added a few words to the documentation to pre-generate the files for shell completion as well. 

@whi-tw wrote something to this topic: https://github.com/StackExchange/dnscontrol/pull/2449#issuecomment-1600825604

> The benefit of not packaging the files is that they can be updated and people will likely automatically start using them without any messing around with local files.

I cant entirely agree with that. A manual change must be done with the existing instructions. This may be possible. But it is not convenient. I cant manually set a completion command for every tool on my system. 

I want to install a tool. And as soon as the tool and the zsh/bash completion are installed, it should work.

A way for pre-generating and packaging is therefore important.

### DIFF
--- a/documentation/getting-started.md
+++ b/documentation/getting-started.md
@@ -60,16 +60,19 @@ Shell completion is available for `zsh` and `bash`.
 
 ### zsh
 
+Enable the required completion in zsh. A good tutorial for this is available at [The Valuable Dev](https://thevaluable.dev/zsh-completion-guide-examples/) <sup>[[archived](https://web.archive.org/web/20231015083946/https://thevaluable.dev/zsh-completion-guide-examples/)]</sup>.
+
 Add `eval "$(dnscontrol shell-completion zsh)"` to your `~/.zshrc` file.
 
-This requires completion to be enabled in zsh. A good tutorial for this is available at
-[The Valuable Dev](https://thevaluable.dev/zsh-completion-guide-examples/) <sup>[[archived](https://web.archive.org/web/20231015083946/https://thevaluable.dev/zsh-completion-guide-examples/)]</sup>.
+Or run `dnscontrol shell-completion zsh > /usr/share/zsh/site-functions/_dnscontrol"`.
 
 ### bash
 
+Install the required `bash-completion` package. See [scop/bash-completion](https://github.com/scop/bash-completion/) for instructions.
+
 Add `eval "$(dnscontrol shell-completion bash)"` to your `~/.bashrc` file.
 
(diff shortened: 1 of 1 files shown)


whi-tw:
I think I may have not quite expressed myself as I intended there: I'm not advocating 'completely not packaging the files'. 'Packaging', there, was referring to not pointing people to download completion scripts from GitHub. Definitely used the wrong word.

Packaging is something I had in mind when working on #2449. The idea would be to update the various packages and write out the completion files as files to be installed with the binary. [A homebrew example](https://github.com/alphagov/homebrew-gds/blob/3b2faf5cdf32150df53fd2b6c78534a233e64644/Formula/gds-cli.rb#L27-L30).

I'm not sure where the packaging scripts for the various package managers come from, but PRs should definitely go in there.

I agree completely that people need an easy way to use the completions and keep them up to date. The most common way of doing this is to have the tool generate the completions itself, and instruct the user to add a line to their shell's `*rc` file. For example: [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#enable-shell-autocompletion), [1password cli (`op`)](https://developer.1password.com/docs/cli/reference/commands/completion/)

I'm not sure instructing a user to write out a completion is the right approach for 'normal' users: this file won't be kept up to date automatically by the tool itself, meaning that syntax / usage could drift, leading to unpredictable behaviour.

Another possible issue here is that the user may not have permission to write to /usr/share, at which point `sudo` and `tee` would likely have to come in to the documentation, which seems super out-of-scope for this project's documentation. `~/.${SHELL}rc` will always(?) be writable, so this works for all users. Those who want to do caching etc will likely have their own preexisting way.

x70b1:
I think our basic assumption is wrong. We assume that the user has to set up his completion for `dnscontrol` himself. But that should already be solved by the package that he is using and all he have to do is to make sure that zsh/bash completion is setup.

Only if that not is the case, he can use one of the other options manually.

I will rewrite this a bit.

whi-tw:
I wonder if the improvement here would be to add a line at the start of the shell completion documentation section along the lines of:

> Currently, shell-completion is not installed as part of the packaging of dnscontrol. We intend to implement this in the future. For now, users installing dnscontrol from their package manager must use the following one-time instructions to enable tab completion on their systems.

whi-tw:
I've submitted a PR to get completions packaged in Homebrew: Homebrew/homebrew-core#155049

x70b1:
WTF

Issue #2639, opened 2023-11-21
NSONE: Value None for field '<obj>.tags' is not of type object


norman-zon:
**Describe the bug**

Since updating from 4.4.1 to 4.6.0 I cannot push changes to NS1 anymore.

For any record type I get:

```
FAILURE! PUT https://api.nsone.net/v1/zones/<zone>/<record>/<type>: 400 Input validation failed (Value None for field '<obj>.tags' is not of type object)
```

I see there already is a [workaround for a similiar problem](https://github.com/StackExchange/dnscontrol/blob/bfa9da98c8642073dffc7c7013bcd151b2d061ae/providers/ns1/ns1Provider.go#L299) in the provider. 

**To Reproduce**

1. Push any change to NS1

**Expected behavior**
The record gets created/updated without issues

**DNS Provider**
 - NS1

costasd:
hm. will look very soon, thanks.

costasd:
ah excellent, I think I did all of these in the past, outside of sharing the secrets with you :boom:. Will go over it again just in case something changed in the config since then and then share the secrets.

tlimoncelli:
If you would like to set up a test account with NS1, I'll add the creds to the CI/CD pipeline so that NS1 gets tested with every PR.  You can use https://transfer.secretoverflow.com/u/tlimoncelli to securely send me credentials.

costasd:
ns1/ns1-go#224

PR #2642, opened 2023-11-21
NS1: Downgrade ns1-go to 2.7.11

costasd:
Latest changes in 2.7.13 seem to be breaking record creation completely, API returns 500.

Whereas 2.7.12 suffers from a bug that breaks creation with 400 due to mishandling for tags/blockedTags.

Let's downgrade to 2.7.11 that is working and passing the integration suite, for now.

Relates to #2639


tlimoncelli:
Thanks!

PR #2644, opened 2023-11-21
NS1: update ns1 for new txt handling

costasd:
* update auditrecords for succeeding tests
  * multiple segments not needed anymore. This was there to cover an edge case of a user setting segments in the dnsconfig file
  * strings can be longer than 255 octets, but not 0.
* update ns1 txt handling. No need for segmenting anymore, ns1 api handles this transparently.

### DIFF
--- a/providers/ns1/ns1Provider.go
+++ b/providers/ns1/ns1Provider.go
@@ -302,7 +302,7 @@ func buildRecord(recs models.Records, domain string, id string) *dns.Record {
 		if r.Type == "MX" {
 			rec.AddAnswer(&dns.Answer{Rdata: strings.Fields(fmt.Sprintf("%d %v", r.MxPreference, r.GetTargetField()))})
 		} else if r.Type == "TXT" {
-			rec.AddAnswer(&dns.Answer{Rdata: r.GetTargetTXTSegmented()})
+			rec.AddAnswer(&dns.Answer{Rdata: []string{r.GetTargetField()}})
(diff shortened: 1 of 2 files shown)


providers/ns1/ns1Provider.go
```diff
@@ -302,7 +302,7 @@ func buildRecord(recs models.Records, domain string, id string) *dns.Record {
 		if r.Type == "MX" {
 			rec.AddAnswer(&dns.Answer{Rdata: strings.Fields(fmt.Sprintf("%d %v", r.MxPreference, r.GetTargetField()))})
 		} else if r.Type == "TXT" {
-			rec.AddAnswer(&dns.Answer{Rdata: r.GetTargetTXTSegmented()})
+			rec.AddAnswer(&dns.Answer{Rdata: []string{r.GetTargetField()}})
```
tlimoncelli, line 305: I've been meaning to write up about this new function and when to use it.  For the meanwhile... trust me :-)

```suggestion
			rec.AddAnswer(&dns.Answer{Rdata: []string{r.GetTargetTXTJoined()}})
```

tlimoncelli (approved):

tlimoncelli (approved):

tlimoncelli (approved):

providers/ns1/ns1Provider.go
costasd: sorry now coming to that - sure :)

PR #2648, opened 2023-11-22
CLOUDFLARE: Update docs WRT domain creation

shodges-vermeer:
Previously it was stated that DNSControl would not create a new domain in Cloudflare unless `dnscontrol create-domains` was executed. That subcommand has now been deprecated and is part of `dnscontrol push`

**Note:** `go vet, go fmt, etc.` were not ran since this is only modifying the documentation and not the code itself.


tlimoncelli:
Thanks for catching that!

Issue #2625, opened 2023-11-15
OVH: provider only supports EU API endpoint


greatislander:
**Is your feature request related to a problem? Please describe.**

I use OVH Canada, which has a different endpoint than OVH in the EU (see [here](https://github.com/ovh/go-ovh/blob/48ddc3f3bc3b62c7b76678184733525b0d06eb14/ovh/ovh.go#L28C2-L30)). Registering an application with OVH as described [here](https://docs.dnscontrol.org/service-providers/providers/ovh#activation) must happen in the region of your OVH account, so as a Canadian client I have to register my application at https://ca.api.ovh.com/createApp rather than https://eu.api.ovh.com/createApp.

However, the OVH provider hardcodes the EU API endpoint, which means that applications registered by Canadian or US users can't be used: https://github.com/StackExchange/dnscontrol/blob/b8096b7b8a90e45b876f3373723fe9cac04dc46c/providers/ovh/ovhProvider.go#L37

**Describe the solution you'd like**
Ideally the provider could be configured to use the CA or US endpoints instead. This would be a non-breaking change as the default could still be the EU endpoint.

**Describe alternatives you've considered**
Canadian and US providers could be created based on the current one, but this seems like a lot of duplication.

**Additional context**

tlimoncelli:
CC @masterzen (OVH provider maintainer)

masterzen:
I'll have a look over the week-end. It doesn't seem difficult to add a backward compatible option to select a non EU endpoint.

masterzen:
@greatislander would you mind testing the PR I just pushed and let me know if that works for accessing OVH CA API?
You would need to add:
```
  "endpoint": "ca"
```
to your `creds.json` file for OVH.

greatislander:
Oops I see it's already been merged. Thanks so much!

cafferata:
@tlimoncelli is a super fast 🚀. Would you (@greatislander) close the GitHub issue when everything works successfully? 🙏

greatislander:
Resolved in #2651.

PR #2651, opened 2023-11-26
OVH: Allow ovh provider to connect to other endpoints than EU (#2625)

masterzen:
Fix for #2625, allowing users to set OVH provider `endpoint` configuration to:
* `eu` (default) to target the EU API endpoint
* `ca` to target the Canada API endpoint
* `us` to target the US API endpoint
* any other url

If omitted, the default is the EU endpoint for backward compatibility.


tlimoncelli:
Thanks!

tlimoncelli:
P.S.  Thank you for a great PR.  This is great!  It fixes a real problem, has tests, and you updated the docs!  I wish all PRs would be so complete!  Thanks again!

PR #2657, opened 2023-11-28
Update docs and move Azure Private DNS to the "community support" list

tlimoncelli:


tlimoncelli:
CC @matthewmgamble for visibility

P.S.  I've moved the provider to the list of community-supported providers.

Issue #2643, opened 2023-11-21
INWX: support null MX records


gvangool:
**Is your feature request related to a problem? Please describe.**
I tried to apply null MX records on a domain at INWX.
But I get an error that it's not supported (https://github.com/StackExchange/dnscontrol/blob/bfa9da98c8642073dffc7c7013bcd151b2d061ae/providers/inwx/auditrecords.go#L14)

**Describe the solution you'd like**
Based on [their documentation](https://kb.inwx.com/en-us/3-nameserver/128-can-i-store-a-null-mx-record-for-my-domain), we can disable the check for `MxNull`.

**Describe alternatives you've considered**
I'm currently using an `IGNORE("@", "MX")` rule to manage it manually in the UI.

**Additional context**
...

tlimoncelli:
CC @patschi (INWX maintainer)

Hi @gvangool !   Would you be willing to submit a PR for this?  Just delete that line and run the integration tests (I don't have an account with INWX so someone else will need to do it). Post the integration test results as a comment, plz.

How to run integration tests: https://docs.dnscontrol.org/developer-info/integration-tests

Tom

gvangool:
> P.S. Being in the wrong directory is a common issue. I'm updating the docs to be more clear. If there are other places I could put this warning please let me know. See #2653

This is already really helpful!

tlimoncelli:
I don't have a test account on INWX but I created https://github.com/StackExchange/dnscontrol/pull/2687. If someone can confirm it works, I'll merge it.

PR #2687, opened 2023-12-05
INWX: Null MX is permitted

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/2643

### DIFF
--- a/providers/inwx/auditrecords.go
+++ b/providers/inwx/auditrecords.go
@@ -11,8 +11,6 @@ import (
 func AuditRecords(records []*models.RecordConfig) []error {
 	a := rejectif.Auditor{}
 
-	a.Add("MX", rejectif.MxNull) // Last verified 2020-12-28
-
 	a.Add("SRV", rejectif.SrvHasNullTarget) // Last verified 2020-12-28
 
 	a.Add("TXT", rejectif.TxtHasBackticks) // Last verified 2021-03-01


tlimoncelli:
[gvangool](https://github.com/gvangool) and @patschi: Please take a look.

tlimoncelli:
That's an odd error message!  The type is wrong?  Maybe no field is being sent (noempty) therefore the type is not "string".

Take a look at `providers/inwx/inwxProvider.go` and add a debug printf something like...

```
  case "MX":
    req.Priority = int(rec.MxPreference)
    req.Content = content[:len(content)-1]
    printer.Printf("DEBUG: inwx mx=%q\n", req.Content)
```

That might shed some light on the situation.

gvangool:
That gives the following (partial) log:
```
=== RUN   TestDNSProviders/phyx-ote.dev/15:MX:Null_MX
    integration_test.go:247:
        + CREATE MX phyx-ote.dev 0 . ttl=300
DEBUG: inwx mx=""
    integration_test.go:252: (2005) Parameter value syntax error. Reason: (The given parameter value for 'content' does not match the expected parameter type 'text') The given parameter value for 'content'
 does not match the expected parameter type 'text'
```

Looking at the way the trailing dot is handled for INWX, I ended up with this patch to fix it
```diff
diff --git i/providers/inwx/inwxProvider.go w/providers/inwx/inwxProvider.go
index c511d85b..23d3a233 100644
--- i/providers/inwx/inwxProvider.go
+++ w/providers/inwx/inwxProvider.go
@@ -180,7 +180,11 @@ func makeNameserverRecordRequest(domain string, rec *models.RecordConfig) *goinw
                req.Content = content[:len(content)-1]
        case "MX":
                req.Priority = int(rec.MxPreference)
-               req.Content = content[:len(content)-1]
+               if content == "." {
+                       req.Content = content
+               } else {
+                       req.Content = content[:len(content)-1]
+               }
        case "SRV":
                req.Priority = int(rec.SrvPriority)
                req.Content = fmt.Sprintf("%d %d %v", rec.SrvWeight, rec.SrvPort, content[:len(content)-1])
@@ -305,7 +309,11 @@ func (api *inwxAPI) GetZoneRecords(domain string, meta map[string]string) (model
                        "PTR":   true,
                }
                if rtypeAddDot[record.Type] {
-                       record.Content = record.Content + "."
+                       if record.Type  == "MX" && record.Content == "." {
+                               // null records don't need to be modified
+                       } else {
+                               record.Content = record.Content + "."
+                       }
                }

                rc := &models.RecordConfig{
```

tlimoncelli:
Excellent detective work!

Would you please submit a PR?

PR #2688, opened 2023-12-05
AZURE_PRIVATE_DNS: Enable building this provider by default

matthewmgamble:
Discovered that by default on the master branch (and the current releases) that the AZURE_PRIVATE_DNS module wasn't getting built by default.  Adding it back to all.go so that it gets included by default.


tlimoncelli:
Oops!  I normally check everything on the checklist but I must have missed this one. I'm surprised much worked with it not in _all.  Odd!

Issue #2652, opened 2023-11-27
CLOUDFLARE: `push` doesn't create records if domain doesn't exist


shodges-vermeer:
**Describe the bug**
We've noticed that `dnscontrol push` will create new domains, but won't actually create any records for said domain. This forces us to run `dnscontrol push` again, or revert to the deprecated `dnscontrol create-domains` then `dnscontrol push`.

Looking through the code (with no experience in writing Go), it doesn't seem that this is intended behavior. 

**To Reproduce**
1. Create a new domain file
2. Add records to new domain
3. Run `dnscontrol push`
4. *Notice DNS records didn't get created, but domain did*
5. Run `dnscontrol push` again

**Expected behavior**
`dnscontrol push` should create the new domain, then create records with a single execution of the command

**DNS Provider**
 - Cloudflare (At minimum, this is untested with other providers)

tlimoncelli:
Thanks for reporting this!

@jpbede : Would you mind taking a look at this?  IIRC (and if git blame is accurate) you worked on that code most recently.

jpbede:
It seems to me that the automatic creation of the domain does not play well with Cloudflare. Cloudflare will not create records if the domain setup was not completed through their portal. As long as my domain was in "Pending Nameserver update" status, they did not accept any records

jpbede:
Ok nvm, user error..

jpbede:
Fixed with PR https://github.com/StackExchange/dnscontrol/pull/2690

shodges-vermeer:
Thank you both!

PR #2690, opened 2023-12-05
CLOUDFLARE: Fixed bug: Zone not populated with records if domain was created in the same run

jpbede:
Reset the `domainIndex` after the domain has been created at Cloudflare so that the following functions can fetch a new list of nameservers.

This fixes https://github.com/StackExchange/dnscontrol/issues/2652


tlimoncelli:
Thanks for the fast fix!

PR #2700, opened 2023-12-08
INWX: support NullMx records

### LINKED ISSUE #2643: INWX: support null MX records

gvangool:
To support the `MxNull` records, we need to avoid the special casing in the INWX provider that let's INWX handle the final dot.

Replaces #2687
Fixes #2643


tlimoncelli:
Yay!  I love it when a plan comes together!

Thanks!

PR #2718, opened 2023-12-11
CHORE: Fix golint warnings about stuttering

tlimoncelli:
CC @blackshadev 

Fix these golint warnings:

```
pkg/dnsgraph/dnsgraph.go:15:6: type name will be used as dnsgraph.DNSGraphEdge by other packages, and that stutters; consider calling this Edge
pkg/dnsgraph/dnsgraph.go:22:6: type name will be used as dnsgraph.DNSGraphEdges by other packages, and that stutters; consider calling this Edges
pkg/dnsgraph/dnsgraph.go:25:6: type name will be used as dnsgraph.DNSGraphNode by other packages, and that stutters; consider calling this Node
```

I used VSCode's "rename symbol" which seems to take care of all the right things.  However I'm splitting this out as its own PR in case we need to roll it back.

### DIFF
--- a/pkg/diff2/diff2.go
+++ b/pkg/diff2/diff2.go
@@ -53,6 +53,7 @@ type Change struct {
 	HintRecordSetLen1 bool
 }
 
+// GetType returns the type of a change.
 func (c Change) GetType() dnsgraph.NodeType {
 	if c.Type == REPORT {
 		return dnsgraph.Report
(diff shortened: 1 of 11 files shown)


tlimoncelli:
@blackshadev I won't merge this until I hear from you.  No rush.

blackshadev:
Go ahead and merge. It is a great suggestion, I love these kind of automated tools the whole reason for the duplicate naming was because it first didn't have it's own package. And when I moved it to it's own package I was busy making things work and then forgot.

tlimoncelli:
👍

PR #2723, opened 2023-12-14
AXFRDDNS: Support separate servers for AXFR and DDNS

midnightveil:
One can specifiy `transfer-server` in `creds.json` which gets used for AXFR, else it falls back to the existing "master" config key.

Use case is if the AXFR server is different to the DDNS server.


cafferata:
Hey @midnightveil!

I see that this is your first PR. That is great! 👏 Welcome! 

Could you also include an code example in the [AXFR+DDNS documentation](https://docs.dnscontrol.org/service-providers/providers/axfrddns)? You can find the Markdown file here: `documentation/providers/axfrddns.md`.

Cheers!

midnightveil:
I added the following section after the existing primary master documentation:

image

cafferata:
Thank you @midnightveil! Looks good. See the GitBook preview deep link URL here:
_https://docs.dnscontrol.org/~/revisions/ATQbKsbCmhZPCQUycGZU/service-providers/providers/axfrddns#transfer-axfr-server_

Small question, I see you [partially removing the trailing dot](https://github.com/StackExchange/dnscontrol/pull/2723/files#diff-c14c7ca0f765d1bddc9965bb295945619860c430a450a18872ada9cace6cb257R110) from the nameservers. Is this correct? Is this desired? If so, would you like to remove it consistently? See [Default nameservers](https://docs.dnscontrol.org/~/revisions/ATQbKsbCmhZPCQUycGZU/service-providers/providers/axfrddns#default-nameservers).

```javascript
var DSP_AXFRDDNS = NewDnsProvider("axfrddns", {
        "default_ns": [
            "ns1.example.com.",
            "ns2.example.com.",
            "ns3.example.com.",
            "ns4.example.com."
        ]
    }
)
```

midnightveil:
> Small question, I see you partially removing the trailing dot from the nameservers. Is this correct? Is this desired? (Maybe it's JavaScript versus JSON?) If so, would you like to remove it consistently? See Default nameservers.

I removed it because in my testing DNS control would error out complaining about a trailing dot, so I assumed the documentation was incorrect. Possibly should be a separate change.

midnightveil:
> provider code leaves trailing dot on nameserver

Is the error I get.

tlimoncelli:
Thanks!

PR #2744, opened 2023-12-26
GCORE: Fix handling very long TXT records

xddxdd:
Fix Gcore provider's error when handling a very long TXT record (>255 chars):

```
FAILURE! 400: bad TXT-record content [...]. TXT-record content must contain exactly 1 element (e.g. some text)
```

Although Gcore's API takes TXT records as a string array, it doesn't accept multiple elements in it. Fixed by passing the entire TXT record as one element instead of using the segmented one.


cafferata:
Ping @xddxdd, the maintainer of this provider.

tlimoncelli:
Thanks for fixing that!  (and apologies for breaking it!)

PR #2753, opened 2024-01-01
NEW REGISTRAR: Dynadot (DYNADOT)

e-im:
I have added support for Dynadot as a registrar. They do not have an official (or any community) go package for their API, so I have implemented a small subset manually. I based the implementation on the existing internet.bs and easyname providers.

[Dynadot API docs](https://www.dynadot.com/domain/api-commands)

They do provide a JSON API (replace `.xml` with `.json` in the url), however it is quite odd and I suspect internally may just be converting XML to JSON after the response has been generated, yielding weird results. As such I have used the XML API.

I have tested this manually. I'm happy to provide a test API key if necessary.


providers/dynadot/api.go
```diff
@@ -0,0 +1,135 @@
...
+type getNsResponse struct {
+	XMLName     xml.Name  `xml:"GetNsResponse"`
+	NsContent   nsContent `xml:"NsContent"`
+	GetNsHeader header    `xml:"GetNsHeader"`
+}
+
+type nsContent struct {
+	Host   []string `xml:"Host"`
+	NsName string   `xml:"NsName"`
+}
```
e-im, line 44: I'm not too familiar with go and couldn't find a better way to do this.
For all responses except `get_ns` I only need the header, except it's under a different key each time.

Please let me know if there is a better way to implement this. A current consequence of my implementations means error checking is fully extracted to each method rather than the general `get()`.

providers/dynadot/api.go
```diff
@@ -0,0 +1,135 @@
...
+		err = xml.Unmarshal(b, &resp)
+		if err != nil {
+			return fmt.Errorf("failed NS add (Dynadot): %s", err)
+		}
+
+		if resp.AddNsHeader.SuccessCode != 0 {
+			// No apparent way to get all existing entries on an account, so filter
+			if strings.Contains(resp.AddNsHeader.Error, "already exists") {
+				continue
+			}
```
e-im, line 84: The Dynadot API requires nameservers be added to a separate list on each account before they can be added to a domain.

I don't see any way with their API to get this full list, so I've just attempted to add every nameserver to it every time, ignoring ones that error due to already existing.

This feels super hacky although I don't see a better way of doing it.

The SuccessCode is always just -1 for a failure, there are no codes for specific cases so checking the error message is required.

providers/dynadot/api.go
tlimoncelli: You did it the way I would have done it!  (So it must be right, eh? :-) )

providers/dynadot/api.go
tlimoncelli: I can't think of anything better.

tlimoncelli (approved):

tlimoncelli:
Thank you for contributing this new registrar!  Please be sure to join the https://groups.google.com/g/dnscontrol-discuss email list so you stay up to date with any big announcements.

tlimoncelli:
provider-DYNADOT created!

PR #2759, opened 2024-01-03
GCLOUD: display all correction messages affecting same label + type in a zone

### LINKED ISSUE #2758: GCLOUD:  correction message strings for multiple changes affecting same label + type are partially dropped
**Describe the bug**
When multiple changes affect the same label + type in a GCLOUD zone, push/preview only displays one of the messages

**To Reproduce**
1. An existing dnsconfig and empty zones created in the GCLOUD account:
```
var DNS_GOOGLE = NewDnsProvider('gcloud');
D("dnscontrol-integration-test.com", REG_NONE,DnsProvider(DNS_GOOGLE),
	NAMESERVER_TTL('21600')
);
```
2. Modify the dnsconfig with changes that affect the same label + type:
```
var DNS_GOOGLE = NewDnsProvider('gcloud');
D("dnscontrol-integration-test.com", REG_NONE,DnsProvider(DNS_GOOGLE),
	NAMESERVER_TTL('21600'),
	MX('@', 10, 'mail.foo.bar.'),
	MX('@', 10, 'mail2.foo.bar.'),
	TXT('@', 'foo'),
	TXT('@', 'bar'),
	A('foo', '127.0.0.1'),
	A('foo', '127.0.0.2')
);
```
3. Run 'dnscontrol preview'
```
$ dnscontrol preview
******************** Domain: dnscontrol-integration-test.com
1 correction (gcloud)
#1: + CREATE MX dnscontrol-integration-test.com 10 mail2.foo.bar. ttl=300
+ CREATE TXT dnscontrol-integration-test.com "foo" ttl=300
+ CREATE A foo.dnscontrol-integration-test.com 127.0.0.2 ttl=300
Done. 1 corrections.
```

**Expected behavior**
Preview/Push should display all changes associated with a label
```
$ dnscontrol preview
******************** Domain: dnscontrol-integration-test.com
1 correction (gcloud)
#1: + CREATE MX dnscontrol-integration-test.com 10 mail.foo.bar. ttl=300
+ CREATE MX dnscontrol-integration-test.com 10 mail2.foo.bar. ttl=300
+ CREATE TXT dnscontrol-integration-test.com "bar" ttl=300
+ CREATE TXT dnscontrol-integration-test.com "foo" ttl=300
+ CREATE A foo.dnscontrol-integration-test.com 127.0.0.1 ttl=300
+ CREATE A foo.dnscontrol-integration-test.com 127.0.0.2 ttl=300
Done. 1 corrections.
```

**DNS Provider**
 - GCLOUD

**Additional context**
The actual API changes are still read and executed correctly so the integration tests are passing.  Will submit PR with fix for the messages.

asn-iac:
fixes #2758

## Release changelog section

* GCLOUD: display all correction messages affecting same label + type in a zone

### DIFF
--- a/providers/gcloud/gcloudProvider.go
+++ b/providers/gcloud/gcloudProvider.go
@@ -280,13 +280,25 @@ func (g *gcloudProvider) GetZoneRecordsCorrections(dc *models.DomainConfig, exis
 
 	changedKeys := map[key]string{}
 	for _, c := range create {
-		changedKeys[keyForRec(c.Desired)] = fmt.Sprintln(c)
+		msg := fmt.Sprintln(c)
+		if k, ok := changedKeys[keyForRec(c.Desired)]; ok {
+			msg = strings.Join([]string{k, msg}, "")
+		}
+		changedKeys[keyForRec(c.Desired)] = msg
 	}
 	for _, d := range toDelete {
-		changedKeys[keyForRec(d.Existing)] = fmt.Sprintln(d)
+		msg := fmt.Sprintln(d)
+		if k, ok := changedKeys[keyForRec(d.Existing)]; ok {
+			msg = strings.Join([]string{k, msg}, "")
+		}
+		changedKeys[keyForRec(d.Existing)] = msg
 	}
 	for _, m := range modify {
-		changedKeys[keyForRec(m.Existing)] = fmt.Sprintln(m)
+		msg := fmt.Sprintln(m)
+		if k, ok := changedKeys[keyForRec(m.Existing)]; ok {
+			msg = strings.Join([]string{k, msg}, "")
+		}
+		changedKeys[keyForRec(m.Existing)] = msg
 	}
 	if len(changedKeys) == 0 {
 		return nil, nil


tlimoncelli:
Ah!  Thanks for spotting this bug!

It makes me wonder if all the providers that use `IncrementalDiff` will have the same problem.

(Fundamentally we need to rewrite this provider to not use the backwards compatibility mode.  Using diff2 would make all that code much cleaner.)

tlimoncelli (approved):

asn-iac:
I'd guess it depends on how the provider is handling the messages coming from `IncrementalDiff`.  GCLOUD used to just copy all of the message strings from `IncrementalDiff` into a separate `[]string` and print them all out as a single correction.  I had actually implemented `differ.ChangedGroups` for the GCLOUD provider in a private fork of 2.x, which worked well, but then reverted to `IncrementalDiff` when I contributed #2482 in an effort to reduce the number of changes and because of the ongoing efforts with diff2 (`ByRecordSet` looked like it could be used for this provider, but based on the docs I'm unsure if the provider API conforms with the intent of that function).  That's where this message display bug was originally introduced.
Thanks for the quick review!

tlimoncelli:
@asn-iac Sounds fair!

By the way... would you be interested in taking a try at rewriting it with diff2?  I'd be glad to collaborate on it.  For example, if  `ByRecordSet` is insufficient, I'd be glad to fix it or provide a new `By` function that matches the API.

PR #2762, opened 2024-01-04
GCLOULD: Convert to diff2 native

tlimoncelli:

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -232,14 +232,14 @@ func makeChanges(t *testing.T, prv providers.DNSServiceProvider, dc *models.Doma
 			t.Fatal(fmt.Errorf("runTests: %w", err))
 		}
 		if tst.Changeless {
-			if count := len(corrections); count != 0 {
+			if count := countActions(corrections); count != 0 {
 				t.Logf("Expected 0 corrections on FIRST run, but found %d.", count)
 				for i, c := range corrections {
-					t.Logf("UNEXPECTED #%d: %s", i, c.Msg)
+					t.Logf("UNEXPECTED a #%d: %s", i, c.Msg)
 				}
 				t.FailNow()
 			}
-		} else if (len(corrections) == 0 && expectChanges) && (tst.Desc != "Empty") {
+		} else if (countActions(corrections) == 0 && expectChanges) && (tst.Desc != "Empty") {
 			t.Fatalf("Expected changes, but got none")
 		}
 		for _, c := range corrections {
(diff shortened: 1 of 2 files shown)


tlimoncelli:
Hi @asn-iac @riyadhalnur!

I got inspired to try converting gcloud to native diff2 and it was easier than I thought.

All the tests work except some of the IGNORE tests.

tlimoncelli:
(and I have no idea why those tests are failing!)

tlimoncelli:
I figured it out!  All tests now pass.

The code is considerably shorter and easier to read.

Any feedback @riyadhalnur?  I'd like to merge this soon.

asn-iac:
@tlimoncelli - after initial review, the only (semi meaningful?) external behavior change I see is with the potential efficiency of the `*gdns.Change`s sent to the gcloud api.  There are conditions where a `Change` is not completely filled with the maximum number of `Additions` and `Deletions`.  However, it seems like there are other cases where this behavior would now be desirable, due to update reordering from dependency graph offered in diff2.

The following

tlimoncelli:
Thanks for that test case!  Very helpful!

Sadly I'm not sure what inefficiency you are pointing out.  Is it that a Change could potentially have (for example) 1000 additions (no deletions), but then the next batch has deletions, which possibly could have been in the first batch?

Tom

P.S.  Interestingly enough... I set "batchMax = 10000" and the test case worked.  I wonder if Google eliminated the 1000-item limit.

asn-iac:
>  Is it that a Change could potentially have (for example) 1000 additions (no deletions), but then the next batch has deletions, which possibly could have been in the first batch?
Yup, you got it - it's a fairly minor issue imo, but wanted to make sure it's understood.
> P.S. Interestingly enough... I set "batchMax = 10000" and the test case worked. I wonder if Google eliminated the 1000-item limit.
That *is* interesting... it does seem like there has been a change, but of course it's not documented anywhere.  From some additional testing it looks like the new max is actually 3000 (of each of additions and deletions) per `gdns.Change`.  The following test will fail if batchMax > 3000
```
testgroup("pager3201",
			only(
				"GCLOUD",
			),
			tc("3200 records", manyA("rec%04d", "1.2.3.4", 3200)...),
			tc("Update 3200 records", manyA("rec%04d", "1.2.3.5", 3200)...),
		),
```

I confirmed 3000 works for both the current diff1-based code and the new diff2-based code.

tlimoncelli:
I'm glad you raised the issue. I'll add a comment about that in the code.

My priority is to be correct first, efficient second.  I don't have an official stats, but it is safe to assume that most `dnscontrol push` runs make very small changes.  An update of >1000 ResourceSets is probably very rare (when initially populating a zone?)

I'm glad we now have a test case that tests the batching algorithm!  I'm always concerned about off-by-one errors in my code, which is why we have the `pager*` tests.  This is a good test to have for similar reasons.

Thanks again!

asn-iac:
> My priority is to be correct first, efficient second. I don't have an official stats, but it is safe to assume that most dnscontrol push runs make very small changes. An update of >1000 ResourceSets is probably very rare (when initially populating a zone?)

I would agree with that approach.  Is it also safe to say diff2 should be the source of truth for the order of the changes, and that providers should honor that?

tlimoncelli:
> I would agree with that approach. Is it also safe to say diff2 should be the source of truth for the order of the changes, and that providers should honor that?

In general, yes.  The new dnsgraph code should order things properly so that (for example) a CNAME's target is always created before the CNAME, etc.  If it doesn't order the updates in a way that is appropriate for the provider, it's probably best to file a bug to see if we can fix the dnsgraph code.

Tom

PR #2669, opened 2023-11-30
DOCS: Fix `R53_` broken links

cafferata:
@jonathanbouvier [commented](https://github.com/StackExchange/dnscontrol/pull/2659#issuecomment-1833834072)

> The link to R53_EVALUATE_TARGET_HEALTH on [R53_ALIAS](https://docs.dnscontrol.org/language-reference/domain-modifiers/service-provider-specific/amazon-route-53/r53_alias) is pointing to https://github.com/StackExchange/dnscontrol/blob/master/documentation/functions/record/R53_EVALUATE_TARGET_HEALTH.md instead of https://docs.dnscontrol.org/language-reference/record-modifiers/service-provider-specific/amazon-route-53/r53_evaluate_target_health
> 
> Any ideas what might be going wrong there?

I fixed the `R53_` broken links by adding a backslash in the file names. See the preview URL:
_https://docs.dnscontrol.org/~/revisions/SMhInsWkrPeuva4GWfmr/language-reference/domain-modifiers/service-provider-specific/amazon-route-53/r53_alias_

_PS. I've never understood when this is necessary. I'm going after it. The question has been submitted to the GitBook support department._

PR #2771, opened 2024-01-06
DOCS: Bunny DNS - GitBook escape the underscore

cafferata:
Enscape the underscore so that GitBook resolves the Markdown file successfully.


ppmathis:
@cafferata Wouldn't this also need to be done for Azure (Private) DNS? That provider was one that served as inspiration and as I saw that `azure_dns.md` was referenced as-is, I did not deem any escaping to be necessary. Or is there like yet another sort of mechanism which makes it work in that case?

cafferata:
I don't have a good idea when GitBook (_the documentation generator_) requires this underscore. 😖 I contacted GitBook support at the time of https://github.com/StackExchange/dnscontrol/pull/2669#issue-2019292905, but the answer I received did not get me much further. For completeness, as it is missing in https://github.com/StackExchange/dnscontrol/pull/2669:

> **Why is the underscore considered a special char? Why does it work in one Markdown file, and not in another Markdown file?**
> 
> This is not specific to one markdown file or another. What happens is that when you create the link in a markdown file it should work just fine, however when you edit that same file through the editor it escapes those automatically during normalization.
> 
> This means that if the underscore is not escaped it'll unfortunately cause the issue you saw there.
> 
> I understand that this is not easily done in an open source project with multiple users, but it would be the best outcome in this case.

PR #2772, opened 2024-01-06
DOCS: Bunny DNS - Be more consistent in ENV variable example

cafferata:
Aligned the environment variable (Bunny provided with DNS) similar to [.github/workflows/pr_test.yml](https://github.com/StackExchange/dnscontrol/blob/961eaa7862984683d652d03047074f95355b43c9/.github/workflows/pr_test.yml#L133).


ppmathis:
@cafferata Certainly can't hurt to have these env variables aligned. :+1: On that note though, I would also like to hint towards the Azure Provider, which once again served as a template for me as it also had a underscore in its name.

There it's also referenced as `AZURE_XXX` instead of `AZURE_DNS_XXX`, which was why I decided to ditch the `_DNS` suffix as well. So if its desired to always have these aligned with the provider name, maybe these could be updated as well.

cafferata:
> So if its desired to always have these aligned with the provider name, maybe these could be updated as well.

Thanks for this feedback! Do you want to prepare a GitHub pull request yourself? Or would you prefer that I make these changes?

ppmathis:
> Thanks for this feedback! Do you want to prepare a GitHub pull request yourself? Or would you prefer that I make these changes?

Feel free to make these changes if you find time, maybe they can even be added to this PR with a more generalized title, I didn't get to check yet if other providers might also be affected from this. Otherwise let me know and I'll open one myself.

Issue #2798, opened 2024-01-15
CLOUDFLAREAPI: LOC supported by CF but not by integration


ConnorMcF:
**Describe the bug**
The LOC record is supported by Cloudflare, but not by the integration.
https://github.com/StackExchange/dnscontrol/blob/56220fba5f94b13c935dea782682f4fd45d6d0a1/providers/cloudflare/cloudflareProvider.go#L46

**Expected behavior**
Cloudflare supports LOC records, I would assume that the DNSControl integration would also support them.
I'm hoping this may be as simple as switching `Cannot` to `Can`.

LOC is listed as a valid type in the Cloudflare API: https://developers.cloudflare.com/api/operations/dns-records-for-a-zone-list-dns-records

**DNS Provider**
 - Cloudflare

cafferata:
> I'm hoping this may be as simple as switching Cannot to Can.

That's right! Do you want to prepare a GitHub pull request yourself? Or shall I do that for you? ☺️

cafferata:
cc: Maintainer `CLOUDFLAREAPI` @tresni for visibility.

cafferata:
> **type** _string_
> Record type.
> 
> Allowed values: `A`, `AAAA`, `CAA`, `CERT`, `CNAME`, `DNSKEY`, `DS`, `HTTPS`, `LOC`, `MX`, `NAPTR`, `NS`, `PTR`, `SMIMEA`, `SRV`, `SSHFP`, `SVCB`, `TLSA`, `TXT`, `URI`

[Deeplink](https://developers.cloudflare.com/api/operations/dns-records-for-a-zone-list-dns-records#Query-Parameters:~:text=any-,type,-string)

tlimoncelli:
I've been thinking about this bug and the associated draft PR https://github.com/StackExchange/dnscontrol/pull/2799.

This bug exposes a fundamental design problem with DNSControl: a record like LOC() has its inputs validated, massaged, etc, in Javascript which is difficult to work with because DNSControl has no unit testing framework, debugging tools, etc. for Javascript.  Also, the Javascript interpreter we use implements an old standard of Javascript.  (Such tools exist for Javascript, but we don't use them.)

[...]

PR #2799, opened 2024-01-16
CLOUDFLAREAPI: Enable CanUseLOC

ConnorMcF:
Enables LOC records with the Cloudflare provider for #2798.

Draft PR as this currently doesn't work.

### DIFF
--- a/commands/types/dnscontrol.d.ts
+++ b/commands/types/dnscontrol.d.ts
@@ -363,7 +363,7 @@ declare function AZURE_ALIAS(name: string, type: "A" | "AAAA" | "CNAME", target:
  * );
  * ```
  *
- * DNSControl contains a [`CAA_BUILDER`](../record/CAA_BUILDER.md) which can be used to simply create `CAA()` records for your domains. Instead of creating each CAA record individually, you can simply configure your report mail address, the authorized certificate authorities and the builder cares about the rest.
+ * DNSControl contains a [`CAA_BUILDER`](CAA_BUILDER.md) which can be used to simply create `CAA()` records for your domains. Instead of creating each CAA record individually, you can simply configure your report mail address, the authorized certificate authorities and the builder cares about the rest.
  *
  * @see https://docs.dnscontrol.org/language-reference/domain-modifiers/caa
  */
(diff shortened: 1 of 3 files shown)


tlimoncelli:
CC @systemcrash

models/record.go
```diff
@@ -95,40 +95,65 @@ type RecordConfig struct {
...
+	LocLatDirection  string            `json:"loclatdirection,omitempty"`
+	LocLongDegrees   uint8             `json:"loclongdegrees,omitempty"`
+	LocLongMinutes   uint8             `json:"lotlongminutes,omitempty"`
+	LocLongSeconds   float32           `json:"loclongseconds,omitempty"`
+	LocLongDirection string            `json:"loclongdirection,omitempty"`
+
+	LocOrigAltitude  int32             `json:"locorigaltitude,omitempty"`
+	LocOrigSize      float32           `json:"locorigsize,omitempty"`
+	LocOrigHorizPre  float32           `json:"locorighorizpre,omitempty"`
+	LocOrigVertPre   float32           `json:"locorigvertpre,omitempty"`
```
tlimoncelli, line 131: Why does this provider need new fields when multiple providers already support LOC without needing the extra fields?

ConnorMcF: The Cloudflare API requires the original degrees, minutes, seconds values whereas to my knowledge the LOC model converts to decimal degrees, which doesn't fit into the request data Cloudflare asks for.
I'm trying to pass through both the DD and the original DMS values to the provider to allow either to be used.

https://developers.cloudflare.com/api/operations/dns-records-for-a-zone-create-dns-record (Body -> LOC Record)

models/record.go
```diff
@@ -163,40 +188,65 @@ func (rc *RecordConfig) UnmarshalJSON(b []byte) error {
...
 		LocLongitude     int               `json:"loclongitude,omitempty"`
 		LocAltitude      uint32            `json:"localtitude,omitempty"`
+
+		LocLatDegrees    uint8             `json:"loclatdegrees,omitempty"`
+		LocLatMinutes    uint8             `json:"lotlatminutes,omitempty"`
+		LocLatSeconds    float32           `json:"loclatseconds,omitempty"`
+		LocLatDirection  string            `json:"loclatdirection,omitempty"`
+		LocLongDegrees   uint8             `json:"loclongdegrees,omitempty"`
+		LocLongMinutes   uint8             `json:"lotlongminutes,omitempty"`
+		LocLongSeconds   float32           `json:"loclongseconds,omitempty"`
```
systemcrash, line 218: I couldn't see the LOC records at cloud flare link. So we assume only seconds will take decimal fractions ie floats. Degrees and minutes are whole integers.

tlimoncelli:
I've done some research and now I think I can give a (hopefully!) more useful recommendation.

Rather than derive those fields and store them in RecordConfig, those fields should be derived "on demand" when needed. That is, the cloudflare createRecDiff2() function should generate those additional fields.

This would have a few benefits:
1. It doesn't increase the memory size of RecordConfig.
2. It better supports other providers, who may have a different set of fields they need.
3. It maintains the ideal that RecordConfig is the desired state, and each provider does what's needed for their native implementation. For example, TXT records are stored as one big string. If a provider wants them as 255-octet chunks, it's their responsibility to do the chunking and un-chunking behind the scenes. This converstion is usually done in functions with names like toNative() and toRC(). In Cloudflare's case, the conversion isn't isolated but it part of createRecDiff2() and other functions.
4. This won't require additional CPU, and might require less CPU since the conversion will only be done if an LOC record changes.
5. If there are helper functions that will be useful to many providers, put them in t_loc.go.

Hope that helps,
Tom

systemcrash:
Evidently cloudflare constructs the content record for you. They probably consider this a tricky record type.

https://developers.cloudflare.com/api-next/resources/dns/subresources/records/models/loc_record/#(schema) 
(page seems a bit buggy for that direct link)

PR #2802, opened 2024-01-18
GCORE: enable ALIAS records

xddxdd:
GCore DNS supports ALIAS record at apex zone level by automatic CNAME flattening: https://gcore.com/docs/dns/dns-records/specify-cname-at-root

Enable ALIAS records for GCORE provider, and transform ALIAS records into CNAME just like CLOUDFLARE provider.


tlimoncelli:
Thanks!

PR #2818, opened 2024-01-26
CICD: Bugfix: Changes introduced while tagging new releases cause git problems

tlimoncelli:
* Correct the `git push` statement used in producing releases.

### DIFF
--- a/documentation/release-engineering.md
+++ b/documentation/release-engineering.md
@@ -24,7 +24,7 @@ git commit -a -m "Update generated files for $VERSION"
 ```shell
 export VERSION=v4.2.0
 git tag -m "Release $VERSION" -a $VERSION
-git push origin --tags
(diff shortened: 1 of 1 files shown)


tlimoncelli:
CC @philpennock

cafferata:
Just a thought, is it an idea to create the git tags via the GitHub interface?

https://github.com/StackExchange/dnscontrol/releases/new?tag=v4.8.3&title=v4.8.3&body=Notes

tlimoncelli:
I prefer command line tools when making a release.  There may be a `gh` command equivalent to that URL.

cafferata:
> There may be a gh command equivalent to that URL.

GitHub CLI is also an option: https://cli.github.com/manual/gh_release_create

philpennock (approved):

PR #2835, opened 2024-02-12
DOCS: Document daily update limits

tlimoncelli:
* Update deSEC to include the 300 updates per day per domain limit
* Remove deSEC from certain tests that would exceed limits


tlimoncelli:
You can see a draft at https://docs.dnscontrol.org/~/revisions/Kn8ghzw8TST6mlns6cwJ/service-providers/providers/desec
(That's the "Check" called "GitBook (documentat) - docs.dnscontrol.org" above).

cafferata (approved):

PR #2855, opened 2024-02-27
AXFRDDNS: Avoid appending dot if TSIG key ID already has a dot suffix

halochou:
AXFRDDNS provider will append a `.` to TSIG key ID no matter if there's already an existing one.

If the key ID provided by user already has a dot suffix, a general `bad rdata` error will be shown without few meaningful clue for troubleshooting.


tlimoncelli:
CC @hnrgrgr (the maintainer of this provider) for visibility.

PR #2874, opened 2024-03-18
IDNA proposal

tlimoncelli:

### DIFF
--- a/documentation/design-idna.md
+++ b/documentation/design-idna.md
@@ -1,0 +1,187 @@
+# DNSControl and Internationalized domain name
+
+This is my proposal for how to make IDNs work better in DNSControl.
+Basically, the UI will accept any format.  Early in the process
+DNSControl will store labels/domains data 4 ways: As received from the
+user (downcased), ASCII, Unicode, and in a "display" format that shows
+both. The converstions already done ahead of time, providers can
+access whatever format they need.  Output from the main program will
+use the "display" format when possible.
+
+# Problem Statement
+
+DNSControl doesn't handle internationalized domain names (IDNs) very
+well.  Coverage is unevent: They work better in some providers than
+others. There are bugs and inconsistencies. Writing a provider that
+handles IDNs properly requires doing most of the work in the provider
+itself, which means every provider maintainer must be an expert in
+IDNs, which is unreasonable.
+
+# Background:
+
+RFC 3490 recommends how applications should handle IDNs. My summary:
+(1) the UI should accept a mix of Unicode and ASCII domains/labels.
+(2) internally translate everything to ASCII (punycode) and do all
+processing in that format, (3) when displaying output, display it as
+the user input it, or Unicode, or ASCII, or give users a choice.
+
+* IDNA: Internationalizing Domain Names in Applications
+* IDNs: Internationalized domain names
+* ACE Prefix: The `xn--` that means "Puny code follows"
+* ASCII: A label or domain is output as ASCII with ACE prefix if needed.
+* Unicode: A label or domain is output as Unicode.
+
+Proposed Outcome:
+
+1. Users should be able to input domains and labels in either ASCII (with ACE prefix if needed), Unicode, or a mix.  This holds for input via `dnsconfig.js` (domain names, labels, and targets); as well as flags such as `--domains`.
+2a. Output should be Unicode or both with the ASCII being in parenthesis. Example: `рф.com (xn--p1ai.com)`
+2b. Or maybe the reverse? Example: `xn--p1ai.com (рф.com)`
+3. DNSControl's main code should create a "paved path" for providers to make it easy for them to do the right thing. It should be easier to do the right thing than the wrong thing.  The default (i.e. "lazy") path should result in the behavior we desire.
+
+Here are some example outputs:
+
+NOTE: Feedback needed!  Do you prefer "a" or "b"?  Is there an even better format I should consider?  Should we use `{}` instead of `()`?
+
+Example 1a: CREATE unicode (ascii)
+
+```
+#1: + CREATE foo.рф.com (foo.xn--p1ai.com) MX 10 рф.com. (xn--p1ai.com) (ttl=14400)
+```
+
+Example 1b: CREATE ascii (unicode)
+
+```
+#2: + CREATE foo.xn--p1ai.com (foo.рф.com) MX 10 xn--p1ai.com. (рф.com.) (ttl=14400)
+```
+
+Example 3a: MODIFY ascii (unicode) -> ascii (unicode)
+
+```
+#3: ± MODIFY foo.xn--p1ai.com (foo.рф.com) (10 xn--p1ai.com. (рф.com.) ttl=14400) -> (10 foo.xn--p1ai.com. (foo.рф.com.) ttl=14400)
+```
+
+Example 3b: MODIFY unicode (ascii) -> unicode (ascii)
+
+```
+#4: ± MODIFY foo.рф.com (foo.xn--p1ai.com) (10 рф.com. (xn--p1ai.com.) ttl=14400) -> (10 foo.рф.com. (foo.xn--p1ai.com.) ttl=14400)
+```
+
+Example 3c: MODIFY ascii
+
+```
+#5: ± MODIFY foo.рф.com (10 рф.com. ttl=14400) -> (10 foo.рф.com. ttl=14400)
+```
+
+Example 3d: MODIFY unicode
+
+```
+#6: ± MODIFY foo.xn--p1ai.com (10 xn--p1ai.com. ttl=14400) -> (10 foo.xn--p1ai.com. ttl=14400)
+```
+
+NOTE: When the ASCII and Unicode versions are the same (i.e.
+everything is plain ASCII) the display would appear as before:
+
+```
+#7: + CREATE foo.example.com MX 10 xn--p1ai.com (рф.com) (ttl=14400)
+#8: ± MODIFY foo.example.com (10 example.com. ttl=14400) -> (10 foo.example.com. ttl=14400)
+```
+
+# Design
+
+In general, DNSControl will store domains and labels in multiple
+formats: (1) in the original
+format the user specified (downcased), in ASCII, and in Unicode, and
+in a format useful for displaying to users.  This way
+providers do not have to do conversions.
+
+When the user used Unicode:
+
+* Original: рф.com
+* ASCII: xn--p1ai.com
+* Unicode: рф.com
+* Display: xn--p1ai.com (рф.com)
+
+When the user used ASCII:
+
+* Original: xn--p1ai.com
+* ASCII: xn--p1ai.com
+* Unicode: рф.com
+* Display: xn--p1ai.com (рф.com)
+
+NOTE: User input is downcased.  If the user input is `D('xn--P1AI.COM')` the Original field would be `xn--p1ai.com` and so on.
+
+Memory usage will be minimized by using Go's slices.  In the above
+example, the Display string would be generated first, the others would
+be slices of that string.
+
+
+```
+models.DomainConfig:
+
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Mentioning a number of people that I suspect have IDNA domains or knowledge about how IDNA works (based on skimmy bugs and PRs)

@adamus1red
@dkim1970
@flz
@juliusrickert
@killerbees19
@Kusado
@louis-lau
@masterzen
@mderriey
@pmoroney
@tresni
@Yannik

I'd love to receive feedback about this proposal!

Yannik:
@tlimoncelli 

I think your example for the asci==unicode exampe might be wrong:
```
#7: + CREATE foo.example.com MX 10 xn--p1ai.com (рф.com) (ttl=14400)
``` 
Was this what you intended?

Yannik:
I noticed you created both CREATE and MODIFY examples for `ascii (unicode)` and `unicode (ascii)`, but only MODIFY examples for `ascii` and `unicode`. How is that to be understood?

tlimoncelli:
> @tlimoncelli
> 
> I think your example for the asci==unicode exampe might be wrong:
> 
> ```
> #7: + CREATE foo.example.com MX 10 xn--p1ai.com (рф.com) (ttl=14400)
> ```
> 
> Was this what you intended?

Ah, good point!

The first one is where the label is ascii==unicode but the target is ascii!=unicode.  I'll update the comment.


Thanks for finding that!

Tom

tlimoncelli:
> I noticed you created both CREATE and MODIFY examples for ascii (unicode) and unicode (ascii), but only MODIFY examples for ascii and unicode. How is that to be understood?

I've added more examples.  I don't think I've covered every combination, but my goal is to show typical examples not every possible example.

I've also added examples where we use {} and `⟬⟭` and `❮❯`.  I think using unicode chars to highlight unicode domains would be cool (maybe too clever?).

adamus1red:
While I'm not against having the ASCII and UTF on the output lines, I do worry it might make the output too busy. 
Wouldn't simply using the `.Name` value be a better since it should then match what is in the dnscontrol configuration?

pmoroney:
LGTM

tlimoncelli:
@adamus1red wrote:
> While I'm not against having the ASCII and UTF on the output lines, I do worry it might make the output too busy.
> Wouldn't simply using the .Name value be a better since it should then match what is in the dnscontrol configuration?

That's an interesting point!  I guess my thought is that showing both versions helps with debugging.

adamus1red:
> That's an interesting point!  I guess my thought is that showing both versions helps with debugging.

@tlimoncelli maybe a compromise would be if the DNS provider or Registrar used.

I know I've had issues where the DNS is using UTF but the registrar is using ASCII.

louis-lau:
The only IDNA domain I have is for fun, so I don't have a strong preference. I'll give my input nonetheless :). If you want to show both, I think I like B better, as it feels more consistent to me. Anything not in brackets will always be ASCII that way.

I'd probably go with showing what the original user input was, with a flag to only show ASCII if needed. It's less information to parse, and the user should be familiar with it as that's the way it's listed in their config. I could see points being made for showing both, but I've always liked things more distraction free and less dense.

I think the Unicode brackets are a little too clever, perhaps even a little confusing ;)

Yannik:
First of all, improving IDNA handling would be a great improvement to dnscontrol.

Regarding output, the one thing I definitely do not like is having the ascii output come first, because it is the one least likely to be understood/mentally associated with the relevant domain.

I think simply using the original user input has merit, pairing that with a toggle to additionally show ascii seems fine to me.  
However, I also wouldn't mind the `unicode (ascii)` output.

dkim1970:
> First of all, improving IDNA handling would be a great improvement to dnscontrol.
> 
> Regarding output, the one thing I definitely do not like is having the ascii output come first, because it is the one least likely to be understood/mentally associated with the relevant domain.
> 
> I think simply using the original user input has merit, pairing that with a toggle to additionally show ascii seems fine to me. However, I also wouldn't mind the `unicode (ascii)` output.

I'm seconding this suggestion, by displaying the "human readable" format I think the barrier for using IDN's with dnscontrol is getting lowered.

Because the IDNA format is not human readable, especially when it comes to non-latinized languages.

tlimoncelli:
This is excellent feedback!  It's getting me excited!

Question: In what situations would people want to see something besides the .Name (the user input) version?

adamus1red:
> Question: In what situations would people want to see something besides the .Name (the user input) version?

What about if the registrar or dns provider use something different than the .Name value, then include the version they are using in brackets?

louis-lau:
Personally, I think whatever the dns provider does isn't relevant to the cli output. Behind the scenes at every provider, it's all punycode anyway.

Yannik:
> Personally, I think whatever the dns provider does isn't relevant to the cli output. Behind the scenes at every provider, it's all punycode anyway.

Agreed, having the display format handled outside of the provider is to be preferred IMO.

masterzen:
I don't have experience with IDNA at all.
My $0.02: I do agree that showing the ascii version is useful only for debugging, what users want to see is if their unicode domain (or however it was entered in dnsconfig.js) is being processed and how.

tlimoncelli:
Hi folks!

2 ideas:

### Support multiple formats?

There's been a lot of discussion about `ascii (unicode)` vs `unicode (ascii)`.  It might be possible to add a command line flag that selected the format.   No promises, but it might be possible.  In that case, I'd recommend the default be `userinput` and add a flag for debugging that shows `unicode (ascii)` or `userinput (ascii)`.

I'll know more if this is possible when I start coding.

### An idea that would break less existing code

Existing code expects .Name to be ASCII (the current code runs dc.Punycode() for all providers, which rewrites .Name to be ASCII).  Rather than require every use of .Name to change to .NameASCII, maybe the names should be:   .Name (ASCII, to be compatible with old code), .NameORIG (how the user input the string), .NameUNICODE, and .NameDisplay.

documentation/design-idna.md
```diff
@@ -0,0 +1,244 @@
...
+
+Memory usage will be minimized by using Go's slices.  In the above
+example, the Display string would be generated first, the others would
+be slices of that string.
+
+
+```
+models.DomainConfig:
+
+   .Name: the name from D() after downcased via unicode.ToLower()
```
flokli, line 178: I'd prefer to always put the punycode variant in here (which is identical to the non-punycode if it's a ASCII-only domain).
This should require the least amount of changes in all providers and "do the right thing" out of the box 99% of the time, without breaking pure-ASCII domains.

It's also least surprising for users - how they write it shouldn't affect how the provider should treat it.
Which encoding the user provided in the `D` call is nothing individual providers need to worry about.

documentation/design-idna.md
```diff
@@ -0,0 +1,244 @@
...
+   .NameASCII: The name stored after calling ToASCII() (with ACE prefix if any Unicode chars are present)
```
flokli, line 179: drop

documentation/design-idna.md
```diff
@@ -0,0 +1,244 @@
...
+   .NameUnicode: The name stored after calling ToUnicode()
```
flokli, line 180: `IDN` seems more fitting (Internationalized Domain Name)

documentation/design-idna.md
```diff
@@ -0,0 +1,244 @@
...
+   .NameDisplay: if .NameASCII != .NameUnicode, store as "ascii (unicode)"
```
flokli, line 181: This is confusing. Shouldn't this be the bigger slice that `Name` and `IDN` would be subslices of?

documentation/design-idna.md
```diff
@@ -0,0 +1,244 @@
...
+Proposed Outcome:
+
+1. Users should be able to input domains and labels in either ASCII (with ACE prefix if needed), Unicode, or a mix.  This holds for input via `dnsconfig.js` (domain names, labels, and targets); as well as flags such as `--domains`.
+2a. Output should be Unicode or both with the ASCII being in parenthesis. Example: `рф.com (xn--p1ai.com)`
+2b. Or maybe the reverse? Example: `xn--p1ai.com (рф.com)`
+3. DNSControl's main code should create a "paved path" for providers to make it easy for them to do the right thing. It should be easier to do the right thing than the wrong thing.  The default (i.e. "lazy") path should result in the behavior we desire.
+
+Here are some example outputs:
+
+NOTE: Feedback needed!  Do you prefer "a" or "b"?  Is there an even better format I should consider?  Should we use `{}` instead of `()`?
```
flokli, line 43: I prefer `b` and `()`. We should also specify `NameDisplay` only is a string with both variants if the domain is non-ASCII.

KaiSchwarz-cnic, line 43: `b`and `()`

documentation/design-idna.md
KaiSchwarz-cnic, line 179: yes, drop!

documentation/design-idna.md
KaiSchwarz-cnic, line 178: punycode variant in here ftw.

documentation/design-idna.md
KaiSchwarz-cnic, line 181: So, it will be either

* `mydomain.com` if it is not about an IDN and 
* `xn--p1ai.com (рф.com)` otherwise.

Sounds good to me.

documentation/design-idna.md
KaiSchwarz-cnic, line 180: While I enjoy playing with short names that make things clear - so `IDN` would be perfect, but from developer perspective, having that property to start also with `NameXXX` makes it probably better accessible via IDE as the `Name` properties appear right after each other.

* `.NameIDN` isn't applicable as it states "name" twice out
* `.NameUnicode` is ok though
* `.NameInternationalized` eventually as well

My favs are by that `.IDN` or `.NameUnicode` while `.NameUnicode` is still my personal pick. `Unicode` and `ASCII` keywords are often used whenever it comes to IDN translation libraries. While `IDN` and `Punycode` keywords are usually used on Domain/DNS Provider side. That's at least what I noticed over the years. Still, that shouldn't give a hint for making a decision.

KaiSchwarz-cnic:
Also, IDN isn't IDN if we compare .de and .com. Some TLD Providers support different IDNA Standards (IDNA2003 vs. IDNA2008, UTS46). Translating an IDN might by that end in a different punycode variant.

Let me provide some example in here from the HEXONET Provider's ConvertIDN API Command:

```
[COMMAND]
COMMAND = ConvertIDN
DOMAIN0 = ärzte.com
DOMAIN1 = ärzte.de
EOF

[RESPONSE]
CODE = 200
DESCRIPTION = Command completed successfully
PROPERTY[ACE][0] = xn--rzte-koa.com
PROPERTY[ACE][1] = xn--rzte-koa.de
PROPERTY[IDN][0] = ärzte.com
PROPERTY[IDN][1] = ärzte.de
EOF
```

No big difference in here. But let us pick one with german special characters:

```
[COMMAND]
COMMAND = ConvertIDN
DOMAIN0 = fußball.com
DOMAIN1 = fußball.de
EOF

[RESPONSE]
CODE = 200
DESCRIPTION = Command completed successfully
PROPERTY[ACE][0] = fussball.com
PROPERTY[ACE][1] = xn--fuball-cta.de
PROPERTY[IDN][0] = fussball.com
PROPERTY[IDN][1] = fußball.de
EOF
 ```

Let us ignore that `.com` is covering that differently and let us use the punycode variant returned for `.de` as a `.com` domain name ` xn--fuball-cta.com`. While the IDN translation is from technical perspective correct, it won't work together with the TLD Provider as of a different supported IDNA Standard. By that, I highly think that this needs to be considered as well when going for a IDNA proposal.

tlimoncelli:
https://github.com/StackExchange/dnscontrol/pull/3879 fixes the problems and, thanks for all of your suggestions, just works a lot better.

PR #2879, opened 2024-03-23
NEW FEATURE: Add RFC4183 support to REV()

tlimoncelli:
CC @misilot
Fixes https://github.com/StackExchange/dnscontrol/issues/2844
Fixes https://github.com/StackExchange/dnscontrol/issues/2845

PR #2887, opened 2024-03-27
DOCS: `REV()` and `REVCOMPAT()`

cafferata:
Documentation (style) fixes for the (open) GitHub pull request https://github.com/StackExchange/dnscontrol/pull/2879.

### DIFF
--- a/commands/types/dnscontrol.d.ts
+++ b/commands/types/dnscontrol.d.ts
@@ -2274,7 +2274,7 @@ declare function PANIC(message: string): never;
  * [`D_EXTEND()`](../global/D_EXTEND.md)
  * function to insert `A` and `PTR` records into previously-defined domains.
  *
- * ```
+ * ```javascript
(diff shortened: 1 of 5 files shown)


tlimoncelli (approved):
Thanks for cleaning up my mess!

question: How do you decide when to use `{% code title="dnsconfig.js" %}` instead of just using `'''javascript`?   (backticks changed to `'` because.... quoting)

cafferata:
> question: How do you decide when to use {% code title="dnsconfig.js" %} instead of just using '''javascript? (backticks changed to ' because.... quoting)

I would always say use `{% code title="xxxxxx.xx" %}` with the correct file name. Why? This gives me as a user a little more context between the different DNSControl files: `helpers.js`, `creds.json`, `dnsconfig.js` and the generated `types-dnscontrol.d.ts`.

cafferata:
> Thanks for cleaning up my mess!

You're welcome. ☺️

tlimoncelli (approved):

PR #2902, opened 2024-04-02
DOCS: Warn that get-certs will be removed without notice

tlimoncelli:
* Make the "get-certs" subcommand hidden.
* Warn that the code may be removed without notice.
* Unify the warning to list July 2025 as the max deadline.


cafferata (approved):

tlimoncelli:
CC @cafferata

PR #2907, opened 2024-04-10
DOCS: Disabling colors via environment variable

allixsenos:
dnscontrol uses fatih/color package which respects the NO_COLOR standard

https://github.com/fatih/color?tab=readme-ov-file#disableenable-color

https://no-color.org

This change touches only docs, no code, so no code testing was performed.


tlimoncelli:
Thanks!!

Issue #2843, opened 2024-02-19
BIND: SOA number doesn't change if zone.js is required


mraspor:
**Describe the bug**
If D() is in the main dnsconfig.js file, the SOA number for the BIND provider will get correctly created and updated (for example from 2024021800 to 2024021900).

If we require a file from the main dnsconfig.js
`require("./example.com.zone.js");`
the resulting .zone file will have its SOA number set to 1 and will not update.

**To Reproduce**
1. Create "zones" directory in the dnscontrol root directory
2. Create file example.com.zone.js in dnscontrol root directory with the following data
`D('example.com', REG_NONE, DnsProvider(DSP_BIND),
    DefaultTTL('5m'),
    NAMESERVER_TTL('1h'),
    A('@', '10.10.10.10'),
    A('localhost', '127.0.0.1')
);`
3. Edit dnsconfig.js and put the following at the end
`require("./example.com.zone.js");`
5. Execute `dnscontrol push`
6. Check the resulting file in zones/example.com.zone
7. Change something in the example.com.zone.js and execute `dnscontrol push`
8. Observe the SOA number haven't changed in the resulting zones/example.com.zone file 

**Expected behavior**
SOA number should change

**DNS Provider**
 - BIND

tlimoncelli:
Hey there!   Thanks for reporting this. 

When I try to reproduce this issue I see erratic behavior.  Sometimes the serial resets to the `...00` serial for the day (2024040300) and sometimes not.

I'll investigate further.

tlimoncelli:
Hey friend!

I've tried to reproduce the bug with no success.  Can you help me figure out what I'm doing differently?

```
==> creds.json <==
{}

==> dnsconfig.js <==
var REG_NONE = NewRegistrar("none");
var DSP_BIND = NewDnsProvider("bind");

require("./example.com.zone.js");

==> example.com.zone.js <==

D('example.com',
    REG_NONE,
    DnsProvider(DSP_BIND),
    DefaultTTL('5m'),
    NAMESERVER_TTL('1h'),
    A('@', '10.10.10.10'),
    A('localhost', '127.0.0.1'),
    A('test', '127.0.0.2'),
END);

==> zones/example.com.zone <==
$TTL 300
; generated with dnscontrol 2024-04-05T15:47:56-04:00
@                IN SOA   DEFAULT_NOT_SET. DEFAULT_NOT_SET. 2024040505 3600 600 604800 1440
                 IN A     10.10.10.10
[...]

PR #2275, opened 2023-04-13
BIND: Enable hermetic zone builds

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/1859

CC [w4](https://github.com/w4)

```
$ rm zones/example.com
$ dnscontrol --diff2 push --domains --bindserial=5444
$ cat zones/example.com
$TTL 14400
; generated with dnscontrol 2023-04-13T16:23:54-04:00
@          300   IN SOA   DEFAULT_NOT_SET. DEFAULT_NOT_SET. 5444 3600 600 604800 1440
...              
...
```

[w4](https://github.com/w4): Is this what you're looking for?

PR #2908, opened 2024-04-10
BIND: BUGFIX: SOA serial number doesn't change if SOA() used

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/2843

The BIND provider has a bug where if the `SOA()` record is explicitly defined, the serial number does not increment.  I have verified this bug with and without `require()`.

To replicate:

dnsconfig.js:

```javascript
var DSP_BIND = NewDnsProvider("bind");

D('example.com', REG_NONE, DnsProvider(DSP_BIND),
  SOA("@", "ns1.example.com.", "hostmaster.example.com.", 3600, 601, 604800, 1440),
  DefaultTTL('5m'),
  NAMESERVER_TTL('1h'),
  A('@', '10.10.10.10'),
END);
```

If you remove the SOA() line, the serial number increments just fine.

The problem is that if a SOA() exists in "desired" (i..e. the dnsconfig.js definition of the zone), the code sets `c.skipNextSoaIncrease = true` and the serial number is not incremented.  I can't see any reason what that code was intended to do in the first place.  In fact, I've removed the skipNextSoaIncrease field and everything related to it and all my tests still work.

CC [w4](https://github.com/w4) since the change touches the feature added in https://github.com/StackExchange/dnscontrol/pull/2275.  Please verify this doesn't break your hermetic builds.


tlimoncelli:
- CC [w4](https://github.com/w4) for visibility
- CC [mraspor](https://github.com/mraspor) for confirmation this fixes the bug you reported

PR #2926, opened 2024-05-01
Enable support for SVCB/HTTPS in GCLOUD and HEDNS

tlimoncelli:


tlimoncelli:
@fritterhoff Ah!  I found the problem. The tests need to have a `requires(providers.CanUseHTTPS)` so that only providers that support that record type will run them.

I also ran "go generate"

fritterhoff:
> @fritterhoff Ah! I found the problem. The tests need to have a `requires(providers.CanUseHTTPS)` so that only providers that support that record type will run them.
> 
> I also ran "go generate"

Thanks for digging into the issue. 

Maybe it makes sense to add the go generate command to the docs? I did not see any task telling me to run that command...

tlimoncelli:
Good point about the docs. Will do!

PR #2941, opened 2024-05-07
NS1: update ns1-go to 2.10.0

### LINKED ISSUE #2639: NSONE: Value None for field '<obj>.tags' is not of type object

costasd:
Fixes #2639

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,7 +55,7 @@ require (
 	golang.org/x/net v0.24.0
 	golang.org/x/oauth2 v0.19.0
 	google.golang.org/api v0.177.0
-	gopkg.in/ns1/ns1-go.v2 v2.7.11
+	gopkg.in/ns1/ns1-go.v2 v2.10.0
 )
 
 require (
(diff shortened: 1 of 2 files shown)


tlimoncelli:
Good things come to those who wait!  (...after filing bug reports with NS1 :-) )

Thanks for taking care of this. It's good to know we can use the latest library now!

PR #2942, opened 2024-05-08
DESEC: Enable support for SVCB/HTTPS

Meroje:
Listed here https://desec.readthedocs.io/en/latest/dns/rrsets.html#supported-types and confirmed on my account that get-zones / preview is clear.

(#1600)


tlimoncelli (approved):

tlimoncelli:
Thanks!

PR #2951, opened 2024-05-13
NS1: add support for DNAME and DHCID record types

costasd:
This changeset enables support for DNAME and DHCID record types in NS1.


tlimoncelli:
Lookin' good!

Please run `go generate ./...` and then check in the `documentation/providers.md` file that it generates.

costasd:
meh always forgetting a thing, thanks!

Issue #2955, opened 2024-05-14
Autocomplete does not work in ZSH when installed as a file


Maetveis:
**Describe the bug**
Autocomplete does not work if the completions are installed as file, instead of directly sourced into the shell.

Adding the output of `dnscontrol shell-completion zsh` to a path that is in `$fpath` as described by
[The Z Shell Manual - 20.2.2 Completion System Autoloaded functions](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Autoloaded-files) should enable shell completions to work, but it does not.

**To Reproduce**

(Tested on Ubuntu 22.04, zsh installed from apt or homebrew).

```zsh
# 1. Write autocompletion script to file
mkdir -p /tmp/dnscontrol
dnscontrol shell-completion zsh >! /tmp/dnscontrol/_dnscontrol

# 2. Add to fpath (by modifiying zshrc)
cp ~/.zshrc ~/.zshrc.bak
echo 'fpath=($fpath[@] /tmp/dnscontrol)' | cat - ~/.zshrc.bak >! ~/.zshrc

# 3. remove cached completions (if using prezto)
rm -r ~/.cache/prezto

# 4. Restart zsh
exec zsh
```

5. Try out completions by hitting TAB after typing `dnscontrol`

6. Undo changes to `.zshrc`: `mv ~/.zshrc.bak ~/.zshrc`.

**Expected behavior**

Shell completions should work: the sub-commands of `dnscontrol` should be completed. For example check or create-domains should be suggested. The completions shoulnd't be the default of zsh, for me when it suggests filenames.

**DNS Provider**
Not relevant.

**Additional context**

Homebrew packages dnscontrol zsh completions like this too, therefore completions don't work out of the box there.

PR #2956, opened 2024-05-14
BUG: ZSH Autocomplete does not work if installed as a file #2955

Maetveis:
Fixes: #2955

The #compdef directive shouldn't have double-quotes. (https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Autoloaded-files)

We need to execute the function the first time when its loaded as described by https://zsh.sourceforge.io/Doc/Release/Functions.html#Autoloading-Functions

> [If the KSH_AUTOLOAD option is not set] the function body
(with no surrounding `funcname() {...}``) is taken to be the complete contents of the file.
This form allows the file to be used directly as an executable shell script.
>
> If processing of the file results in the function being re-defined, the function itself is not re-executed. To force the shell to perform initialization and then call the function defined, the file should contain initialization code (which will be executed then discarded) in addition to a complete function definition (which will be retained for subsequent calls to the function), and a call to the shell function, including any arguments, at the end.

To keep direct sourcing possible, `zsh_eval_context` is used (https://zsh.sourceforge.io/Doc/Release/Parameters.html#Parameters-Set-By-The-Shell) to detect wether the script is being auto-loaded.

### DIFF
--- a/commands/completion-scripts/completion.zsh.gotmpl
+++ b/commands/completion-scripts/completion.zsh.gotmpl
@@ -1,4 +1,4 @@
-#compdef "{{.App.Name}}"
+#compdef {{.App.Name}}
 
 _dnscontrol() {
   local -a opts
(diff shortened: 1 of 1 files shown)


tlimoncelli:
CC @whi-tw

whi-tw (approved):
LGTM! Good catch, I've not spotted the behaviour myself, but that is definitely what the spec says!

Seems to work fine on my end :)

tlimoncelli:
Thanks!  All the zsh users thank you too!

PR #2960, opened 2024-05-16
DNSIMPLE: enable Null MX records

philpennock:
DNSimple sent out an announcement 2024-05-16 that they now support Null MX records.

### DIFF
--- a/providers/dnsimple/auditrecords.go
+++ b/providers/dnsimple/auditrecords.go
@@ -11,7 +11,7 @@ import (
 func AuditRecords(records []*models.RecordConfig) []error {
 	a := rejectif.Auditor{}
 
-	a.Add("MX", rejectif.MxNull) // Last verified 2023-03
+	// DNSimple sent email 2024-05-16 announcing that they now support Null MX records.
 
 	a.Add("TXT", rejectif.TxtLongerThan(1000)) // Last verified 2023-12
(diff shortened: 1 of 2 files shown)


providers/dnsimple/auditrecords.go
```diff
@@ -11,7 +11,7 @@ import (
 func AuditRecords(records []*models.RecordConfig) []error {
 	a := rejectif.Auditor{}
 
-	a.Add("MX", rejectif.MxNull) // Last verified 2023-03
+	// DNSimple sent email 2024-05-16 announcing that they now support Null MX records.
```
cafferata, line 14: If you ask me, I would personally leave this comment out. And just delete the original line. 

```suggestion
```

philpennock: I think that when things change, and people are swearing they thought X was the case, a comment providing a clue helps.  Changed in observed behavior, and when, are ideal use-cases for comments explaining both what is present and what is deliberately not present.

tlimoncelli: Delete the line and trust that the PR description is documentation enough. 

If we were taking away a feature, more documentation would be warranted. However I've never seen anyone complain about losing restrictions.

cafferata (approved):

PR #2966, opened 2024-05-20
AXFRDDNS: Canonicalize key name as per RFC 2845 3.4.2

nemunaire:
Hi,

In the AXFRDDNS provider, it appears we can use a key like: `hmac-sha256:MyTsIgKey:Base64EncodedSecret=`

But the key name (here `MyTsIgKey`) is not canonicalized, however it is required by the [RFC 2845](https://datatracker.ietf.org/doc/html/rfc2845#page-7).

This PR calls `CanonicalName`, which also adds a leading `.` if it is missing.


tlimoncelli:
CC @hnrgrgr for approval

tlimoncelli:
friendly ping @hnrgrgr ?

PR #2971, opened 2024-05-21
DOCS: [get-zones] Removed old reference to `convertzone`

cafferata:
Removed the old reference to `convertzone` within the '[get-zones](https://docs.dnscontrol.org/commands/get-zones)' page.


cafferata:
`convertzone` is not mentioned anywhere else.

_https://docs.dnscontrol.org/?q=convertzone_

PR #2972, opened 2024-05-21
DOCS: Several improvements

cafferata:
An older git branch with more (random) documentation improvements.


documentation/why-the-dot.md
```diff
@@ -46,7 +46,7 @@ How are they ambiguous?
...
-  * What is the CNAME target is "www.bar.com" and the domain is "bar.com".  Then It is reasonable to infer the user's intent, right?  "www.bar.com.bar.com." would be silly, right?  Maybe. What if we are copying 100 lines of dnsconfig.js from one `D()` to another. Buried in the middle is this one CNAME that means something entirely different when in a new $DOMAIN. That would be bad.  We've seen this in production and want to prevent this kind of error.
+  * What is the CNAME target is "www.bar.com" and the domain is "bar.com".  Then It is reasonable to infer the user's intent, right?  "www.bar.com.bar.com." would be silly, right?  Maybe. What if we are copying 100 lines of `dnsconfig.js` from one `D()` to another. Buried in the middle is this one CNAME that means something entirely different when in a new $DOMAIN. That would be bad.  We've seen this in production and want to prevent this kind of error.
```
tlimoncelli, line 49:
```suggestion
  * What if the CNAME target is "www.bar.com" and the domain is "bar.com"?  Then It is reasonable to infer the user's intent, right?  `www.bar.com.bar.com.` would be silly, right?  Maybe. What if we are copying 100 lines of `dnsconfig.js` from one `D()` to another. Buried in the middle is this one CNAME that means something entirely different when in a new $DOMAIN. That would be bad.  We've seen this in production and want to prevent this kind of error.
```

tlimoncelli:
Thanks!

PR #2983, opened 2024-05-30
GCORE: add SVCB and HTTPS support

xddxdd:
Add SVCB and HTTPS support for GCore provider. Ran integration test and tested with own config.

### DIFF
--- a/providers/gcore/convert.go
+++ b/providers/gcore/convert.go
@@ -45,6 +45,11 @@ func nativeToRecords(n gcoreRRSetExtended, zoneName string) ([]*models.RecordCon
 				return nil, fmt.Errorf("unparsable record received from G-Core: %w", err)
 			}
 
+		case "SCVB": // GCore mistypes "SVCB" as "SCVB"
+			if err := rc.PopulateFromString("SVCB", value.ContentToString(), zoneName); err != nil {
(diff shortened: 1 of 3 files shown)


cafferata (approved):

providers/gcore/convert.go
```diff
@@ -45,6 +45,11 @@ func nativeToRecords(n gcoreRRSetExtended, zoneName string) ([]*models.RecordCon
 				return nil, fmt.Errorf("unparsable record received from G-Core: %w", err)
 			}
 
+		case "SCVB": // GCore mistypes "SVCB" as "SCVB"
```
tlimoncelli, line 48: Maybe support both in case they correct their typo?

```suggestion
		case "SCVB", "SVCB": // GCore mistypes "SVCB" as "SCVB"
```

xddxdd: If GCore fixes their typo, it will be handled by the default case, which has the exact same logic as the `SCVB` case, including the type name (both passing `SVCB`).

tlimoncelli: Oh!  Of course!  Sorry, I didn't look too closely.

tlimoncelli (approved):

PR #2954, opened 2024-05-13
AKAMAIEDGEDNS: new maintainer: Ed Lynes

svernick:
The new AKAMAIEDGEDNS maintainer is Ed Lynes (edglynes).

PR #2940, opened 2024-05-07
AKAMAIEDGEDNS: new maintainer: Chris Dornin

svernick:
AKAMAIEDGEDNS has a new maintainer: Chris Dornin (cdornin)

PR #2427, opened 2023-06-08
INWX: Update providers.md maintainer

patschi:
As requested in https://github.com/StackExchange/dnscontrol/issues/1668#issuecomment-1579739374

PR #2997, opened 2024-06-08
BUILD: Generate OWNERS files

cafferata:
`go generate ./...` will now also generate the `OWNERS` file.

### DIFF
--- a/build/generate/ownersFile.go
+++ b/build/generate/ownersFile.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"github.com/StackExchange/dnscontrol/v4/providers"
+	"os"
+	"sort"
+	"strings"
+)
+
+func generateOwnersFile() error {
+	maintainers := providers.ProviderMaintainers
+	sortedProviderNames := getSortedProviderNames(maintainers)
+
+	var ownersData strings.Builder
+	for _, providerName := range sortedProviderNames {
+		providerMaintainer := maintainers[providerName]
+		if providerMaintainer == "NEEDS VOLUNTEER" {
+			ownersData.WriteString("# ")
+		}
+		ownersData.WriteString("providers/")
+		ownersData.WriteString(getProviderDirectory(providerName))
+		ownersData.WriteString(" ")
+		ownersData.WriteString(providerMaintainer)
+		ownersData.WriteString("\n")
+	}
+
+	file, err := os.Create("OWNERS")
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	_, err = file.WriteString(ownersData.String())
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getProviderDirectory(providerName string) string {
+	// Strip the underscores from the providerName constants
+	providerDirectory := strings.ToLower(
+		strings.ReplaceAll(
+			providerName, "_", "",
+		),
+	)
+
+	// These providers use a different directory name
+	if providerDirectory == "cloudflareapi" {
+		providerDirectory = "cloudflare"
+	}
+	if providerDirectory == "dnsoverhttps" {
+		providerDirectory = "doh"
+	}
+
+	return providerDirectory
+}
+
+func getSortedProviderNames(maintainers map[string]string) []string {
+	providerNameSorted := make([]string, 0, len(maintainers))
+	for providerNameKey := range maintainers {
+		providerNameSorted = append(providerNameSorted, providerNameKey)
+	}
+	sort.Strings(providerNameSorted)
+
+	return providerNameSorted
+}
--- a/providers/providers.go
+++ b/providers/providers.go
@@ -72,6 +72,15 @@ func RegisterDomainServiceProviderType(name string, fns DspFuncs, pm ...Provider
 	unwrapProviderCapabilities(name, pm)
 }
 
+var ProviderMaintainers = map[string]string{}
+
+func RegisterMaintainer(
+	providerName string,
+	gitHubUsername string,
+) {
+	ProviderMaintainers[providerName] = gitHubUsername
+}
+
 // CreateRegistrar initializes a registrar instance from given credentials.
 func CreateRegistrar(rType string, config map[string]string) (Registrar, error) {
 	var err error
--- a/build/generate/generate.go
+++ b/build/generate/generate.go
@@ -6,6 +6,9 @@ func main() {
 	if err := generateFeatureMatrix(); err != nil {
 		log.Fatal(err)
 	}
+	if err := generateOwnersFile(); err != nil {
+		log.Fatal(err)
+	}
 	funcs, err := generateFunctionTypes()
 	if err != nil {
 		log.Fatal(err)
--- a/documentation/writing-providers.md
+++ b/documentation/writing-providers.md
@@ -224,8 +224,6 @@ an automated way to test for this bug.  The manual steps are here in
   * Add the provider to the "Providers" list.
 * Create `documentation/provider/PROVIDERNAME.md`:
   * Use one of the other files in that directory as a base.
-* Edit `OWNERS`:
-  * Add the directory name and your GitHub username.
 
 {% hint style="success" %}
 **Need feedback?** Submit a draft PR!  It's a great way to get early feedback, ask about fixing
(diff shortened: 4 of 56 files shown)


tlimoncelli:
I have some random thoughts about this.  I don't have a final decision but here are my thoughts...


I agree that OWNERS getting out of date is a problem and some automation would greatly help.  I appreciate you taking the time to promote a solution.

I'm not sure how I feel about moving the souce-of-truth into the source code.

On one hand, it has benefits: OWNERS gets updated automatically with the creation and deletion of providers.  On the other hand, it embeds people's Github account names in the source code, and therefore in the binary.

An alternative would be to put this data in a per-provider yaml file (maybe `$git/providers/PROVIDERNAME/meta.yaml` ?).  The generator would have more work to do, but it isn't a lot more work.

A much more simple design would be to simply have the the CI/CD system cross-check that any subdirectory in $git/providers is mentioned in OWNERS and out if something is missing. This adds no new systems, datastructures, etc.  However it kicks the can down the road as far as other things we could do with ownership data.

The bottom line: I would like a solution that is as big as the problem. Is this a big problem?  I think we've had only 1-2 typos in OWNERS in the last year.

tlimoncelli:
How about this instead?  It's more simple and keeps the OWNERS file the source of truth?
 https://github.com/StackExchange/dnscontrol/pull/3005

cafferata:
> I'm not sure how I feel about moving the souce-of-truth into the source code.

Personally, I am a great fan of single source of truth (SSOT). Preferably in the source code. Well, that is of course a personal preference.

> On the other hand, it embeds people's Github account names in the source code, and therefore in the binary.

I understand what you're saying, what would you say is the objection here?

> The bottom line: I would like a solution that is as big as the problem. Is this a big problem? I think we've had only 1-2 typos in OWNERS in the last year.

I would say, no, it's certainly not a big problem. It was one of the ideas from https://github.com/StackExchange/dnscontrol/issues/1998 (_and also a fun writing GoLang experiment for me_). I wasn't really concerned about typo either. These are issues such as maintainer changes:
1. https://github.com/StackExchange/dnscontrol/pull/2954
2. https://github.com/StackExchange/dnscontrol/pull/2940
3. https://github.com/StackExchange/dnscontrol/pull/2427
4. https://github.com/StackExchange/dnscontrol/pull/1692
5. https://github.com/StackExchange/dnscontrol/pull/1403

Where one or the other place is forgotten. By transferring everything to the source code, the [providers page](https://docs.dnscontrol.org/getting-started/providers) below can ultimately be generated too.

> How about this instead? It's more simple and keeps the OWNERS file the source of truth?

Is this simpler? I am not a bash guru myself. This means a new technology on top of the existing GoLang within DNSControl CI/CD and/of GitHub workflows.

I am curious about your explanation and whether you would like to continue with this GitHub pull request or whether we choose to close this GitHub pull request. No hard feelings at all! ♥️

tlimoncelli:
> > I'm not sure how I feel about moving the souce-of-truth into the source code.
> 
> Personally, I am a great fan of single source of truth (SSOT). Preferably in the source code. Well, that is of course a personal preference.

That's what makes this so difficult (for me).  I wasn't sure.

After some thought, I realize it is a personal preference until we have a solid technical reason.

So I starting thinking about the technical benefits.

My conclusion is that we should adopt a goal of having each provider be self-contained in a single place (`$git/providers/name`).  A PR that creates a new provider should only have to add files in `$git/providers/name`, and deleting a provider should be as simple as removing `$git/providers/name` (plus removing it from `$git/providers/_all`)

If we agree to this goal, this PR moves us towards it.

By the way... other things we should correct:
* Documentation for provider-specific rtypes
* Small bits of code in models/* that are provider-specific
* Integration tests that are provider-specific (this is more difficult, and might not be worthwhile?)

> I am curious about your explanation and whether you would like to continue with this GitHub pull request or whether we choose to close this GitHub pull request. No hard feelings at all! ♥️

Yes, let's keep this PR and close the other one.  I'll review the code soon.

Best
Tom

providers/providers.go
```diff
@@ -72,6 +72,15 @@ func RegisterDomainServiceProviderType(name string, fns DspFuncs, pm ...Provider
 	unwrapProviderCapabilities(name, pm)
 }
 
+var ProviderMaintainers = map[string]string{}
+
+func RegisterMaintainer(
```
tlimoncelli, line 77: What happens if a provider forgets to call this function?  We're back in the same situation where OWNERS gets out of date.

Instead of a function, how about adding the gitHubUsername as a parameter to `providers.RegisterRegistrarType` and `providers.RegisterDomainServiceProviderType`?

Benefits:
* An author can't forget to add it.
* "Volunteer needed" could be represented by `nil` or `""` (for strings, they are equiv).  `generateOwnersFile()` can generate the string "VOLUNTEER NEEDED" if the field is "".

tlimoncelli:
I forgot about this.  Oops!

Let's merge this.  Confirm?

cafferata:
> Let's merge this. Confirm?

Tomorrow/the day after tomorrow I will update this GitHub pull request and submit it so that it can be merged.

- [ ] Rebase on main
- [ ] https://github.com/StackExchange/dnscontrol/pull/2997#discussion_r1636535963
- [ ] https://github.com/StackExchange/dnscontrol/pull/2997#discussion_r1636743815
- [ ] https://github.com/StackExchange/dnscontrol/pull/2997/conflicts

build/generate/ownersFile.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	return providerDirectory
+}
+
+func getSortedProviderNames(maintainers map[string]string) []string {
+	providerNameSorted := make([]string, 0, len(maintainers))
+	for providerNameKey := range maintainers {
+		providerNameSorted = append(providerNameSorted, providerNameKey)
+	}
```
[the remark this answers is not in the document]
cafferata: This is a Go `1.23.x` featute, but we are currently still using `1.22.1`:
 https://github.com/StackExchange/dnscontrol/blob/0ed31efdc9276d423e7832813255ada467816d8e/go.mod#L3

providers/providers.go
cafferata: > Instead of a function, how about adding the gitHubUsername as a parameter to `providers.RegisterRegistrarType` and `providers.RegisterDomainServiceProviderType`?

This was my original implementation. However, this also has disadvantages. You must maintain double administration for each provider that has a `RegisterRegistrarType()` and a `RegisterDomainServiceProviderType()`.

**Separate function `RegisterMaintainer()`**

```diff
providers.RegisterDomainServiceProviderType(providerName, fns, features)
providers.RegisterRegistrarType(providerName, newRoute53Reg)
+providers.RegisterMaintainer(providerName, providerMaintainer)
```

**Parameter to `providers.RegisterRegistrarType` and `providers.RegisterDomainServiceProviderType`**

```diff
-providers.RegisterDomainServiceProviderType(providerName, fns, features)
+providers.RegisterDomainServiceProviderType(providerName, providerMaintainer, fns, features)
-providers.RegisterRegistrarType(providerName, newRoute53Reg)
+providers.RegisterRegistrarType(providerName, providerMaintainer, newRoute53Reg)
```

Personally, I find the separate `RegisterMaintainer()` function clearer/more organized.

cafferata:
The git rebase has taken place. I made one additional commit to define the Huawei Cloud provider maintainer: https://github.com/StackExchange/dnscontrol/pull/2997/commits/a634f78abe98128bc79aac5f53deb393eea0bdf6.

Ready to (re)review.

providers/providers.go
tlimoncelli: Good point!

Ok, here's a more radical solution 

An more "modern golang" pattern is to pass a struct with a number of options:

```
providers.Register(providers.RegisterOptions{
     Name: "POWERDNS",
     Maintainer: "@jpbede",
     Features: features,
     RegistrarInitializer: newRoute53Reg,
     DnsInitializer: newDSP,
     AuditRecords: AuditRecords,
})
```

If something is not a registrar, leave out RegistrarInitializer; if it is not a dns provider, leave out DnsInitializer.

This could be in a separate PR. It seems unrelated to generating the OWNERS file.

tlimoncelli (approved):
All good. Let me know if you're ready to merge.

cafferata:
Ready! 🥸

Issue #2313, opened 2023-05-03
CLOUDFLAREAPI: FEATURE REQUEST: Support Cloudflare Redirects (via Product Rules instead of Page Rules)


pro-sumer:
It's great that DNSControl [supports redirects via Cloudflare Page Rules](https://docs.dnscontrol.org/service-providers/providers/cloudflareapi) using `CF_TEMP_REDIRECT` and `CF_REDIRECT`!

Unfortunately we only get 3 rules on free plans. Cloudflare recently introduced [Redirects via Product Rules](https://developers.cloudflare.com/rules/url-forwarding/) (currently in beta) with more relaxed limits.

It would be great if you can support these new redirects as well.

tlimoncelli:
Sounds like a useful thing to support. Can you suggest a name (like CF_BULK_REDIRECT()) and a syntax?

hmoffatt:
The bulk redirects are account-wide, not per domain. You need a business or enterprise subscription to use wildcards.

The single redirects look like a useful alternative to page rules here though as they're per domain. You can match on a lot more than source URL though so the syntax would be complicated; you'd need to be able to specify a whole match expression (eg `(http.request.uri.path eq "/*" and ip.geoip.country eq "AF")`) and the target.

aglasson:
I believe the below image is what OP is talking about. These seem to have been intended to supplanted page rules while page rules have retained their ability to do URL redirects.

My suggestion, without fully understanding the potential impacts or underlying Cloudflare intentions, is to change both `CF_TEMP_REDIRECT` and `CF_REDIRECT` to manage the new rule types instead. Obviously there needs to be some conversion handling process or manual intervention.
A hypothetical conversion approach might be:
<pre>CF_TEMP_REDIRECT("foo.bar.co/*", "https://foofoo.bar.co/", <b>{'cloudflare_ruletype_redirect': 'true'}</b>)`</pre>


> You can match on a lot more than source URL though so the syntax would be complicated  

While the new rules have a lot more functionality I don't see any reason why DNSControl needs to make use of any of the additional rule functionality, stick to doing only what it does now. I don't think DNSControl has ever intended to be a complete Cloudflare management system, if a cloudflare user wants more than simple redirects, manage those outside of DNSControl.

![image](https://github.com/StackExchange/dnscontrol/assets/11340849/7e868837-5dd1-4f8f-ada3-921ee66f6b64)

tlimoncelli:
[...]

PR #3002, opened 2024-06-11
CLOUDFLAREAPI: CF_REDIRECT/CF_TEMP_REDIRECT should dtrt using Single Redirects

tlimoncelli:
Relates to https://github.com/StackExchange/dnscontrol/issues/2313


tlimoncelli:
CC @tresni for visibility

At this point I'm just working out a general idea of how it would work.  It does not yet use the API to download Single Redirects.

PR #3046, opened 2024-07-10
DOCS: Fix markdown links in Cloudflare provider docs

svenluijten:
This PR fixes 2 links in the Cloudflare API provider documentation. These were incorrectly formatted previously, and would not be rendered as links on the website.

- https://docs.dnscontrol.org/provider/cloudflareapi#old-style-vs-new-style-redirects


tlimoncelli:
Thanks for catching that!

PR #3071, opened 2024-08-05
BUNNY_DNS: run goimports on convert.go

tlimoncelli:
Format this with "goimports" instead of "gofmt" so that the imports are consistent with the rest of the code.

CC @ppmathis


ppmathis:
@tlimoncelli Thanks! 😄 I happened to notice the same recently and would have bundled it along with another upcoming PR of mine for fixing the Null-MX record handling, as I noticed that is broken right now. I'll rebase and send along a PR with just the bugfix in the near future.

PR #3072, opened 2024-08-05
PORKBUN: Remove unused function

tlimoncelli:
CC: @imlonghao

staticcheck reports:
```
api.go:103:27: func (*porkbunProvider).ping is unused (U1000)
```

Also fixed the import order by using "goimports" instead of "gofmt".  This is more consistent with other code.


imlonghao:
How about just remove the unused function?

tlimoncelli:
I've removed it.  Can you run the integration tests to verify that everything still works?

Thanks!

imlonghao:
Integration tests LGTM

imlonghao (approved):

tlimoncelli:
Thanks!

PR #3080, opened 2024-08-13
POWERDNS: Use canonical domain in calls to PowerDNS

kytta:
In PowerDNS' `GetZoneRecords()` as well as the two methods in `diff.go`, the domain name is not canonicalized, like elsewhere in the module. For me, this results in an error when querying and/or updating my zones. This PR fixes that


tlimoncelli:
CC @jpbede

Thanks for the PR!

I'm surprised that the integration tests work with and without this change.  ...but they do!

Is there a test we could add that would detect this bug?

kytta:
> I'm surprised that the integration tests work with and without this change.  ...but they do!

I was also surprised! [According to the docs](https://doc.powerdns.com/authoritative/http-api/zone.html?highlight=trailing%20dot#objects), the zones _must_ have a trailing dot 🤔

What version of PowerDNS are you testing against? Can it be that the behaviour changed in a major release?

tlimoncelli:
@jpbede: What version is that server?

jpbede:
It is running PDNS 4.4.0

kytta:
Okay, I might have some ideas where this is coming from, which also makes this PR rather invalid. What I didn't know is that there are two ways to identify a zone in PowerDNS:`id` and `name`.

`name` is the canonical domain name, the one with the trailing dot. It should be used when creating a zone, and it's the single point of truth for what the domain should be resolved under.

`id` is a string that is "assigned by the server" and "should not be interpreted by the application". It gets returned upon creating a zone, and, in my tests, it still has a trailing dot. But, apparently, one shouldn't rely on it matching the domain name.

As such, my canonicalization of the domain name in the requests is wrong, because I assumed those are `name`s, not `id`s. But, using the code of today is also semi-wring, because there is seemingly no guarantee that the domain name is the zone ID.

(not that I am talking about the declared protocol here, not the actual PDNS behaviour)

kytta:
> `id` is a string that is "assigned by the server" and "should not be interpreted by the application". It gets returned upon creating a zone, and, in my tests, it still has a trailing dot. But, apparently, one shouldn't rely on it matching the domain name.

I misunderstood it, apparently. The only difference between zone ID and zone name is the fact that zone ID only has ASCII characters, whereas zone name can be anything: https://github.com/PowerDNS/pdns/blob/auth-4.4.0/pdns/ws-api.cc#L337-L365

Matter of fact, the trailing dot is always added there.

When querying the zones, the ID gets unparsed back to the domain name. If it didn't have `=` signs in it, it is basically unmodified: https://github.com/PowerDNS/pdns/blob/auth-4.4.0/pdns/ws-api.cc#L290-L335. So, if there was no trailing dot, it won't be added. And, if there is no trailing dot, the `DNSName` object cannot be constructed: https://github.com/PowerDNS/pdns/blob/auth-4.4.0/pdns/dnsname.hh#L59

**TL;DR:** I have become even more confused as to why the tests work without canonicalizing the name :D

@tlimoncelli is there a way to run the integration tests for one provider locally?

tlimoncelli:
Yup!  Integration test instructions are here: https://docs.dnscontrol.org/developer-info/integration-tests

kytta:
Okay, I want to apologize for a confusion I've had regarding this. 

The thing is, I am not using PowerDNS directly, but my requests are proxied. The proxy in question followed the PowerDNS docs (as well as the standards), and it did not allow domain names without the trailing dot.

After having tested this with PowerDNS directly, we found out that it, despite what its docs say, will always normalize the domain name to include the trailing dot. This is why the integration tests pass either way.

So, the proxy was disallowing requests which were actually valid. The proxy authors fixed this issue, and now, for my personal usage, I do not need this PR to be merged.

However, I would still want this PR to be merged, just because the changed code now matches the advertised behaviour of PowerDNS as well as the RFCs behind DNS. You are free to close this, though, as it apparently is not a bug :)

tlimoncelli:
I'm in favor of merging this.  However it is up to @jpbede (the owner of this provider).  What's your opinion?

jpbede (approved):
Yes, I'm in favor of merging this too. The code looks good to me.

tlimoncelli:
I love it when a plan comes together!

PR #3084, opened 2024-08-15
HUAWEICLOUD: Document DNSSEC capability and improve docs, reliability

huihuimoe:
This pr is intended to make 3 changes
- Huawei Cloud supports DNSSEC but doesn't provide a corresponding API, so I've labelled it as Unimplemented.
- Huawei Cloud has changed the link to the document, so I've updated it here.
- Fixed a potential de-duplication bug


tlimoncelli:
Thanks!

Issue #3092, opened 2024-08-22
Update Oracle SDK


fabienmazieres:
**Describe the bug**

Oracle SDK in use is version 32, released 12 Jan 2021.
Current version available is 65.72.0 hence I propose to update the SDK version to latest available.

Also, new version is capable of handling retry within the SDK, reducing the need for custom code to handle retry on HTTP 429.

**To Reproduce**

N/A

**Expected behavior**

N/A

**DNS Provider**

Oracle

**Additional context**

N/A

PR #3094, opened 2024-08-23
Oracle: Use SDK v65 and default retry policy

fabienmazieres:
- Update Oracle SDK to v65
- Use HTTP error handling and retry from SDK

See #3092


tlimoncelli (approved):

tlimoncelli:
I'm not sure why "go generate" output that error. When I do it on my system I didn't get an error, and no files were updated.  I wouldn't worry about it.

tlimoncelli:
Thanks!  It's been a while since we've had this provider so up-to-date!  The community appreciates it!

PR #3098, opened 2024-08-28
AUTODNS:  Enable CanUsePTR

christianbur:
AUTODNS: Allow to add PTR records


tlimoncelli:
Thanks!

CC @arnoschoon for visibility.

PR #3105, opened 2024-09-03
POWERDNS: Improve preformance by using new batched add/remove functions in 0.6.6 API

Veratil:
The go-powerdns API has been updated to allow batching `zones.ResourceRecordSet`s, greatly speeding up the ability to add/remove records. This PR updates PowerDNS provider to batch the changes into two sets: changed and deleted (ignoring the `diff2.REPORT` type that is). The changed set can include add/modify/delete whereas the delete set only includes full record set deletes.

Also, I don't know of any limits to PowerDNS' API itself, so none have been put in place (comparing to Route53's batching). It might be useful in the future to add the ability to limit, but as of now I didn't want to try.

And finally, this does bump the go-powerdns to 0.6.6. I saw there was another PR opened by dependabot for 0.6.5 which only added the add record batching. 0.6.6 adds the remove record batching.

---
Example of the batched changes with a locally running powerdns docker container. Unfortunately I do know it still says `1 corrections` and I don't know how that would be adjusted to accurately tell how many changes there actually are, or if that is actually something that is important.
```
var PROV_PDNS = NewDnsProvider("powerdns");

var RRSETS = [];

var alphabet = ["a","b","c","d","e","f","g","h","i","j","k","l","m","n","o","p","q","r","s","t","u","v","w","x","y","z"];
var offset = 0;
while (alphabet[offset] != 'z') {
        RRSETS.push(A(alphabet[offset], "127.0.0." + (offset + 1).toString()));
        offset += 1;
};

D("test2.internal", REG_NONE, DnsProvider(PROV_PDNS),
        RRSETS,
END);
```
with batching:
```
$ dnscontrol preview
******************** Domain: test2.internal
1 correction (powerdns)
#1:
+ CREATE a.test2.internal A 127.0.0.1 ttl=300
+ CREATE b.test2.internal A 127.0.0.2 ttl=300
+ CREATE c.test2.internal A 127.0.0.3 ttl=300
+ CREATE d.test2.internal A 127.0.0.4 ttl=300
+ CREATE e.test2.internal A 127.0.0.5 ttl=300
+ CREATE f.test2.internal A 127.0.0.6 ttl=300
+ CREATE g.test2.internal A 127.0.0.7 ttl=300
+ CREATE h.test2.internal A 127.0.0.8 ttl=300
+ CREATE i.test2.internal A 127.0.0.9 ttl=300
+ CREATE j.test2.internal A 127.0.0.10 ttl=300
+ CREATE k.test2.internal A 127.0.0.11 ttl=300
+ CREATE l.test2.internal A 127.0.0.12 ttl=300
+ CREATE m.test2.internal A 127.0.0.13 ttl=300
+ CREATE n.test2.internal A 127.0.0.14 ttl=300
+ CREATE o.test2.internal A 127.0.0.15 ttl=300
+ CREATE p.test2.internal A 127.0.0.16 ttl=300
+ CREATE q.test2.internal A 127.0.0.17 ttl=300
+ CREATE r.test2.internal A 127.0.0.18 ttl=300
+ CREATE s.test2.internal A 127.0.0.19 ttl=300
+ CREATE t.test2.internal A 127.0.0.20 ttl=300
+ CREATE u.test2.internal A 127.0.0.21 ttl=300
+ CREATE v.test2.internal A 127.0.0.22 ttl=300
+ CREATE w.test2.internal A 127.0.0.23 ttl=300
+ CREATE x.test2.internal A 127.0.0.24 ttl=300
+ CREATE y.test2.internal A 127.0.0.25 ttl=300
WARNING: No nameservers declared; skipping registrar. Add {no_ns:'true'} to force.
Done. 1 corrections.
```
```
$ time dnscontrol push
******************** Domain: test2.internal
1 correction (powerdns)
--snip--
Done. 1 corrections.

real    0m0.126s
user    0m0.021s
sys     0m0.041s
```
without batching:
```
$ dnscontrol preview
******************** Domain: test2.internal
25 corrections (powerdns)
#1: + CREATE a.test2.internal A 127.0.0.1 ttl=300
#2: + CREATE b.test2.internal A 127.0.0.2 ttl=300
#3: + CREATE c.test2.internal A 127.0.0.3 ttl=300
#4: + CREATE d.test2.internal A 127.0.0.4 ttl=300
#5: + CREATE e.test2.internal A 127.0.0.5 ttl=300
#6: + CREATE f.test2.internal A 127.0.0.6 ttl=300
#7: + CREATE g.test2.internal A 127.0.0.7 ttl=300
#8: + CREATE h.test2.internal A 127.0.0.8 ttl=300
#9: + CREATE i.test2.internal A 127.0.0.9 ttl=300
#10: + CREATE j.test2.internal A 127.0.0.10 ttl=300
#11: + CREATE k.test2.internal A 127.0.0.11 ttl=300
#12: + CREATE l.test2.internal A 127.0.0.12 ttl=300
#13: + CREATE m.test2.internal A 127.0.0.13 ttl=300
#14: + CREATE n.test2.internal A 127.0.0.14 ttl=300
#15: + CREATE o.test2.internal A 127.0.0.15 ttl=300
#16: + CREATE p.test2.internal A 127.0.0.16 ttl=300
#17: + CREATE q.test2.internal A 127.0.0.17 ttl=300
#18: + CREATE r.test2.internal A 127.0.0.18 ttl=300
#19: + CREATE s.test2.internal A 127.0.0.19 ttl=300
#20: + CREATE t.test2.internal A 127.0.0.20 ttl=300
#21: + CREATE u.test2.internal A 127.0.0.21 ttl=300
#22: + CREATE v.test2.internal A 127.0.0.22 ttl=300
#23: + CREATE w.test2.internal A 127.0.0.23 ttl=300
#24: + CREATE x.test2.internal A 127.0.0.24 ttl=300
#25: + CREATE y.test2.internal A 127.0.0.25 ttl=300
WARNING: No nameservers declared; skipping registrar. Add {no_ns:'true'} to force.
Done. 25 corrections.
```
```
$ time dnscontrol push
******************** Domain: test2.internal
25 corrections (powerdns)
--snip--
Done. 25 corrections.

real    0m1.727s
user    0m0.079s
sys     0m0.023s
```


cafferata:
_Ping @jpbede, the maintainer of the PowerDNS provider._

cafferata:
Close https://github.com/StackExchange/dnscontrol/pull/3100

Veratil:
> But I actually don't know a way to change this, I guess we need a way to tell dnscontrol to run corrections in a batch.

My idea would be adding a count field to `models.Correction`. Then instead of adding the length of the corrections array you'd just add `corrections.Count`. But this would be a big change that all provider maintainers would need to address. But this is also a separate discussion to be had. :)

My assumption is that this also affects Route53 batches, only showing the total number of batched corrections.

tlimoncelli:
In hindsight, it's silly to have each provider try to count the number of corrections.  diff2.By*() functions should do the counting and pass that value to the provider as an opaque value that is passed along to the ppreview/ppush code. (I'm just realizing this now).  This would require a change to all providers, but a minor one.

In fact, it is diff2.diffTargets() that knows the real number of changes.  Everything is is just reverse-engineering to figure out what information was lost.  (at least for changes to the zone. Changes to delegations need to be counted elsewhere).

I suspect that gcloud and rout42 report # of changes incorrectly because they batch things too.

tlimoncelli:
https://github.com/StackExchange/dnscontrol/pull/3108 start that refactors the change count to be calculated once, then used everywhere else.  It's a big change.  It might be better to return a struct, but that would be an even bigger change.

Veratil:
So other than the change count, how's this PR? 😂

tlimoncelli:
The code looks good!

Issue 1:

Sadly one integration test is failing.  The one called `TypeChangeHard`, which appears as test 14.

```
export POWERDNS_DOMAIN=dnscontrol-example.com
export POWERDNS_APIKEY=[redacted]
export POWERDNS_APIURL=[redacted]
export POWERDNS_SERVERNAME=localhost
cd integrationTest
go test -v -verbose -provider POWERDNS -start 14 -end 14
```

Since a label with a CNAME can not have any other DNS records, if one is changing from `foo CNAME bar` to `foo A 1.2.3.4`, the CNAME must be deleted first.

The diff2 pkg carefully orders the creates/modifies/deletes.  By splitting them into creates and deletes, we lose that ordering. 

Maybe the author (@blackshadev) has some suggestions on how to handle this. (Ideally the PowerDNS server shouldn't make such checks until all batches are processed).

Issue 2:

Are there other places POWERDNS requires ordering?  For example, when adding `foo NS bar` does PowerDNS require that "bar" is created first?  There's no integration test for this (I think).

Tom

blackshadev:
Hi guys, I can provide suggestions for sure. But without much context I am unaware what to do. And to be enterly honest, I am not willing to start guessing, deducing and running the code, sorry for my lazyness. 

If you have concreet questions, problems  with how the reordering is handeld, or even suggestions how we can change it so it would help you: let me know and I will try and help you out. There could even be a bug in there since most DNS providers are really liniant about accepting broken DNS records.

Regards,
Vincent

Veratil:
> Since a label with a CNAME can not have any other DNS records, if one is changing from `foo CNAME bar` to `foo A 1.2.3.4`, the CNAME must be deleted first.
> 
> The diff2 pkg carefully orders the creates/modifies/deletes. By splitting them into creates and deletes, we lose that ordering.

Ah! That's a good point. We could batch everything into a single set then, that shouldn't be an issue. I'll work on changing this.

> Are there other places POWERDNS requires ordering? For example, when adding `foo NS bar` does PowerDNS require that "bar" is created first? There's no integration test for this (I think).

I can test take a look at the docs and do some tests with the API direct.

Veratil:
Actually, a single set can't work. It will have to be multiple, but batching until it needs to change to a new batch is better than nothing. 👍

tlimoncelli:
A simple solution would be to only batch if they are all the same type (i.e. all deletes).  Not sure if that's worth the complexity.

Tom

Veratil:
I pushed this just testing 14.. running the full test now. But I realized just swapping the deletes first would solve the issue. :)

tlimoncelli:
Ah yes!  Delete then create!  So simple yet should always work!  (Like "break before make" in elecronics)

Test 42 works for me.  Maybe a temporary problem?

I did notice there's an extra blank line in the integration test output.  I couldn't easily see where it is getting added. Does it affect `preview` too?

Veratil:
> I did notice there's an extra blank line in the integration test output. I couldn't easily see where it is getting added. Does it affect `preview` too?

Ah, that's on purpose so the [batch lines line up](https://github.com/StackExchange/dnscontrol/pull/3105/files#diff-e8bf792945f47f78757e76a00657924e7982a2da1072a75daf5ab35dd2f4f054R25). Maybe that's causing the issue? I can replace the blank line with a report correction saying "Batching" of some sort.


> Test 42 works for me. Maybe a temporary problem?

I can only hope so.

tlimoncelli:
> Ah, that's on purpose so the [batch lines line up](https://github.com/StackExchange/dnscontrol/pull/3105/files#diff-e8bf792945f47f78757e76a00657924e7982a2da1072a75daf5ab35dd2f4f054R25). Maybe that's causing the issue? I can replace the blank line with a report correction saying "Batching" of some sort.

Ah, I understand.  Yes, a "Batching" message might be better.  

You might also check that the output of `dnscontrol push -i` looks as expected.

Tom

tlimoncelli:
great!

Anything else before I merge?

Veratil:
As long as everyone is good with it, I think it's good to go. Nothing else from me that I can think of.

Issue #3124, opened 2024-09-24
CLOUDNS: API error messages leak credentials


hmoffatt:
**Describe the bug**
If a CLOUDNS API error occurs, the credentials are leaked. This is because the credentials are passed in the query part of the URL, and the whole URL is printed in the error.

**To Reproduce**
Seen while debugging ClouDNS changes, but it would occur if you had the credentials wrong.

**Expected behavior**
Credentials should not be leaked in the error text.

**DNS Provider**
 - ClouDNS

**Additional context**

cafferata:
_Ping @pragmaton, the maintainer of the ClouDNS provider._

hmoffatt:
I think the simplest change is not to print the URL - change is in the referenced PR. Otherwise a sanitized URL could be printed, if this is desired.

cafferata:
@hmoffatt, can this issue be closed with https://github.com/StackExchange/dnscontrol/pull/3126? Or are there other ideas/implementations besides https://github.com/StackExchange/dnscontrol/pull/3126 that would require this GitHub issue to be kept open?

PR #3126, opened 2024-09-24
CLOUDNS: don't print full URL on error (#3124)

hmoffatt:
Don't print the full request URI on API error, as it contains the authentication parameters.


tlimoncelli (approved):

tlimoncelli:
Good catch!  Thanks for fixing this!

PR #3127, opened 2024-09-24
CLOUDNS: add support for LOC records

hmoffatt:
This PR adds support for NAPTR and LOC records on CLOUDNS.


tlimoncelli (approved):

tlimoncelli:
Thanks!

tlimoncelli (approved):

tlimoncelli:
Please add a comment with the output of the integration tests (redact as needed)

providers/cloudns/cloudnsProvider.go
```diff
@@ -340,13 +341,33 @@ func toRc(domain string, r *domainRecord) *models.RecordConfig {
 	case "CLOUD_WR":
 		rc.Type = "WR"
 		rc.SetTarget(r.Target)
+	case "NAPTR":
+		rc.SetTargetNAPTRStrings(r.NaptrOrder, r.NaptrPref, r.NaptrFlag, r.NaptrParams, r.NaptrRegexp, r.NaptrReplace+".")
+	case "LOC":
+		loc := fmt.Sprintf("%s %s %s %s %s %s %s %s %s %s %s %s",
```
systemcrash, line 347: This part should be redundant now.

providers/cloudns/cloudnsProvider.go
```diff
@@ -388,6 +409,27 @@ func toReq(rc *models.RecordConfig) (requestParams, error) {
...
+		parts := strings.Fields(rc.GetTargetCombined())
+		req["lat-deg"] = parts[0]
+		req["lat-min"] = parts[1]
+		req["lat-sec"] = parts[2]
+		req["lat-dir"] = parts[3]
+		req["long-deg"] = parts[4]
+		req["long-min"] = parts[5]
+		req["long-sec"] = parts[6]
+		req["long-dir"] = parts[7]
+		req["altitude"] = strings.Split(parts[8], "m")[0]
```
systemcrash, line 429: And this.

hmoffatt:
Oops, I forgot the unit tests. This showed up an interesting quirk with CLOUDNS's handling of the size/precision fields - it quietly substitutes its own default values if you pass "0.00", even though those are documented as valid! I've put in a workaround.

providers/cloudns/cloudnsProvider.go
hmoffatt: I think specifically the part that strips the decimal part of altitude is redundant? The rest has to remain as I need to assemble the LOC record from the individual API parts.

Better to call `calculateLOCFields` directly, but then I would have to duplicate the string to int/float conversions as the API returns strings for everything, despite being JSON.

providers/cloudns/cloudnsProvider.go
hmoffatt: I'll wait for your change to be merged then rebase this. Thanks.

hmoffatt:
All tests are passing now except the NAPTR test. The CLOUDNS API is rejecting records with both `regexp` and `replacement` set, which the test cases use. I couldn't find anywhere in the RFC that explicitly said this wasn't allowed, but none of the examples use both fields.

```
=== RUN   TestDNSProviders/dnscontrol-test.dev/35:NAPTR:NAPTR_record
    integration_test.go:248: 
        + CREATE NAPTR test.dnscontrol-test.dev 100 10 "U" "E2U+sip" "!^.*$!sip:customer-service@example.com!" example.foo.com. ttl=300
    integration_test.go:253: failed create record (ClouDNS): ClouDNS API error: You need to enter either Regular Expression or Replacement, not both
```

providers/cloudns/cloudnsProvider.go
systemcrash: > Better to call `calculateLOCFields` directly, but then I would have to duplicate the string to int/float conversions as the API returns strings for everything, despite being JSON.

🤦 

I don't know what to say. But coercing as early as possible is safest, especially when working with numbers and maths.

systemcrash:
> Oops, I forgot the unit tests. This showed up an interesting quirk with CLOUDNS's handling of the size/precision fields - it quietly substitutes its own default values if you pass "0.00", even though those are documented as valid! I've put in a workaround.

Can you supply nothing?

`helpers.js` assumes 0 if the data source is DMS or DMM using a domain.js. But no assumptions or changes are made for records already deployed in a domain.

systemcrash:
> All tests are passing now except the NAPTR test. The CLOUDNS API is rejecting records with both `regexp` and `replacement` set, which the test cases use. I couldn't find anywhere in the RFC that explicitly said this wasn't allowed, but none of the examples use both fields.
> 
> ```
> === RUN   TestDNSProviders/dnscontrol-test.dev/35:NAPTR:NAPTR_record
>     integration_test.go:248: 
>         + CREATE NAPTR test.dnscontrol-test.dev 100 10 "U" "E2U+sip" "!^.*$!sip:customer-service@example.com!" example.foo.com. ttl=300
>     integration_test.go:253: failed create record (ClouDNS): ClouDNS API error: You need to enter either Regular Expression or Replacement, not both
> ```

Seems pretty wild. Both are necessary for a functioning NAPTR pointer. Flag an error with their support. 

Recommend that you separate out NAPTR changes to a separate PR if you think this could take a while to resolve.

hmoffatt:
> Can you supply nothing?

It turns out that the weird behaviour with "0.00" also applies to the altitude field. If I supply nothing, I get the default of 10,000m. :confused:

hmoffatt:
> Recommend that you separate out NAPTR changes to a separate PR if you think this could take a while to resolve.

Done. Once your #3130 is merged I will submit this PR.

hmoffatt:
I've stripped out the NAPTR code for now, and updated this for @systemcrash 's decimal LOC changes. 
[cloudns-test.log](https://github.com/user-attachments/files/17171453/cloudns-test.log)

All the unit tests are passing, so I think this is ready to go! Thanks all for review and assistance.

systemcrash:
Sorry lads, I've no experience with cloud but your approach looks sound. 🤓

hmoffatt:
Hi @tlimoncelli I think this is ready to go now. Thanks.

tlimoncelli:
Thanks!

systemcrash:
@hmoffatt have your changes proved beneficial?

hmoffatt:
My interest in this was academic tbh, just learning my way around Go and the dnscontrol codebase.

PR #3019, opened 2024-06-23
PORKBUN: fix JSON unmarshal and add retry on rate limit

imlonghao:
1. fix JSON unmarshal

I removed lots of unused struct fields to fixed #3016

API responses `securityLock` field with number 0 for OFF and string "1" for ON.

2. remove token check on initialize

>  Providers shouldn't talk to the network in a setup function

Ref: https://github.com/StackExchange/dnscontrol/issues/3014#issuecomment-2176727082

3. add retry on rate limit

They are using [AWS WAF](https://docs.aws.amazon.com/waf/latest/developerguide/waf-captcha-and-challenge-actions.html) for rate limit control, the following HTTP response will occur when hit the limit.

```
HTTP/2 202
...
server: awselb/2.0
x-amzn-waf-action: challenge
...
```

I don't know the extra rate limit rules besides the 1 req per second. As I observed during the integration test, sending more than 100 requests will ban you for some minutes.

My retry strategy implementation doesn't always work, sometimes after getting HTTP 202, the API response 400 Bad Request and I don't know why...

---

Integration test logs

1. [FULL](https://pb.esd.cc/license-amount-depend)
2. [RETRY 13](https://pb.esd.cc/window-express-sunset)
3. [RETRY 14](https://pb.esd.cc/arrange-suspect-vintage)
4. [RETRY 22](https://pb.esd.cc/energy-wheat-subject)
5. [RETRY 29](https://pb.esd.cc/drift-year-brush)
6. [RETRY 38](https://pb.esd.cc/curve-six-mother)
7. [RETRY 39](https://pb.esd.cc/beauty-crash-finish)
8. [RETRY 76](https://pb.esd.cc/elegant-equal-aware)
9. [RETRY 77](https://pb.esd.cc/hole-vehicle-umbrella)

PR #3153, opened 2024-10-11
PORKBUN: Update porkbun API endpoint per documentation / email

plttn:
As of December 1, 2024, Porkbun will require API hits to use `api.porkbun.com`, rather than `porkbun.com`.

Tested successfully and validated that information was obtained as expected.

https://porkbun.com/api/json/v3/documentation
 
> Please note that porkbun.com currently works and has historically been the correct hostname for our API. However, we will be migrating away from that hostname and the API will no function using it at some point in the future.
> 
> Deadline: 2024-12-01 00:00:00 UTC
> 

CleanShot 2024-10-11 at 14 39 30@2x

### DIFF
--- a/providers/porkbun/api.go
+++ b/providers/porkbun/api.go
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	baseURL = "https://porkbun.com/api/json/v3"
+	baseURL = "https://api.porkbun.com/api/json/v3"
 )
 
 type porkbunProvider struct {


cafferata (approved):

imlonghao (approved):
It's a great change.
No more AWSALB, nor the weird rate limit.
I run the integration test and all tests success all at once, not hitting any rate limit compare to #3019.

PR #3154, opened 2024-10-11
Update Porkbun base API URL

railworks2:
Moving from porkbun.com to api.porkbun.com as part of an API change at Porkbun sent via email.

Old endpoint will fail on 1st December 2024 @ midnight UTC.


railworks2:
Giving way to #3153 for being first.

PR #3161, opened 2024-10-20
GCORE: Add support for GeoDNS/Failover

xddxdd:
Add metadata fields for configuring Gcore's GeoDNS and Healthcheck/Failover functionalities, as well as documentations for them.

Tested manually with my own config.

As a side note, I didn't add any automated tests because I'm not sure how we should do so for custom metadata fields. Should I add them as testgroups in `integrationTest/integration_test.go`?


tlimoncelli:
Neat stuff!  Thanks for the PR!

Yes, please add integration tests.  Just limit them to GCORE similar to how Clouldflare's proxy tests are just for CLOUDFLAREAPI.

See this example around line 1964:

```
testgroup("CF_PROXY A create",
      only("CLOUDFLAREAPI"),
```

xddxdd:
Added integration test, and made some small changes to the code to make it play nice with integ test.

providers/gcore/convertMetadata.go
```diff
@@ -0,0 +1,403 @@
...
+	result := map[string]string{}
+
+	if len(rrset.Filters) > 0 {
+		result[metaFilters] = serializeRecordFilter(rrset.Filters)
+	}
+
+	// RRSet only supports failover field
+	if rrset.Meta != nil && rrset.Meta.Failover != nil {
+		failover := *rrset.Meta.Failover
+		// FIXME: maybe we can reduce duplication by using reflection here
```
tlimoncelli, line 52: > 		// FIXME: maybe we can reduce duplication by using reflection here

It's fine like it is.  A little repetition is better than a lot of complexity.

tlimoncelli (approved):

tlimoncelli:
Great new feature!  I love the `withMeta()` function that makes testing so easy.

I've rebased.  Anything else before merging?

xddxdd:
Looks good to me.

Issue #3162, opened 2024-10-21
M365_BUILDER Support for Second Level Domains


CJFelto:
**Describe the bug**
Using the production version of the M365_BUILDER in v4.14.0 produces incorrect output with second level domains

**To Reproduce**
Below is an example of an affected record
```
M365_BUILDER("example.co.nz", {
	initialDomain: 'exampleconz.onmicrosoft.com'
	, dkim: true
	, skypeForBusiness: true
	, mdm: true
	, mx: true
	, autodiscover: true
})
```
This produces the following record
`selector2-example-co.nz._domainkey.exampleconz.onmicrosoft.com.`

**Expected behavior**
Microsoft expects the record to be formatted as below
`selector2-example-co-nz._domainkey.exampleconz.onmicrosoft.com.`

**Workaround**
I have overridden the M365_BUILDER function in my build of DNSControl with the only change being the highlighted line below
```
    if (!value.domainGUID) {
        // Does not work with dashes in domain name.
        // Microsoft uses its own, (probably) deterministic algorithm to transform these domains.
        // Unfortunately, underlying algorithm is not known to us.
        if (name.indexOf('-') !== -1) {
            throw (
                'M365_BUILDER requires domainGUID for domains with dashes: ' +
                name
            );
        }

        value.domainGUID = name.replace(/\./g, '-'); <- Changed from value.domainGUID = name.replace('.', '-');
    }
```
Changing to the Regex pattern now produces the expected output below
`selector2-example-co-nz._domainkey.exampleconz.onmicrosoft.com.`

**DNS Provider**
Cloudflare

**Additional context**
I have also tested this for domains deeper than second level, I.E (marketing.example.co.nz)
With the regex above this will produce `selector2-marketing-example-co-nz._domainkey.exampleconz.onmicrosoft.com.`
This matched what Microsoft requested

ucdelerium:
This was also an issue for me.

tlimoncelli:
CC @juliusrickert (original author)

@CJFelto Would you please submit a PR with your change so we can review it?

juliusrickert:
[`replaceAll`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) seems to do the trick (if it's available in [otto](https://github.com/robertkrimen/otto)).

You can always provide the domain's GUID (IDN, dashes, etc.) for non-trivial domains.

I'm looking forward to a PR.

[...]

PR #3165, opened 2024-10-21
M365_BUILDER: Support second level domains, fix domainGUID Generation

CJFelto:
Update M365_BUILDER to use a global match regex

### DIFF
--- a/pkg/js/helpers.js
+++ b/pkg/js/helpers.js
@@ -1850,7 +1850,7 @@ function M365_BUILDER(name, value) {
             );
         }
 
-        value.domainGUID = name.replace('.', '-');
+        value.domainGUID = name.replace(/\./g, '-');
     }
 
     if (value.dkim && !value.initialDomain) {


CJFelto:
Related to Issue: #3162

pkg/js/helpers.js
```diff
@@ -1850,7 +1850,7 @@ function M365_BUILDER(name, value) {
             );
         }
 
-        value.domainGUID = name.replace('.', '-');
+        value.domainGUID = name.replace(/\./g, '-');
```
juliusrickert, line 1853: Would [`replaceAll`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) work too? That would be more idiomatic to me.

CJFelto: I agree replaceAll would be the more appropriate change, however replaceAll does not appear to be included within Otto

The Otto repo mentions it targets ES5 while replaceAll was added later in ES2021

As this regex is extremely simple, I do not believe we would see any unintended consequences or performance impact

PR #3167, opened 2024-10-22
NS1: Ignore RRSets of type REDIRECT

kabenin:
NS1 replaces URLFWD with REDIRECTs. The latter supports HTTPS, which is nice, but it is not a 'real' DNS Record from the configuration point of view. To create a redirect, there is a separate API Endpoint. Creating a redirect may also involve Certificate management, if a user wants HTTPS support.
However, NS1 returns names, for which there is a redirect configured in n.Zones.Get(), with a stub answers section 'NS1 MANAGED RECORD'. RRType of REDIRECT confuses dnscontrol, and entire sync fails.

This commit ignores REDIRECTs coming from NS1 API.


tlimoncelli:
CC @costasd

costasd:
Thank you. LGTM!

This is new, was waiting for this to emerge so that I can filter it the same way as docs were close to non-existent and API wasn't returning any redirects yet, the last time I checked.

Will check if the docs/api/sdk are more mature now, so that we can add some proper support (if possible)

Issue #1400, opened 2022-02-05
Deprecate "get-certs" (Let's Encrypt) support


tlimoncelli:
We are considering removing this feature in the future.

Why:

* The "get-certs" command (renews certs via Let's Encrypt) has no maintainer.  If it breaks, we can't really support it.
* There are other projects that do a better job. 

Post your thoughts in the comments.

IzumiSenaSora:
Its really a good feature!... We can manage all domain related things like (manage dns + certificate) in one place!... Please for now don't remove it!.... Unless it breaks!

juliusrickert:
I didn't even know this feature existed.
What were the motivations behind including this functionality into DNSControl?
 
I think this should not be the concern of DNSControl. For me, DNSControl is an application that is intended to be primarily run in CI. Thus, I don't see how issuing certificates would be beneficial. This is a concern of a load balancer or any other public-facing server.

There are other projects that are better at fulfilling this need, e.g., [cert-manager](https://github.com/cert-manager/cert-manager).

On that note: I'm also not quite sure whether I like the deep integration with Cloudflare's redirects and workers. But I can see how that's related to DNSControl's goal. Both are redirects/routes in a way.

tlimoncelli:
> Please for now don't remove it!.... Unless it breaks!

Removing it when it breaks sounds like a terrible idea.  People would be stranded.  We'd rather give people advanced warning so they can remove it on their schedule. (Which is why we're starting the discussion now.)

tlimoncelli:
> I didn't even know this feature existed. What were the motivations behind including this functionality into DNSControl?

The motivation was:  Let's Encrypt DNS01 requires the ability to talk to your DNS Provider, and dnscontrol already does that, it seems like a good match.

IzumiSenaSora:
> > Please for now don't remove it!.... Unless it breaks!
> 
> Removing it when it breaks sounds like a terrible idea. People would be stranded. We'd rather give people advanced warning so they can remove it on their schedule. (Which is why we're starting the discussion now.)

Oh Sorry i'm not giving any idea!... i'm just saying if no one step forward to maintain it!... Then maybe removing it is only option!
BTW I'm bad in English

IzumiSenaSora:
> I didn't even know this feature existed. What were the motivations behind including this functionality into DNSControl?
> 
[...]

PR #3169, opened 2024-10-23
GET-CERTS:  Bump github.com/go-acme/lego to v4 to resolve security issues with go-jose

jauderho:
This is intended to resolve GHSA-c5q2-7r4c-mv6g and referenced in #1400. 

I updated github.com/go-acme/lego to v4. There was an error thrown for https://github.com/StackExchange/dnscontrol/blob/main/pkg/acme/acme.go#L144 initially which was resolved by adding an additional blank string. I'm not 100% sure there is a specific reason that needs to be provided but did not see any such requirement when I looked further. 

Fixing that results in a working build.


tlimoncelli:
I've tested this solution in our staging area.  That's not a complete test but it looks good so far.

Thanks so much for pursuing this!  I really appreciate it!

PR #3191, opened 2024-11-04
 FEATURE: M365_BUILDER add support for alternative MX hosts

### LINKED ISSUE #3185: Support for mx.microsoft. in M365_BUILDER
[Microsoft has recently introduced support for inbound SMTP DANE in Exchange Online](https://techcommunity.microsoft.com/t5/exchange-team-blog/announcing-general-availability-of-inbound-smtp-dane-with-dnssec/ba-p/4281292), and with it [comes a new MX record target](https://learn.microsoft.com/en-gb/purview/how-smtp-dane-works#inbound-smtp-dane-with-dnssec) that will be in a DNSSEC signed zone.

Implementation would involve changing MX record target from `example-com.protection.outlook.com.` to `example-com.<random-id>.mx.microsoft.`  
Where `<random-id>` may be different per-domain. To mitigate DNSSEC performance issues with large zones Microsoft are spreading MX targets across many zones.

Currently, [M365_BUILDER](https://docs.dnscontrol.org/language-reference/domain-modifiers/m365_builder) supports setting an MX record following the `example-com.protection.outlook.com.` pattern, by using `mx: true`.

It would be good to also support the `example-com.<random-id>.mx.microsoft.` pattern.  

It could be done by using an additional label of `mxHost`, defaulting to `protection.outlook.com` if undefined. This would allow the user to specify `mxHost: "<random-id>.mx.microsoft"`, but could also be used to accommodate the sovereign clouds like [Office 365 DOD](https://learn.microsoft.com/en-us/microsoft-365/enterprise/dns-records-for-office-365-dod?view=o365-worldwide#exchange-online), or 21Vianet - as they all use different MX target hosts.

Another way would be changing the behaviour of `mx` to support `true`, `false`, or `"<any MX target here>"`.  
This would avoid needing another label and wouldn't break backwards compatibility, but I don't know what the opinion on mixing types in a label is.

For domains that we want to use inbound SMTP DANE with, we can set the M365_BUILDER to `mx: false` and set an MX record manually for `example-com.<random-id>.mx.microsoft.`

ajh0912:
Closes #3185

Could you review and advise if this approach is the right way?

I've added a new example for `M365_BUILDER`, with all services enabled and using the new `mxHost` parameter to point to an example DNSSEC enabled MX host.

### DIFF
--- a/documentation/language-reference/domain-modifiers/M365_BUILDER.md
+++ b/documentation/language-reference/domain-modifiers/M365_BUILDER.md
@@ -66,6 +66,7 @@ This sets up Mobile Device Management only.
 
 * `label` The label of the Microsoft 365 domain, useful if it is a subdomain (default: `"@"`)
 * `mx` Set an `MX` record? (default: `true`)
+* `mxHost` Set your MX host for Exchange Online (default: `mail.protection.outlook.com`)
 * `autodiscover` Set Autodiscover `CNAME` record? (default: `true`)
 * `dkim` Set DKIM `CNAME` records? (default: `true`)
 * `skypeForBusiness` Set Skype for Business/Microsoft Teams records? (default: `false`)
--- a/documentation/language-reference/domain-modifiers/M365_BUILDER.md
+++ b/documentation/language-reference/domain-modifiers/M365_BUILDER.md
@@ -42,7 +42,7 @@ END);
 
 This sets up `MX` records, Autodiscover, and DKIM.
 
-### Advanced example
+### Example with MDM only
 
 {% code title="dnsconfig.js" %}
 ```javascript
--- a/documentation/language-reference/domain-modifiers/M365_BUILDER.md
+++ b/documentation/language-reference/domain-modifiers/M365_BUILDER.md
@@ -62,6 +62,29 @@ END);
 
 This sets up Mobile Device Management only.
 
+### Example with all services and DNSSEC MX
+
+{% code title="dnsconfig.js" %}
+```javascript
+D("example.com", REG_MY_PROVIDER, DnsProvider(DSP_MY_PROVIDER),
+  M365_BUILDER("example.com", {
+      mx: true, // Can be omitted as default: true
+      mxHost: "o-v1.mx.microsoft", // Override the default mail.protection.outlook.com
+      autodiscover: true, // Can be omitted as default: true
+      dkim: true, // Can be omitted as default: true
+      skypeForBusiness: true,
+      mdm: true,
+      initialDomain: "example.onmicrosoft.com",
+  }),
+END);
+```
+{% endcode %}
+
+This sets up MX (custom), AutoDiscover, DKIM, Skype for Business/Microsoft Teams and Mobile Device Management records.
+
+Instead of the default MX target ending in `mail.protection.outlook.com`, this example uses `<random-id>.mx.microsoft` which is a DNSSEC signed zone.
+`<random-id>` is obtained from Microsoft after [enabling DNSSEC functionality for the domain](https://learn.microsoft.com/purview/how-smtp-dane-works#inbound-smtp-dane-with-dnssec) within Exchange Online.
+
 ### Parameters
 
 * `label` The label of the Microsoft 365 domain, useful if it is a subdomain (default: `"@"`)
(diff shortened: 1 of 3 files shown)


tlimoncelli:
CC @juliusrickert @nikolamilekic for feedback

documentation/language-reference/domain-modifiers/M365_BUILDER.md
```diff
@@ -62,10 +63,34 @@ END);
...
+D("example.com", REG_MY_PROVIDER, DnsProvider(DSP_MY_PROVIDER),
+  M365_BUILDER("example.com", {
+      mx: true, // Can be omitted as default: true
+      mxHost: "o-v1.mx.microsoft", // Override the default mail.protection.outlook.com
+      autodiscover: true, // Can be omitted as default: true
+      dkim: true, // Can be omitted as default: true
+      skypeForBusiness: true,
+      mdm: true,
+      initialDomain: "example.onmicrosoft.com",
+  }),
```
juliusrickert, line 79: The trailing comma should be removed, I think.

pkg/js/helpers.js
```diff
@@ -1879,15 +1879,14 @@ function M365_BUILDER(name, value) {
...
     // MX (default: true)
     if (value.mx) {
-        r.push(
-            MX(
-                value.label,
-                0,
-                value.domainGUID + '.mail.protection.outlook.com.'
-            )
-        );
+        r.push(MX(value.label, 0, value.domainGUID + '.' + value.mxHost + '.'));
```
juliusrickert, line 1889: The [PowerShell(?) commands](https://learn.microsoft.com/en-gb/purview/how-smtp-dane-works#inbound-smtp-dane-with-dnssec:~:text=Success-,contosotest%2Dcom.o%2Dv1.mx.microsoft,-The%20success%20response) (link to highlight, might not work in all browsers) returns the whole MX hostname: `contosotest-com.o-v1.mx.microsoft`.

Therefore, and to make the helper more flexible in general, I advise to do one of the following:

1. Document how to disable the MX and set the Microsoft-provided hostname. (**preferred**)
2. Allow any MX to be set, possibly multiple MXs

Documentation for the first option could look like this:

```
D("example.com", REG_MY_PROVIDER, DnsProvider(DSP_MY_PROVIDER),
  M365_BUILDER("example.com", {
      mx: false,
      …
  }),
  MX('@', 0, 'example-com.<random>.mx.microsoft.') // Note the trailing dot.
);
```

juliusrickert (changes requested):
In short, I advise against creating a custom solution for this.
Instead, I'd like to see documentation on how to solve the situation using existing mechanisms. (See code comment.)

nikolamilekic:
@tlimoncelli
I wish I could give constructive feedback on this, but I'm not sure I can.

I understand the challenges of supporting complex macros, but at the same time, I highly value the ability to glance at a DNSControl file and quickly get the idea of what is configured and what isn't (and correctly). For complex configurations such as that of M365 (or MTA-STS), I believe it is especially important to have a macro because I only need to read a single line of code and know what it pertains to instead of 20.

I think it will be of enormous help for the adoption of various complex DNS conventions and standards (such as MTA-STS) if there were an easy way to configure them. With the DMARC and SPF builders, DNSControl is already well on the way of becoming just such a tool. Wouldn't it be great if we went all the way? If the cost is community-supported maintenance of complex builders, would that be so bad? I'd much rather place trust on a community of experts to maintain a set of macros, than risk copy/paste mistakes from documentation.

tlimoncelli:
@nikolamilekic 

Thank you for taking the time to write up your thoughts.

"Builders" are perhaps one of DNSControl's most unique and powerful features.  They are complex to maintain, and I'd like to improve that.

One idea I had was to remove them from the system, so that people can maintain them independently.  However I think that just takes a big problem and moves it elsewhere.

I'd rather make them easier to manage:

1. Testable: Unit tests so we don't break existing use-cases.
2. Isolated: Isolate the code to 1 place for easier management
3. Versioned: Either semver or a policy of "no breaking changes until major releases"

I don't know how to do those things in JavaScript, but I do know how to do them all in Go.  I have an idea for a way to make it possible to write these in Go, with proper testing, isolation, and versioning.

That's long term. As far as M365_BUILDER, I defer to @juliusrickert and would prefer documentation (rather than major changes) at this time.

PR #3193, opened 2024-11-07
feat(import_transform): skip records with import_transform_skip set

tlimoncelli:
Problem:
* IMPORT_TRANSFORM/IMPORT_TRANSFORM_STRIP need to be able to skip certain records.

Solution:
* With this PR, if the metadata for a record includes "import_transform_skip" (any value), that record is skipped


tlimoncelli:
@costasd Hi Costas!  It looks like NS1 has finally disabled URLFWD: https://github.com/StackExchange/dnscontrol/actions/runs/11731300660/job/32681410264

costasd:
@tlimoncelli ah nice, will prepare a PR to remove it completely.

PR #3202, opened 2024-11-15
NAMECHEAP: Enable cmode concurrent

tlimoncelli:
CC @willpower232

* Enable Concurrent mode for Namecheap

I did a limited amount of testing and it seems to work.  I don't see any fields namecheap's struct that are updated after initialization.


tlimoncelli:
> Thanks @tlimoncelli I'll run some tests tomorrow if you like

That'd be great!

A good test would be:
* Compile with the `-race` flag: `go install -race`
* Use a dnsconfig.js file with multiple namecheap domains (at least 3... the more the better)
* Run the "race" version of the binary: `dnscontrol preview --cmode concurrent`

Thanks!
Tom

willpower232:
> CONCURRENTLY gathering 5 zone(s)
> SERIALLY gathering 18 zone(s)

I have 5 zones using both Namecheap DNS and registrar so looking good to me! I even ran it in a loop to offend the rate limit and that worked nicely too.

I'm using 3 other DNS providers but they don't have concurrency ticked on the docs so everything lines up.

Anything else you need me to try?

tlimoncelli:
How about testing the other providers? :-)

willpower232:
Ah of course, I have tried netlify, hetzner, and digital ocean with no problem. The preview sure is fast now!

> CONCURRENTLY gathering 23 zone(s)
> SERIALLY gathering 0 zone(s)

tlimoncelli:
Would you mind submitting a PR for each? That way the CHANGELOG will highlight each one.

willpower232:
can do, I could probably figure out how to try route53 and cloudflare too if that helps?

tlimoncelli:
> can do, I could probably figure out how to try route53 and cloudflare too if that helps?

Thanks but those 2 are already concurrent:

```
$ grep providers.CanConcur providers/route53/route53Provider.go providers/cloudflare/cloudflareProvider.go 
providers/route53/route53Provider.go:	providers.CanConcur:              providers.Can(),
providers/cloudflare/cloudflareProvider.go:	providers.CanConcur:              providers.Can(),
```

willpower232:
three PRs, get em while they're hot 🫡

tlimoncelli:
> three PRs, get em while they're hot 🫡

Thank you!

PR #3206, opened 2024-11-19
HETZNER: Enable cmode concurrent

willpower232:
As per #3202, I believe Hetzner works with concurrency


tlimoncelli:
CC @das7pad (provider owner)

Looking at the code, this should work fine.

If you'd like to verify empirically, please do the following:

1.  Compile with the `-race` flag:  `go install -race`
2. Run with a `dnsconfig.js` that has multiple domains from this provider (4 or more is preferable)
3. Run "dnscontrol preview --cmode concurrent` two or three times.  You should get zero race condition errors

das7pad:
Hi! Sorry for the delay in getting back to you. I've opened https://github.com/StackExchange/dnscontrol/pull/3234 with the necessary changes to make the provider thread-safe.

PR #3207, opened 2024-11-19
DIGITALOCEAN: Enable cmode concurrent

willpower232:
As per #3202, I believe DigitalOcean works with concurrency


tlimoncelli:
CC @Deraen (provider owner)

(I merge this prematurely.  Sorry!  Anyway... if you have time, please verify this change.)

Looking at the code, this should work fine.

If you'd like to verify empirically, please do the following:

1.  Compile with the `-race` flag:  `go install -race`
2. Run with a `dnsconfig.js` that has multiple domains from this provider (4 or more is preferable)
3. Run "dnscontrol preview --cmode concurrent` two or three times.  You should get zero race condition errors

willpower232:
@tlimoncelli I forgot the docs sorry, if you've reverted should I make a fresh PR with the docs?

tlimoncelli:
Don't worry about the docs. I run "go generate" as part of the release process.

PR #3215, opened 2024-11-27
Add notifications with Shoutrr

### LINKED ISSUE #2994: Notifications via Shoutrrr
Currently, there is a limited number of notification types. Adding a new one is simple, but each notification type would require their own configuration parameters. Sometimes it's a token, sometimes it's a URL, sometimes it's both.

We could consider using [Shoutrrr](https://containrrr.dev/shoutrrr/v0.8/), which is a Go library for notifications. Unlike the current solution, it uses a single URL for a notification type, and this URL encodes all information needed, be it tokens or other metadata. It supports all current notification types natively except for Bonfire, but since it also has webhook support, Bonfire is also easy to implement.

An obvious alternative would be to implement all notification types from scratch

**Additional context**
N/A

hmoffatt:
I need to update the documentation, but this seems to work for me. Resolves #2994 .

Ideally the notifiers would all use a common formatter (or at least a common format), as there's slight differences in the format of the messages between each currently.

The custom notifiers for Slack/Teams/Telegram could be refactored to use Shoutrrr too, though there's so little code it's probably not worth it.


pkg/notifications/shoutrrr.go
```diff
@@ -0,0 +1,33 @@
...
+func init() {
+	initers = append(initers, func(cfg map[string]string) Notifier {
+		if url, ok := cfg["shoutrrr_url"]; ok {
+			return shoutrrrNotifier(url)
+		}
+		return nil
+	})
+}
+
+// bonfire notifier for stack exchange internal chat. String is just url with room and token in it
```
tlimoncelli, line 18: Suggestion: Delete this comment.

kytta:
I've tested this with Matrix and ntfy, and it works like a charm!

hmoffatt:
I've added documentation now.

documentation/notifications.md
```diff
@@ -94,6 +95,12 @@ This is Stack Overflow's built in chat system. This is probably not useful for m
 
 Configure `bonfire_url` to be the full url including room and api key.
 
+### Others
+
+DNSControl can use many other notification methods via Shoutrrr, such as email (SMTP), Discord, Pushover and others. See the [Shoutrrr documentation](https://containrrr.dev/shoutrrr/v0.8/services/overview/) for a list of supported methods and configuration instructions.
```
cafferata, line 100: I would like to propose to change this version '0.8' to 'latest'. From the MkDocs generator a redirect takes place to the actual latest version.

```suggestion
DNSControl can use many other notification methods via Shoutrrr, such as email (SMTP), Discord, Pushover and others. See the [Shoutrrr documentation](https://containrrr.dev/shoutrrr/latest/services/overview/) for a list of supported methods and configuration instructions.
```

cafferata (changes requested):

cafferata (approved):

documentation/notifications.md
```diff
@@ -94,6 +95,12 @@ This is Stack Overflow's built in chat system. This is probably not useful for m
 
 Configure `bonfire_url` to be the full url including room and api key.
 
+### Others
```
tlimoncelli, line 98: Suggestion: Move this to the top of the list
```suggestion
### Shoutrrr (email, Discord, Pushover, etc.)
```

tlimoncelli:
Thank you for adding this!  I didn't know Shoutrr existed.  This should be the way we do all notifications in the future.

Issue #2474, opened 2023-07-07
CLOUDFLAREAPI: Exit code 0 even if ending with ERROR due to invalid API token


apoeteo:
**Describe the bug**
With CLOUDFLAREAPI, even if an invalid API token is used and `preview` ends with "ERROR", the exit code is 0.
That's especially confusing when using in a pipeline.

**To Reproduce**
1. Use an invalid API token in "creds.json"
2. Run `docker run --rm -it -v $(pwd)/dnsconfig.js:/dns/dnsconfig.js -v $(pwd)/creds.json:/dns/creds.json -v $(pwd)/spfcache.updated.json:/dns/spfcache.updated.json -v $(pwd)/spfcache.json:/dns/spfcache.json -v $(pwd)/zones/:/dns/zones/ -u $(id -u ${USER}):$(id -g ${USER}) stackexchange/dnscontrol preview`
3. The output is "ERROR: ERROR: failed fetching domain list from cloudflare(""): Invalid access token (9109)Done. 0 corrections." but the exit code is 0.

Same issue happens if you forget to whitelist your IP on CloudFlare and attempt to connect there. The output is "ERROR: ERROR: failed fetching domain list from cloudflare(""): Cannot use the access token from location: XXX.XXX.XXX.XXX (9109)Done. 0 corrections." but the exit code is 0.

If you try to run the command without a token (token field empty), the exit code is 1.

**Expected behavior**
There should be an error exit code, the same like if you run with empty token field.

**DNS Provider**
CLOUDFLAREAPI

**Additional context**
When used in a GitLab pipeline, it's confusing when the preview job ends with success even though it was no successful.

TomOnTime:
CC @tresni

PR #3216, opened 2024-11-27
BUG: Errors while listing zones do not propagate through the system (#2474)

hmoffatt:
If an error occurred listing zones (eg invalid credentials), the preview/push commands didn't return an error code.

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -88,7 +88,7 @@ jobs:
         Write-Host "Integration test providers: $Providers"
         echo "integration_test_providers=$(ConvertTo-Json -InputObject $Providers -Compress)" >> $env:GITHUB_OUTPUT
       env:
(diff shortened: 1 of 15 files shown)


hmoffatt:
Also discovered that CLOUDFLAREAPI caches an empty zone list if the fetch fails.

tlimoncelli:
Good catch!

CC @tresni for approval.

tlimoncelli:
Providers shouldn't generate any network traffic just at initialization time.  I guess I missed that when I reviewed ClouDNS.

We should probably add a `.CheckCreds()` method or something.  But so far nobody has complained not getting an error until the first domain of that provider.

hmoffatt:
I've pushed another change for CLOUDNS to delay fetching the domain list.

tlimoncelli:
Thanks for fixing this!!!

PR #3226, opened 2024-12-07
CNR: Disable pager101 integration test

tlimoncelli:
Problem:

The pager101 integration test files for CNR.

Solution:

* I suspect our account has a limit in place.
* This test is safe to ignore for most providers as most don't send API results 1 page at a time.


tlimoncelli:
CC @KaiSchwarz-cnic  @AsifNawaz-cnic for visibility.  If you want to debug this, please open a bugid. I'd be glad to collaborate.

PR #3228, opened 2024-12-09
PORKBUN: increase req interval and retry on 503

### LINKED ISSUE #3225: PORKBUN: Rate limits are not respected, causing many errors
**Describe the bug**
I have about 20 domains on PorkBun, and every single run of `preview` or `push` includes multiple errors while attempting to read data from PorkBun's API. Here are some examples:

```
ERROR
Error getting corrections (porkbun): failed parsing nameserver list from porkbun: invalid character '<' looking for beginning of value
```

```
ERROR
Error getting corrections (porkbun): failed parsing record list from porkbun: invalid character '<' looking for beginning of value
```

```
ERROR
Error getting corrections (porkbun): failed parsing url forwarding record list from porkbun: invalid character '<' looking for beginning of value
```

These errors appear to be purely rate-limiting, because it is always possible to manually run just the one domain that failed and have it work successfully.

**To Reproduce**
1. Add several PorkBun domains to your `dnsconfig.js` file. Adding URL forwarding records seems to make things worse, but it always seems to happen around 20 or 30 domains total.

This can also be reproduced with `curl` against the API. Within 10 or 20 requests, the API starts returning a 503 response, with this body:

```
<html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body>
<center><h1>503 Service Temporarily Unavailable</h1></center>
<hr><center>openresty</center>
</body>
</html>
```

**Expected behavior**
I expect API requests to PorkBun to respect the PorkBun request limit, and either limit requests so the rate limit is not hit, or back off and retry on errors.

**DNS Provider**
 - PorkBun

**Additional context**

PorkBun apparently has a global rate limit of 1 request per second, as recounted in [this issue on the terraform porkbun provider](https://github.com/cullenmcdermott/terraform-provider-porkbun/issues/23#issuecomment-1366859999), so it is easy to hit with just a couple dozen domains.

> BTW, I received the response from the Porkbun developers confirming my thoughts:
"About 60 / minute or 2 / second. There are some bursting capabilities up to 5 / second. The best bet would be to keep things to 1 / second if there are constant commands being issued and to wait / resend if a threshold is reached."

imlonghao:
1. Increase the delay between each request from 0.5s to 1s.
2. Retry on status code 503
3. Retry on 10s, 20s, 30s, 40s

close #3225


imlonghao:
@indirect can you try this pr and see whether it works for you?

indirect:
This worked with no errors, thank you @imlonghao 👍🏻

tlimoncelli:
Awesome!!

tlimoncelli:
YAY!  I love rate limiting!!

Thanks, folks!

PR #3232, opened 2024-12-10
CLOUDNS: enable concurrent access

hmoffatt:
This enables concurrent access on CLOUDNS. Tested with my 24 zones.

PR #3233, opened 2024-12-10
DOCS: Add instructions for increasing the timeout on integration tests

hmoffatt:
The CLOUDNS tests take longer than the default 10 minute timeout to run. This change adds a note to the testing instructions on how to increase or disable the timeout.

### DIFF
--- a/documentation/integration-tests.md
+++ b/documentation/integration-tests.md
@@ -47,6 +47,12 @@ go test -v -verbose -provider ROUTE53 -end 5
 go test -v -verbose -provider ROUTE53 -start 16 -end 20
 ```
 
+For some providers it may be necessary to increase the test timeout using `-test`. The default is 10 minutes.
+
+```shell
+go test -v -verbose -provider CLOUDNS -timeout 0
+```
+
 The actual tests are in the file `integrationTest/integration_test.go`.  The
 tests are in a little language which can be used to describe just about any
 interaction with the API.  Look for the comment `START HERE` or the line
--- a/documentation/integration-tests.md
+++ b/documentation/integration-tests.md
@@ -47,12 +47,13 @@ go test -v -verbose -provider ROUTE53 -end 5
 go test -v -verbose -provider ROUTE53 -start 16 -end 20
 ```
 
-For some providers it may be necessary to increase the test timeout using `-test`. The default is 10 minutes.
+For some providers it may be necessary to increase the test timeout using `-test`. The default is 10 minutes.  `0` is "no limit".  Typical Go durations work too (`1h` for 1 hour, etc).
 
 ```shell
-go test -v -verbose -provider CLOUDNS -timeout 0
+go test -timeout 0 -v -verbose -provider CLOUDNS 
 ```
 
+FYI: The order of the flags matters.  Flags native to the Go testing suite (`-timeout` and `-v`) must come before flags that are part of the DNSControl integration tests (`-verbose`, `-provider`). Yeah, that sucks and is confusing.
 The actual tests are in the file `integrationTest/integration_test.go`.  The
 tests are in a little language which can be used to describe just about any
 interaction with the API.  Look for the comment `START HERE` or the line
--- a/documentation/integration-tests.md
+++ b/documentation/integration-tests.md
@@ -54,6 +54,7 @@ go test -timeout 0 -v -verbose -provider CLOUDNS
 ```
 
 FYI: The order of the flags matters.  Flags native to the Go testing suite (`-timeout` and `-v`) must come before flags that are part of the DNSControl integration tests (`-verbose`, `-provider`). Yeah, that sucks and is confusing.
+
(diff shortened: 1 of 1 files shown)


documentation/integration-tests.md
```diff
@@ -47,6 +47,12 @@ go test -v -verbose -provider ROUTE53 -end 5
 go test -v -verbose -provider ROUTE53 -start 16 -end 20
 ```
 
+For some providers it may be necessary to increase the test timeout using `-test`. The default is 10 minutes.
+
+```shell
+go test -v -verbose -provider CLOUDNS -timeout 0
```
tlimoncelli, line 53:
```suggestion
go test -timeout 0 -v -verbose -provider CLOUDNS 
```

documentation/integration-tests.md
```diff
@@ -47,6 +47,12 @@ go test -v -verbose -provider ROUTE53 -end 5
...
+```
+
```
tlimoncelli, line 55:
```suggestion

FYI: The order of the flags matters.  Flags native to the Go testing suite (`-timeout` and `-v`) must come before flags that are part of the DNSControl integration tests (`-verbose`, `-provider`). Yeah, that sucks and is confusing.
```

documentation/integration-tests.md
```diff
@@ -47,6 +47,12 @@ go test -v -verbose -provider ROUTE53 -end 5
 go test -v -verbose -provider ROUTE53 -start 16 -end 20
 ```
 
+For some providers it may be necessary to increase the test timeout using `-test`. The default is 10 minutes.
```
tlimoncelli, line 50:
```suggestion
For some providers it may be necessary to increase the test timeout using `-test`. The default is 10 minutes.  `0` is "no limit".  Typical Go durations work too (`1h` for 1 hour, etc).
```

tlimoncelli (commented):
Some nitpicks :)

tlimoncelli:
Here's a thought: Are there tests that take a long time but don't provide much value?

For example:

* The `pager101` test is irrelevant to providers that don't send replies one page at a time.
* Other tests that take a long time to run can be skipped using the "not()" function if they are irrelevant (or just add a comment to add this test when a full test is needed)

hmoffatt:
> Here's a thought: Are there tests that take a long time but don't provide much value?

That's a good question. I posted my full test summary for CLOUDNS at https://github.com/StackExchange/dnscontrol/pull/3232#issuecomment-2530428497 just yesterday, so I'll take a look. 

The full run took 1328 seconds, more than twice the timeout. How long is it taking on other providers?

tlimoncelli:
Here's how many minutes each took in [a recent run](https://github.com/StackExchange/dnscontrol/actions/runs/12208339671/job/34061589146).

2 BIND
4 GCLOUD
4 HEXONET
4 ROUTE53
4.5 DO
4.5 SAKURA
5 NAMEDOTCOM
5 NS1
6 GANDI
7.5 AZURE_DNS
8.5 HEDNS
9 CLOUDFLARE
9 CNR
11.4 TRANSIP
520 CSCGLOBAL (not a typo. It takes 7 hours thus we only run it when we need to)

The median is about 5 minutes.  Nearly half are close to hitting 10 minutes.

Tom

hmoffatt:
This CLOUDNS result is weird; why did the test take 77s if skipped?

```
        --- PASS: TestDNSProviders/dnscontrol-test.dev/30:pager601_***SKIPPED(disabled_by_only)***:Empty (76.89s)
```

tlimoncelli:
Ok, that explains it!  My test host is in the U.S.  I guess we'll just have to live with the latency.  At least we know the problem.

hmoffatt:
Adding the timeout instructions to this documentation will help me next time I need to run this test suite :)

documentation/integration-tests.md
```diff
@@ -47,6 +47,13 @@ go test -v -verbose -provider ROUTE53 -end 5
 go test -v -verbose -provider ROUTE53 -start 16 -end 20
 ```
 
+For some providers it may be necessary to increase the test timeout using `-test`. The default is 10 minutes.  `0` is "no limit".  Typical Go durations work too (`1h` for 1 hour, etc).
+
+```shell
+go test -timeout 0 -v -verbose -provider CLOUDNS 
+```
+
+FYI: The order of the flags matters.  Flags native to the Go testing suite (`-timeout` and `-v`) must come before flags that are part of the DNSControl integration tests (`-verbose`, `-provider`). Yeah, that sucks and is confusing.
```
tlimoncelli, line 56:
```suggestion
FYI: The order of the flags matters.  Flags native to the Go testing suite (`-timeout` and `-v`) must come before flags that are part of the DNSControl integration tests (`-verbose`, `-provider`). Yeah, that sucks and is confusing.

```

tlimoncelli:
Thanks!  The best kind of of doc updates are clarifications of the non-obvious!

PR #3234, opened 2024-12-10
HETZNER: enable concurrency capability

das7pad:
Closes #3206 

This PR is making the Hetzner provider thread-safe and enabling the concurrency feature flag.

I've tweaked the rate-limiting to play better with concurrency by not overshooting the rate-limit reset point.

cc @tlimoncelli @willpower232


willpower232:
I'm presuming you have already run this against your accounts so let me know if you'd like me to run it as well, otherwise thanks for making it even better :sunglasses:

tlimoncelli:
Thanks so much!  Excellent mutex'ing!!!

Best,
Tom

Issue #3259, opened 2024-12-17
A Name.com TXT entry produces quoted value


tkurki:
**Describe the bug**
[This configuration](https://github.com/SignalK/dns/blob/main/dnsconfig.js#L21-L23) produces a quoted TXT record value:
<img width="968" alt="image" src="https://github.com/user-attachments/assets/1b09664c-0b7b-49f3-a2a1-16a6535546b1" />

If I manually remove the quotes in name.com UI subsequent `push` operations do nothing .

**To Reproduce**
See linked repo.

Log of 2 push runs: 
```
CONCURRENTLY gathering 0 zone(s)
SERIALLY gathering 1 zone(s)
Serially Gathering: "signalk.org"
******************** Domain: signalk.org
1 correction (name.com)
#1: + CREATE TXT _atproto.signalk.org "did=did:plc:b6o3a5sifu6lp7if47yabyvp" ttl=300
SUCCESS!
Done. 1 corrections.

====> THE VALUE APPEARS QUOTED. MANUALLY REMOVED QUOTES FROM THE VALUE HERE <====

signalk/dns(main|✔) %   ./run.sh push
CONCURRENTLY gathering 0 zone(s)
SERIALLY gathering 1 zone(s)
Serially Gathering: "signalk.org"
******************** Domain: signalk.org
Done. 0 corrections.

```
**Expected behavior**
The value is created without quotes. All push operations effect changes so that the actual dns is like the donscontrol configuration.

**DNS Provider**
 - name.com

**Additional context**
The record is for Bluesky.social domain verification.

tlimoncelli:
This sounds a lot like https://docs.dnscontrol.org/developer-info/testing-txt-records 

The bad news is that the NAMEDOTCOM provider doesn't have a maintainer at this time.

The good news is that while writing some tips about how to debug this on your own, I think I found the bug.

Can you build https://github.com/StackExchange/dnscontrol/pull/3260 and see if that fixes the problem?

tkurki:
If you tell me how 😁 - pointers?

tlimoncelli:
hi @tkurki !  Glad to help.  I put the instructions in https://github.com/StackExchange/dnscontrol/pull/3260#issuecomment-2549290654

Issue #3227, opened 2024-12-08
IGNORE()'d record deleted when other modifications are made (provider: MYTHICBEASTS)


rmc47:
**Describe the bug**
When `dnscontrol push` is used and results in modification to a zone (at least using MythicBeasts as the provider), a record which is marked with `IGNORE()` is also deleted.

**To Reproduce**
1. Establish zone with MythicBeasts as the DNS provider
2. In the zone definition, specify records: `IGNORE("testignore"), A("testdefined", "9.9.9.9")`
3. `dnscontrol push`
4. Outside of DNSControl, create an A record for `testignore` (for example, a record used for dynamic DNS, via `https://ipv4.api.mythic-beasts.com/dns/v2/dynamic/testignore.example.com`)
5. `dnscontrol push` again - no modifications are made, and both records remain in place as seen in provider's control panel
6. Modify the definition of `testdefined` to point to a different address, and run `dnscontrol push` again, resulting in the output below
7. Observe that `testignore` is now absent in the provider's control panel and DNS zone

Same zone definition:
```
D("example.com", REG_NONE, DnsProvider(DNS_MYTHIC),
  IGNORE("testignore"),
  A("testdefined", "8.8.8.8")
);
```

Output of DNSControl push when modifying record is:
```
******************** Domain: example.com
----- Getting nameservers from: mythic
----- DNS Provider: mythic...
1 correction (mythic)
#1: 1 records not being deleted because of IGNORE*():
    testignore.example.com. A 8.8.8.8
± MODIFY testdefined.example.com A (9.9.9.9 ttl=300) -> (8.8.8.8 ttl=300)
SUCCESS!
----- Registrar: none...
0 corrections (none)
```

**Expected behavior**
The `IGNORE()`d record should still exist after the other modification is applied.

**DNS Provider**
 - MYTHICBEASTS

Docker image version: https://github.com/stackexchange/dnscontrol/pkgs/container/dnscontrol/302287665?tag=4.14.3

cafferata:
Ping @tomfitzhenry, the maintainer of the Mythic Beasts provider.

tlimoncelli:
I don't think this is a problem with just mythicbeasts.  This might be a problem for any provider that uses the ByZone() function (that is, any updates require uploading the entire zone).  I can think of some other edge-cases that aren't covered.

We may have to disable IGNORE() for anything that uses ByZone() [autodns, bind, mythicbeasts, realtimeregister, sakuracloud].  

[...]

PR #3260, opened 2024-12-17
NAMEDOTCOM: BUGFIX: TXT records add unneeded quotes

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/3259


tlimoncelli:
Gladly!

If you are on a Mac, Linux, or WIndows host, you'll need to install Git and Go.  Git has different install instructions for different OSs.   Go is on https://go.dev/ 

cd to somewhere like /tmp or an otherwise unused directory.

```
git clone https://github.com/StackExchange/dnscontrol.git
cd dnscontrol
git checkout tlim_namedotcom
go install
```

The last command (go install) will install the dnscontrol binary in your $HOME/bin directory.  You should be able to run it by just running "dnscontrol".

Change to the directory with your dnsconfig.js file and run it as normal.

Feel free to ping me with any questions!

tkurki:
Was hoping for a one command docker based build, as I prefer not to taint my host with one time use tools. But thanks, I’ll figure that part out myself if i get around to it.

tlimoncelli:
A docker-based build would be excellent. Sadly, we're not there yet.  (CC @cafferata )

rmc47:
@tkurki while testing #3227, I ran the tests within a Docker container - `docker run -it golang` grabs the base Go container and pops you into an interactive shell within that environment, then you can run Tom's suggested commands within that.

tkurki:
The changes work: with these changes dnscontrol correctly updated an existing record that had unneeded quotes and created from a scratch the original problematic record in #3259.

Thank you for looking into this so quickly.

tlimoncelli:
Glad it worked!  I'll include this change in the next release, due in early January.

PR #3269, opened 2024-12-23
AXFRDDNS: Add support for LOC records

gucci-on-fleek:
I've tested this and it seems to work on the DNS server that I'm running (Knot 3.4.3), so presumably it works on other DNS servers too.

---

Aside: I'm willing to donate secrets for AXFR to the project if you're interested. I'm just running it on [a personal server](https://ns.maxchernoff.ca/), so it should be up most of the time, but it might be down for a couple of hours whenever I'm upgrading something. Or alternatively, it's pretty easy to run [Knot](https://www.knot-dns.cz/), so I could try and set it up to run a server inside each workflow.


cafferata:
Thank you for testing and for your generous offer to donate secrets for AXFR – that’s greatly appreciated! 🙌 We’d be happy to make use of it. You can find all the details on how to share the credentials in our documentation [Donate secrets to the project](https://docs.dnscontrol.org/developer-info/byo-secrets#donate-secrets-to-the-project).

As the holiday season is upon us, I wanted to take a moment to wish you and your loved ones a Merry Christmas and a joyful holiday season. May the days ahead be filled with warmth, happiness, and maybe even a little bit of time to relax.

Looking forward to collaborating more in the new year! 🎄✨

PR #3278, opened 2024-12-30
CICD: Update docs/comments after file renames

tomfitzhenry:
Here's some small fixes for things I noticed while doing https://github.com/StackExchange/dnscontrol/pull/3277 .

I raised these in a separate PR so they're easier to review.

Descriptions for each fix in the commit messages.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -1264,12 +1264,6 @@ func makeTests() []*TestGroup {
 			tc("Create a TXT/SPF", txt("foo", "v=spf1 ip4:99.99.99.99 -all")),
 
 			// Nobody needs this and many APIs don't allow it.
-			tc("TXT with 1 backslash", txt("fooosbs1", `1back\slash`)),
-			tc("TXT with 2 backslash", txt("fooosbs2", `2back\\slash`)),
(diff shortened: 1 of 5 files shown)


cafferata (approved):
Thank you, @tomfitzhenry, for your contribution! 🎉 You’ve given the documentation a great touch-up by fixing those small imperfections. It’s these little details that really make everything shine. 👌🚀

integrationTest/integration_test.go
```diff
@@ -1263,12 +1263,6 @@ func makeTests() []*TestGroup {
 			// Vultr syntax-checks TXT records with SPF contents.
 			tc("Create a TXT/SPF", txt("foo", "v=spf1 ip4:99.99.99.99 -all")),
 
-			// Nobody needs this and many APIs don't allow it.
```
tlimoncelli, line 1266: Why remove this?  I think the comment "Nobody needs this" was sarcasm.  If an API doesn't allow it, the auditrecords.go file should indicate if this needs to be skipped.

tlimoncelli: AH, now I see!  I really shouldn't do reviews from my phone :)

Issue #3279, opened 2025-01-01
Add AXFRDDNS to automated tests


gucci-on-fleek:
## Standard Boilerplate

I would like to volunteer to maintain the credentials and account for the AXFR+DDNS provider.

> - The account/credentials should only access the test domain.

[Yes](https://github.com/gucci-on-fleek/maxchernoff.ca/blob/8cf8a5cbd057746e876b0efac885c66247966fa4/web/knot/config/knot.conf#L62-L64).

> - Renew the domain in a timely manner.

Not applicable, uses an internal domain.

On the more general “availability” topic, the server should be up most of the time, but this is just a single server that hosts my personal website, so it may be down occasionally for updates. Uptime is generally pretty good though, just not quite as good as you'd expect from a commercial provider.

> - Don't do anything that will get you in trouble with your employer, like charging it to your employer without permission.

Not applicable, I'm unemployed.

> To securely send the credentials to the project, use this link: https://transfer.secretoverflow.com/u/tlimoncelli

Done (item `D8N1-RJUK`).

## Configuration

AXFR+DDNS is a little special in that it's a protocol and _not_ a specific provider, so there are quite a few different valid configurations here. I'll go through the [configuration options](https://docs.dnscontrol.org/provider/axfrddns#configuration) and list the valid values for my server. You can either choose to test just one possible valid configuration or multiple combinations of different options.

### `update-mode` and `transfer-mode`

`tcp` and `tcp-tls` both work; `udp` will _not_.

### `transfer-key` and `update-key`

Both are _required_, and both use the same keys. I've generated keys for as many algorithms as possible, so the valid values are:

- `hmac-md5:dnscontrol-testing-md5:[REDACTED]`
- `hmac-sha1:dnscontrol-testing-sha1:[REDACTED]`
- `hmac-sha224:dnscontrol-testing-sha224:[REDACTED]`
- `hmac-sha256:dnscontrol-testing-sha256:[REDACTED]`
- `hmac-sha384:dnscontrol-testing-sha384:[REDACTED]`
- `hmac-sha512:dnscontrol-testing-sha512:[REDACTED]`

### `master`

Must be `ns.maxchernoff.ca`.

### `nameservers` and `transfer-server`

Not necessary, but you can set them to `ns.maxchernoff.ca` if you want.

## Details

Currently, the server is running [Knot](https://www.knot-dns.cz/) v3.4.3; to see what the version is in the future, you can run

```shell
[...]

PR #3290, opened 2025-01-01
AXFRDDNS: Enable automated testing

tlimoncelli:
Fixes https://github.com/StackExchange/dnscontrol/issues/3279

Issue

* Add AXFRDDNS and AXFRDDNS_DNSSEC to the list of automated tests
* integration_test.go doesn't handle testing the same provider twice with different inputs
* AXFRDDNS shouldn't warn about "buggy-cname" setting.

Resolution

* integration_test.go: permit providers.json to list the same provider multiple times; as long as they all set the TYPE correctly
* providers.json: Add AXFRDDNS_DNSSEC stanza
* axfrddnsProvider.go: Add missing field to initAxfrDdns


tlimoncelli:
CC [gucci-on-fleek](https://github.com/gucci-on-fleek)

tlimoncelli:
It turns out the integration tests don't accept 2 profiles with the same provider. I hacked around a bit to make it work.

PR #3294, opened 2025-01-02
AXFRDDNS: Modernize, Update supported capabilities

### LINKED ISSUE #2581: AXFR: message too large
Hello (again),

we are using dnscontrol for a large bunch of zones. The most zones are updated using AXFR. In case of large updates (e.g. changing the ttl for the complete zone) results pretty often in the error "message too large". From my point of view it maybe would make sense to catch this error and retry the operation in small batches? 

Before implementing this change I wanted to discuss if there is maybe any corner case I did not consider?

Thanks!

### LINKED ISSUE #3292: AXRFDNS: CNAME tests fail when DNSSEC is enabled
CC @hnrgrgr and @gucci-on-fleek 

**Describe the bug**

The CNAME tests fail when AXRFDNS uses DNSSEC.

**To Reproduce**

```
go test -v -verbose -profile AXFRDDNS_DNSSEC -start 4 -end 14
...
=== RUN   TestDNSProviders/with-dnssec.dnscontrol.internal/Clean_Slate:Empty
=== RUN   TestDNSProviders/with-dnssec.dnscontrol.internal/04:CNAME:Create_a_CNAME
=== RUN   TestDNSProviders/with-dnssec.dnscontrol.internal/04:CNAME:Change_CNAME_target
    integration_test.go:306: Expected 0 corrections on second run, but found 1.
    integration_test.go:308: UNEXPECTED #0: DDNS UPDATES to 'with-dnssec.dnscontrol.internal' (primary master: 'ns.maxchernoff.ca:53'). Changes:
        + CREATE testcname.with-dnssec.dnscontrol.internal CNAME www.yahoo.com. ttl=300
=== RUN   TestDNSProviders/with-dnssec.dnscontrol.internal/Clean_Slate:Empty#01
...
=== RUN   TestDNSProviders/with-dnssec.dnscontrol.internal/Clean_Slate:Empty#10
=== RUN   TestDNSProviders/with-dnssec.dnscontrol.internal/14:TypeChangeHard:Create_a_CNAME
=== RUN   TestDNSProviders/with-dnssec.dnscontrol.internal/14:TypeChangeHard:Change_to_A_record
=== RUN   TestDNSProviders/with-dnssec.dnscontrol.internal/14:TypeChangeHard:Change_back_to_CNAME
    integration_test.go:306: Expected 0 corrections on second run, but found 1.
    integration_test.go:308: UNEXPECTED #0: DDNS UPDATES to 'with-dnssec.dnscontrol.internal' (primary master: 'ns.maxchernoff.ca:53'). Changes:
        + CREATE foo.with-dnssec.dnscontrol.internal CNAME google2.com. ttl=300
--- FAIL: TestDNSProviders (7.49s)
```

**Expected behavior**
tests should pass.

**DNS Provider**
 - AXFRDDNS_DNSSEC

**Additional context**

To circumvent the issue, I've added code to skip those tests. These lines should be removed when the tests work.

```
                testgroup("CNAME",
+                       not("AXFRDDNS_DNSSEC"),
                        tc("Create a CNAME", cname("testcname", "www.google.com.")),
                        tc("Change CNAME target", cname("testcname", "www.yahoo.com.")),
                ),
@@ -1115,6 +1116,7 @@ func makeTests() []*TestGroup {
                // changes properly for you. Let's verify that we got it right!
 
                testgroup("TypeChangeHard",
+                       not("AXFRDDNS_DNSSEC"),
                        tc("Create a CNAME", cname("foo", "google.com.")),
                        tc("Change to A record", a("foo", "1.2.3.4")),
                        tc("Change back to CNAME", cname("foo", "google2.com.")),
```

gucci-on-fleek:
Here are the commit messages; let me know if you'd rather that I submit these each as individual pull requests. I've also never written anything in Go before, so please let me know if you want me to make any changes. Thanks!

---

## Remove `buggy-cname`

It's semantically invalid for a CNAME record to coexist with other (non-DNSSEC) records, and RFC 2136 §3.4.2.2 requires DDNS recipients to ignore any such updates. Therefore, it should always be safe to unconditionally delete any records for a given name before we add a CNAME record for that name.

`buggy-cname` was only needed as a workaround for a Knot DNS bug/feature that caused it to ignore an additional followed by a deletion to the same CNAME record. However, if we delete *all* records for a given name (instead of just the CNAME records), Knot DNS will process the update as expected. Therefore, we no longer need `buggy-cname`.

This has only been tested with Knot DNS, but given the semantics of CNAME records, this _should_ work with any RFC-compliant DNS server.

Fixes #3292.

## Add support for sha224 and sha384 TSIG algorithms

I doubt that these are used by anyone, but they are supported by `github.com/miekg/dns` and supporting them is trivial.

## Change dummy nameserver addresses

A NS record must be a domain name and not an IP address.

## Change the default value of `update-mode` to `tcp`

AXFR is only defined over TCP, and DNSControl always requests an AXFR before sending an UPDATE, therefore any server supported by DNSControl is already capable of handling TCP connections. The only advantage to using UDP for DNS is that it has lower latency, but that's only relevant for queries, not updates. By using TCP for updates, we should be more resilient to network issues, and we should be able to handle larger updates.

(Maybe?) fixes #2581.

## Update supported capabilities

- `CanConcur -> Can`: The integration tests all pass, so presumably that means that this is fine?

- `CanUseDNAME -> Can`: Supported by the DDNS protocols and all the integration tests pass. I've only tested this with Knot DNS, so it might not work with some other server implementations.

- `CanUseDS -> Can`: Ditto.

- `CanUseAlias -> Cannot`: Impossible to support using the DDNS protocol; see the code comments for more details.

- `CanUseSOA -> Cannot`: Supported by the DDNS protocol, but doesn't work right now.

- `CanUseDNSKEY -> Cannot`: Ditto, but this might just be a bug in the Knot DNS server.

- `DocCreateDomains -> Cannot`: Possible to add support via catalog zones, but not implemented yet.

- `CanGetZones -> Cannot`: Ditto.


tlimoncelli:
CC @fritterhoff  @hnrgrgr for visibility.

This is fantastic!  Thanks for modernizing the provider!

> Here are the commit messages; let me know if you'd rather that I submit these each as individual pull requests. I've also never written anything in Go before, so please let me know if you want me to make any changes. Thanks!

Yes, big PRs are ok.  I only split things out if I feel that there may be a need to revert specific changes.

> ## Remove `buggy-cname`
> It's semantically invalid for a CNAME record to coexist with other (non-DNSSEC) records, and RFC 2136 §3.4.2.2 requires DDNS recipients to ignore any such updates. Therefore, it should always be safe to unconditionally delete any records for a given name before we add a CNAME record for that name.
> 
> `buggy-cname` was only needed as a workaround for a Knot DNS bug/feature that caused it to ignore an additional followed by a deletion to the same CNAME record. However, if we delete _all_ records for a given name (instead of just the CNAME records), Knot DNS will process the update as expected. Therefore, we no longer need `buggy-cname`.
> 
> This has only been tested with Knot DNS, but given the semantics of CNAME records, this _should_ work with any RFC-compliant DNS server.
> 
> Fixes #3292.

Is there a chance that some software other than Knot DNS would need this workaround? (Or... How many CNAMEs would not Knot delete if a not Knot not could not CNAMEs?  j/k)

> ## Make `transfer-mode`/`update-mode` configurable for integration tests
> ## Add support for sha224 and sha384 TSIG algorithms
> I doubt that these are used by anyone, but they are supported by `github.com/miekg/dns` and supporting them is trivial.

Concurr.

> ## Change dummy nameserver addresses
> A NS record must be a domain name and not an IP address.

Good point!

> ## Change the default value of `update-mode` to `tcp`
> AXFR is only defined over TCP, and DNSControl always requests an AXFR before sending an UPDATE, therefore any server supported by DNSControl is already capable of handling TCP connections. The only advantage to using UDP for DNS is that it has lower latency, but that's only relevant for queries, not updates. By using TCP for updates, we should be more resilient to network issues, and we should be able to handle larger updates.
> 
> (Maybe?) fixes #2581.

Sounds good!

> ## Update supported capabilities
> * `CanConcur -> Can`: The integration tests all pass, so presumably that means that this is fine?

Should there be a mutex around `.rand`?

> * `CanUseDNAME -> Can`: Supported by the DDNS protocols and all the integration tests pass. I've only tested this with Knot DNS, so it might not work with some other server implementations.
> * `CanUseDS -> Can`: Ditto.
> * `CanUseAlias -> Cannot`: Impossible to support using the DDNS protocol; see the code comments for more details.
> * `CanUseSOA -> Cannot`: Supported by the DDNS protocol, but doesn't work right now.
> * `CanUseDNSKEY -> Cannot`: Ditto, but this might just be a bug in the Knot DNS server.
> * `DocCreateDomains -> Cannot`: Possible to add support via catalog zones, but not implemented yet.
> * `CanGetZones -> Cannot`: Ditto.

Sounds good!

Thanks!

gucci-on-fleek:
> > ## Remove `buggy-cname`
> > [...]
> 
> Is there a chance that some software other than Knot DNS would need this workaround? (Or... How many CNAMEs would not Knot delete if a not Knot not could not CNAMEs? j/k)

Maybe, but I doubt it. RFC 2136 is somewhat vague and CNAMEs have such weird semantics that I'm not entirely sure if Knot's behaviour here is a bug or intentional, and looking through Knot's source didn't clear anything up.

 I _think_ that the reason that this was needed is that Knot only processes the first CNAME update for a name, so when we would send the update

```zone
test-cname  0    NONE  CNAME old-value.example.  ;; delete this value
test-cname  300  IN    CNAME new-value.example.  ;; add this value
```

Knot would see that as two CNAME records for the `test-cname` name, so it ignores the second value. But if we delete all the records, then we're sending this as an update

```zone
test-cname  0    ANY   ANY    old-value.example.  ;; delete all records
test-cname  300  IN    CNAME  new-value.example.  ;; add this value
```

so a CNAME doesn't appear twice. This behaviour kinda makes sense, so I would blindly guess that all implementations would work with either the new way or the old way and that `buggy-cname` shouldn't be needed any more. Probably.

> > ## Update supported capabilities
> > 
> > * `CanConcur -> Can`: The integration tests all pass, so presumably that means that this is fine?
> 
> Should there be a mutex around `.rand`?

`crypto/rand` should be thread-safe (I think?), and `github.com/miekg/dns` defaults to that if we omit the `Id`, so I removed that entirely in 808e1d17.

tlimoncelli:
Sounds good!

For someone that has never written anything in Go before, you've done a bang-up job!  Congrats!

tlimoncelli (approved):

fritterhoff:
Hi guys! Sorry for the (too) late response. 

The changes look great but in my case they cause some problems.
I get more or less randomly errors. @tlimoncelli is there a chance that I send you an email with details? Would like to avoid pasting the complete log here...

I would guess this comes do to the usage of several AXFR providers concurrently. Any way what I noticed is that an error in any concurrent action does not result in an exit code != 0. From my point of view before using concurrent actions an failure resulted in an exit code 1 and thus "broke" our CI tooling.

gucci-on-fleek:
@fritterhoff

> The changes look great but in my case they cause some problems. [...] I would guess this comes do to the usage of several AXFR providers concurrently.

Ok, so I guess I was a little too ambitious in my changes:

> ## 24bdc758d0ea1e894789d56c2de4949c9aeb4bc4 Update supported capabilities
> 
>  * `CanConcur -> Can`: The integration tests all pass, so presumably that means that this is fine?
> 
> [...]

Does switching back to non-concurrent fix things for you?

```diff
diff --git a/providers/axfrddns/axfrddnsProvider.go b/providers/axfrddns/axfrddnsProvider.go
index 2183306a..7e2e5d45 100644
--- a/providers/axfrddns/axfrddnsProvider.go
+++ b/providers/axfrddns/axfrddnsProvider.go
@@ -37,7 +37,7 @@ var features = providers.DocumentationNotes{
 	// The default for unlisted capabilities is 'Cannot'.
 	// See providers/capabilities.go for the entire list of capabilities.
 	providers.CanAutoDNSSEC:          providers.Can("Just warn when DNSSEC is requested but no RRSIG is found in the AXFR or warn when DNSSEC is not requested but RRSIG are found in the AXFR."),
-	providers.CanConcur:              providers.Can(),
+	providers.CanConcur:              providers.Cannot(),
 	providers.CanUseCAA:              providers.Can(),
 	providers.CanUseDHCID:            providers.Can(),
 	providers.CanUseDNAME:            providers.Can(),
```

fritterhoff:
Will check tomorrow if changing the BIND9 settings works for our use case of if this has any other (negativ) impact. 

@tlimoncelli are there maybe some plans to provide some limiting (e.g. maximum amount of parallel actions?)

providers/axfrddns/axfrddnsProvider.go
```diff
@@ -52,32 +50,39 @@ var features = providers.DocumentationNotes{
 	providers.CanUseSSHFP:            providers.Can(),
 	providers.CanUseSVCB:             providers.Can(),
 	providers.CanUseTLSA:             providers.Can(),
-	providers.DocCreateDomains:       providers.Cannot(),
 	providers.DocDualHost:            providers.Cannot(),
 	providers.DocOfficiallySupported: providers.Cannot(),
+	// Possible to support via catalog zones (RFC 9432), but those are not
+	// directly supported by DNSControl right now (although nothing is stopping
+	// you from manually updating a catalog zone using DNSControl if you wish).
+	providers.CanGetZones:      providers.Cannot(),
```
hnrgrgr, line 58: What's the meaning of  `CanGetZones` ? If it means that the `get-zone` command is available, then, it is available with `AXFR`, isn't it ?

integrationTest/providers.json
```diff
@@ -18,21 +18,23 @@
   },
   "AXFRDDNS": {
     "TYPE": "AXFRDDNS",
-    "buggy-cname": "$AXFRDDNS_BUGGY_CNAME",
     "domain": "$AXFRDDNS_DOMAIN",
     "master": "$AXFRDDNS_MASTER",
     "nameservers": "ns.example.com",
     "transfer-key": "$AXFRDDNS_TRANSFER_KEY",
-    "update-key": "$AXFRDDNS_UPDATE_KEY"
+    "transfer-mode": "AXFRDDNS_TRANSFER_MODE",
```
hnrgrgr, line 25: There is a missing `$` here, I've opened #3317 to fis this one.

hnrgrgr (commented):
Thanks for this PR and for simplifying the handling of `CNAME`. I was AFK the last week and unable to review the changes. I opened #3317 with minor fixes.

providers/axfrddns/axfrddnsProvider.go
tlimoncelli: CanGetZones  means that the provider is able to get a list of zones at the server, and therefore know if a zone doesn't exist and needs to be created.

Each provider tends to have a bespoke way to do that. Since AXFRDDNS is generic and works with many providers, I think it is best to leave that as Cannot().  (Though, I'd be glad to be wrong!)

tlimoncelli:
[fritterhoff](https://github.com/fritterhoff) Re:  BIND's limits

Rate limiting is a bit of a challenge. 

The code uses [sync.WaitGroup](https://pkg.go.dev/sync#WaitGroup) which is pretty simple. All concurrent sessions start at the same time and we leave it up to the provider code to detect errors and 429 "please slow down" errors and respond by pausing then retrying.

Two ideas:

1. Update AXFRDDNS to detect BIND's error; then pause and re-try.  (Search for "retry" in the providers for example code)
2. Add a command line flag that restricts the concurrency either globally or on a per-provider basis.  (I'm not an expert in  Go's concurrency and locking features but... how hard could it be?)
3.  Or just set `providers.CanConcur()` to `providers.Cannot()` and revisit this some other time.

Would you please open a new bug to discuss this? (if one doesn't exists yet)

fritterhoff:
Mhm. BIND does not respond with sensefull errors. We simply get a denied so detecting a config error vs. ratelimit is more or less impossible

tlimoncelli:
> Mhm. BIND does not respond with sensefull errors. We simply get a denied so detecting a config error vs. ratelimit is more or less impossible

Oh drat.   That means that if we add a maxconcurrency flag, we could still have the same problem.

tlimoncelli:
There are 2 ways to do this: a global flag or a per-provider flag. 

The global flag would be the easiest.  You can then use https://github.com/nozzle/throttler

Per-provider flag would be difficult.  I recommend using key:value in the creds.json file to pass the max concurrency limit to the provider. It would appear in dc.Metadata[].  Rewriting the scheduler to handle each provider differently would be a big change.

tlimoncelli:
I don't have access to a test system for AXFRDDNS. I would appreciate it if someone could test the change and report back.

Please test https://github.com/StackExchange/dnscontrol/pull/3764 (branch `tlim/edge102_connlimit`).  It adds a flag called `--cmax` which limits concurrency.  I recommend `--cmax 5` then work upwards.

Instructions on testing a branch is here: https://docs.dnscontrol.org/developer-info/test-a-branch

PR #3265, opened 2024-12-19
DOCS: Add dep updates steps to rel-eng

tlimoncelli:
Issue:

* The docs to update dependencies are later in the doc.

Resolution:

* Copy them to "step 0" for easier finding.

PR #3306, opened 2025-01-05
DOCS: Release improvements

cafferata:
Thanks for pointing this out! As you mentioned https://github.com/StackExchange/dnscontrol/pull/3301:

> “Release engineering docs should clarify that they should start in main. Also add fmtjson instructions.”

Initially, the docs did clarify that releases should start in main. However, this changed when `Step 0` was added via https://github.com/StackExchange/dnscontrol/pull/3265.

To resolve this, I suggest moving the duplicate instructions into a “Prepare the Git Repository” section to streamline the process and avoid redundancy. You can check the current state of the docs in the preview here: [Release Engineering Docs Preview](https://docs.dnscontrol.org/~/revisions/VD64bb6k0kj9TcNrp3Cc/release/release-engineering).

Let me know if that works for you, or if you have any other suggestions!

### DIFF
--- a/documentation/release-engineering.md
+++ b/documentation/release-engineering.md
@@ -7,10 +7,16 @@ GitHub Actions (GHA) will do most of the work for you. You will need to edit the
 Please change the version number as appropriate.  Substitute (for example)
 `v4.2.0` any place you see `$VERSION` in this doc.
 
-## Step 0. Update dependencies
+## Prepare the git repository
 
 ```shell
 git checkout main
+git pull
+```
+
+## Step 0. Update dependencies
+
+```shell
 git checkout -b update_deps
 go install github.com/oligot/go-mod-upgrade@latest
 go-mod-upgrade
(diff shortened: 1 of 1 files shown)


tlimoncelli:
I see your point but this situation benefits from some duplication.

Each step needs to not depend on the previous steps.

* The steps are often used independently, outside of the official release process.
* When these steps *are* used for an official release, they often happen over many hours and I can't expect to be in the right branch when I come back to my computer to continue.

I have had recent mess-ups due to being in the wrong branch.  I want to prevent that in the future.

Having the "git checkout main" at the top also makes copying and pasting the commands as a batch much easier.  If I don't need the checkout, I can copy the other commands.  Having to type one command manually breaks my rhythm.

Thank you for understanding.

Tom

cafferata:
Thank you for your detailed response! I understand your perspective, and it makes sense to keep the steps independent for flexibility and to avoid potential issues during the release process.

If the current version works well for you, that’s great to hear. I’ll go ahead and close the GitHub pull request. Please don’t hesitate to reach out if there’s anything else I can assist with in the future.

Issue #3007, opened 2024-06-12
Preview of records if zone doesn't exist yet


Koopzington:
Could we make `preview` display the records that would be created even if the zone doesn't exist yet?
Right now (at least for the Hetzner DNS Provider if that's relevant) i only get to see the created records when i `push` and let dnscontrol automatically create the zone. It would be nice if we could review them in the `preview` output first though.

tlimoncelli:
Hi there!

I'm having difficulty reproducing this.  Can you send me sample output?

Also, please include the output of `dnscontrol version`


Thanks!

Koopzington:
```
var REG_NONE = NewRegistrar('none')
var DSP_HETZNER = NewDnsProvider('hetzner')

D('example.org', REG_NONE, DnsProvider(DSP_HETZNER),
    A('@', '127.0.0.1')
)
```

![image](https://github.com/StackExchange/dnscontrol/assets/3015529/b9b38713-36fa-4bbc-a8dc-f6f74a73ddaf)

tlimoncelli:
Would you please try again with the `ppreview` command?  I have a hunch this bug may not appear in the new code.

(ppreview is the future replacement for preview. it is a rewrite from scratch with many new features)

Koopzington:
![image](https://github.com/StackExchange/dnscontrol/assets/3015529/ae41ae89-0b65-4ef1-b5cd-f7ca837ce6e0)

tlimoncelli:
CC @das7pad

das7pad:
On the phone, I can check in detail when I'm back late next week.

@Koopzington can you try again with the latest version of dnscontrol and a different, random domain name?

tlimoncelli:
I just tested this with the BIND provider and I get a full preview.   I wanted to verify that this isn't a global problem.

I look forward to @Koopzington 's results.  I'm optimistic.

das7pad:
I can reproduce the report with a random domain. It looks like something is not quite right in dnscontrol.

```js
var REG_NONE = NewRegistrar('none')
var DSP_HETZNER = NewDnsProvider('HETZNER')

D('testing-2024-12-27.dev', REG_NONE, DnsProvider(DSP_HETZNER),
    A('@', '127.0.0.1')
)
```
HEAD (6d5bfe952041979bd915830df18c317c25368e72)
```
$ dnscontrol preview
CONCURRENTLY gathering 1 zone(s)
SERIALLY gathering 0 zone(s)
[...]

PR #3328, opened 2025-01-08
ROUTE53: make caching of zones thread-safe

das7pad:
Hi @tresni!
While reviewing all the `ZoneCreator` implementations, I noticed that the ROUTE53 provider has an unsafe caching implementation for zones and `EnsureZoneExists` has a race-condition bug. `EnsureZoneExists` resets the cache before making the API call for creating a given zone. This might run concurrently to the processing of other zones and in turn could result in unexpected behavior: the cache could get re-populated before the creation of the zone completes and the new zone could be missing from the newly populated cache. This PR is making all access to the zone cache thread-safe and fixing the aforementioned race-condition. Would you mind giving this a try and let me know how it goes? Thanks!

Bonus: The zone cache is no longer reset when creating zones. Instead, the cache is populated with the zone that is included in the API response from creating the zone.

### DIFF
--- a/providers/route53/route53Provider.go
+++ b/providers/route53/route53Provider.go
@@ -9,14 +9,10 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 	"unicode/utf8"
 
-	"github.com/StackExchange/dnscontrol/v4/models"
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Friendly reminder to rebase so that this can be merged.  Thanks!

tlimoncelli:
Thanks!

PR #3330, opened 2025-01-08
CLOUDFLARE: fix handling of zone cache when creating zone

das7pad:
Hi @tresni,

While reviewing all the `ZoneCreator` implementations, I noticed that the CLOUDFLARE provider has a bug in the clearing of the zone cache in `EnsureZoneExists`. `EnsureZoneExists` uses `clear` on the map that is used as zone cache, which merely deletes all the entries in the map. Any following nil-checks on the cache will deem the cache as populated still and not trigger populating of the cache again. This PR is fixing this bug and also avoiding the reset by using the zone as included in the API response from creating the zone. Would you mind giving this a try and let me know how it goes? Thanks!

### DIFF
--- a/providers/cloudflare/cloudflareProvider.go
+++ b/providers/cloudflare/cloudflareProvider.go
@@ -12,14 +12,15 @@ import (
 
 	"golang.org/x/net/idna"
 
+	"github.com/cloudflare/cloudflare-go"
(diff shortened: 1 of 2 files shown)


das7pad:
Closing in favor of https://github.com/StackExchange/dnscontrol/pull/3373

PR #3334, opened 2025-01-08
INWX: populate zone cache after creating zone

das7pad:
Hi @patschi!
While reviewing all the `ZoneCreator` implementations, I noticed that the INWX provider has an incomplete caching implementation for zones. The provider is populating the cache once on first access. Any zones that are created will not be readable in the same life-cycle of dnscontrol. This PR is populating the zone cache after creating a zone. Would you mind giving this a try and let me know how it goes? Thanks!

### DIFF
--- a/providers/inwx/inwxProvider.go
+++ b/providers/inwx/inwxProvider.go
@@ -7,12 +7,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/nrdcg/goinwx"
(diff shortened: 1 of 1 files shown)


eliheady:
I've tested this also - I have found no problems with the change.

das7pad:
Thanks for giving this a try, @eliheady!

@tlimoncelli this is ready to merge.

PR #3335, opened 2025-01-08
LUADNS: populate zone cache after creating zone

das7pad:
Hi @riku22!

While reviewing all the `ZoneCreator` implementations, I noticed that the LUADNS provider has an incomplete caching implementation for zones. The provider is populating the cache once on first access. Any zones that are created will not be readable in the same life-cycle of dnscontrol. This PR is populating the zone cache after creating a zone. Would you mind giving this a try and let me know how it goes? Thanks!


riku22:
Thank you for the fix.

Lgtm.

tlimoncelli:
@das7pad Please rebase.

PR #3336, opened 2025-01-08
HETZNER: populate zone cache after creating zone

das7pad:
Similar to the other providers, the HETZNER provider is now populating the zone cache after creating a zone as well (and saving a round-trip to the provider by dropping the need to fill the zone cache again).


das7pad:
Closing in favor of https://github.com/StackExchange/dnscontrol/pull/3372

PR #3364, opened 2025-01-14
FEATURE: Extend PTR magic handling to support RFC4183 names

tomhughes:
This extends magic mode in PTR records to support RFC4183 names so that things like this will work:

```js
REVCOMPAT("rfc4183");

D(REV("9.9.9.128/25"), REGISTRAR, DnsProvider(BIND),
  PTR("9.9.9.129", "first.example.com."),
);
```

and generate the name `129` in the `128-25.9.9.9.in-addr.arpa` zone.


tlimoncelli:
Fantastic!  Thanks for keeping us up to date!

PR #3365, opened 2025-01-15
FEAT: Add ZoneCache primitive

das7pad:
This PR is kicking off a shared, thread-safe "zone cache" implementation for all the drivers to use instead of the various implementations out there today. The new zone cache uses Go-generics to let each driver choose their zone struct and retain type-safety. See https://github.com/StackExchange/dnscontrol/pull/3337 and the below commits for more details.

I've left out the provider changes. Let me know if you want to ship them here already:
- HETZNER driver c7789ccd886d317f98ba5a042d87a517a5f6f297 (chained onto tweaks in https://github.com/StackExchange/dnscontrol/pull/3336)
- CLOUDFLARE driver 85e796586f61690ffc39aa19eebad2b04b06662c (chained onto fixes in https://github.com/StackExchange/dnscontrol/pull/3330)

### DIFF
--- a/pkg/zoneCache/zoneCache.go
+++ b/pkg/zoneCache/zoneCache.go
@@ -1,0 +1,86 @@
+package zoneCache
+
+import (
+	"errors"
+	"sync"
+)
+
+func New[Zone any](fetchAll func() (map[string]Zone, error)) ZoneCache[Zone] {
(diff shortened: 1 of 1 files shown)


tlimoncelli (approved):

tlimoncelli:
This looks great! Thanks!

I'll merge this and we can do the other providers in other PRs.

PR #3368, opened 2025-01-15
CLOUDFLAREAPI: No longer treat TTL=300 as special

tlimoncelli:
# Issue

Cloudflare used to treat TTL=1 and TTL=300 as "magic".  300 is no longer magic.

# Resolution

Do not treat TTL=300 as special.

### DIFF
--- a/documentation/provider/cloudflareapi.md
+++ b/documentation/provider/cloudflareapi.md
@@ -413,6 +413,6 @@ A TTL of 0 tells DNSControl to use the default TTL for that provider, which is 1
 In summary:
 * TTL of 0 and 1 are the same ("auto TTL").
 * TTL of 2-60 are all the same as 60.
-* TTL of 61-299, and 301 to infinity are not magic.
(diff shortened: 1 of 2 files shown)


documentation/provider/cloudflareapi.md
```diff
@@ -404,15 +404,14 @@ When `-cfworkers=false` is set, tests related to Workers are skipped.  The Accou
...
 
 If the TTL isn't set to 1, Cloudflare has a minimum of 1 minutes.
 
 A TTL of 0 tells DNSControl to use the default TTL for that provider, which is 1.
 
 In summary:
-* TTL of 0, 1 and 300 are all the same ("auto TTL").
+* TTL of 0 and 1 are the same ("auto TTL").
 * TTL of 2-60 are all the same as 60.
 * TTL of 61-299, and 301 to infinity are not magic.
```
SukkaW, line 416:
```suggestion
* TTL of 61 to infinity is not magic.
```

Also update this line~

PR #3376, opened 2025-01-15
GCLOUD: Re-try on 502 errors

tlimoncelli:
# Issue

502 errors are retry-able after 30 seconds

Rate limit logic error: only kicks in when `if err != nil {` which I believe is reversed.

# Resolution

* retry 502 errors exactly once
* Check for the need to rate limit if `if err == nil {`


tlimoncelli:
CC @riyadhalnur

tlimoncelli:
Hi @riyadhalnur  

I added rate limiting in db6ad08f5e8fb28bf1aa1196d2ad25ed1f83e448 and I think I got the logic wrong. It only kicks in if there was NO error.  Does the new code look more correct?

Thanks!

tlimoncelli:
I've observed it kick in and it seems to work.  

Sorry for the rush but I'm going to merge this. I'm trying to get a bunch of PRs merged so I can do a release tomorrow.  I can fix any problems in follow-up PRs.

PR #3391, opened 2025-01-17
CNR: Initial Performance improvement; golint review

KaiSchwarz-cnic:
Hey friends ;-)

* refactored the code to fix issues reported by golint
* dropped one API Request which was in place to identify a exact number of resource records available to fetch all of them in the next request. now the max limit is hardcoded with 10000 which should be fine for integration tests
* refactored EnsureZoneExists method to immediately request a dnszone creation and to deal with the result instead of prechecking and adding the zone. Again, one less api call. This should not have an impact on the run time of integration tests.
* reviewed debugmode handling and documentation

Integration Tests runtime on our end:

* originally: 325.560s
* this PR: 241.808s

Thanks!

### DIFF
--- a/documentation/provider/cnr.md
+++ b/documentation/provider/cnr.md
@@ -19,7 +19,11 @@ Example:
     "apilogin": "your-cnr-account-id",
     "apipassword": "your-cnr-account-password",
     "apientity": "LIVE", // for the LIVE system; use "OTE" for the OT&E system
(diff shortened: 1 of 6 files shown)


tlimoncelli:
This ran in 6m 54s!  Fantastic!

Thank you so much for the fast response!

KaiSchwarz-cnic:
@tlimoncelli you guys may update the issue with this run time - that leaves already a far better impression to our provider plugin to readers :-)

PR #3441, opened 2025-02-18
refactor(RecordConfig): Use ChangeType() instead of assignment

tlimoncelli:
# Issue

Code that changes ALIAS to CNAME, or SPF to TXT, etc. is incompatible with future RecordConfig changes.

# Resolution

Refactor such changes to use a function rc.ChangeType() which will be rewritten when RecordConfig changes.


tlimoncelli:
CC @ppmathis (bunnydns), @tresni (cloudflare), @vojtad (dnsmadeeasy), @TomOnTime (gandiv5), @xddxdd (gcore), @rblenkinsopp (hedns), @systemcrash (loopia), (nobody) (namedotcom)

This is a no-op change right now.  In a future release, the ChangeType() method will do this in a way that is compatible with a change we plan on making to how RecordConfig is handled.

PR #3444, opened 2025-02-19
Improve tagged domain handling in support of Split Horizon feature

eliheady:
This change corrects the pattern matching of `--domains=` to use the unique name of each domain, ensuring that tagged domains are matched when specified in the filter. Tests for whichZonesToProcess are included.


tlimoncelli:
Looks great!  Thank you for writing unit tests!

Next step: Update the `documentation/preview-push.md` and `documentation/language-reference/top-level-functions/D.md` (if needed)

Is there a need to match all tags?  I was thinking `example.com!` or maybe `example.com!*` might be good syntax for that.  (The docs say that `example.com!` and `example.com` are equivalent, but my testing says they aren't.)

eliheady:
I added the wildcard tag matching feature, and adjusted the checking to ensure `'example.com!'` equals `'example.com'` as stated in the docs (so `example.com!*` is the wildcard pattern, and we aren't introducing selection ambiguity when `example.com` and `example.com!foo` both exist in the configuration)

commands/ppreviewPush_test.go
```diff
@@ -0,0 +1,146 @@
...
+				filter: "example.com!george,example.com!john",
+			},
+			want: []*models.DomainConfig{dcTaggedGeorge, dcTaggedJohn},
+		},
+		{
+			name: "testMultiFilterTaggedNoMatch",
+			why:  "Should return nothing",
+			args: args{
+				dc:     allDC,
+				filter: "example.com!ringo",
```
tlimoncelli, line 54: Poor Ringo is always being left out! :-)

documentation/language-reference/top-level-functions/D.md
```diff
@@ -86,11 +86,12 @@ may have noticed this mistake, but will your coworkers?  Will you in
 six months? You get the idea.
 
 DNSControl command line flag `--domains` matches the full name (with the "!").  If you
-define domains `example.com!george` and `example.com!john` then:
+define domains `example.com!george`, `example.com!paul` and `example.com!john` then:
 
 * `--domains=example.com` will not match either domain.
 * `--domains='example.com!george'` will match only match the first.
```
tlimoncelli, line 92: minor nit: Change 'first' to 'george'

documentation/language-reference/top-level-functions/D.md
```diff
@@ -86,11 +86,12 @@ may have noticed this mistake, but will your coworkers?  Will you in
 six months? You get the idea.
 
 DNSControl command line flag `--domains` matches the full name (with the "!").  If you
-define domains `example.com!george` and `example.com!john` then:
+define domains `example.com!george`, `example.com!paul` and `example.com!john` then:
 
 * `--domains=example.com` will not match either domain.
```
tlimoncelli, line 91:
```suggestion
* `--domains=example.com` will not match any of the three.
```

documentation/language-reference/top-level-functions/D.md
```diff
@@ -86,11 +86,12 @@ may have noticed this mistake, but will your coworkers?  Will you in
 six months? You get the idea.
 
 DNSControl command line flag `--domains` matches the full name (with the "!").  If you
-define domains `example.com!george` and `example.com!john` then:
+define domains `example.com!george`, `example.com!paul` and `example.com!john` then:
```
tlimoncelli, line 89: As a Beatles fan, I request the ordering be !john then !paul then !george.

```suggestion
define domains `example.com!john`, `example.com!paul` and `example.com!george` then:
```

documentation/preview-push.md
```diff
@@ -49,10 +49,13 @@ OPTIONS:
...
-    Typically all domains are included in `preview`/`push`. Wildcards are not
-    permitted except `*` at the start of the entry. For example, `--domains
+    Typically all domains are included in `preview`/`push`. Wildcards are 
+    permitted `*` at the start of the entry. For example, `--domains
     example.com,*.in-addr.arpa` would include `example.com` plus all reverse lookup
-    domains.
+    domains. A wildcard suffix is permitted if the domain contains a tag, and
+    indicates all configured tags of that domain should be selected. A wildcard
+    tagged domain `--domains=example.com!*` would include `example.com!foo` and
+    `example.com!bar` but not `example.com`.
```
tlimoncelli, line 58: Suggestion: Improve readability by splitting it into multiple bullets:

```
* Specifies a comma-separated list of domains to include. Example: `--domains example.com,myexample.net`
* Domains may include a wildcard at the beginning.  For example, `--domains example.com,*.in-addr.arpa` would include `example.com` plus all IPv4 reverse lookup domains.
* Matching includes tags. If the domains are `example.com!foo` and `example.com!bar`, then `--domains example.com!foo` would match the first one, and `--domains example.com` will not match either.
* A wildcard tag is permitted and indicates all configured tags of that domain should be selected. Example: `--domains=example.com!*` would match `example.com!foo` and  `example.com!bar` but not `example.com`.
* If `--domains` is not specified, the default is all domains.
```

commands/ppreviewPush_test.go
```diff
@@ -0,0 +1,146 @@
...
+			name: "testFilterEmptyTag",
+			why:  "Should return untagged domain",
+			args: args{
+				dc:     allDC,
+				filter: "example.com!",
+			},
+			want: []*models.DomainConfig{dcNoTag},
+		},
+		{
+			name: "testFilterNoTagNoMatch",
```
tlimoncelli, line 77: Move this to the end. That way all the `dc: allDC,` items are grouped together.

commands/ppreviewPush_test.go
```diff
@@ -0,0 +1,146 @@
...
+			args: args{
+				dc:     allDC,
+				filter: "example.com",
+			},
+			want: []*models.DomainConfig{dcNoTag},
+		},
+		{
+			name: "testFilterTaggedNoMatch",
+			why:  "Should return nothing",
+			args: args{
```
tlimoncelli, line 97: Move this to the end. That way all the `dc: allDC,` items are grouped together.

documentation/language-reference/top-level-functions/D.md
eliheady: Of course. I don't know what I was thinking

eliheady:
I have some changes still to work out around the handling of `example.com! == example.com` 

> The docs say that example.com! and example.com are equivalent, but my testing says they aren't.

Same here. Should they be treated as equivalent?  Should this be a doc bug (ie, change the docs) or config parsing bug (ie, fix the implementation)? Maybe this used to work and there are users that haven't noticed the problem yet?

Current behavior:

```
D("example.com",
  DnsProvider(r53),
  A("@", "198.51.100.2");
);

D("example.com",
  DnsProvider(r53),
);
```

Leads to the intuitive `executing dnsconfig.js: example.com is declared more than once`

```
D("example.com",
  DnsProvider(r53),
  A("@", "198.51.100.2");
);

D("example.com!",
  DnsProvider(r53),
);
```
Does not treat the two as the same domain and will try to operate on the empty tagged domain - in this example deleting the A record.

tlimoncelli:
I don't have a strong opinion on this. I don't personally use the split horizon feature.  That said...

Is there anything bad about treating "example.com" and "example.com!" as equivalent?

Whatever we decide, it should be consistent.  The same behavior for D() warning "duplicate domain" and `--domains` filtering should act the same way.

eliheady:
I'm not using it either but I can see how it will be useful.

I'm exploring the config parsing to see where enforcement of this rule should go to make sure the configuration is consistent.

[This seems to work](https://github.com/StackExchange/dnscontrol/commit/3f5a1697c1228f492ae36d71aab77fc8bdd88252#diff-0aaa8d01222a16e928693eac5547154bf5452e7b68dbde90de40e8cb4565e444L151-R163), but I'm not sure it is the right place for the duplicate domain check. I'll keep exploring.

```
diff --git a/pkg/js/helpers.js b/pkg/js/helpers.js
index c5ea9f07..3ae20a1a 100644
--- a/pkg/js/helpers.js
+++ b/pkg/js/helpers.js
@@ -148,8 +148,19 @@ function D(name, registrar) {
         var m = arguments[i];
         processDargs(m, domain);
     }
-    if (conf.domain_names.indexOf(name) !== -1) {
-        throw name + ' is declared more than once';
+
+    if (name.slice(-1) === "!") {
+        taggedName = name.slice(0, name.length - 1);
+    } else {
+        taggedName = "";
+    }
+
+    if (conf.domain_names.indexOf(name) !== -1 || conf.domain_names.indexOf(taggedName) !== -1) {
+        message = name + ' is declared more than once';
+        if (taggedName && taggedName.length > 0) {
+            throw message += ' (' + taggedName + ')';
+        }
+        throw message;
     }
     conf.domains.push(domain);
     conf.domain_names.push(name);
```

eliheady:
Putting a simple suffix trim in [validation.go](https://github.com/StackExchange/dnscontrol/commit/c49b1ccc8b4ca2b6f212f74a7b65941e07add635#diff-066235539e86f0951b67023d3d557b7407f741b3aa6fe0a034912aead53cea38R567) to remove the "!" when comparing for uniqueness gets the expected error when both `example.com` and `example.com!` are in the configuration. This is what one sees with the adjusted validation check in validate.go

```
2025/02/20 16:25:12 1 Validation errors:
2025/02/20 16:25:12 ERROR: duplicate domain name: "example.com"
exiting due to validation errors
```

And this is the error if the check is also in helpers.js (plus it seems the error happens prior to any provider interaction)

```
executing dnsconfig.js: intarwebs.com! is declared more than once (example.com)
```

It is maybe a bit belt-and-suspenders, but as I don't know in what cases the input could slip by the check in helpers.js it seems good to include both validation checks.

[helpers.js](https://github.com/StackExchange/dnscontrol/commit/3f5a1697c1228f492ae36d71aab77fc8bdd88252#diff-0aaa8d01222a16e928693eac5547154bf5452e7b68dbde90de40e8cb4565e444L151-R163)
[validate.go](https://github.com/StackExchange/dnscontrol/commit/c49b1ccc8b4ca2b6f212f74a7b65941e07add635#diff-066235539e86f0951b67023d3d557b7407f741b3aa6fe0a034912aead53cea38R567)

If this looks ok I can include it in this PR unless it should be submitted on its own (as it is more general than the original scope of this `--domains` filtering issue)

tlimoncelli:
Wow!  Thanks for researching that!  Yes, it is belt-and-suspenders but that's ok. (Validate is kind of a mess that needs to be refactored some day, so what's a little redundancy between friends).

Yes, please add that to the PR.

eliheady:
This is sprawling a bit. I don't think it is quite ready though and I plan to work on it a bit more sometime in the next couple of days.

pkg/js/parse_tests/037-splithor.json
eliheady, line 1: A lot of noise here ... I tried installing 'fmtjson' node package and it doesn't seem to come with a cli utility. Maybe the guide meant https://github.com/mohae/fmtjson? I'll try it also... fwiw this is what `jsonlint` does to the parse test json files.

eliheady: 🤦 I found it. https://github.com/StackExchange/dnscontrol/blob/main/bin/fmtjson

tlimoncelli:
`fmtjson` is in the "bin" directory of the repo.  I will clarify that in the docs.

The sprawl is ok as long as it doesn't totally derail you.  This is the kind of ambiguity we need to fix and I appreciate you taking a stab at it!

eliheady:
The last changes cause 4 tests in TestParsedFiles to fail: 028, 029, 031, and 037. I will investigate.

tlimoncelli (approved):

tlimoncelli:
Thank you so much for the excellent unit tests!

eliheady:
Thanks @tlimoncelli for the review and feedback!

PR #3454, opened 2025-02-24
DOCS: Update migrating.md to match BIND's current behavior

udf2457:
`migrating.md` is out of sync with `dnscontrol` functionality.

As per the BIND docs, `directory` and `filenameformat` are no longer hardcoded.  But as presently worded `migrating.md` implies that it is.


tlimoncelli:
Good catch!  Thank you for filing a PR!

PR #3470, opened 2025-02-27
Update axfrddns.md

udf2457:
Put important note at the top to save others many hours of head-bashing.


cafferata:
CC @hnrgrgr @fritterhoff

fritterhoff:
Hi. So we are running dnscontrol only in dockerized Environments...

Issue #2097, opened 2023-02-27
Refactor Punycode (IDNA/ACE) handling


tlimoncelli:
IDNA is something that needs to be fixed in DNSControl.  Currently it takes the string from dnsconfig.js and convert it by calling `dc.Punycode()`.  At that point we lose the original.

A better way would be: The DomainConfig struct should store the original string from dnsconfig.js and the string converted to punycode.  (Maybe call the punycode version `.Name` and the Unicode version `.NameUnicode`?)  It should then use the original for displaying to the users and the punycode for everything else.

Or, if we're lazy we can just output "Domain {.NameUnicode} being converted to {.Name}" once and refer to the punycode for all other output.

tlimoncelli:
The IDNA part of dnscontrol is kind of a mess.

First, I must confess that I'm not an expert here. I had to look up what ACE-encoding it just now.  Please correct me if I use any terminology incorrectly.

The way things work now: Users can put Unicode in dnsconfig.js.  The providers call dc.Punycode() which converts everything (zone names, labels, targets) to ACE-encoded (i.e. xn--bcher-kva) strings. After that all code assumes everything is ASCII.  The "preview/push" output is all ASCII (I think).

The problem I see are:
* some providers call dc.Punycode() and others do not.
* preview/push output doesn't try to convert ACE-encoded to Unicode.  As a result, a user could input a domain in Unicode but then be confused that the preview/push output is ACE-encoded.

I'm not 100% sure what the best way forward is. Some thoughts:
* It would be less confusing to users if preview/push messages output ascii when the user input was ascii, and Unicode when the user input was Unicode. (This would be a simple matter of recording (bool) if the input was Unicode, and using that to convert back during preview/push output.)
* I think it would be better if all providers called dc.Punycode().  It would be better if the conversion was done by the system, so that each provide didn't have a choice.

tlimoncelli:
[dkim1970](https://github.com/dkim1970) writes:
> I would say this RFC covers the "opinions" about how IDN should be handled in applications. https://datatracker.ietf.org/doc/rfc3490/
> 
> More about IDN can be found by this author: https://datatracker.ietf.org/person/paf@paftech.se

tlimoncelli:
Thanks, @dkim1970, for those URLs.  I read rfc3490 and now I feel much more confident about what we should do:

My proposal:
* Users should be permitted to use unicode or ascii in D("") statements.
[...]

PR #3479, opened 2025-03-02
DESEC: Properly support punycode domains

networkException:
This pull request makes some adjustments to the deSEC provider to properly handle unicode domains by encoding them as punycode when necessary.

### DIFF
--- a/providers/desec/convert.go
+++ b/providers/desec/convert.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/StackExchange/dnscontrol/v4/models"
 	"github.com/StackExchange/dnscontrol/v4/pkg/printer"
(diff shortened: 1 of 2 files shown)


cafferata:
Ping @D3luxee, the maintainer of the provider.

flokli:
Hm, I ran into a similar issue with the AUTODNS provider. Initially opened https://github.com/StackExchange/dnscontrol/issues/3566, but then stumbled over https://github.com/StackExchange/dnscontrol/issues/2097, proposing to solve this in a general fashion inside dnscontrol itself, so providers would always see punycode.

While patching punycode-encoding into each provider should be a no-op if it's already punycoded, shouldn't this be handled more generally in dnscontrol itself, not inside each provider?

PR #3483, opened 2025-03-07
Bug: HASH() outputs debug info

jpmens:
I note that when using `HASH("SHA1", ..)` the value and hash are printed on _stdout_:

```console
$ dnscontrol preview
Reading dnsconfig.js or equiv.
example.com.
22b77bea9d7a8cf2147bb0616aaa388b0d152013
example.
9133ab60caa1c7be379f21f1ba2968365a54bf36
```

If I use `SHA256` instead, the hashes aren't emitted (but the values are).

Looking at this code, it would appear as though two `Printf()` were forgotten. This small patch removes them.


tlimoncelli:
Thanks for fixing that!

PR #3485, opened 2025-03-08
FEATURE: Support ignoring the `ech=` parameter in `HTTPS`/`SVCB` RR types

gucci-on-fleek:
[ECH (Encrypted Client Hello)](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni) is a new-ish TLS feature that lets clients communicate with a server, without revealing the SNI (desired domain) in plaintext. To solve the bootstrapping problem (how do you encrypt data to a server without knowing its public key?), you typically need to [publish the server's public key in the DNS](https://datatracker.ietf.org/doc/html/draft-ietf-tls-svcb-ech). This effectively leaves us with 3 options:

1. Manage the DNS exclusively with DNSControl. This means that you'll have to manually generate and rotate keys for all of your servers, which is both crypographically-unwise (since key management can't be automated), and a giant pain (since you'll need to manually copy-and-paste random base64-encoded strings into your `dnsconfig.js`).

2. Ignore the `HTTPS`/`SVCB` records by placing `IGNORE(<name>, "HTTPS")` into your `dnsconfig.js`. This means that if you want to change the TTL, target, or other SvcParams, you'll need to do this outside of DNSControl. But it's typically [web servers that will be setting the `ech=` key](https://github.com/caddyserver/caddy/releases/tag/v2.10.0-beta.1), and since most web servers don't provide an interface to set arbitrary DNS records, there isn't any good way to change these parameters.

3. Modify DNSControl to support ignoring the `ech=` key (while still allowing to change everything else). That's what this PR implements.

I've tested this PR and it works as expected on my server, but the implementation is a bit of a mess. Some possible questions:

1. What should happen in the case of multiple `HTTPS`/`SVCB` RRs for a single FQDN?
2. Should this implementation be extended to support setting _any_ SvcParam key to `IGNORE`?
3. Is `diff2.diffTargets()` the correct place to implement this logic?
4. Is there a better way to do this than using regexes on strings?
5. How should I test this? I had to comment out a few of my added tests because they were failing (even though I think that they should work?).

Thanks!

### DIFF
--- a/commands/types/dnscontrol.d.ts
+++ b/commands/types/dnscontrol.d.ts
@@ -1181,6 +1181,8 @@ declare function HASH(algorithm: "SHA1" | "SHA256" | "SHA512", value: string): s
  *
  * Modifiers can be any number of [record modifiers](https://docs.dnscontrol.org/language-reference/record-modifiers) or JSON objects, which will be merged into the record's metadata.
  *
+ * If you set the parameter `ech` to the special value `IGNORE`, DNSControl will ignore the contents of that parameter when updating a zone.
+ *
  * ```javascript
  * D("example.com", REG_MY_PROVIDER, DnsProvider(DSP_MY_PROVIDER),
  *   HTTPS("@", 1, ".", "ipv4hint=123.123.123.123 alpn=h3,h2 port=443"),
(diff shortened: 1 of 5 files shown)


gucci-on-fleek:
@tlimoncelli Could you please take a look at this? No problem at all if you're busy though.

gucci-on-fleek:
Ok, I've rebased onto `main`, and all the tests still pass. This isn't ready to be merged quite yet, but I'd like some feedback on the implementation, so I'm going to mark this as “Ready for review”.

Issue #2856, opened 2024-02-28
INWX can provide ALIAS and LOC 


gotjoshua:
I want to use ALIAS records and noticed that some other ? in the provider matrix seem inaccurate

**Describe the solution you'd like**
ALIAS, LOC, and if [possible DNSSEC](https://kb.inwx.com/en-us/3-nameserver/131-how-do-i-set-up-dnssec-for-my-domain)

**Additional context**
cli error:
`domain tld.in uses ALIAS records, but DNS provider type INWX does not support them`

Seems possible via web ui:
![Screenshot from 2024-02-28 09-43-03](https://github.com/StackExchange/dnscontrol/assets/8686454/fc68cc3d-31fd-4fc4-903b-9f895b3f0772)
![Screenshot from 2024-02-28 09-42-52](https://github.com/StackExchange/dnscontrol/assets/8686454/41fb7a9e-cb57-462f-b06f-6c808d0e90dd)

tlimoncelli:
CC @patschi (maintainer of the INWX provider)

eliheady:
@gotjoshua regarding the DNSSEC capability, managed DNSSEC support was added in v4.19.0

ALIAS RR support was also included in that release but there is a minor bug that causes the dnscontrol config and the published zone to remain out of sync, manifesting as dnscontrol detecting a change after pushing an ALIAS RR (#3564)

eliheady:
As of the 4.20.0 release, ALIAS support is working without the issue described in #3564 

This issue title should probably be revised to reflect the outstanding feature request is LOC records

tlimoncelli:
I've updated the title.  Thanks!

TomOnTime:
As per policy, I'm closing any issue older than 6 months. Please re-open if it is still relevant.

PR #3499, opened 2025-03-22
INWX: support ALIAS records

tennox:
see https://github.com/StackExchange/dnscontrol/issues/2856#issuecomment-1969153718

I have tested to confirm it works.


tennox:
One weird thing though - it will update the ALIAS record on every push:

```
#2: ± MODIFY ALIAS tam.ma: (ipfs.target.com.domain.org. ttl=600) -> (ipfs.target.com. ttl=600)
```

Which I'm guessing is an error in the read value?

eliheady:
Hi @tennox

> One weird thing though - it will update the ALIAS record on every push:

It is because the INWX API removes the trailing dot in API responses. You have to tell the provider to add it back before calculating zone corrections. This may be all that is needed to resolve the bad change detection:

```
@@ -311,6 +311,7 @@ func (api *inwxAPI) GetZoneRecords(domain string, meta map[string]string) (model
                        "NS":    true,
                        "SRV":   true,
                        "PTR":   true,
+                       "ALIAS": true,
                }
                if rtypeAddDot[record.Type] {
                        if record.Type == "MX" && record.Content == "." {
```

This will add the trailing dot to ALIAS records after pulling the zone data.

eliheady:
This [commit](https://github.com/egon0/dnscontrol/commit/ebb7dea17ecc831a6f221738443e57490d59b83e) by @egon0 implements changes that get ALIAS working for INWX without the unnecessary repeated change on push.

@tennox are you interested in pulling those additional changes into your PR?

Issue #2800, opened 2024-01-17
INWX: support changing from null MX to non-null MX


synotna:
NOTE: Have a general question?  You'll get a better response on the [dnscontrol-discuss](https://groups.google.com/g/dnscontrol-discuss) email list!

**Describe the bug**
With INWX, when changing from null MX to non-null MX records, the ordering matters because INWX enforces that you cannot add further MX records while you have a null MX record

dnscontrol does not seem to handle this, leading to try to add non-null MX records before the null MX record is modified

**To Reproduce**
1. Have an INWX domain that has null MX
2. Change to having non-null MX records e.g. the standard set for Google
3. The additions are attempted before the modification of the existing null MX record, causing them to fail

#1: + CREATE MX example.org 10 alt3.aspmx.l.google.com. ttl=3600
FAILURE! (2308) Data management policy violation
#2: + CREATE MX example.org 10 alt4.aspmx.l.google.com. ttl=3600
FAILURE! (2308) Data management policy violation
#3: + CREATE MX example.org 5 alt1.aspmx.l.google.com. ttl=3600
FAILURE! (2308) Data management policy violation
#4: + CREATE MX example.org 5 alt2.aspmx.l.google.com. ttl=3600
FAILURE! (2308) Data management policy violation
...
#6: ± MODIFY MX example.org: (0 . ttl=3600) -> (1 aspmx.l.google.com. ttl=3600)
SUCCESS!

**Expected behavior**
The existing null MX record should be modified first so that the subsequent additions do not fail

**DNS Provider**
INWX

tlimoncelli:
CC @blackshadev (author of dnsgraph, which decides the ordering of updates)
CC @patschi (INWX maintainer)

blackshadev:
I am interested to know if any other provider has the same issue. I do not (yet) have any exceptions in the ordering flow for any provider. And it seems weird to me to make an exception since technically you can depent on empty records  and the diff shown below should work in most systems. 

I don't know of the INWX maintainer (@patschi) agrees, but it seems INWX specific and thus is should be best handled as an exception in the provider. My suggestion would be to detect the case in the provider : modify-ing an empty MX record. And treat is as a Delete + Create. The TransIP also have some weird edgecases were I ended up deleting and recreating the record instead.

tlimoncelli:
I'd like to see if its possible to create an integration test that detects this situation.

I created this: https://github.com/StackExchange/dnscontrol/pull/2801

[...]

PR #3501, opened 2025-03-22
INWX: move to diff2 for record corrections

eliheady:
Related to #2800 but does not fix the null MX to non-null MX issue raised there

This refactors the INWX provider zone records corrections calculation to use diff2 methods.


tlimoncelli:
Fantastic!  Thanks so much for moving this to diff2!

Issue #3507, opened 2025-03-25
Typescript autocomplete puts errors on require calls


labrown:
**Describe the bug**
After running  `dnscontrol  write-types` I noticed that the block of `require` calls in my main dnscontrol.js file was marked with errors in vscode:

![Image](https://github.com/user-attachments/assets/7f4f5dcf-4b6c-4e33-892c-ab20b2d6d51c)

When I mouse over the marked text I see this error message:

![Image](https://github.com/user-attachments/assets/0077452e-9a6a-404b-9f7e-dc39dc56f9df)

I expect this is fallout from re-using the `require` command from javascript as part of the dnscontrol DSL.

**To Reproduce**
1. Install dnscontrol
2. Follow instructions to install the typescript autocomplete files and activate them in a dnscontro.js file
3. Add a `require("file")` call to your dnscontro.js
4. See the incorrect error text

A clear and concise description of what you expected to happen:

The typescript autocomplete should handle the `require` top-level command properly, not through an inappropriate error message.

huihuimoe:
`require` is a special keyword; using the require function causes typescript to infer that this file is a module, thus losing the ability to infer global variables.
If you need this functionality, here's a temporary solution:

```javascript
var require2 = require;
require2("./zones/example.com.js");
```

labrown:
I'm aware `require` is a special keyword, which is why I mentioned the bit about it being a javascript command.

I tried putting your tsconfig.json file in the root dir of my dnscontrol setup and then loaded dnscontrol.js into vscode but it is still showing the incorrect error.  I am not a javascript/typescript programmer so I'm just banging around here.  How does one use a tsconfig.json file?

huihuimoe:
just a reference for you.

```
my-domains
├── .gitignore
├── .gitlab-ci.yml
├── creds.json
├── dnsconfig.js
├── spfcache.json
├── tsconfig.json
├── types-dnscontrol.d.ts
└── zones
    ├── example.com.js
    ├── other...
```

`tsconfig.json` should include all your js/ts code.
```json
...
  "include": [
    "zones/**/*.js",         // <- all zones
    "dnsconfig.js",          // <- entry point
    "types-dnscontrol.d.ts", // <- types file
  ]
}
```

`dnsconfig.js`
```javascript
var REG_NONE = NewRegistrar("none");
var DSP_CLOUDFLARE = NewDnsProvider("cloudflare");
require_glob("./zones/", false);
```

`example.com.js`
[...]

PR #3508, opened 2025-03-26
META: typescript: Mark recursive parameter of require_glob optional

labrown:
From #3507 
* DOCS: mark `recursive` parameter to `require_glob` optional

tested as per Tom's suggestion in the issue. Works.


tlimoncelli:
Perfect!

Thanks so much!

PR #3524, opened 2025-04-11
INWX: Fix Null MX handling

### LINKED ISSUE #2800: INWX: support changing from null MX to non-null MX

eliheady:
This adds special handling of MX records for INWX. The API rejects record updates that violate RFC 7505: if a Null MX RR is present in the zone, new regular MX records are rejected. The inverse is also rejected - if MX records exist in the zone, a Null MX record can not be added.

Separately from the Null MX issue, changes to MX preference from non-zero to zero are accepted by the API but not applied, leading to inconsistent state between the dnscontrol configuration and the published zone.

Related to #2800

### DIFF
--- a/providers/inwx/inwxProvider.go
+++ b/providers/inwx/inwxProvider.go
@@ -4,10 +4,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 	"time"
 
+	"github.com/fatih/color"
 	"github.com/nrdcg/goinwx"
 	"github.com/pquerna/otp/totp"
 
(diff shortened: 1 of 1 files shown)


tlimoncelli (approved):
Thanks for this!  Good job of handling all the edge cases!

Issue #3379, opened 2025-01-16
Provider request: Vercel DNS


SukkaW:
[Vercel](https://vercel.com) has its own DNS service (https://vercel.com/docs/projects/domains/managing-dns-records). Though Vercel doesn't have the official Go API SDK, it does maintain the official terraform provider at https://github.com/vercel/terraform-provider-vercel.

It does have an API client at [`vercel/terraform-provider-vercel/client/client.go`](https://github.com/vercel/terraform-provider-vercel/blob/main/client/client.go), and DNS related methods at [`vercel/terraform-provider-vercel/client/dns_record.go`](https://github.com/vercel/terraform-provider-vercel/blob/main/client/dns_record.go).

Please record this provider request and hopefully, someone could implement this. I myself only have very limited Go knowledge, but I could give it a shot if no one else is doing this.

PR #3542, opened 2025-04-20
feat(#3379): implement Vercel DNS Provider

SukkaW:
The PR is still WIP. Many methods are still not implemented. Some are blocked by Vercel's Go SDK (which I will create a PR toward them first).

Also, I am not familiar with Go at all. All kinds of reviews are welcome.

### DIFF
--- a/providers/vercel/vercelProvider.go
+++ b/providers/vercel/vercelProvider.go
@@ -1,0 +1,93 @@
+package vercel
+
+/*
+Vercel DNS provider (vercel.com)
+
+Info required in `creds.json`:
+	- account_id
+	- api_token
+*/
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+
+	"github.com/StackExchange/dnscontrol/v4/models"
+	"github.com/StackExchange/dnscontrol/v4/providers"
+	vercelClient "github.com/vercel/terraform-provider-vercel/client"
+)
+
+var defaultNameservers = []string{
+	"ns1.vercel-dns.com",
+	"ns2.vercel-dns.com",
+}
+
+var features = providers.DocumentationNotes{
+	// The default for unlisted capabilities is 'Cannot'.
+	// See providers/capabilities.go for the entire list of capabilities.
+	providers.CanAutoDNSSEC:          providers.Cannot(),
+	providers.CanGetZones:            providers.Cannot(),
+	providers.CanConcur:              providers.Unimplemented(),
+	providers.CanUseAlias:            providers.Can(),
+	providers.CanUseCAA:              providers.Can(),
+	providers.CanUseDHCID:            providers.Cannot(),
+	providers.CanUseDS:               providers.Cannot(),
+	providers.CanUseDSForChildren:    providers.Cannot(),
+	providers.CanUseLOC:              providers.Cannot(),
+	providers.CanUseNAPTR:            providers.Cannot(),
+	providers.CanUsePTR:              providers.Cannot(),
+	providers.CanUseSOA:              providers.Cannot(),
+	providers.CanUseSRV:              providers.Can(),
+	providers.CanUseSSHFP:            providers.Cannot(),
+	providers.CanUseTLSA:             providers.Cannot(),
+	providers.DocCreateDomains:       providers.Unimplemented(),
+	providers.DocDualHost:            providers.Cannot(),
+	providers.DocOfficiallySupported: providers.Cannot(),
+}
+
+// hednsProvider stores login credentials and represents and API connection
+type vercelProvider struct {
+	client vercelClient.Client
+}
+
+func init() {
+	const providerName = "Vercel"
+	const providerMaintainer = "@SukkaW"
+	fns := providers.DspFuncs{
+		Initializer:   newProvider,
+		RecordAuditor: AuditRecords,
+	}
+	providers.RegisterDomainServiceProviderType(providerName, fns, providers.CanUseSRV, docNotes)
+	providers.RegisterMaintainer(providerName, providerMaintainer)
+}
+
+func newProvider(creds map[string]string, meta json.RawMessage) (providers.DNSServiceProvider, error) {
+	if creds["account_id"] == "" || creds["api_token"] == "" {
+		return nil, errors.New("api_token required for ns1")
+	}
+
+	// Enable Sleep API Rate limit strategy - it will sleep until new tokens are available
+	// see https://help.ns1.com/hc/en-us/articles/360020250573-About-API-rate-limiting
+	// this strategy would imply the least sleep time for non-parallel client requests
+	c := vercelClient.New(
+		creds["api_token"],
+	)
+
+	ctx := context.Background()
+
+	team, err := c.Team(ctx, creds["account_id"])
+	if err != nil {
+		return nil, err
+	}
+
+	c = c.WithTeam(team)
+	return &vercelProvider{client: c}, nil
+}
+
+// GetNameservers returns the default Vercel nameservers.
+// Though Vercel RESTful API supports getting "intendedNameServers", but it is not implemented in the Go SDK
+// Let's hard-coded this for now
+func (c *vercelProvider) GetNameservers(_ string) ([]*models.Nameserver, error) {
+	return models.ToNameservers(defaultNameservers)
+}
--- a/go.mod
+++ b/go.mod
@@ -120,6 +120,7 @@ require (
 	github.com/gopherjs/gopherjs v1.17.2 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
+	github.com/hashicorp/go-hclog v1.6.3 // indirect
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
 	github.com/hashicorp/go-retryablehttp v0.7.7 // indirect
 	github.com/hashicorp/go-rootcerts v1.0.2 // indirect
@@ -127,6 +128,7 @@ require (
 	github.com/hashicorp/go-secure-stdlib/strutil v0.1.2 // indirect
 	github.com/hashicorp/go-sockaddr v1.0.2 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
+	github.com/hashicorp/terraform-plugin-log v0.9.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/kolo/xmlrpc v0.0.0-20220921171641-a4b6fa1dd06b // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
@@ -148,6 +150,7 @@ require (
 	github.com/sony/gobreaker v0.5.0 // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
 	github.com/tjfoc/gmsm v1.4.1 // indirect
+	github.com/vercel/terraform-provider-vercel v1.14.1 // indirect
 	github.com/xrash/smetrics v0.0.0-20240521201337-686a1a2994c1 // indirect
 	go.mongodb.org/mongo-driver v1.13.1 // indirect
 	go.opentelemetry.io/auto/sdk v1.1.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -127,6 +127,7 @@ github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7
 github.com/exoscale/egoscale v0.102.4 h1:GBKsZMIOzwBfSu+4ZmWka3Ejf2JLiaBDHp4CQUgvp2E=
 github.com/exoscale/egoscale v0.102.4/go.mod h1:ROSmPtle0wvf91iLZb09++N/9BH2Jo9XxIpAEumvocA=
 github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
+github.com/fatih/color v1.13.0/go.mod h1:kLAiJbzzSOZDVNGyDpeOxJ47H46qBXwg5ILebYFFOfk=
 github.com/fatih/color v1.18.0 h1:S8gINlzdQ840/4pfAwic/ZE0djQEH3wM94VfqLTZcOM=
 github.com/fatih/color v1.18.0/go.mod h1:4FelSpRwEGDpQ12mAdzqdOukCy4u8WUtOY6lkT/6HfU=
 github.com/fatih/structs v1.1.0 h1:Q7juDM0QtcnhCpeyLGQKyg4TOIghuNXrkL32pHAUMxo=
@@ -251,6 +252,8 @@ github.com/hashicorp/go-sockaddr v1.0.2 h1:ztczhD1jLxIRjVejw8gFomI1BQZOe2WoVOu0S
 github.com/hashicorp/go-sockaddr v1.0.2/go.mod h1:rB4wwRAUzs07qva3c5SdrY/NEtAUjGlgmH/UkBUC97A=
 github.com/hashicorp/hcl v1.0.0 h1:0Anlzjpi4vEasTeNFn2mLJgTSwt0+6sfsiTG8qcWGx4=
 github.com/hashicorp/hcl v1.0.0/go.mod h1:E5yfLk+7swimpb2L/Alb/PJmXilQ/rhwaUYs4T20WEQ=
+github.com/hashicorp/terraform-plugin-log v0.9.0 h1:i7hOA+vdAItN1/7UrfBqBwvYPQ9TFvymaRGZED3FCV0=
+github.com/hashicorp/terraform-plugin-log v0.9.0/go.mod h1:rKL8egZQ/eXSyDqzLUuwUYLVdlYeamldAHSxjUFADow=
 github.com/hashicorp/vault/api v1.16.0 h1:nbEYGJiAPGzT9U4oWgaaB0g+Rj8E59QuHKyA5LhwQN4=
 github.com/hashicorp/vault/api v1.16.0/go.mod h1:KhuUhzOD8lDSk29AtzNjgAu2kxRA9jL9NAbkFlqvkBA=
 github.com/huaweicloud/huaweicloud-sdk-go-v3 v0.1.142 h1:9iOJ8tfNLw8uSiR5yx7VcHEYSOajJq5hb9SXF0BCUdA=
@@ -298,7 +301,9 @@ github.com/mailru/easyjson v0.0.0-20190614124828-94de47d64c63/go.mod h1:C1wdFJiN
 github.com/mailru/easyjson v0.0.0-20190626092158-b2ccc519800e/go.mod h1:C1wdFJiN94OJF2b5HbByQZoLdCWB1Yqtg26g4irojpc=
 github.com/matryer/moq v0.0.0-20190312154309-6cfb0558e1bd/go.mod h1:9ELz6aaclSIGnZBoaSLZ3NAl1VTufbOrXBPvtcy6WiQ=
 github.com/mattn/go-colorable v0.0.9/go.mod h1:9vuHe8Xs5qXnSaW/c/ABM9alt+Vo+STaOChaDxuIBZU=
+github.com/mattn/go-colorable v0.1.9/go.mod h1:u6P/XSegPjTcexA+o6vUJrdnUu04hMope9wVRipJSqc=
 github.com/mattn/go-colorable v0.1.11/go.mod h1:u5H1YNBxpqRaxsYJYSkiCWKzEfiAb1Gb520KVy5xxl4=
+github.com/mattn/go-colorable v0.1.12/go.mod h1:u5H1YNBxpqRaxsYJYSkiCWKzEfiAb1Gb520KVy5xxl4=
 github.com/mattn/go-colorable v0.1.13 h1:fFA4WZxdEF4tXPZVKMLwD8oUnCTTo08duU7wxecdEvA=
 github.com/mattn/go-colorable v0.1.13/go.mod h1:7S9/ev0klgBDR4GtXTXX8a3vIGJpMovkB8vQcUbaXHg=
 github.com/mattn/go-isatty v0.0.3/go.mod h1:M+lRXTBqGeGNdLjl/ufCoiOlB5xdOkqRJdNxMWT7Zi4=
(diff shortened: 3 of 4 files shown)


SukkaW (commented):
So, Vercel has some automatically created records (like immutable NS, automatically created CAA records, etc.) that can not be deleted. How do I implement this?

~~Also, Vercel enforces a minimum TTL of 60, how do I incorporate that?~~ Done in `GetZoneRecordsCorrections`.

tlimoncelli:
> So, Vercel has some automatically created records (like immutable NS, automatically created CAA records, etc.) that can not be deleted. How do I implement this?

That's an interesting problem.  Luckily we've seen this before.

It should be a simple matter of "pretend those records don't exist".  `GetZoneRecords()` can skip over them rather than convert them (line 122 `for _, r := range records {`).

Then everything else should "just work" because you are using diff2.ByRecord() and future add/change/deletes will not affect the existing records.

Hope that helps!

providers/vercel/httpsRecord.go
```diff
@@ -0,0 +1,101 @@
...
+// HTTPS record structure for Vercel API
+type httpsRecord struct {
+	Priority uint16 `json:"priority"`
+	Target   string `json:"target"`
+	Params   string `json:"params,omitempty"`
+}
+
+type CreateHTTPSDNSRecordRequest struct {
+	Name    string       `json:"name"`
+	TTL     int64        `json:"ttl,omitempty"`
```
tlimoncelli: Please confirm that TTL should be int64.  Most systems use a uint32 for TTL, since that's how large they are in the packet format.

providers/vercel/httpsRecord.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	var result vercelClient.DNSRecord
+	return c.doRequest(clientRequest{
+		ctx:    ctx,
+		method: "PATCH",
+		url:    url,
+		body:   mustMarshal(payload),
+	}, &result)
+}
```
tlimoncelli: Please don't panic().  This function should return an error, which would then cause updateHTTPSRecord to return the same error. 

```suggestion
		var result vercelClient.DNSRecord
		payloadJSON, err := json.Marshal(payload)
		if err != nil {
		     return err
		}    
		return c.doRequest(clientRequest{
		ctx:    ctx,
		method: "PATCH",
		url:    url,
		body:   payloadJSON,
	}, &result)
}
```

(My formatting is probably off)

providers/vercel/listVercelDnsRecords.go
```diff
@@ -0,0 +1,82 @@
...
+// listDNSRecords retrieves all DNS records for a domain, handling pagination.
+// It replaces the client.ListDNSRecords method which does not support pagination.
+// The official Vercel client's ListDNSRecords is a test helper that limits results to 100
+// and does not implement the pagination logic required for production use with large zones.
+func (c *vercelProvider) listDNSRecords(domain string) ([]domainRecord, error) {
+	var allRecords []domainRecord
+	var nextTimestamp int64
+
+	// Vercel API limit is max 100
+	limit := 100
```
tlimoncelli:
```suggestion
	const limit = 100
```

providers/vercel/vercelProvider.go
```diff
@@ -0,0 +1,338 @@
...
+		Initializer:   newProvider,
+		RecordAuditor: AuditRecords,
+	}
+	providers.RegisterDomainServiceProviderType(providerName, fns, providers.CanUseSRV, features)
+	providers.RegisterMaintainer(providerName, providerMaintainer)
+}
+
+func newProvider(creds map[string]string, meta json.RawMessage) (providers.DNSServiceProvider, error) {
+	if creds["team_id"] == "" || creds["api_token"] == "" {
+		return nil, errors.New("api_token required for ns1")
```
tlimoncelli:
```suggestion
		return nil, errors.New("api_token required for VERCEL")
```

providers/vercel/listVercelDnsRecords.go
```diff
@@ -0,0 +1,82 @@
...
+	vercelClient "github.com/vercel/terraform-provider-vercel/client"
+)
+
+// domainRecord is a helper struct to unmarshal the JSON response.
+// It embeds vercelClient.DNSRecord to reuse the upstream type,
+// but adds fields to handle API inconsistencies (type vs recordType, mxPriority).
+type domainRecord struct {
+	vercelClient.DNSRecord
+	Type       string `json:"type"`
+	MXPriority int64  `json:"mxPriority"`
```
tlimoncelli: MxPreference can only be 0 to 64k.  Is this acceptable or does the SDK enforce an int64?

```suggestion
	MXPriority uint16  `json:"mxPriority"`
```

providers/vercel/listVercelDnsRecords.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// pagination represents the pagination object in Vercel API responses.
+type pagination struct {
+	Count int64  `json:"count"`
+	Next  *int64 `json:"next"`
+	Prev  *int64 `json:"prev"`
```
tlimoncelli: Do these need to be "int64"?  I would think that "int" is sufficient. (Most systems int64 == int but calling these out as a specific size seems to be over-specifying)

tlimoncelli (changes requested):
For someone "not familiar with Go at all" this is quite excellent!  I only had a couple minor suggestions!

providers/vercel/httpsRecord.go
SukkaW: Since Vercel's official API SDK uses int64 for the type. I thought we'd better be consistent on this. I will add a comment about this.

providers/vercel/listVercelDnsRecords.go
SukkaW: Since Vercel's official API SDK uses int64 for `Priority`'s type. I thought we'd better be consistent on this. And there will be a type conversion (`uint16OrZero`, which I copied from Cloudflare provider) when mapping the API response to the records. I will add a comment about this.

providers/vercel/listVercelDnsRecords.go
SukkaW: From the API docs and the actual JSON response, the cursor appears to be some kind of timestamps. But since pagination is not implemented in the official Go SDK, I just use int64 to be safe here.

providers/vercel/httpsRecord.go
SukkaW: comment added in [`0ef392a` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/0ef392ad2b651831d77e75da5ac46d5a08580279).

providers/vercel/httpsRecord.go
SukkaW: Addressed in [`0ef392a` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/0ef392ad2b651831d77e75da5ac46d5a08580279)

providers/vercel/listVercelDnsRecords.go
SukkaW: Addressed in [`0ef392a` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/0ef392ad2b651831d77e75da5ac46d5a08580279)

providers/vercel/vercelProvider.go
SukkaW: Typo fixed in [`0ef392a` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/0ef392ad2b651831d77e75da5ac46d5a08580279)

providers/vercel/listVercelDnsRecords.go
SukkaW: comment added in [`0ef392a` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/0ef392ad2b651831d77e75da5ac46d5a08580279)

providers/vercel/listVercelDnsRecords.go
SukkaW: comment added in [`0ef392a` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/0ef392ad2b651831d77e75da5ac46d5a08580279)

SukkaW:
Also, I noticed that the Vercel API has somewhat strict API rate limiting, something like 60 modifications per minute and 500 pages of records retrieval per minute. That would be enough for small/personal domains (Vercel's DNS is not meant for serious deployment anyway), but how would I finish intergration tests though.

providers/vercel/httpsRecord.go
tlimoncelli: I concur.

providers/vercel/listVercelDnsRecords.go
tlimoncelli: I concur.

tlimoncelli:
> Also, I noticed that the Vercel API has somewhat strict API rate limiting, something like 50 modifications/deletions per minute, 100 creations per minute, and 500 records retrieval per minute. That would be enough for small/personal domains (Vercel's DNS is not meant for serious deployment anyway), but how would I finish intergration tests though?

That's an interesting problem.

In the short term, we can just run the integration tests a few at a time using the `-start` and `-end` flags. If the code fails at test 30, you can do the next group by running `-start 30 -end 60`, etc.

In the long term, we should detect the error that Vercel gives when we hit the limit, sleep, then try again.  There are some modules that can help with that. Some providers code their own, others use github.com/failsafe-go/failsafe-go

SukkaW:
> That's an interesting problem.
> 
> In the short term, we can just run the integration tests a few at a time using the `-start` and `-end` flags. If the code fails at test 30, you can do the next group by running `-start 30 -end 60`, etc.
> 
> In the long term, we should detect the error that Vercel gives when we hit the limit, sleep, then try again. There are some modules that can help with that. Some providers code their own, others use github.com/failsafe-go/failsafe-go

I ended up skipping the pager test, as the preparation took an hour, which is way too much. Vercel, in the end, is not a serious DNS provider, and I have already ensured that I have handled the pagination (which you can only trust me on this).

On the other hand, I copied and modified the rate limit implementation from Hetnezer's provider, as it behaves mostly the same as Vercel's API, and can easily be ported.

SukkaW:
@tlimoncelli 

Currently, I am failing `CAA:CAA_many_records` integration test, but it is because Vercel doesn't believe `;` in `128 issuewild ";" ttl=300` is a valid FQDN. Which `rejectif` should I use here in my audit record?

providers/vercel/api.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	payloadJSON, err := json.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.doRequest(clientRequest{
+		ctx:    ctx,
+		method: "POST",
```
nekomeowww:
```suggestion
		method: http.MethodPost,
```

providers/vercel/api.go
```diff
@@ -0,0 +1,170 @@
...
+type createDNSRecordRequest struct {
+	vercelClient.CreateDNSRecordRequest
+	HTTPS *httpsRecord `json:"https,omitempty"`
+}
+
+// CreateDNSRecord creates a DNS record.
+func (c *vercelProvider) CreateDNSRecord(ctx context.Context, req createDNSRecordRequest) (*vercelClient.DNSRecord, error) {
+	url := fmt.Sprintf("https://api.vercel.com/v4/domains/%s/records", req.Domain)
+	if c.teamID != "" {
+		url += fmt.Sprintf("?teamId=%s", c.teamID)
```
nekomeowww:
```suggestion
		url += "?teamId=" + c.teamID
```

fmt.Sprintf has extra overhead when running, direct string concat is recommended.

providers/vercel/api.go
```diff
@@ -0,0 +1,170 @@
...
+type updateDNSRecordRequest struct {
+	vercelClient.UpdateDNSRecordRequest
+	HTTPS *httpsRecord `json:"https,omitempty"`
+}
+
+// UpdateDNSRecord updates a DNS record.
+func (c *vercelProvider) UpdateDNSRecord(ctx context.Context, recordID string, req updateDNSRecordRequest) (*vercelClient.DNSRecord, error) {
+	url := fmt.Sprintf("https://api.vercel.com/v4/domains/records/%s", recordID)
+	if c.teamID != "" {
+		url += fmt.Sprintf("?teamId=%s", c.teamID)
```
nekomeowww:
```suggestion
		url += "?teamId=" + c.teamID
```

providers/vercel/api.go
```diff
@@ -0,0 +1,170 @@
...
+
+	var result vercelClient.DNSRecord
+	payloadJSON, err := json.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.doRequest(clientRequest{
+		ctx:    ctx,
+		method: "PATCH",
```
nekomeowww:
```suggestion
		method: http.MethodPatch,
```

providers/vercel/api.go
```diff
@@ -0,0 +1,170 @@
...
+		url:    url,
+		body:   string(payloadJSON),
+	}, &result, c.updateLimiter)
```
nekomeowww:
```suggestion
	payloadJSON, err := json.Marshal(req)
	if err != nil {
		return nil, err
	}

	var result vercelClient.DNSRecord
	err = c.doRequest(clientRequest{
		ctx:    ctx,
		method: "PATCH",
		url:    url,
		body:   string(payloadJSON),
	}, &result, c.updateLimiter)
```

declare here when used.

providers/vercel/api.go
```diff
@@ -0,0 +1,170 @@
...
+
+	return &result, err
+}
+
+// DeleteDNSRecord deletes a DNS record.
+func (c *vercelProvider) DeleteDNSRecord(ctx context.Context, domain string, recordID string) error {
+	url := fmt.Sprintf("https://api.vercel.com/v2/domains/%s/records/%s", domain, recordID)
+	if c.teamID != "" {
+		url += fmt.Sprintf("?teamId=%s", c.teamID)
```
nekomeowww:
```suggestion
		url += "?teamId=", c.teamID
```

providers/vercel/api.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	return c.doRequest(clientRequest{
+		ctx:    ctx,
+		method: "DELETE",
```
nekomeowww:
```suggestion
		method: http.MethodDelete,
```

providers/vercel/auditrecords.go
```diff
@@ -0,0 +1,72 @@
...
+func rejectifCaaTargetContainsUnsupportedFields(rc *models.RecordConfig) error {
+	target := rc.GetTargetField()
+	if !strings.Contains(target, ";") {
+		return nil
+	}
+
+	parts := strings.Split(target, ";")
+	// The first part is the domain, which we only check length for now
+	if len(parts[0]) < 1 {
+		return fmt.Errorf("caa target domain is empty")
```
nekomeowww:
```suggestion
		return errors.New("caa target domain is empty")
```

fmt.Errorf is way slower than new error, use it only when you need to inherit/wrap previous/inner returned got error with `%w`

providers/vercel/request.go
```diff
@@ -0,0 +1,298 @@
...
+	httpClient := &http.Client{
+		Timeout: 5 * 60 * time.Second,
+	}
+
+	for {
+		r, err := req.toHTTPRequest()
+		if err != nil {
+			return err
+		}
+		r.Header.Add("Authorization", fmt.Sprintf("Bearer %s", c.apiToken))
```
nekomeowww:
```suggestion
		r.Header.Add("Authorization", "Bearer " + c.apiToken)
```

Perhaps

providers/vercel/request.go
```diff
@@ -0,0 +1,298 @@
...
+
+		if rl == nil {
```
nekomeowww: We could check this before running into any `for`, can we?

providers/vercel/request.go
```diff
@@ -0,0 +1,298 @@
...
+		resp, err := httpClient.Do(r)
+		if err != nil {
+			return fmt.Errorf("error doing http request: %w", err)
+		}
+
+		// Handle rate limiting and retries, 429 is handled here
+		retry, err := rl.handleResponse(resp)
+
+		if err != nil {
+			resp.Body.Close()
```
nekomeowww: you could use `defer resp.Body.Close()` if suitable, optional suggestion.

providers/vercel/vercelProvider.go
```diff
@@ -0,0 +1,394 @@
...
+	listLimiter   *rateLimiter
+}
+
+// uint16Zero converts value to uint16 or returns 0, use wisely
+//
+// Vercel's Go SDK implies int64 for almost everything, but since Vercel doesn't actually
+// implement their own NS and instead uses NS1 / Constellix (previously), we'd assume if
+// TTL and Priority are int64, they are in fact uint16 and otherwise be rejected by upstream
+// providers. Under this assumption, we'd convert int64 to uint16 as wells.
+func uint16Zero(value interface{}) uint16 {
```
nekomeowww, line 70:
```suggestion
func uint16Zero[T ~float64 | ~uint16](value T) uint16 {
```

Generic could be used here.

providers/vercel/vercelProvider.go
```diff
@@ -0,0 +1,397 @@
...
+		}
+	case "CAA":
+		value := fmt.Sprintf(`%v %s "%s"`, rc.CaaFlag, rc.CaaTag, rc.GetTargetField())
+		req.Value = &value
+	}
+
+	return req, nil
+}
+
+// ptrInt64 returns a pointer to an int64
```
nekomeowww, line 394:
```suggestion
	req.TTL = lo.ToPtr(int64(rc.TTL))
	req.Comment = ""

	switch rc.Type {
	case "MX":
		req.MXPriority = ptrInt64(int64(rc.MxPreference))
	case "SRV":
		req.SRV = &vercelClient.SRVUpdate{
			Priority: ptrInt64(int64(rc.SrvPriority)),
			Weight:   ptrInt64(int64(rc.SrvWeight)),
			Port:     ptrInt64(int64(rc.SrvPort)),
			Target:   &value,
		}
		req.Value = nil // SRV uses the SRV struct, not Value
	case "TXT":
		txtValue := rc.GetTargetTXTJoined()
		req.Value = &txtValue
	case "HTTPS":
		req.HTTPS = &httpsRecord{
			Priority: int64(rc.SvcPriority),
			Target:   rc.GetTargetField(),
			Params:   rc.SvcParams,
		}
	case "CAA":
		value := fmt.Sprintf(`%v %s "%s"`, rc.CaaFlag, rc.CaaTag, rc.GetTargetField())
		req.Value = &value
	}

	return req, nil
}
```

If acceptable by dnscontrol maintainers, github.com/samber/lo is recomended here for this helper with Generics.

providers/vercel/vercelProvider.go
SukkaW: This was copied from the Cloudflare provider, and the idea is that we would accept any arbitrary value. Probably should rename this to `UNSAFE_uint16Zero` though.

providers/vercel/vercelProvider.go
SukkaW: `ptrInt64` is also copied from somewhere else. But I'd love to have something that can inline ptr.

providers/vercel/api.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/api.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/api.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/api.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/request.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/request.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/api.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/api.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/auditrecords.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/api.go
SukkaW: Done in [`bdc5cee` (#3542)](https://github.com/StackExchange/dnscontrol/pull/3542/commits/bdc5cee9a7964c6723bd14d35950219f1c409744).

providers/vercel/request.go
SukkaW: I did update to use defer, not sure if it is ok though. cc @tlimoncelli what do you think?

providers/vercel/vercelProvider.go
tlimoncelli: I'm ok with leaving this to match the Cloudflare code.

providers/vercel/vercelProvider.go
tlimoncelli: I'm ok with this mirroring the Cloudflare code.

tlimoncelli:
> Sure! I would like to create a new draft PR to see if the intergration tests pass from my forked repo first, and I can donate those credentials after that.

perfect!

PR #3544, opened 2025-04-24
INWX: Add rules for TXT RRs containing double quotes and backslash

eliheady:
INWX will not apply these RR changes and also does not error on the record content.

qoutes end up double-escaped. next preview:
  `TXT ("in\\\"side" ttl=300) -> ("in\"side" ttl=300)`

odd-numbered backslashes get stripped. next preview:
  `TXT ("1backslash" ttl=300) -> ("1back\\slash" ttl=300)`
  `TXT ("3back\\\\slash" ttl=300) -> ("3back\\\\\\slash" ttl=300)`

## Release changelog section

Some examples:
* INWX: filter unsupported TXT RR content containing unbalanced quotes and backslashes


patschi:
I agree, let's keep the code base "simple" and not overcomplicate things. Should there ever be any valid use case or need having such TXT records, we can re-consider.

PR #3548, opened 2025-04-26
[wip] CI/CD: Add provider AuditRecords validation test

eliheady:
This [wip] change is to add a new test mode to the integration tests to allow validating a provider's AuditRecord rules.

This exposes a new test mode flag in the integration test suite `-auditrecords` (bool)

false (default): The tests execute according to established behavior.
Providers' AuditRecords rules are used to skip broken tests

true: The test case selection is inverted and normally working tests are
skipped. Tests that are normally excluded by AuditRules are executed and
success is treated as failure to highlight the change in API capabilities

The AuditRecords rules that each provider contains are checked during integration tests. If a rule is present that establishes the provider doesn't support some specific variation of record attributes, the integration test suite will skip tests that are expected to fail. 

Over time, providers may add support for these records. This new test mode is meant to identify rules that can be removed from the DNSControl provider's AuditRecords function and allow users to use the newly supported records.


eliheady:
I'm leaving this up in draft for now, as I think there might be better ways to do accomplish the periodic check of the filtering rules. I'm still interested in feedback if this seems like the right approach

eliheady:
I'm closing this with a plan to break the proposed GHA refactoring out into a different change, apart from reworking this with suggestions from Tom regarding the changes to the integration tests

PR #3551, opened 2025-04-27
AXFRDDNS: Ignore ZONEMD records

gucci-on-fleek:
`ZONEMD` records are DNSSEC-like, so we'll handle them like the other DNSSEC records and simply ignore them. See https://datatracker.ietf.org/doc/html/rfc8976 for more information.

I've enabled `ZONEMD` generation on the zone used in the `AXFRDDNS_DNSSEC` integration test server, so any other PRs are likely to fail until this is merged. I have _not_ enabled it on the `AXFRDDNS` (no DNSSEC) zone, so any failures there should be unrelated to this PR.


tlimoncelli:
CC @hnrgrgr

tlimoncelli:
Thanks for fixing that!

Issue #2742, opened 2023-12-25
Add ALIAS support for INWX provider


ngrie:
**Is your feature request related to a problem? Please describe.**
While migrating my DNS zones to INWX, I noticed that there is no ALIAS support for that provider in DNSControl:
```
printIR.go:113: ERROR: domain xxxx.com uses ALIAS records, but DNS provider type INWX does not support them
```

INWX however does support ALIAS records using the [web interface](https://kb.inwx.com/en-us/3-nameserver/130-how-do-i-create-alias-records) and the [API](https://www.inwx.de/en/help/apidoc/f/ch03.html#type.recordtype).

**Additional context**
From briefly scanning the code, I would guess that updating this line:
https://github.com/StackExchange/dnscontrol/blob/8f18f6e7a1942fcf25bb2169eba05ea6df0a29f6/providers/inwx/inwxProvider.go#L46

as well as this part:
https://github.com/StackExchange/dnscontrol/blob/8f18f6e7a1942fcf25bb2169eba05ea6df0a29f6/providers/inwx/inwxProvider.go#L172-L180

could be enough to add support? (I am not used to do Go programming so ...)

cafferata:
Ping @patschi, the maintainer of this provider.

eliheady:
The [commit](https://github.com/egon0/dnscontrol/commit/ebb7dea17ecc831a6f221738443e57490d59b83e) by @egon0 referenced above seems to resolve this issue and passes integration tests.

There is an [open PR](https://github.com/StackExchange/dnscontrol/pull/3499) that needs the additional changes - @tennox and @egon0 would you be interested in collaborating on the change?

eliheady:
ALIAS support was added in v4.19.0 and improved in v4.20.0

I believe this issue can be closed.

patschi:
Thank you very much for taking care about this, @eliheady!

PR #3564, opened 2025-05-04
INWX: BUGFIX: ALIAS RRs updates are broken due to trailing dot issue

eliheady:
#3499 was an incomplete fix for #2742 and #2856 

The INWX API strips trailing dots on ALIAS RRs so the provider needs to handle this similarly to NS and CNAME. The problem manifests as an unnecessary change detected after pushing an ALIAS RR.

## Release changelog section

* INWX: fix ALIAS RR trailing dot

### DIFF
--- a/providers/inwx/inwxProvider.go
+++ b/providers/inwx/inwxProvider.go
@@ -189,7 +189,7 @@ func makeNameserverRecordRequest(domain string, rec *models.RecordConfig) *goinw
 	   Records with empty targets (i.e. records with target ".")
 	   are allowed.
 	*/
-	case "CNAME", "NS":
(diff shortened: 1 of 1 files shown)


tlimoncelli:
How urgent is this? Should I make a 4.19.1 release?

eliheady:
I think it is not terribly urgent: support for ALIAS was just added so this isn't really a regression that would affect existing users. New ALIAS records can be pushed and changed, old ones can be removed. The error is not causing unexpected zone changes, just line noise for anyone adopting the new RR support ... I would say it should go into the next release, not necessarily trigger a new release.

Issue #3493, opened 2025-03-18
Route53 - Changing from R53_Alias to CNAME


misilot:
**Describe the bug**
When trying to change a R53_ALIAS to a CNAME, 

**To Reproduce**
1. Create R53 Alias
2. Create CNAME record for same R53 Alias, push changes

```
******************** Domain: example.net
2 corrections (aws)
#1: + CREATE redmine.services.example.net CNAME itsupport22.example.net. ttl=300
- DELETE redmine.services.example.net R53_ALIAS itsupport.example.net. atype=CNAME zone_id=ZONEID evaluate_target_health=false ttl=300
FAILURE! operation error Route 53: ChangeResourceRecordSets, https response error StatusCode: 400, RequestID: 6b39d0d7-0a2f-4c6e-97b2-55f3cea43af8, InvalidChangeBatch: [The request contains an invalid set of changes for a resource record set 'CNAME redmine.services.example.net.']
```

**Expected behavior**
Delete the R53_ALIAS first and than create the CNAME

**DNS Provider**
 - Route53

**Additional context**

cafferata:
Ping @tresni, the maintainer of the provider.

cmsmith7:
@tlimoncelli @cafferata @tresni 

We are experiencing this issue too.

If you have a CNAME and change this to a R53_ALIAS the push works correctly (deletes CNAME and creates R53_ALIAS - correct logical order - **delete before create**).

however ..

If you have a r53_ALIAS and change to a CNAME the push fails (tries to create the CNAME which fails as R53_ALIAS exists and has not been deleted - incorrect logical order - **create before delete**).

Manual workaround which is far from ideal is to manually delete the R53_ALIAS and re-push the zone containing the new CNAME. (we are having to do this on 6 zones).

Just to add that the preview does show that it is going to create the record before deleting.

We were using version 4.16.0 .. and have now tested with the latest version 4.19.0 but doing a preview only we can still see the logic is incorrect. (create before delete). So no point in merging this to do a push.

Please can we bump the priority of this issue as we manage **5500+ domains** and manual intervention should not be required as the binary should have the correct logic.

Many thanks in advance.

krzysztofdrys:
Hoping it will make things move a little quicker I created this PR: https://github.com/StackExchange/dnscontrol/pull/3565

I am afraid, I haven't been able to test it. Also, if there are any automated tests I could add for this change, please let me know.

tlimoncelli:
[...]

PR #3565, opened 2025-05-09
ROUTE53: Move R53_ALIAS deletion to the head of instructions 

krzysztofdrys:
We need to perform deletion before the creation. If deletion comes after the create we get error "The request contains an invalid set of changes for a resource record set".

### DIFF
--- a/providers/route53/route53Provider.go
+++ b/providers/route53/route53Provider.go
@@ -414,15 +414,18 @@ func (r *route53Provider) GetZoneRecordsCorrections(dc *models.DomainConfig, exi
 //     this, we simply move all R53_ALIAS instructions to the end of the list, thus
 //     guaranteeing they will happen after the records they refer to have been
 //     reated.
+//   - We don't move DELETE instructions for R53_ALIAS records, because they must be
+//     before other instructions for the records they refer to. For example, DELETE for R53_ALIAS(example.com) must
+//     come before CREATE for CNAME(example.com)..
 func reorderInstructions(changes diff2.ChangeList) diff2.ChangeList {
 	var main, tail diff2.ChangeList
 	for _, change := range changes {
-		// Reports should be early in the list.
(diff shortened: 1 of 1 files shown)


krzysztofdrys:
Hi @blackshadev! Thanks for the feedback!

I understand that putting all deletes in front might not work. However, I am not sure what solution you are recommending.

Is it not to move DELETES at all (neither to the front nor to the end)? This is how I updated my PR. I think that should solve our problem. And hopefully after `dnssort.SortUsingGraph` the changes should be in a reasonable order anyway, right?

Or are you saying that the whole `reorderInstructions` isn't needed? It would make sense, because again after `dnssort.SortUsingGraph` should be in a correct order.

Anyway, please let me know how to approach this. I realise there may be some edge cases here, so I appreciate your help on this!

blackshadev:
Thanks for the work. After some digging, I now understand where you are coming from. How I see it we can take one of 2 approaches. 
- Using your `reorderInstructions` , where you both need to  account for deleting aliases and creating them. Which is the essence of this MR
- Using dnssort, dnssort should always ensure things can happen in the correct order.  If that does not work for R53_alias we should investigate in there and find out why it doesn't work. dnssort only applies to "certain" records, but R53_ALIAS is in the list. See `models/record.go:486` in `GetDependencies`. 

If it turns out dnssort should fix this and it doesn't we should create a test case for it and fix the issue. But my initial guess is your solution predates `dnssort` and thus maybe you can remove it now all together.

tlimoncelli:
Here's a PR that reproduces the bug in an integration test: https://github.com/StackExchange/dnscontrol/pull/3567

tlimoncelli:
Moving discussion of the bug itself to https://github.com/StackExchange/dnscontrol/issues/3493.  

Note that PR https://github.com/StackExchange/dnscontrol/pull/3567 is a competing solution to this PR.

PR #3568, opened 2025-05-11
AUTODNS: Enable "get-zones" (ListZones, EnsureZoneExists, GetRegistrarCorrections)

flokli:
Best reviewed commit-by commit.

## ListZones 
We already advertised `providers.CanGetZones`, but `ZoneLister`'s `ListZones()` wasn't implemented yet. Adding support is very simple, we just need to `POST zone/_search` without filters. I made some of the error paths more concise (`if err := …; err != nil { … }`)

## EnsureZoneExists
This allows to create new zones. We do some error wrapping before to be able to distinguish a zone not existing from other failures, and then create the zone with dummy values, using the `POST /zone` endpoint.

## Registrar support
Use the `GET domain/{name}` endpoint to retrieve nameserver config of a domain, and use `PUT domain/{name}` to update it. Then `GetRegistrarCorrections` can simply compare the list of nameservers retrieved on `getDomain` with the desired values, and if different, do a (sparse) request to update these values.

I had to juggle some of the provider initialization around, to be able to use the same code for DomainServiceProvider and RegistrarType. We'll end up with two instances, but considering both deal with rate limiting it should be fine, and much more complicated if we'd try to aggregate clients based on the same settings, as rate-limits are per user.

### DIFF
--- a/providers/autodns/api.go
+++ b/providers/autodns/api.go
@@ -114,6 +114,24 @@ func (api *autoDNSProvider) findZoneSystemNameServer(domain string) (*models.Nam
 	return systemNameServer, nil
 }
 
+func (api *autoDNSProvider) createZone(domain string, zone *Zone) (*Zone, error) {
+	responseData, err := api.request("POST", "zone", zone)
+	if err != nil {
+		return nil, err
+	}
+
+	var responseObject JSONResponseDataZone
+	if err := json.Unmarshal(responseData, &responseObject); err != nil {
+		return nil, err
+	}
+
+	if len(responseObject.Data) != 1 {
+		return nil, errors.New("Zone " + domain + " not returned")
+	}
+
+	return responseObject.Data[0], nil
+}
+
 func (api *autoDNSProvider) getZone(domain string) (*Zone, error) {
 	systemNameServer, err := api.findZoneSystemNameServer(domain)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


flokli:
cc @arnoschoon

tlimoncelli:
Code looks good to me.  @arnoschoon, do you approve?

tlimoncelli (approved):

arnoschoon (approved):
As I said managing DNS records works properly, but I'm confused about the domain management (registrar) part but that might be a limit of my test setup since the used domain is not one I actually registered.

tlimoncelli:
> As I said managing DNS records works properly, but I'm confused about the domain management (registrar) part but that might be a limit of my test setup since the used domain is not one I actually registered.

Sounds like AUTODNS doesn't support 0-length TXT records. This will fix it:

```
--- a/providers/autodns/auditrecords.go
+++ b/providers/autodns/auditrecords.go
@@ -12,6 +12,7 @@ func AuditRecords(records []*models.RecordConfig) []error {
        a := rejectif.Auditor{}
 
        a.Add("MX", rejectif.MxNull)      // Last verified 2022-03-25
+       a.Add("TXT", rejectif.TxtIsEmpty) // Last verified 2025-05-13
 
        return a.Audit(records)
 }
```

flokli:
pushed, thanks!

Issue #3305, opened 2025-01-04
The "Provider Features" table is too wide; users don't naturally see the right-most columns


tlimoncelli:
**Describe the bug**

The "Provider Features" table is too wide; users don't naturally see the right-most columns

This page: https://docs.dnscontrol.org/getting-started/providers

**To Reproduce**
1. Go to `https://docs.dnscontrol.org/getting-started/providers`
2. You can't (without scrolling) see the right column.
3. When you do scroll, you don't see the row heads (provider names).

**Expected behavior**
Users should always see the row headers.

Some ideas:
* Replace the rtype columns (ALIAS, CAA, HTTPS, LOC, NAPTR, PTR, SOA, SRV, SSHFP, SVCB, TLSA, DS, DHCID, DNAME, DNSKEY) with a single column that lists the names of the supported records: A, MX, CNAME, LOC, PTR, TLSA.
  * The list should be a fixed-width, and word-wrap.
  * The list doesn't have to list records that all providers support: A, MX, CNAME
* Alternatively, make a 2nd table that just lists which record types are supported.

**DNS Provider**
 - n/a
 
**Additional context**
CC @cafferata

cafferata:
Thanks for raising this issue! I agree that the current “[Provider Features](https://docs.dnscontrol.org/getting-started/providers#provider-features)” table is quite wide, and it might not offer the best user experience.

However, I’m starting to wonder if this table should even be supported going forward. I’ve always questioned what this table really tells me, especially when I’m not using the specific providers listed. Instead, I previously suggested in https://github.com/StackExchange/dnscontrol/issues/1998 that we move this information to the individual provider pages. This way, everything can be consolidated in one place for each provider, including:

* Configuration
* Metadata
* Supported features
* Usage examples
* Limitations/Known issues
* Maintainer info
* And more…

What do you think of this idea? I believe it would provide a cleaner, more useful structure.

Looking forward to hearing your thoughts!

tlimoncelli:
Thank you for the feedback!

I see the value of putting this information on the individual provider pages, but it is also useful to have a table.  There is marketing value to being able to point people to a big list of providers, with indications of which have official support, and which are registrar-only, dns-only, or both.  It helps people know if DNSControl is right for them.  It also (hopefully) shames vendors who's support is limited.

Tom

cafferata:
As soon as I have the time, I will work on preparing a suitable layout.

eliheady:
[...]

PR #3579, opened 2025-05-22
DOCS: generate smaller tables for feature matrix

chicks-net:
# context

#3305 notes that the provider feature table does not display well in the GitBook-based documentation site.

This has also been brought up in a couple of the monthly live calls.

# progress

This PR splits the features into different groups for different tables.  It generates the tables with H3 headings.  The changes to the generator and the documentation are included in this PR.

Some additions to `/.gitignore` are also included.

# future

Updating individual provider pages (#3584) is coming next.

Maybe file some bug reports with GitBook for the issues found.

# prior "discussion" questions

- Does this split of things into three tables seem reasonable?
    - This seems to be the easiest fix.  Other things would require get patches in GitBook.
- Would it be better to leave the DomainModifiers out of the top level table and add that information into individual provider pages (with more automation of course)?
    - This will happen in a future PR
- If we're going with 3 tables at the top level, would there be a better grouping of the domain modifiers or any of it?
    - This grouping has been improved.


build/generate/featureMatrix.go
```diff
@@ -125,6 +125,36 @@ func matrixData() *FeatureMatrix {
...
+			ProviderThreadSafe,
+			DualHost,
+			CreateDomains,
+			// NoPurge,
+			GetZones,
+		},
+		Features2nd: []string{
+			DomainModifierAlias,
+			DomainModifierCaa,
+			DomainModifierDnssec,
```
eliheady, line 141: maybe move `DomainModifierDnssec` to be grouped with `DomainModifierDs` and `DomainModifierDnskey`

eliheady:
I don't have much of an opinion either way on keeping the scrolling table or breaking it up. Using three tables would solve the horizontal scrolling problem at least.

As for grouping, maybe the RR types should be rearranged to keep related types (and meta-types) together. For instance, `AutoDNSSEC`, `DS`, `DNSKEY` would be good to put together.

I think the data should definitely be included on each provider page as a supported features table with just two columns: feature and green/red/? status.

chicks-net:
> I don't have much of an opinion either way on keeping the scrolling table or breaking it up. Using three tables would solve the horizontal scrolling problem at least.

😁 

> As for grouping, maybe the RR types should be rearranged to keep related types (and meta-types) together. For instance, `AutoDNSSEC`, `DS`, `DNSKEY` would be good to put together.

That sounds good.  I will try that.

> I think the data should definitely be included on each provider page as a supported features table with just two columns: feature and green/red/? status.

This was my original goal and I still think it is a great idea.  I'm guessing that will be a different PR though.  Fixing the existing table has helped me understand what I will need to do that.

I will clean up this code and request more feedback once it is ready for a review.  (It will probably be next week, since we're babysitting this weekend.)

build/generate/featureMatrix.go
chicks-net: I've redone this structure to be easier to work with and regrouped things as suggested.

eliheady: 🙌

chicks-net:
I've moved this out of draft - please review.

Looking at the [preview](https://docs.dnscontrol.org/~/revisions/HhgRfSFUjFwWGRnq1CQd/provider/index) in GitBook is encouraging, but I'm guessing we should aim for 4 tables.  I'll work on that next.

eliheady:
I'm not sure this is better than the side-scrolling table, but it's not really worse either. It's just a lot of data to fit. I wish we had better control over the Gitbook layout and styles so something like sticky headers and columns or even column filtering could be used. I found an interesting [hack using SVG](https://github.com/sindresorhus/css-in-readme-like-wat/blob/d123714bbbadaa7ce854759e660eacc7c0629fb1/readme.md) to stuff css styles that works in GitHub markdown rendering. I didn't try it in Gitbook.

If there are to be multiple tables, the 'Table 1', 'Table 2' section headings should be omitted or used to highlight which features are included in the tables (not sure how to best summarize that for a terse heading though).

chicks-net:
New question: do any of the `DomainModifier`s make sense for registry-only providers, like OpenSRS?  Could we presume unimplemented for all of those?  Or should there be a new category of "not applicable" with an emoji like ✖️ ?

chicks-net:
[preview](https://docs.dnscontrol.org/~/revisions/xb6pKWgnlW1gwc18w5G9/provider/index) is looking good to me.  6 tables looks ok.  The "DNS extensions" table has the most columns and the most word wrapping, but it is still pretty readable.

Does anybody know why GitBook doesn't handle Markdown paragraphs well?  It was keeping the line endings from the file and not flowing the paragraph unless i had the whole paragraph on one line.  GitBook says they support Commonmark which is [pretty clear](https://spec.commonmark.org/0.31.2/#paragraphs) about how paragraphs in Markdown should work.

chicks-net:
[yet another preview](https://docs.dnscontrol.org/~/revisions/9XjoqvnU6HmZsBXhuU2c/provider/index) to document the dual-host column.  For some reason it tries to link back to github - which 404s since this is not merged yet.  

This also effects the Concurrency Verified column on the live site.  So it looks like a GitBook bug and I'm not sure why GitBook is acting this way.  I'll see if I can fix this.  If not, I'll make a new github issue for this problem.

build/generate/featureMatrix.go
```diff
@@ -11,34 +12,43 @@ import (
...
 
 	var tableData [][]string
 	for _, providerName := range allProviderNames() {
 		featureMap := matrix.Providers[providerName]
 
 		var tableDataRow []string
 		tableDataRow = append(tableDataRow, "[`"+providerName+"`]("+strings.ToLower(providerName)+".md)")
-		for _, featureName := range matrix.Features {
+		for _, featureName := range matrix.FeatureTables[tableNumber] {
 			tableDataRow = append(tableDataRow, featureEmoji(featureMap, featureName))
```
eliheady, line 52: This might be overengineering, but there are a lot of providers that have unknown support for all types in some tables. We could shave some vertical scrolling if any provider that has `?` for all types in a table is just skipped. I don't know if that is a good optimization though - then we are leaving the 'unknown' status for those types implicit and it might not be clear to someone not already familiar with the project.

totally untested, but something like this would skip over provider rows that have no known status for any types in the given table.

```diff
diff --git a/build/generate/featureMatrix.go b/build/generate/featureMatrix.go
index 4b52bc2f..1bfbd2cb 100644
--- a/build/generate/featureMatrix.go
+++ b/build/generate/featureMatrix.go
@@ -41,7 +41,15 @@ func markdownTable(matrix *FeatureMatrix) (string, error) {
                for _, featureName := range matrix.Features {
                        tableDataRow = append(tableDataRow, featureEmoji(featureMap, featureName))
                }
-               tableData = append(tableData, tableDataRow)
+               skipThisRow := true
+               for status := range tableDataRow[1:] {
+                       if tableDataRow[status+1] != "❔" {
+                               skipThisRow = false
+                       }
+               }
+               if !skipThisRow {
+                       tableData = append(tableData, tableDataRow)
+               }
        }
 
        markdownTable, err := markdown.NewTableFormatterBuilder().
```

eliheady:
This is looking great @chicks-net ! I think the way you have categorized and broken up the tables makes it _much_ easier to find the bit of information needed with less noise to ignore.

chicks-net:
> This is looking great @chicks-net ! I think the way you have categorized and broken up the tables makes it _much_ easier to find the bit of information needed with less noise to ignore.

Thanks for your comments, encouragement, and code.  🙏   It was fun to do some golang again after focusing on other languages recently.

I also left https://gist.github.com/chicks-net/1b85b88851812075130bfc67d465bb3c out of this PR, but if anyone is interested in seeing how helpful `just` can be, this is a very small example.  My personal web sites have `justfile`s that have grown much larger, and it is still manageable.  😁

chicks-net:
Gave up on fixing merge conflicts and opened #3589 .

PR #3588, opened 2025-05-29
GCORE: Implement ListZones to allow auto create zones

### LINKED ISSUE #3586: GCORE: push fails if domain does not already exist
**Describe the bug**
The `dnscontrol push` command does not create DNS zones if they do not already exist. Instead, it fails with a 404 error.
However, running `dnscontrol create-domains` beforehand successfully creates the missing domains. A subsequent `dnscontrol push` then works as expected.

**To Reproduce**
1. setup a sample domain that doesn't already exist in GCORE
1. `dnscontrol push` -> ❌ fails
2. `dnscontrol create-domains` -> ✅ works
3. `dnscontrol push` -> ✅ works

**DNS Provider**
 - GCORE

xddxdd:
Fix #3586

Also enable concurrency for GCORE, seems to be working for me.


tlimoncelli:
Nice!

Thanks so much!

PR #3592, opened 2025-05-31
CNR: Update docs with proper Nameserver example

mejofi:
'NAMESERVER' values should be specified with a trailing dot, and, according to the documentation and the 'CNR' provider code, there are three known NS records;

https://kb.centralnicreseller.com/dns/keydns/keydns-upgrade-to-anycast


cafferata (approved):

tlimoncelli:
Thanks for fixing that!

PR #3600, opened 2025-06-02
NS1: Add missing ListZones()

### LINKED ISSUE #3587: NS1- zone creation bug
Hi, 
According to the documentation, NS1 should automatically create zones during dnscontrol push. However, I receive a "zone does not exist" error unless I manually create the zone in the NS1 dashboard or explicitly run dnscontrol create-domains beforehand.

#To Reproduce

Add a new NS1 zone file (e.g., zones/ns1/example.com.js)

Run dnscontrol preview or dnscontrol push

#Expected behavior
The zone should be created automatically during the push step.

#DNS Provider
NS1

#Actual Output

INFO#1: Domain "example.com" provider ns1 Error: zone does not exist

#Additional context

*Running dnscontrol create-domains example.com works and successfully creates the zone.

*Using the latest DNSControl version.

costasd:
The ListZones() API call was missing, leading to dnscontrol being unable to understand if a zone should be created via EnsureZoneExists().

Let's add the missing method.

Fixes #3587

### DIFF
--- a/providers/ns1/ns1Provider.go
+++ b/providers/ns1/ns1Provider.go
@@ -75,6 +75,26 @@ func newProvider(creds map[string]string, meta json.RawMessage) (providers.DNSSe
 	)}, nil
 }
 
+func (n *nsone) ListZones() ([]string, error) {
+	var zones []string
+
+	for rtr := 0; ; rtr++ {
+		zs, httpResp, err := n.Zones.List()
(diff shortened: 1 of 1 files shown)


sandeep901290:
Hi!
I’ve tested this fix using the latest build from the main branch and can confirm that it works correctly now. The NS1 zones are created automatically as expected when running dnscontrol push.

Thank you for your quick response and for fixing this! Looking forward to seeing it in the next official release. 🙏

Issue #3607, opened 2025-06-07
DOCS: cleanup provider table


chicks-net:
## Describe the bug

#3589 reorganized the [generated provider feature table in the docs](https://docs.dnscontrol.org/provider/index).  But @cafferata noted:

> Just a quick heads-up:
>
> The links/paths in the Markdown files currently point to GitHub URLs. This is likely because the new documents (concurrency-verified.md and dual-host.md) haven’t yet been added to the SUMMARY.md file.
https://docs.dnscontrol.org/provider/index#provider-api
> Once they’re included in SUMMARY.md, they’ll be part of the DNSControl documentation site structure, and the links can (and should) reference the documentation site instead.
> Let me know if you need any help updating this!

I was also [puzzles about this](https://github.com/StackExchange/dnscontrol/pull/3579#issuecomment-2906395432) and it should be easy to fix.

[Release 4.21.0](https://github.com/StackExchange/dnscontrol/discussions/3605) led to a small and reasonable request from @willpower232 :

> what are the odds of getting a position sticky on those table headers to help with their incredible length?

The odds are getting better with getting this issue filed.

## To Reproduce

1. Look at the [generated provider feature table in the docs](https://docs.dnscontrol.org/provider/index). 😮 

## Meta-administrivia

Feel free to assign this to me and give it the docs tag.

willpower232:
@chicks-net had a quick look and the tables are slightly different but mostly "easily" improved with sticky headers

![Image](https://github.com/user-attachments/assets/a2419257-9ada-4c22-a1ec-4ad16dc5dff3)

not sure how chaotic removing those overflow settings is for mobile though

PR #3619, opened 2025-06-13
DOCS: add links to help jump to provider table anchors

chicks-net:
## Context

Fixes: #3607

## Does

Since the anchors were automatically provided by GitBook I just needed to:

- create an inline table of contents for the generated tables

This information is also in the sidebar on the right, but having it inline seems nicer.

## Maybe

Should we provide links to jump back to the top?  I'm leaving it out for now.

## Verify

[The preview](https://docs.dnscontrol.org/~/revisions/tqzh8eb8YmxphuX5Ufbw/provider/index) lets me jump around to the different tables with the inline links and the sidebar links.

### DIFF
--- a/build/generate/featureMatrix.go
+++ b/build/generate/featureMatrix.go
@@ -15,7 +15,7 @@ func generateFeatureMatrix() error {
 	var replacementContent string = "Jump to a table:\n\n"
 	matrix := matrixData()
 
-	for i, tableTitle := range matrix.FeatureTablesTitles; i++ {
+	for _, tableTitle := range matrix.FeatureTablesTitles {
 		var jumptotableContent string = ""
 
 		var anchor string = strings.ToLower(tableTitle)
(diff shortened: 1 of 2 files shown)


build/generate/featureMatrix.go
```diff
@@ -12,9 +12,20 @@ import (
 )
 
 func generateFeatureMatrix() error {
-	var replacementContent string = ""
+	var replacementContent string = "Jump to a table:\n\n"
 	matrix := matrixData()
 
+	for i := 0; i < len(matrix.FeatureTables); i++ {
+		var tableTitle = matrix.FeatureTablesTitles[i]
```
tlimoncelli:
```suggestion
	for i := range matrix.FeatureTables {
		var tableTitle = matrix.FeatureTablesTitles[i]
```

or (unsure if this will work as it uses .FeatureTablesTitles instead of FeatureTables)

```suggestion
	for i, tableTitle := range matrix.FeatureTablesTitles; i++ {
```

PR #3622, opened 2025-06-14
NS1: code reorganization

costasd:
This changeset splits the existing provider code into multiple files to enhance readability.

While there, fix also a couple minor tidbits observed during the split:
* Use the GetZone() that handles api timeouts in GetNameservers
* Merge a runaway capability to the rest of the capabilities


tlimoncelli:
Very interesting!

Just the other day I was thinking that it would be useful to create a standard set of filenames to use for providers.  I really like your name scheme.

The name scheme I had come up with was based on interfaces, not type of work.  (see https://github.com/StackExchange/dnscontrol/pull/3611 ).

I'd be interested in any feedback about what the new standard should be.

tlimoncelli:
After much consideration... I prefer your format.  Merging.

Issue #3509, opened 2025-03-29
DKIM_BUILDER


mweinelt:
**Is your feature request related to a problem? Please describe.**
DKIM is a key-value TXT record with lenghty values, which can make it messy fast.

**Describe the solution you'd like**
I think having a `DKIM_BUILDER` that allow structured configuration of DKIM TXT records similar to what already exists for SPF and DMARC would complete this useful triangle.

**Describe alternatives you've considered**
None, which means sticking with unstructured strings.

**Additional context**
Same vein as https://github.com/StackExchange/dnscontrol/issues/1022

cafferata:
Thanks for the suggestion — I think that’s a great idea! It would make the support for these DNS records feel more complete. Curious to hear what @tlimoncelli thinks about this too.

PR #3627, opened 2025-06-17
NEW FEATURE: DKIM_BUILDER() adds a DKIM record builder

costasd:
This changeset adds a simple `DKIM_BUILDER()` function under `pkg/js/helpers.js` and the relevant documentation.

Relates to #3509

### DIFF
--- a/documentation/language-reference/domain-modifiers/DKIM_BUILDER.md
+++ b/documentation/language-reference/domain-modifiers/DKIM_BUILDER.md
@@ -1,0 +1,92 @@
+---
+name: DKIM_BUILDER
+parameters:
+  - label
+  - selector
+  - pubkey
+  - flags
+  - hashtypes
+  - keytype
+  - servicetypes
+  - note
+  - ttl
+parameters_object: true
+parameter_types:
+  label: string?
+  selector: string
+  pubkey: string
+  flags: string[]?
+  hashtypes: string[]?
+  keytype: string?
+  servicetypes: string[]?
+  note: string?
+  ttl: Duration?
+---
+
+DNSControl contains a `DKIM_BUILDER` which can be used to simply create
+DKIM policies for your domains.
+
+
+## Example
+
+### Simple example
+
+{% code title="dnsconfig.js" %}
+```javascript
+D("example.com", REG_MY_PROVIDER, DnsProvider(DSP_MY_PROVIDER),
+  DKIM_BUILDER({
+    selector: "s1",
+    pubkey: "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDC5/z4L"
+  }),
+);
+```
+{% endcode %}
+
+This yield the following record:
+
+```text
+s1._domainkey   IN  TXT "v=DKIM1; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDC5/z4L"
+```
+
+### Advanced example
+
+{% code title="dnsconfig.js" %}
+```javascript
+D("example.com", REG_MY_PROVIDER, DnsProvider(DSP_MY_PROVIDER),
+  DKIM_BUILDER({
+    label: "alerts",
+    selector: "k2",
+    pubkey: "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDC5/z4L",
+    flags: ['y'],
+    hashtypes: ['sha256'],
+    keytype: 'rsa',
+    servicetypes: ['email'],
+    ttl: 150
+  }),
+);
+```
+{% endcode %}
+
+This yields the following record:
+
+```text
+
+k2._domainkey.alerts    IN  TXT "v=DKIM1; k=rsa; s=email; t=y; h=sha256; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDC5/z4L" ttl=150
+
+```
+
+### Parameters
+
+* `label:` The DNS label for the DKIM record (`[selector]._domainkey` prefix is added; default: `'@'`)
+* `selector:` Selector used for the label. e.g. `s1` or `mail`
+* `pubkey:` Public key `p` to be used for DKIM.
+* `keytype:` Key type `k`. Defaults to `'rsa'` if omitted (optional)
+* `flags:` Which types `t` of flags to activate, ie. 'y' and/or 's'. Array, defaults to 's' (optional)
+* `hashtypes:` Acceptable hash algorithms `h` (optional)
+* `servicetypes:` Record-applicable service types (optional)
+* `note:` Note field `n` for admins. Avoid if possible to keep record length short. (optional)
+* `ttl:` Input for `TTL` method (optional)
+
+### Caveats
+
+* DKIM (TXT) records are automatically split using `AUTOSPLIT`.
--- a/pkg/js/helpers.js
+++ b/pkg/js/helpers.js
@@ -1739,6 +1739,77 @@ function CAA_BUILDER(value) {
     return r;
 }
 
+// DKIM_BUILDER takes an object:
+// label: The DNS label for the DKIM record ([selector]._domainkey prefix is added; default: '@')
+// selector: Selector used for the label. e.g. s1 or mail
+// pubkey: Public key (p) to be used for DKIM.
+// keytype: Key type (k). Defaults to 'rsa' if missing (optional)
+// flags: Which types (t) of flags to activate, ie. 'y' and/or 's'. Array, defaults to 's' (optional)
+// hashtypes: Acceptable hash algorithma (h) (optional)
+// servicetypes: Record-applicable service types (optional)
+// note: Note field fo admins. Avoid if possible to keep record length short. (optional)
+// ttl: The time for TTL, integer or string. (default: not defined, using DefaultTTL)
+
+function DKIM_BUILDER(value) {
+    if (!value) {
+        value = {};
+    }
+    kvs = [];
+
+    if (!value.selector) {
+        throw 'DKIM_BUILDER selector cannot be empty';
+    }
+
+    if (!value.pubkey) {
+        throw 'DKIM_BUILDER pubkey cannot be empty';
+    }
+
+    // build the label
+    if (!value.label) {
+        value.label = '@';
+    }
+
+    if (value.label !== '@') {
+        value.label = value.selector + '._domainkey' + '.' + value.label;
+    } else {
+        value.label = value.selector + '._domainkey';
+    }
+
+    kvs.push('v=DKIM1');
+    if (value.keytype) {
+        kvs.push('k=' + value.keytype);
+    }
+
+    if (value.servicetypes) {
+        kvs.push('s=' + value.servicetypes);
+    }
+
+    if (value.flags && value.flags.length > 0) {
+        kvs.push('t=' + value.flags.join(':'));
+    }
+
+    if (value.hashtypes && value.hashtypes.length > 0) {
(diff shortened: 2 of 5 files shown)


cafferata (approved):
The documentation looks great. Thanks for the clear work. I’m approving this.

One optional suggestion: it might be helpful to define TypeScript types to improve clarity and robustness.

That said, feel free to merge it as it is. Thanks again for the pull request, @costasd.

documentation/language-reference/domain-modifiers/DKIM_BUILDER.md
```diff
@@ -0,0 +1,92 @@
...
+  }),
+);
+```
+{% endcode %}
+
+This yields the following record:
+
+```text
+
+k2._domainkey.alerts    IN  TXT "v=DKIM1; k=rsa; s=email; t=y; h=sha256; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDC5/z4L" ttl=150
```
cafferata, line 74: **Question**: The advanced example currently shows the output as:

```text
k2._domainkey.alerts    IN  TXT "v=DKIM1; k=rsa; s=email; t=y; h=sha256; p=..." ttl=150
```

Shouldn't the TTL appear before `IN`, like this?

```text
k2._domainkey.alerts 150 IN TXT "v=DKIM1; k=rsa; s=email; t=y; h=sha256; p=..."
```

Just checking to confirm what's technically correct.

Thanks!

pkg/js/helpers.js
```diff
@@ -1739,6 +1739,77 @@ function CAA_BUILDER(value) {
...
+    kvs.push('p=' + value.pubkey);
+
+    var DKIM_TTL = function () {};
+    if (value.ttl) {
+        DKIM_TTL = TTL(value.ttl);
+    }
+
+    r = []; // The list of records to return.
+    r.push(
+        TXT(value.label, kvs.join('\; '), DKIM_TTL)
```
cafferata, line 1816: **Question**: shouldn't this be:

```diff
-kvs.join('\; ')
+kvs.join('; ')
```

It looks like the backslash is being included literally in the output, which isn't required (or valid) in DNS TXT records. Just checking to confirm.

documentation/language-reference/domain-modifiers/DKIM_BUILDER.md
```diff
@@ -0,0 +1,92 @@
...
+
+* `label:` The DNS label for the DKIM record (`[selector]._domainkey` prefix is added; default: `'@'`)
+* `selector:` Selector used for the label. e.g. `s1` or `mail`
+* `pubkey:` Public key `p` to be used for DKIM.
+* `keytype:` Key type `k`. Defaults to `'rsa'` if omitted (optional)
+* `flags:` Which types `t` of flags to activate, ie. 'y' and/or 's'. Array, defaults to 's' (optional)
+* `hashtypes:` Acceptable hash algorithms `h` (optional)
+* `servicetypes:` Record-applicable service types (optional)
+* `note:` Note field `n` for admins. Avoid if possible to keep record length short. (optional)
+* `ttl:` Input for `TTL` method (optional)
```
cafferata, line 88: **Textual suggestions**:

```suggestion
* `label:` Optional DNS label under which to place the record (e.g. `alerts`). Defaults to `'@'`. The full name becomes `[selector]._domainkey.[label]`.
* `selector:` Required selector (e.g. `'s1'`). Forms the name of the TXT record as `[selector]._domainkey`.
* `pubkey:` Required public key (`p=` value) to publish.
* `keytype:` Optional key type (`k=`). Defaults to `'rsa'`.
* `flags:` Optional array of DKIM flags (`t=`), e.g. `['y', 's']`. Defaults to `['s']`.
* `hashtypes:` Optional array of accepted hash algorithms (`h=`), e.g. `['sha256']`.
* `servicetypes:` Optional array of service types (`s=`), usually `['email']`.
* `note:` Optional administrative note (`n=`). Be cautious of TXT record length limits.
* `ttl:` Optional TTL for the record (number or `Duration` object).
```

documentation/language-reference/domain-modifiers/DKIM_BUILDER.md
```diff
@@ -0,0 +1,92 @@
...
+k2._domainkey.alerts    IN  TXT "v=DKIM1; k=rsa; s=email; t=y; h=sha256; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDC5/z4L" ttl=150
+
+```
+
+### Parameters
+
+* `label:` The DNS label for the DKIM record (`[selector]._domainkey` prefix is added; default: `'@'`)
+* `selector:` Selector used for the label. e.g. `s1` or `mail`
+* `pubkey:` Public key `p` to be used for DKIM.
+* `keytype:` Key type `k`. Defaults to `'rsa'` if omitted (optional)
```
cafferata, line 83: I noticed that the rsa key type is not explicitly set in the DKIM record. This causes a DNSControl diff due to the missing `k=rsa` attribute.

Here’s a local example from cafferata.dev showing the difference:

```diff
± MODIFY mandrill._domainkey.cafferata.dev TXT ("v=DKIM1; k=rsa; p=...") -> ("v=DKIM1; p=...")
```

It may be worth explicitly including the k=rsa to prevent unnecessary diffs, even if it’s the default.

pkg/js/helpers.js
```diff
@@ -1739,6 +1739,77 @@ function CAA_BUILDER(value) {
...
+        value = {};
+    }
+    kvs = [];
+
+    if (!value.selector) {
+        throw 'DKIM_BUILDER selector cannot be empty';
+    }
+
+    if (!value.pubkey) {
+        throw 'DKIM_BUILDER pubkey cannot be empty';
```
cafferata, line 1772: Just a small clarification regarding DKIM: although an empty `p=` value in a DKIM record might look invalid at first glance, it is actually allowed by the DKIM specification (RFC 6376, section 3.6.1).

A record like this:

```text
v=DKIM1; p=
```

is valid and signals that the selector should not be trusted. In practice, I use this to **deactivate a DKIM selector** or to **prevent spoofing** on a domain or subdomain by making sure no valid key is published.

This technique is a deliberate choice, especially for domains or subdomains that should not be sending email at all.

tlimoncelli:
As an experiment I converted one of our DKIM records:

```
DKIM_BUILDER({
selector: "dk",
pubkey: "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC3Gnks18obM9XO0ttwJZzF4KZ5ItqV+1mj6G6YMYHs+t37vx9anqDYLCIBaBlwdmNN3c0EeBKLSsOYvdfC7veNTAbNO8cZbphsnbKt/omVJNUZuzHqMt1QdssB0Eo8COGcJPTE7psVE1fIz3lCFvaxeXGgaQCq0KoMe6dfEKYfXwIDAQAB",
 }),
```

and `preview` output this:

```
#1: ± MODIFY dk._domainkey.serverfault.com TXT 
("k=rsa; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC3Gnks18obM9XO0ttwJZzF4KZ5ItqV+1mj6G6YMYHs+t37vx9anqDYLCIBaBlwdmNN3c0EeBKLSsOYvdfC7veNTAbNO8cZbphsnbKt/omVJNUZuzHqMt1QdssB0Eo8COGcJPTE7psVE1fIz3lCFvaxeXGgaQCq0KoMe6dfEKYfXwIDAQAB" ttl=300) -> 
("v=DKIM1; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC3Gnks18obM9XO0ttwJZzF4KZ5ItqV+1mj6G6YMYHs+t37vx9anqDYLCIBaBlwdmNN3c0EeBKLSsOYvdfC7veNTAbNO8cZbphsnbKt/omVJNUZuzHqMt1QdssB0Eo8COGcJPTE7psVE1fIz3lCFvaxeXGgaQCq0KoMe6dfEKYfXwIDAQAB" ttl=300) id=0cb3de7742092cbeaf61b703c1a447ac
```

Is that an equivalent string?   It seems like it would be if `k` defaults to `rsa` and if `v` defaults to `DKIM1`.

pkg/js/helpers.js
costasd: good catch!

documentation/language-reference/domain-modifiers/DKIM_BUILDER.md
costasd: agreed, will fix.

costasd:
> As an experiment I converted one of our DKIM records:
> 
> ```
> DKIM_BUILDER({
> selector: "dk",
> pubkey: "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC3Gnks18obM9XO0ttwJZzF4KZ5ItqV+1mj6G6YMYHs+t37vx9anqDYLCIBaBlwdmNN3c0EeBKLSsOYvdfC7veNTAbNO8cZbphsnbKt/omVJNUZuzHqMt1QdssB0Eo8COGcJPTE7psVE1fIz3lCFvaxeXGgaQCq0KoMe6dfEKYfXwIDAQAB",
>  }),
> ```
> 
> and `preview` output this:
> 
> ```
> #1: ± MODIFY dk._domainkey.serverfault.com TXT 
> ("k=rsa; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC3Gnks18obM9XO0ttwJZzF4KZ5ItqV+1mj6G6YMYHs+t37vx9anqDYLCIBaBlwdmNN3c0EeBKLSsOYvdfC7veNTAbNO8cZbphsnbKt/omVJNUZuzHqMt1QdssB0Eo8COGcJPTE7psVE1fIz3lCFvaxeXGgaQCq0KoMe6dfEKYfXwIDAQAB" ttl=300) -> 
> ("v=DKIM1; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC3Gnks18obM9XO0ttwJZzF4KZ5ItqV+1mj6G6YMYHs+t37vx9anqDYLCIBaBlwdmNN3c0EeBKLSsOYvdfC7veNTAbNO8cZbphsnbKt/omVJNUZuzHqMt1QdssB0Eo8COGcJPTE7psVE1fIz3lCFvaxeXGgaQCq0KoMe6dfEKYfXwIDAQAB" ttl=300) id=0cb3de7742092cbeaf61b703c1a447ac
> ```
> 
> Is that an equivalent string? It seems like it would be if `k` defaults to `rsa` and if `v` defaults to `DKIM1`.

Good one. It's an equivalent. I think this is relevant to @cafferata's https://github.com/StackExchange/dnscontrol/pull/3627#discussion_r2159653177 comment above. Some fields have defaults and if missing are assumed. It feels to me though that DKIM_BUILDER should always set them and never omit them (ie. be explicit not implicit). I'll run a few experiments.

n-rodriguez:
> suggestion: Rename the `pubkey` parameter to be `p` (and similar for other parameters). By having it be different, the user will always have to look up the parameter's name. It becomes more user-friendly to have the names match. (or at least accept `p` as an alias)

I'm not sure this is a good idea. Per https://docs.dnscontrol.org/advanced-features/code-tricks#caveats-about-getting-too-fancy (YSMFN) having just `p` instead of `pubkey` might be confusing (is this the private key? the public key? the priority?)

Being explicit in this case is better than implicit.

costas-at-bunny:
hmm, I'll follow up with the suggested fixes before the next release. Sorry for dropping the ball here, summer vacations happened.

PR #3638, opened 2025-06-22
NEW PROVIDER: ADGUARDHOME

ishanjain28:
This PR adds support for AdGuardHome as a DNS Provider using the rewrite rules feature in AGH. 

I added the provider and a few basic test cases but AGH rewrite rules only support few record types(vs the whole basic set of rtypes) so tests don't _really_ work for it). Theoretically, one can implement support for more record types using filtering rules instead of rewrite rules but it will be more complicated and it feels unnecessary so for now, this provider will return an error for all but the 6(4 standard, 2 specific to AGH) rtypes.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -165,7 +165,6 @@ func makeTests() []*TestGroup {
 		// correctly!
 
 		testgroup("MX",
-			not("ADGUARDHOME"),
 			tc("Create MX", mx("testmx", 5, "foo.com.")),
 			tc("Change MX target", mx("testmx", 5, "bar.com.")),
 			tc("Change MX p", mx("testmx", 100, "bar.com.")),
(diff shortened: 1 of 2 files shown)


ishanjain28:
Hi @tlimoncelli in the tests, How do I ignore TTL parameter? Adguard home doesn't really support that at all. 
I also don't have a way to include it as metadata with a record and just return the same value that was sent in the create/modify requests.

ishanjain28:
Figured it out. 


@tlimoncelli  Addressed both concerns in the two commits. Added `ADGUARDHOME_` prefix to both custom records and updated the tests to exclude AdguardHome when testing rtypes that are not supported by it

ishanjain28:
Hii @tlimoncelli no need for apologies :) 
It is a fair ask and I agree with your suggestion. I'll update the PR to reject unsupported rtypes using auditrecords and revert the changes to the tests in a few minutes

ishanjain28:
@tlimoncelli Applied the suggestions and it is ready for review again ✅

ishanjain28:
@tlimoncelli Rebased ✅

cafferata:
@tlimoncelli can you add label `ADGUARDHOME`?

tlimoncelli:
Done!

PR #3652, opened 2025-07-08
PORKBUN: Improve retry handling, mark as concurrent

### LINKED ISSUE #2995: Porkbun does not work concurrently (`ppreview`)
**Describe the bug**
Testing the new concurrency feature (`ppreview`) on domains handled by the Porkbun provider it fails to handle those:
```
INFO#1: Zone "travelociraptor.xyz" does not exist. Can not create because "porkbun" does not implement ZoneCreator
```

But the zone exists ... all works well when running the normal preview.

**To Reproduce**
1. Have (multiple) domains in dnscontrol setup that use Porkbun
2. `dnscontrol ppreview --cmode all`
3. Multiple `Zone "example.com" does not exist. Can not create because "porkbun" does not implement ZoneCreator` issues.

**Expected behavior**
For Porkbun to work concurrently.

**DNS Provider**
 - Porkbun

**Additional context**
Tested with dnscontrol v4.11.0.

I've checked the "Concurrency Verified" column in https://docs.dnscontrol.org/getting-started/providers ... but I was not sure if the ❌ means "verified & failed" or "not verified".

At least in my testing, when I use `dnscontrol ppreview --cmode all --domains example1.com,example2.com` it still works (most of the time). It only starts throwing issues regularly when I test it with at least 3 different zones.

CC @imlonghao (maintainer of the porkbun provider)

jamesog:
### PORKBUN: Improve retry handling, mark as concurrent (9ba2c08e)

The Porkbun API frequently responds with HTTP 503. It seems like they
use this instead of 429.

The provider previously would cause the HTML response from the 503 to
leak through to the caller, causing JSON decode to fail: the provider
only checked for response codes 202 and 503. Before checking the
response code it had already read the body. If the API responded with
anything other than 202 or 503, the response body was returned as-is,
even if the response was not 200. Per the Porkbun API docs:

    Any HTTP response code other than 200 is an error

But the provider was not checking this. If there is a 5xx error, the
response body contains HTML, which causes the caller to output a warning
such as

    INFO#1: Domain "example.com" provider porkbun Error: failed parsing url forwarding record list from porkbun: invalid character '<' looking for beginning of value

Now the Failsafe HTTP library is used, which provides automatic handling
for 429 and any 5xx-class response. When a response has a Retry-After
header, the retry policy takes this into account and delays by an
acceptable amount. The retry policy is configured with the same number
of retries as before (5), with exponential backoff between 1 and 10
seconds, with a jitter of +/- 100ms.

Because the retry policy is configured as an HTTP client transport there
is no chance of leaking HTML to the caller because the post() method
sees this as an HTTP client error and fails earlier.

With these changes the Porkbun provider is now safe to use concurrently.

Fixes #2995

### CHORE: Add stringer as a tool (7e2a0ddc)

The stringer tool is needed by `go generate` and when adding new rtypes.

We now use the Go toolchain's ability to mark an external command as a
tool, and run it using `go tool` instead of requiring it to be installed
on the developer's machine.

### DIFF
--- a/documentation/advanced-features/adding-new-rtypes.md
+++ b/documentation/advanced-features/adding-new-rtypes.md
@@ -16,12 +16,10 @@ Our general philosophy is:
 -   Anywhere we have a special case for a particular Rtype, we use a `switch` statement and have a `case` for every single record type, usually with a `default:` case that calls `panic()`. This way developers adding a new record type will quickly find where they need to add code (the panic will tell them where). Before we did this, missing implementation code would go unnoticed for months.
 -   Keep things alphabetical. If you are adding your record type to a case statement, function library, or whatever, please list it alphabetically along with the others when possible.
 
-Step 2 requires `stringer`.
+Step 2 requires `stringer` which is available using
 ```shell
-go install golang.org/x/tools/cmd/stringer@latest
+go tool stringer
 ```
-You may need to symlink stringer into your PATH.
-
 ## Step 1: Update `RecordConfig` in `models/record.go`
 
 If the record has any unique fields, add them to `RecordConfig`.
@@ -61,7 +59,7 @@ a minimum.
 
 ```shell
 pushd; cd providers/;
-stringer -type=Capability
+go tool stringer -type=Capability
 popd
 ```
 alternatively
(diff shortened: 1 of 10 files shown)


tlimoncelli:
Looks great! Thanks!

PR #3653, opened 2025-07-08
BUILD: Add version package

### LINKED ISSUE #3651: Regression in version string set at compile time

**Describe the bug**
The changes to the version string made in #2737 and #2761 seem to have broken the version string for the two providers that use it to set a User-Agent header. I discovered this when wanting to add the same behaviour to another provider.

Previously there was a `package version` which provided a version string to all packages. The above pull requests removed that package and replaced uses of it in all files with:

```go
// GoReleaser: version
var version = "dev"
```

As best I can tell, the comment does nothing (it's not a documented build-time replacement that I can see) and the `version` variable is only set by GoReleaser in package main.

This also affects commands/writeTypes.go, but it looks like this is unused.

**To Reproduce**
1. In any provider, set a version variable, as above
2. Set a request header that uses the version variable
3. Add a debug print to print out the request headers prior to making the request
4. Observe the header has value "dev"

**Expected behavior**
The version should have the version as provided by linker flags.

**DNS Provider**
N/A - affects any package that wishes to use the version.

**Additional context**
I think restoring `package version` and having GoReleaser set the version there would be the right thing to do. This is a fairly common pattern.

Looking at the code that used to be in that package it would be useful as a fallback in case a version wasn't set using linker flags.

jamesog:
Commit 9873f9f4b2fece7a46053eaa375fe24070971e09 (v4.7.4) removed the version package in favour of using GoReleaser to set the version, which it only sets in main and is not passed down to other packages.

This means that if any other package, notably two providers which use the version for setting a User-Agent, wants access to the version, they no longer have it.

This change reintroduces the version package, setting GoReleaser to set the version here, and refactors all uses of the version to use the package.

Note that if you run `go run main.go version` the version output by Go for the main module is "(devel)". Using build/build.go gives the commit hash.

Fixes #3651

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -22,7 +22,7 @@ builds:
         goarch: "386"
     ldflags:
       - -linkmode=internal -s -w
-      - -X main.version={{ .Version }}
+      - -X github.com/StackExchange/dnscontrol/v4/pkg/version.version={{ .Version }}
 before:
   hooks:
     - go fmt ./...
--- a/build/build.go
+++ b/build/build.go
@@ -15,7 +15,7 @@ var goos = flag.String("os", "", "OS to build (linux, windows, or darwin) Defaul
 
 func main() {
 	flag.Parse()
-	flags := fmt.Sprintf(`-s -w -X "main.version=%s"`, getVersion())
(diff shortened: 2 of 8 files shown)


tlimoncelli:
After I merged this I realized that "go run main.go version" outputs a blank line.  Can you double-check and submit a PR if that's a bug?

PR #3657, opened 2025-07-09
CHORE: Remove "get-certs" command

tlimoncelli:
# Issue

The `get-certs` command can now be removed since its only user no longer uses it.

This was announced about a year ago.  See discussion in [issues/1400](https://github.com/StackExchange/dnscontrol/issues/1400)

# Resolution

Remove it.


tlimoncelli:
@cafferata FYI: I'm leaving a stub file for get-certs (`documentation/commands/get-certs.md`) but removing it elsewhere.

PR #3664, opened 2025-07-14
chore: update go-powerdns to v0.6.7 and adjust SOAEditAPI type

jpbede:
Bump go-powerdns to v0.6.7 and adjust SOAEditAPI type. This replaces #3647


tlimoncelli:
Thanks so much!

PR #3669, opened 2025-07-16
DOCS: Remove get-certs and MSDNS from release notes

gucci-on-fleek:
Both of these features have already been removed, so it makes no sense to mention that they will be removed in the future. I noticed this in the v4.22.0 release notes, so this PR should fix this for any future releases.


cafferata (approved):

tlimoncelli:
Thanks!

PR #3675, opened 2025-07-21
DNSIMPLE: preview data is now gathered concurrently (CanConcur)

philpennock:
The `dnsimpleProvider` is a simple object of 3 strings with no state.

Using `dnscontrol preview --cmode all` in a configuration with 12 DNSimple zones, and the binary built with race detection, results in no warnings.

Fixed `concurrency-verified.md` to state "the provider in question" instead of an early provider to support this.  (Also some end-of-line cleanups for Markdown; trailing whitespace is semantically significant in block mode but not AFAIK for lists, so I think this is a safe change; this was my text-editor honoring the project's `.editorconfig` settings, which don't have an EOL whitespace exemption for markdown).

Index doc regenerated.

### DIFF
--- a/documentation/advanced-features/concurrency-verified.md
+++ b/documentation/advanced-features/concurrency-verified.md
@@ -2,15 +2,15 @@
 name: CONCURRENCY_VERIFIED
 ---
 
-✅  - A checkmark means "this has been tested, and will run concurrently".  
-❔  - The questionmark means "it hasn't been tested, safety unknown"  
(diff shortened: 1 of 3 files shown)


tlimoncelli (approved):

tlimoncelli:
Thanks!

weppos:
@philpennock I just happened to review the release note for recent versions, and I noticed this set of improvements around concurrency for DNSimple. Thanks a lot for you improvements 🙏

PR #3678, opened 2025-07-22
DNSIMPLE/DOCS: handle multiple provider instances

philpennock:
The previous work to use `sync.Once` unintentionally broke (in theory) anyone
using multiple providers of type `DNSIMPLE`, as only one would have been
initialized.

Checked carefully to make sure it's safe to use a `sync.Once` inside the
provider object.

Updated documentation with guidance for other provider maintainers.


philpennock:
I woke up this morning and before my eyes were even willing to stay open, this realization came to me -- apparently out of nowhere -- and I would have facepalmed but it would have been too much effort.

PR #3687, opened 2025-07-24
TESTING: Exempt 4 providers from NS_only_APEX test

tlimoncelli:
# Issue

GANDI_V5, NAMEDOTCOM, SAKURACLOUD and TRANSIP do not allow manipulating NS records at the apex.

# Resolution

Exempt them from tests.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -403,9 +403,13 @@ func makeTests() []*TestGroup {
 
 		testgroup("NS only APEX",
 			not(
-				"DNSIMPLE", // Does not support NS records nor subdomains.
-				"EXOSCALE", // Not supported.
-				"NETCUP",   // NS records not currently supported.
+				"DNSIMPLE",    // Does not support NS records nor subdomains.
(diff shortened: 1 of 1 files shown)


vatsalyagoel:
https://github.com/StackExchange/dnscontrol/commit/a815bdcf4533610eb65c2f7272564f1ed90a9d3a

New test added but looks like exclusions weren't set properly. 

* GANDI Does not support changing apex NS Records
* TRANSIP Does not support having NS for @ Records
* SAKURA is creating a change on the second run (Thus most likely ignoring the first API call to add @ NS Record)
* NAMEDOTCOM Ignores @ for NS records

tlimoncelli:
New exclusions added.  Thanks, @vatsalyagoel !

tlimoncelli:
@chicks-net Does that capabilities code give you what you wanted?

chicks-net:
> @chicks-net Does that capabilities code give you what you wanted?

I think so.  I can't test for a few days, but this looks reasonable.

Issue #3530, opened 2025-04-14
Reduce new-user pain-points via better documentation and defaults


tlimoncelli:
# Issue

At the March 2025 video call, these pain-points were mentioned. It seems a little bit of documentation would really help.

- Using “@” as the apex.
  - The expectation is that the empty string would work.
  - Suggestion 1: The getting started doc should explain that "@" is used for the apex
  - Suggestion 2: When the user specifies `""`, the error message should suggest using `"@"`.
  - Suggestion 3: When the user specifies "example.com" (the apex domain), the error message should suggest using `"@"`

- Converting zones: Doing the first “push” is nervous-making.  How to migrate a zone safely?
  - The docs should suggest some testing methods to calm fears.
  - The docs should remind people that "get-zones" isn't perfect, but produces "a good first draft" you can edit.
  - Suggest trying iterations of `preview` and editing dnsconfig.js until no "changes" are listed. 

- The fact that “preview” creates zones is surprising.  Normally “preview” would be non-destructive.
  - A user might accidentally create many zones at a provider; this can cost money.
  - `--populate-on-preview` defaults to true.  Should probably default to false. https://github.com/StackExchange/dnscontrol/issues/3399

schlegel11:
Totally agree with at least `--populate-on-preview` behavior should be changed to false. I did not know about this flag and was:
1. Totally surprised to see changes while I'm still using preview.
2. Thought at first the provider implementation had a bug or something because I was unaware of the `--populate...` option.

Issue #3399, opened 2025-01-22
New zones are not created in push with --populate-on-preview=false set


Veratil:
**Describe the bug**
When you do not want to create a zone on preview and set `--populate-on-preview=false`, and then change to a push keeping the option there, the zones will not be created.

**To Reproduce**
1. Create a config with a new zone
2. push with `--populate-on-preview=false`
3. See records failed to be created since the zones weren't created

**Expected behavior**
Zones are created even if `--populate-on-preview=false` is set on a push

**DNS Provider**
 - POWERDNS

**Additional context**
I'm using an ansible playbook (builtin.shell, along with an `-e method=[preview|push]`) to run the dnscontrol docker container, and it's easier to keep the `--populate-on-preview=false` as an argument to shell instead of trying to hand wave it in there when the method variable changes.

tlimoncelli:
CC @das7pad 

Suggestion (but only if it's not a huge lift):  Should we be rethinking the flags?  We have `--no-populate` and `--populate-on-preview`.   Maybe there should be a single flag that accepts values like: "preview" (create on preview), "push" (create on push), "both" (do both), "none" (The default).

das7pad:
I've opened https://github.com/StackExchange/dnscontrol/pull/3402 with a quick one-line fix while we discuss the flags.

> Should we be rethinking the flags? We have `--no-populate` and `--populate-on-preview`. Maybe there should be a single flag that accepts values like: "preview" (create on preview), "push" (create on push), "both" (do both), "none" (The default).

Consolidating these (and while we are at it, removing the unused `--depopulate` flag) sounds like a good idea. Personally, I would go for "preview" as the default for ease of use and allowing users to see the "actual diff" right away.

How would the deprecation look like? Do we warn users when they specified both old and the new flag? Which "wins" in that case? (I would lean towards the new one being authoritative.)

tlimoncelli:
Excellent questions!

We generally remove flags with 3-4 months notice. We open a bug id that lists the timeline.  We mention the change in the footer of release announcements so people see reminders many times until the change happens.

Tom

PR #3693, opened 2025-07-26
The --populate-on-preview flag should default false, which is safer

tlimoncelli:
# Issue

As reported in https://github.com/StackExchange/dnscontrol/issues/3530 ...

- The fact that “preview” creates zones is surprising.  Normally “preview” would be non-destructive.
  - A user might accidentally create many zones at a provider; this can cost money.
  - `--populate-on-preview` defaults to true.  Should probably default to false. https://github.com/StackExchange/dnscontrol/issues/3399

# Resolution

Change the default of `--populate-on-preview` to be false, which is safer.

This may be a bit of a surprise to people relying on this feature as it violates semver, but considering that it reduces risk, I don't mind this.  It is not worth bumping the major version.


tlimoncelli:
CC @das7pad for visibility

PR #3694, opened 2025-07-28
BUG: Improve error handling during zone population and gathering

### LINKED ISSUE #2474: CLOUDFLAREAPI: Exit code 0 even if ending with ERROR due to invalid API token

gvengel:
There were some errors that occurred early in zone population and gathering which were only being returned as correction messages. While the messages were printed for the user, they did not affect the exit code of the program. This prevented CI/CD from detecting some critical failures like bad login credentials, connectivity issues, etc.

This change bubbles these errors up to the main push/preview command so it can exit appropriately.

Fixes #2474.

### DIFF
--- a/commands/ppreviewPush.go
+++ b/commands/ppreviewPush.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/StackExchange/dnscontrol/v4/models"
 	"github.com/StackExchange/dnscontrol/v4/pkg/bindserial"
@@ -217,6 +218,7 @@ func prun(args PPreviewArgs, push bool, interactive bool, out printer.CLI, repor
 	var totalCorrections int
 	var reportItems []*ReportItem
 	var anyErrors bool
+	var concurrentErrors atomic.Bool
 
 	// Populate the zones (if desired/needed/able):
 	if !args.NoPopulate {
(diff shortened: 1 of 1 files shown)


tlimoncelli:
Wow!  Thanks for fixing this!  You really dug into the code and figured out what's wrong!  Thanks!

PR #3683, opened 2025-07-23
Warn if IGNORE() is unreliable for this provider

tlimoncelli:
# Issue

IGNORE() is unreliable for providers that haven't adopted the diff2 library.

A few years ago we replaced the diffing engine ("diff") with a new one that handled IGNORE() and many features much better.  Most of the providers were rewritten to use the new library.  Since rewriting the provider was such a burden (and in some cases the maintainer of the provider had disappeared), I created a backwards compatibility mode called diff.NewCompat().  While it had the benefit of only requiring 2-3 lines of code changes, it doesn't implement IGNORE() reliably in all cases.  Other, newer, features also don't work.

# Resolution

1. Add a new capability called `CanOnlyDiff1Features` to mark which providers have not been upgraded yet. It is not automatically generated, thus we should be careful when converting providers to use diff2.

2. Alphabetize the "features" list for affected providers.  It seems that CanGetZones() wasn't added in the right place.

3. Output a warning if any of the following features are used:

* ENSURE_ABSENT
* NO_PURGE
* IGNORE() and friends
* DISABLE_IGNORE_SAFETY_CHECK

FYI:  Affected providers:

```
providers/akamaiedgedns/akamaiEdgeDnsProvider.go
providers/cloudns/cloudnsProvider.go
providers/cnr/records.go
providers/cscglobal/dns.go
providers/desec/desecProvider.go
providers/digitalocean/digitaloceanProvider.go
providers/dnsimple/dnsimpleProvider.go
providers/dnsmadeeasy/dnsMadeEasyProvider.go
providers/domainnameshop/dns.go
providers/exoscale/exoscaleProvider.go
providers/hetzner/hetznerProvider.go
providers/hexonet/records.go
providers/hostingde/hostingdeProvider.go
providers/linode/linodeProvider.go
providers/loopia/loopiaProvider.go
providers/namecheap/namecheapProvider.go
providers/namedotcom/records.go
providers/netcup/netcupProvider.go
providers/netlify/netlifyProvider.go
providers/oracle/oracleProvider.go
providers/packetframe/packetframeProvider.go
providers/rwth/dns.go
providers/softlayer/softlayerProvider.go
```

PR #3695, opened 2025-07-28
DIGITALOCEAN: Use diff2 to implement digitalocean provider

Deraen:
Using diff2 with ByRecord was a pretty simple change. Based on #3683 I guess switching to the new implementation makes sense.

Somewhat unrelated note, I think DocDualHost capability could be enabled, it works on the API and looks OK on the Digitalocean UI, but I don't know if there is a chance it won't really work in a real use.

### DIFF
--- a/providers/digitalocean/digitaloceanProvider.go
+++ b/providers/digitalocean/digitaloceanProvider.go
@@ -10,7 +10,7 @@ import (
 	"time"
 
 	"github.com/StackExchange/dnscontrol/v4/models"
-	"github.com/StackExchange/dnscontrol/v4/pkg/diff"
+	"github.com/StackExchange/dnscontrol/v4/pkg/diff2"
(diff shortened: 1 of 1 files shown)


tlimoncelli:
> Using diff2 with ByRecord was a pretty simple change. Based on #3683 I guess switching to the new implementation makes sense.

Looks great!  Will merge.

> Somewhat unrelated note, I think DocDualHost capability could be enabled, it works on the API and looks OK on the Digitalocean UI, but I don't know if there is a chance it won't really work in a real use.

Set it to Can() and run this:

    go test -run TestDualProviders -v -verbose -profile DIGITALOCEAN

If that works, create a PR and we're good to go!


Thanks for upgrading to diff2!  This really helps users of this provider!  Now they have access to a ton of new features!

PR #3713, opened 2025-08-06
DOCS: AXFR+DDNS uses TCP by default

Mynacol:
The observed behavior of AXFR+DDNS in DNSControl 4.20.0 is that the DDNS update uses TCP by default as well.


tlimoncelli:
@hnrgrgr: do you agree?

PR #3718, opened 2025-08-07
NEW RECORD TYPE: OPENPGPKEY (basic)

tacerus:
# Issue

OPENPGPKEY is not supported.  This makes it difficult to import zones that have such records.

A full implementation is not needed. A good first step is to simply implement the recored requiring raw input. Later we can add more useful encoding/presentation features.

# Resolution

Add basic functionality. 

This is a follow-up to the efforts in:
https://github.com/StackExchange/dnscontrol/issues/2370
https://github.com/StackExchange/dnscontrol/pull/2374

Whilst full parsing and conversion of address and key material would be preferable and more in line with the ideals of dnscontrol, the logic needed requires significant development effort.

In order not to further completely prevent zones containing OPENPGPKEY records from being managed by dnscontrol, add basic support for the record type which enables a 1:1 mapping of OPENPGPKEY records from dnscontrol to enabled providers.


tacerus:
I tested this with PowerDNS.

systemcrash (commented):
At this stage, it implements basic support for the record type which will work, but the UTF-8 normalisation steps are absent, and assumes that the user will has correctly 'encoded' the `x._openpgpkey` portion, of which the help md document makes no mention. 

This translates to availability but to users who know how to utilise the record properly.

The greatest challenge to this record type is knowing from where the `x._openpgpkey` ingresses: via an existing DNS record in the system, or via a js file? If these can be distinguished, then where to put the normalisation steps (written in Go) will be clear. I think this was what I got stuck on.

PR #3733, opened 2025-08-18
security fix: Prevent exposure of creds in debug output

tlimoncelli:
# Issue

Per https://github.com/StackExchange/dnscontrol/security/advisories/GHSA-pxw8-rhf9-gh9x#event-479298

CodeQL is reporting that in debug mode some creds may be output to the console.  That said, anyone running in debug mode would have access to the `creds.json` file, thus they're not gaining anything they don't already have.

# Resolution

Comment out the code.  It can be un-commented when needed.

CC @systemcrash


systemcrash:
I agree with this assessment. The XML bodies are visible during debugging for obvious reasons, where credentials from the user side are visible during auth.

Issue #3737, opened 2025-08-22
Provider Request: Yandex Cloud DNS


klemensn:
[DNS](https://yandex.cloud/en/docs/dns/) is one of many services on their platform.

Its [HTTP API](https://yandex.cloud/en/docs/dns/api-ref/) supports REST and gRPC.

PR #3738, opened 2025-08-22
DOCS: Added requested provider Yandex Cloud DNS

klemensn:
Added Yandex Cloud DNS #3737 to the list of requested providers.


klemensn:
Following https://docs.dnscontrol.org/developer-info/provider-request#id-4.-commit-your-changes resulted in a conflict, so rebased onto main instead.

PR #3746, opened 2025-08-27
PORKBUN: Configurable retry parameters

jamesog:
The Porkbun provider can now configure the API retry mechanism by setting max_attempts and/or max_duration in the credentials configuration. This aims to help users who have many domains configured and run in to Porkbun API limits with the current retry mechanism.

The max_attempts setting controls how many attempts (in total, not retries) will be made for each domain. The default is 5.

The max_duration allows the retry mechanism to keep retrying up to this duration, for each domain.

The current behaviour is to do exponential backoff between 1 second and 10 seconds, with 5 retries (6 total attempts), with jitter of +/-100ms (max, it will vary). The jitter means the provider will often retry sooner than 1 second, which the Porkbun API doesn't like. To counter this, the minimum retry time is now 1.2 seconds.

With the new behaviour the max backoff time is still 10 seconds, but if a long max_duration is set it allows the provider to keep retrying every 10 seconds.

### DIFF
--- a/providers/porkbun/porkbunProvider.go
+++ b/providers/porkbun/porkbunProvider.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/StackExchange/dnscontrol/v4/models"
 	"github.com/StackExchange/dnscontrol/v4/pkg/diff2"
@@ -23,6 +24,10 @@ const (
 	metaType        = "type"
 	metaIncludePath = "includePath"
 	metaWildcard    = "wildcard"
+)
+
+const (
+	defaultMaxAttempts = 5
 )
 
 // https://kb.porkbun.com/article/63-how-to-switch-to-porkbuns-nameservers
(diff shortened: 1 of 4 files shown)


cafferata:
Ping @imlonghao, the maintainer of the Porkbun provider.

stefanwascoding:
This works (for me).
It seems like setting `max_attemps` to `0` might indicate "infinite" retries (`5` fails, `10` works, `0` works), so maybe add that info to the docs.

providers/porkbun/porkbunProvider.go
```diff
@@ -43,14 +48,31 @@ func newDsp(conf map[string]string, metadata json.RawMessage) (providers.DNSServ
...
 	c.apiKey, c.secretKey = m["api_key"], m["secret_key"]
 
 	if c.apiKey == "" || c.secretKey == "" {
 		return nil, errors.New("missing porkbun api_key or secret_key")
 	}
 
+	if maxAttempts, ok := m["max_attempts"]; ok && maxAttempts != "" {
+		i, err := strconv.Atoi(maxAttempts)
+		if err != nil {
+			return nil, errors.New("porkbun: invalid max_attempts: must be a whole number")
```
tlimoncelli: Please output the invalid value too.  It makes it easier for the user to find what the change in `creds.json`.

providers/porkbun/porkbunProvider.go
```diff
@@ -43,14 +48,31 @@ func newDsp(conf map[string]string, metadata json.RawMessage) (providers.DNSServ
...
+		}
+		c.maxAttempts = i
+	}
+	if maxDuration, ok := m["max_duration"]; ok && maxDuration != "" {
+		d, err := time.ParseDuration(maxDuration)
+		if err != nil {
+			return nil, errors.New("porkbun: invalid max_duration: valid units are ns, us, ms, s, m, h")
```
tlimoncelli: ditto

providers/porkbun/porkbunProvider.go
jamesog: Updated. It now outputs

```
porkbun: invalid max_attempts "a": must be a whole number
```

providers/porkbun/porkbunProvider.go
jamesog: Updated. It now outputs

```
porkbun: invalid max_duration "2": valid units are ns, us, ms, s, m, h
```

tlimoncelli: great!

PR #3762, opened 2025-09-04
FEATURE: Support RFC 9495 issuemail tag in CAA records

networkException:
This pull request adds support for the `issuemail` `CAA` tag introduced in [RFC 9495](https://datatracker.ietf.org/doc/html/rfc9495)


networkException:
Given the range of other [registered tags](https://www.iana.org/assignments/pkix-parameters/pkix-parameters.xhtml#caa-properties) it might be best to drop checking entirely here.

tlimoncelli (approved):

tlimoncelli:
Thanks for adding this!

PR #3768, opened 2025-09-12
DOCS: Add tooltips and distinction between not supported and not implemented to feature matrix

louis-lau:
I found the feature matrix to be a bit unclear (Unknown vs unimplemented). It also seemed a shame that the comments built into the codebase were not shown at all, so now they're inside tooltips :)

Feels better to me, but lmk if you have other thoughts.


tlimoncelli:
I, too, can't test it locally. I just keep pushing new commits and wait for the PR to update with a link to the new draft gitbook.

@cafferata, can you assist here?  Adding tooltips seems like a really good idea.

cafferata:
I also think tooltips are a good idea, but as @louis-lau already noticed they are unfortunately not supported in GitBook. As an alternative you could look into GitBook annotations: 

You can write content as [Markdown footnotes](https://www.markdownguide.org/extended-syntax/#footnotes) to add them as annotations in GitBook.

```md
Here's a simple footnote,[^1] and here's a longer one.[^bignote]

[^1]: This is the first footnote.

[^bignote]: Here's one with multiple paragraphs and code.

    Indent paragraphs to include them in the footnote.

    `{ my code }`

    Add as many paragraphs as you like.
```

_https://gitbook.com/docs/creating-content/formatting/inline#annotations_

louis-lau:
The difference between unknown and unimplemented is already pretty nice on its own I'd say, together with the tiny bit of extra info at the top.

I kind of dislike footnotes for this as it will just be a giant list of them at the bottom, but it doesn't look like a choice. I can take a crack at it when I get the time, though my experience with golang is kind of surface level, you may be able to do it much quicker.

TomOnTime:
Friendly ping (now that the org move is complete).

PR #3800, opened 2025-10-22
fix(DKIM_BUILDER): allow empty pubkey

matteotrubini:
Support scenarios where a DKIM public key may be intentionally empty, for example to protect domains without email services.

See: https://www.cloudflare.com/learning/dns/dns-records/protect-domains-without-email/


tlimoncelli:
Thanks for fixing that!

PR #3803, opened 2025-10-23
POWERDNS: Map dnscontrol Tags to powerdns Variants

pefeigl:
Map example.com!internal to example.com..internal for the relevant pdns HTTP API calls.


tlimoncelli:
I'd like to include this in the release we're doing this week.  I'm going to merge this. Please submit a separate PR with the "create-zones" fix (or just update the docs to say that newly created zones won't abide by the tag).  Thanks!

eratzlaff:
I opened a PR related to this feature, to be able to turn it off or on by adding a config to the provider. 
https://github.com/StackExchange/dnscontrol/pull/3829

I looked into how to fix `EnsureZoneExists()`. One alternative is to extend the Providers interface and add a new `EnsureZoneExistsWithMetadata()`. Let me know what do you think about it @tlimoncelli and @pefeigl

PR #3805, opened 2025-10-25
INWX: Let the API (not DNSControl) enforce the RFC 7505 prohibition of mixed regular/null MX records

### LINKED ISSUE #3796: moving from 4.22.0 to 4.26.0 result in an error job
**Describe the bug**
This message was not making the pipeline job to fail in 4.22.0
INFO#1: Domain "domain.com" provider inwx Error: desired configuration contains both Null MX and regular MX records

**To Reproduce**
1. Use the dnscontrol docker image 4.26.0
2. Run a job with for example in this case null MX records 
`/usr/local/bin/dnscontrol preview --full | tee dnscontrol.preview.before.txt`
3. Watch it fail

**Expected behavior**
This is an INFO log, not an ERROR one, it should not fail.

**DNS Provider**
 - inwx

eliheady:
This change simplifies null MX handling and stops enforcing RFC 7505 prohibition of mixed regular/null MX records.

This fixes #3796 

The null MX integration tests are still passing with this change.

* INWX: stop enforcing RFC 7505 locally and let the error from the API inform users of unsupported config


blackshadev:
Ahh that is a shame. Than I don't think the DNSSort will help you out.

PR #3815, opened 2025-10-30
POWERDNS: New record type: LUA

eratzlaff:
Add LUA record type for PowerDNS provider. Follow the same logic as TXT rtypes.

### DIFF
--- a/documentation/language-reference/domain-modifiers/LUA.md
+++ b/documentation/language-reference/domain-modifiers/LUA.md
@@ -10,6 +10,7 @@ parameter_types:
   rtype: string
   contents: string | string[]
   "modifiers...": RecordModifier[]
+provider: POWERDNS
 ---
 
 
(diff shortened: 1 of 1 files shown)


documentation/language-reference/domain-modifiers/LUA.md
```diff
@@ -0,0 +1,117 @@
+---
```
cafferata, line 1: Can you add this item to the `Table of contents`?: 

https://github.com/StackExchange/dnscontrol/blob/70c1febe23b624a4ed1f928b780bbc4569753fc1/documentation/SUMMARY.md#L61-L65

cafferata (changes requested):

documentation/language-reference/domain-modifiers/LUA.md
eratzlaff: I added it below the "Service Provider specific"  LUA is only available to PowerDNS.

PR #3830, opened 2025-11-04
POWERDNS: Zone creation now abides by variant tags

eratzlaff:
Add metatada to the EnsureZoneExists interface


jpbede (approved):
Nice, thanks a lot! This LGTM

PR #3838, opened 2025-11-16
DEPS: Switch to maintained fork of shoutrrr

gvengel:
Upgrade from [containerrr/shoutrrr](https://github.com/containrrr/shoutrrr) v0.8.0 to [nicholas-fedor/shoutrrr](https://github.com/nicholas-fedor/shoutrrr) v0.12.0.

The shoutrrr documentation links now point at the version specific site, and will need to be updated when shoutrrr is updated in go.mod. This should help prevent confusion if nicholas-fedor/shoutrrr adds new features, but dnscontrol hasn't upgraded yet.


gvengel:
I've tested and confirmed Slack still works with Shoutrrr v0.12.0, but not all the other services. Reading through the release notes between 0.8.0 and 0.12.0, there doesn't appear to be any breaking changes, so I except everything else should just work too.

tlimoncelli:
Thanks for helping save this feature!

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/StackExchange/dnscontrol/v4
 
-go 1.25
+go 1.25.3
```
das7pad, line 3: Was that change intentional?

The latest version is 1.25.4 as of Nov 5: https://go.dev/doc/devel/release#go1.25.4

gvengel: No, go mod updated it due to the requirement from shoutrrr.
```
% go mod tidy -go=1.25  
go: github.com/nicholas-fedor/shoutrrr@v0.12.0 requires go@1.25.3, but 1.25 is requested
```

das7pad: I've opened a PR there to split the minimum required Go version from the Go toolchain selection:
- https://github.com/nicholas-fedor/shoutrrr/pull/464

das7pad: The fix has been merged and released as 0.12.1. I've opened a PR in dnscontrol to bump the version and remove the toolchain pinning:

- https://github.com/StackExchange/dnscontrol/pull/3858

PR #3843, opened 2025-11-21
Remove unused --depopulate flag

das7pad:
`PPreviewArgs.DePopulate` is unused and the command line flag `--depopulate` is wired up to the wrong field (`PPreviewArgs.NoPopulate`). This PR is removing both references.


tlimoncelli (approved):

tlimoncelli:
Thanks for cleaning that up!

Issue #3842, opened 2025-11-19
IDNA handling for records is incomplete in dnscontrol


das7pad:
It looks like IDNA handling for records is incomplete in dnscontrol. Take this minimal config, note the `ä` in the domain:

```js
var REG_NONE = NewRegistrar('none')
var DSP = NewDnsProvider('HETZNER_V2')

D('testing-2025-11-17-ä.dev', REG_NONE, DnsProvider(DSP),
  A('@', '127.0.0.1'),
  A('foo', '127.0.0.1')
)
```
Logging the `first.Name` here will print the record name with origin:

```
CREATE first.Name="xn--testing-2025-11-17--ywb.dev"
CREATE first.Name="foo.xn--testing-2025-11-17--ywb.dev"
DONE
******************** Domain: testing-2025-11-17-ä.dev
2 corrections (HETZNER_V2)
#1: + CREATE xn--testing-2025-11-17--ywb.dev A 127.0.0.1 ttl=300
SUCCESS!
#2: + CREATE foo.xn--testing-2025-11-17--ywb.dev A 127.0.0.1 ttl=300
SUCCESS!
```

And the next listing will find the records with the FQDN
```
.Type="NS" input="@" -> .Name="@" .NameFQDN="xn--testing-2025-11-17--ywb.dev"
.Type="A" input="xn--testing-2025-11-17--ywb.dev" -> .Name="xn--testing-2025-11-17--ywb.dev" .NameFQDN="xn--testing-2025-11-17--ywb.dev.xn--testing-2025-11-17--ywb.dev"
.Type="A" input="foo.xn--testing-2025-11-17--ywb.dev" -> .Name="foo.xn--testing-2025-11-17--ywb.dev" .NameFQDN="foo.xn--testing-2025-11-17--ywb.dev.xn--testing-2025-11-17--ywb.dev"
``` 

Which we can confirm with a DNS query:
```
$  dig foo.xn--testing-2025-11-17--ywb.dev @hydrogen.ns.hetzner.com
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 11474

$ dig foo.xn--testing-2025-11-17--ywb.dev.xn--testing-2025-11-17--ywb.dev @hydrogen.ns.hetzner.com
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 8090
[...]

PR #3845, opened 2025-11-21
BUGFIX: IDN support is broken for domain names

tlimoncelli:
# Issue

Fixes https://github.com/StackExchange/dnscontrol/issues/3842

CC @das7pad

# Resolution

Convert domain.Name to IDN earlier in the pipeline.  Hack the --domains processing to convert everything to IDN.

* Domain names are now stored 3 ways: The original input from dnsconfig.js, canonical IDN format (`xn--...`), and Unicode format.  All are downcased.  Providers that haven't been updated will receive the IDN format instead of the original input format.  This might break some providers but only for users with unicode in their D("domain.tld"). PLEASE TEST YOUR PROVIDER.
* BIND filename formatting options have been added to access the new formats.

# Breaking changes

* BIND zonefiles may change. The default used the name input in the D() statement. It now defaults to the IDN name + "!tag" if there is a tag.
* Providers that are not IDN-aware may break (hopefully only if they weren't processing IDN already)

### DIFF
--- a/documentation/provider/bind.md
+++ b/documentation/provider/bind.md
@@ -22,7 +22,8 @@ Example:
 {
   "bind": {
     "TYPE": "BIND",
-    "directory": "myzones"
+    "directory": "myzones",
+    "filenameformat": "%U.zone"
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


commands/commands.go
```diff
@@ -308,14 +309,19 @@ func (args *FilterArgs) flags() []cli.Flag {
...
 		}
 		if strings.HasPrefix(item, "*") && strings.HasSuffix(domain, item[1:]) {
 			return true
 		}
 		filterDom, filterTag, isFilterTagged := strings.Cut(item, "!")
 		splitDom, domainTag, isDomainTagged := strings.Cut(domain, "!")
+
+		splitDom = downgradeIDNA(splitDom)
+		filterDom = downgradeIDNA(filterDom)
+		//fmt.Printf("###DEBUG: does %q == %q? %v\n", splitDom, filterDom, splitDom == filterDom)
```
das7pad:
```suggestion
```
This looks like a left-over debug print.

tlimoncelli:
> WDYT about printing the domain using unicode again?

I looked into the code. It's more complex than you'd think.  Domains may be tagged (`example.com!tag123`) and now we need to do a lot of extra parsing to work around the "!".  The full fix requires much more thought.

In fact, I'm not sure if I should merge this PR because even this change seems to have "surprises" like Exoscale, BIND filenames, etc.

This aspect of the code has very little testing, which makes it brittle to change.  The integration tests, for example, don't test IDN in domain names (just labels) because it would require us to register 2x the test domains.

> And perhaps use unicode for the filename in BIND as well to avoid the breaking change there?

On one hand, I don't like breaking changes.  On the other hand, I like this because:
* It is future-proofed. If the user changes the value in dnsconfig.js, the filename will always work
* It complies with IETF recommendations (I think 5890?) that says convert to IDN early and use the IDN strings internally... only displaying Unicode in UIs.

PR #3857, opened 2025-11-30
CLOUDFLARE: Add LOC support

### LINKED ISSUE #2798: CLOUDFLAREAPI: LOC supported by CF but not by integration

kevinji:
Fixes #2798.

I tested this locally and it seems to update the `LOC` record correctly.

### DIFF
--- a/models/t_loc.go
+++ b/models/t_loc.go
@@ -99,26 +99,20 @@ func (rc *RecordConfig) calculateLOCFields(d1 uint8, m1 uint8, s1 float32, ns st
 ) error {
 	// Crazy hairy shit happens here.
 	// We already got the useful "string" version earlier. ¯\_(ツ)_/¯ code golf...
-	const LOCEquator uint64 = 0x80000000       // 1 << 31 // RFC 1876, Section 2.
-	const LOCPrimeMeridian uint64 = 0x80000000 // 1 << 31 // RFC 1876, Section 2.
-	const LOCHours uint32 = 60 * 1000
-	const LOCDegrees = 60 * LOCHours
-	const LOCAltitudeBase int32 = 100000
-
-	lat := uint64((uint32(d1) * LOCDegrees) + (uint32(m1) * LOCHours) + uint32(s1*1000))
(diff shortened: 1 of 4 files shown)


models/t_loc.go
```diff
@@ -99,26 +99,20 @@ func (rc *RecordConfig) calculateLOCFields(d1 uint8, m1 uint8, s1 float32, ns st
...
-	const LOCEquator uint64 = 0x80000000       // 1 << 31 // RFC 1876, Section 2.
-	const LOCPrimeMeridian uint64 = 0x80000000 // 1 << 31 // RFC 1876, Section 2.
-	const LOCHours uint32 = 60 * 1000
-	const LOCDegrees = 60 * LOCHours
-	const LOCAltitudeBase int32 = 100000
-
-	lat := uint64((uint32(d1) * LOCDegrees) + (uint32(m1) * LOCHours) + uint32(s1*1000))
-	lon := uint64((uint32(d2) * LOCDegrees) + (uint32(m2) * LOCHours) + uint32(s2*1000))
+	lat := uint32(d1)*dns.LOC_DEGREES + uint32(m1)*dns.LOC_HOURS + uint32(s1*1000)
+	lon := uint32(d2)*dns.LOC_DEGREES + uint32(m2)*dns.LOC_HOURS + uint32(s2*1000)
```
kevinji, line 103: I removed the uint64 cast here because we recast back to uint32 immediately afterwards.

Issue #420, opened 2018-11-28
Provider request: Alibaba cloud 


Marcus-James-Adams:
More information at
https://www.alibabacloud.com/help/product/34269.htm?spm=a3c0i.228936.1218286.3.2054221eXUxW1G

tlimoncelli:
Thank you for requesting this provider!

I've tagged this issue as a provider-request.  It will (soon) be listed as a "requested provider" on the provider list web page: 
https://stackexchange.github.io/dnscontrol/provider-list

I will now close the issue.  I know that's a bit confusing, but it will remain on the "requested provider" list.

If someone would like to volunteer to implement this, please re-open this issue and add the tag `has-pr`.

We encourage you to try adding this provider yourself.  We've tried to
make the process as friendly as possible.  Many people have reported
that adding a provider was their first experience writing Go.  The
process is documented here:
https://stackexchange.github.io/dnscontrol/writing-providers
If you need assistance, please speak up in this issue and someone will get back to you ASAP.

PR #3878, opened 2025-12-02
ALIDNS: Implement ALIDNS Provider

bytemain:
https://github.com/StackExchange/dnscontrol/issues/420

Please create the GitHub label 'provider-ALIDNS'

### DIFF
--- a/commands/types/dnscontrol.d.ts
+++ b/commands/types/dnscontrol.d.ts
@@ -1773,6 +1773,13 @@ declare function IGNORE(labelSpec: string, typeSpec?: string, targetSpec?: strin
  *
  * ## Caveats
  *
+ * ### One per domain
+ *
+ * Only one `IGNORE_EXTERNAL_DNS()` should be used per domain. If you call it multiple
+ * times, the last prefix wins. If you have multiple external-dns instances with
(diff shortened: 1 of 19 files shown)


bytemain:
Yeah, there is an api can get the domain info: 
[https://www.alibabacloud.com/help/en/dns/api-alidns-2015-01-09-describedomaininfo](https://www.alibabacloud.com/help/en/dns/api-alidns-2015-01-09-describedomaininfo?scm=20140722.S_help%40%40%E6%96%87%E6%A1%A3%40%402357283._.ID_help%40%40%E6%96%87%E6%A1%A3%40%402357283-RL_DescribeDomain-LOC_doc%7EUND%7Eab-OR_ser-PAR1_6a0b3e9017647782742893955da48b-V_4-PAR3_r-RE_new5-P0_1-P1_0&spm=a2c63.p38356.help-search.i20#api-detail-40), So we can set the limit automatically.

But there is another problem, We can't get domain version info inside the AuditRecords function, because AuditRecords is called without provider context.
So I think we can run TTL validation at the provider level in GetZoneRecordsCorrections, What do you think?

We can implement the check before the request send.

PR #3884, opened 2025-12-03
TEST (DO NOT MERGE)

tlimoncelli:


bytemain:
The original issue has been resolved, and now alidns will also automatically set the minimum TTL, so this can be turned off.

PR #3889, opened 2025-12-04
DOCS: replace all `NewRegistrar('ThirdParty')`

SukkaW:
`NewRegistrar("ThirdParty");` has been replaced w/ `NewRegistrar("none");`

The PR updates the docs.


SukkaW:
@tlimoncelli Nowadays, if you do `NewRegistrar("ThirdParty");`, `ThirdParty` will be treated as a `creds.json` entry, and end up in the following error:

```
ERROR: creds.json is missing an entry called "ThirdParty". Suggestion: "ThirdParty": { "TYPE": "FILL_IN_PROVIDER_TYPE" }, (See https://docs.dnscontrol.org/commands/creds-json#missing)
```

It has to be `none`.

PR #3957, opened 2026-01-05
CLOUDNS: allow SRV record with null target

RobinDaugherty:
I confirmed today that ClouDNS does actually support SRV records with null (".") targets.


tlimoncelli:
Thanks for the confirmation!

CC @pragmaton

Issue #3956, opened 2026-01-04
OPENPGPKEY failing for Cloudflare


tlimoncelli:
CC:  @Georg who submitted the original OPENPGPKEY code.

# Issue

Cloudflare supports OPENPGPKEY but our test data isn't real enough:

https://github.com/StackExchange/dnscontrol/actions/runs/20697926237/job/59415967275

```
=== FAIL: integrationTest TestDNSProviders/***-cf.com/90:OPENPGPKEY:OPENPGPKEY_record (0.49s)
    helpers_integration_test.go:174: 
[...]

PR #3959, opened 2026-01-06
TEST: replace OPENPGPKEY data

tacerus:
Apparently the original test data was considered invalid by CLOUDFLARE. Replace it with data generated from a new key:

```
pub   ed25519 2026-01-06 [SC] [expires: 2105-12-18]
      345BAEC799CDC052162BDD578AEAED3064904410
uid           [ultimate] TESTING TESTING TESTING (DO NOT TRUST) <nobody@example.com>
```

The test for change operations additionally uses data generated from another key (an older one, for variety):

```
pub   rsa3072 2022-07-23 [SC]
      686EFBA4A120228EE928A4BDF2D8B1BAB5A28E6A
uid           [ultimate] TESTING TESTING TESTING (DO NOT TRUST) <nobody@example.com>
```

The documentation examples are updated too to keep the data aligned and easy to search for.


tacerus:
@tlimoncelli Could you test if this helps? I don't have API credentials for the provider.

cafferata:
> Ah! I see the problem. It's a bug in the Cloudflare provider.
> 
> I'm not going to try to fix it right now. However I will make a note about what the problem was. I may ping you when I get around to fixing that.

@tlimoncelli given this explanation, can GitHub issue https://github.com/StackExchange/dnscontrol/issues/3956 be closed, or would you prefer to keep it open until the Cloudflare provider bug is addressed?

PR #3960, opened 2026-01-06
DOCS: Expanded record acronyms and added RFC links.

grahammiln:
Also added inter-page links and minor style fixes.


tlimoncelli:
Please rebase.

PR #3961, opened 2026-01-06
CLOUDNS: add registrar support

RobinDaugherty:
ClouDNS uses the same API for DNS and registrar-related operations. Following the pattern I saw in other provider/registrar combos (autodns, cscglobal), I implemented the registry call in the `cloudnsProvider` object type, and when used as a registry it creates an additional instance of the Provider.

_Incidentally this causes rate-limiting functionality not to function optimally because each of the instances—the provider and registrar—have their own requestLimit object, causing the rate limit to be tracked and applied within each separately. I'm going to follow up with a PR that will improve the API rate-limit implementation for ClouDNS._

The ClouDNS nameserver API responds with different structures apparently depending on the number of name servers that the domain has. I observed three on the same domain:
1. When there are 2 domain names for non-ClouDNS name servers, the response is an array of name server strings `["abc.ns.cloudflare.net", "def.ns.cloudflare.net"]`
2. When there are 4 ClouDNS name servers, the response is an object with string keys containing the 4 entries `{"1":"pns1.cloudns.net","2":"pns2.cloudns.net","3":"pns3.cloudns.net","4":"pns4.cloudns.net"}`
3. When there are 5 name servers, the response is an object with string keys containing 8 entries included 3 blank values `{"1":"pns1.cloudns.net","2":"pns2.cloudns.net","3":"pns3.cloudns.net","4":"pns4.cloudns.net","5":"abc.ns.cloudflare.net","6":"","7":"","8":""}`

This covers all of those cases, plus the possible case where the array contains blank strings.

The new `setNameservers` API call needed to send multiple copies of the same key `nameservers[]` in the query value, which is supported by the http query values object, but was not possible with the existing call pattern because the `get` function was built to accept a string map instead of a query values object. So I refactored `get` into a wrapper function that converts the map into a query values object, which calls the new `getWithQuery` function with the query values object.

### DIFF
--- a/providers/cloudns/cloudnsProvider.go
+++ b/providers/cloudns/cloudnsProvider.go
@@ -12,7 +12,6 @@ import (
 	"github.com/StackExchange/dnscontrol/v4/pkg/diff"
 	"github.com/StackExchange/dnscontrol/v4/pkg/providers"
 	"github.com/miekg/dns/dnsutil"
-	"golang.org/x/time/rate"
 )
 
 /*
@@ -24,7 +23,7 @@ Info required in `creds.json`:
 
 func NewCloudns(m map[string]string) (*cloudnsProvider, error) {
 	c := &cloudnsProvider{}
-	c.requestLimit = rate.NewLimiter(10, 10)
+	c.requestLimit = NewAdaptiveLimiter(10, 10)
 
 	c.creds.id, c.creds.password, c.creds.subid = m["auth-id"], m["auth-password"], m["sub-auth-id"]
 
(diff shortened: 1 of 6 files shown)


tlimoncelli:
CC @pragmaton 

Wow! This is great!

By the way.... in the future I plan on merging things so that one initializer is used whether the providers is a registrar or a DNS server.

providers/cloudns/cloudnsProvider.go
```diff
@@ -239,6 +248,36 @@ func (c *cloudnsProvider) GetZoneRecordsCorrections(dc *models.DomainConfig, exi
...
+	// 2. Get desired nameservers from config
+	desired := models.NameserversToStrings(dc.Nameservers)
+	sort.Strings(desired)
+	desiredStr := strings.Join(desired, ",")
+
+	// 3. Compare and return correction if needed
+	if existingStr != desiredStr {
+		return []*models.Correction{
+			{
+				Msg: fmt.Sprintf("Update nameservers from '%s' to '%s'", existingStr, desiredStr),
```
tlimoncelli, line 269: Use %q instead of `%s`

PR #3962, opened 2026-01-06
CLOUDNS: pause when API fails due to rate limit

RobinDaugherty:
There was already a [Limiter](https://pkg.go.dev/golang.org/x/time/rate) in use here to keep the rate of requests below the apparent limit.

The ClouDNS API doesn't give any sort of proper API response when rate limit is reached. It's a 200 status code and an error message in the JSON body, and no headers that would help to track or back off for the right amount of time.

There was a comment in the implementation that mentions an undocumented 10-per-second limit, while the error message they give today says that the limit is 20 per second. I kept the settings on the Limiter the same since 10 per second should be plenty fast.

But it will now retry the request when the rate limit is reached. At the same time, it "steals" some reservations on `rate.Limiter` to quiet other concurrent ClouDNS API calls for about half a second. This seems to be plenty to fix my rate-limit issues. (I tested with 20 domains with ClouDNS as both registrar and DNS provider using the functionality in #3961.)

When rate limit is reached, it emits a warn-level message. This follows a pattern I see in `adguardhome` and `desec` providers but I don't love it—it's less important and less actionable than other warn-level messages in the project.

### DIFF
--- a/commands/types/dnscontrol.d.ts
+++ b/commands/types/dnscontrol.d.ts
@@ -3572,7 +3572,7 @@ declare function SOA(name: string, ns: string, mbox: string, refresh: number, re
  *
  * ## Advanced Technique: Define once, use many
  *
- * In some situations we define an SPF setting once and want to reuse
+ * In some situations we define an SPF setting once and want to re-use
  * it on many domains. Here's how to do this:
  *
  * ```javascript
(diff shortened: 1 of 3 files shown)


tlimoncelli:
CC @pragmaton

PR #3975, opened 2026-01-10
CLOUDNS: fix format string for nameserver diff

RobinDaugherty:
Fix a format string that used `%s` where `%q` is more appropriate, to escape the string fully.

This addresses the feedback in #3961 that I didn't get done in time to be merged.

### DIFF
--- a/providers/cloudns/cloudnsProvider.go
+++ b/providers/cloudns/cloudnsProvider.go
@@ -296,7 +296,7 @@ func (c *cloudnsProvider) GetRegistrarCorrections(dc *models.DomainConfig) ([]*m
 	if existingStr != desiredStr {
 		return []*models.Correction{
 			{
-				Msg: fmt.Sprintf("Update nameservers from '%q' to '%q'", existingStr, desiredStr),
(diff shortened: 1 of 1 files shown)


providers/cloudns/cloudnsProvider.go
```diff
@@ -296,7 +296,7 @@ func (c *cloudnsProvider) GetRegistrarCorrections(dc *models.DomainConfig) ([]*m
 	if existingStr != desiredStr {
 		return []*models.Correction{
 			{
-				Msg: fmt.Sprintf("Update nameservers from '%s' to '%s'", existingStr, desiredStr),
+				Msg: fmt.Sprintf("Update nameservers from '%q' to '%q'", existingStr, desiredStr),
```
tlimoncelli: No need for the single quotes.  %q outputs a string with double quotes.  For example `fmt.Sprintf("my %q phone", "blue")` would output `my "blue" phone` 

```suggestion
				Msg: fmt.Sprintf("Update nameservers from %q to %q", existingStr, desiredStr),
```

PR #3979, opened 2026-01-12
DIGITALOCEAN: Disable SOA support

### LINKED ISSUE #3965: DIGITALOCEAN: SOA support?
# Issue

https://github.com/StackExchange/dnscontrol/pull/3963 disabled SOA support in DIGITALOCEAN because it was failing tests.

# Resolution

Investigate if DO stopped supporting SOA records.

chicks-net:
## Summary

Fixes #3965

Investigation revealed that DigitalOcean's SOA support is too limited to be useful in DNSControl:

- While the API documentation mentions SOA records exist, only the TTL can be modified
- The SOA record cannot be deleted
- This makes it impractical to manage via DNSControl

Changes made:
- Disabled `CanUseSOA` capability in the DigitalOcean provider (with comment "despite docs")
- Updated provider documentation to remove SOA from supported record types
- Added clarification explaining why SOA is not supported
- Updated the provider feature matrix table to reflect SOA is not supported

## Test plan

- [x] Reviewed the DigitalOcean API documentation
- [x] Confirmed the limitation matches the test failures from #3963
- [x] Updated all relevant documentation files
- [x] Provider capability correctly set to `Cannot()`

### DIFF
--- a/providers/digitalocean/digitaloceanProvider.go
+++ b/providers/digitalocean/digitaloceanProvider.go
@@ -86,7 +86,7 @@ var features = providers.DocumentationNotes{
 	providers.CanUseLOC:              providers.Cannot(),
 	providers.CanUseNAPTR:            providers.Cannot(), // per docs
 	providers.CanUsePTR:              providers.Cannot(), // per docs
-	providers.CanUseSOA:              providers.Cannot(), // despite docs
+	providers.CanUseSOA:              providers.Cannot("Technically SOA is supported but in reality the API only permits updates to the TTL. That is insufficient for DNSControl to claim 'support'"),
 	providers.CanUseSRV:              providers.Can(),
 	providers.CanUseSSHFP:            providers.Cannot(), // per docs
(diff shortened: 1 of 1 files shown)


providers/digitalocean/digitaloceanProvider.go
```diff
@@ -86,7 +86,7 @@ var features = providers.DocumentationNotes{
 	providers.CanUseLOC:              providers.Cannot(),
 	providers.CanUseNAPTR:            providers.Cannot(), // per docs
 	providers.CanUsePTR:              providers.Cannot(), // per docs
-	providers.CanUseSOA:              providers.Can(),
+	providers.CanUseSOA:              providers.Cannot(), // despite docs
```
tlimoncelli:
```suggestion
	providers.CanUseSOA:              providers.Cannot("Technically SOA is supported but in reality the API only permits updates to the TTL. That is insufficient for DNSControl to claim 'support'"),
```

PR #3988, opened 2026-01-19
DOCS: Fix incorrect hetzner links

willpower232:
I went to look at the Hetzner V2 documentation and found they were based on an external link which wasn't working. I think this is the fix but I might be wrong.

Also I don't think any of the external links are working on the providers index page https://docs.dnscontrol.org/provider/index 

image

I wonder if these are all dead links which create external links when the site is compiled?


tlimoncelli:
@cafferata can you please advise?

cafferata:
> @cafferata can you please advise?

Sure! I see you (@tlimoncelli) reverted theses changes in https://github.com/StackExchange/dnscontrol/pull/3990/changes/7db205524810e7df2e23a225e62010ca354f9331 🤓

willpower232:
Oh I forget thats a generated file so this was a relatively pointless PR :sweat_smile: 

Do the files have to contain underscores to match the name of the provider?

PR #4015, opened 2026-01-27
BUGFIX: avoid impossible comparison (staticcheck/SA4003)

grahammiln:
Use of unsigned integer conflicts with assumptions of `stringer` generated code.


pkg/providers/capabilities.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 // Capability is a bitmasked set of "features" that a provider supports. Only use constants from this package.
-type Capability uint32
+type Capability int32
```
tlimoncelli:
```suggestion
type Capability int
```

PR #4025, opened 2026-01-28
CLOUDFLAREAPI: Add support for record comments and tags

### LINKED ISSUE #1862: CF comments and tags
So cloudflare just announced a new set of features around tagging and commenting on DNS records.
https://blog.cloudflare.com/dns-record-comments/

Any thoughts on if this will be exposed via dnscontrol? We currently use dnscontrol for our workflow and i find myself adding comments in the config file that powers the thousands of records we have across our domains. Would be rather empowering for these comments and such to make their way through to cloudflare.

Would just like a ruling on if it can be an expected feature. I understand if not because it's special to them but figured it wouldn't be the first thing you've had special for them (ie proxying).

allixsenos:
Introduced the following Cloudflare-specific modifiers:

- CF_MANAGE_COMMENTS - Domain modifier to opt into comment syncing
- CF_MANAGE_TAGS - Domain modifier to opt into tag syncing
- CF_COMMENT(comment) - Set a comment on a record (works on all plans)
- CF_TAGS(tag1, tag2, ...) - Set tags on a record (requires paid plan)

Features:
- Comments and tags are read from API and displayed in get-zones output
- Validation ensures CF_COMMENT/CF_TAGS require management flags enabled
- Tags with reserved 'cf-' prefix are rejected
- get-zones --format=js/djs auto-adds management flags when needed
- Integration tests for comments (always run) and tags (-cftags flag)

Fixes #1862

### DIFF
--- a/commands/getZones.go
+++ b/commands/getZones.go
@@ -285,10 +285,10 @@ func GetZone(args GetZoneArgs) error {
 			hasComments := false
 			hasTags := false
 			for _, rec := range recs {
-				if comment, ok := rec.Metadata["cloudflare_comment"]; ok && comment != "" {
+				if rec.Metadata["cloudflare_comment"] != "" {
 					hasComments = true
 				}
-				if tags, ok := rec.Metadata["cloudflare_tags"]; ok && tags != "" {
+				if rec.Metadata["cloudflare_tags"] != "" {
 					hasTags = true
 				}
 				if hasComments && hasTags {
@@ -344,10 +344,10 @@ func GetZone(args GetZoneArgs) error {
 						cfmeta += ",cloudflare_cname_flatten=on"
 					}
 				}
-				if comment, ok := rec.Metadata["cloudflare_comment"]; ok && comment != "" {
+				if comment := rec.Metadata["cloudflare_comment"]; comment != "" {
 					cfmeta += ",cloudflare_comment=" + comment
 				}
-				if tags, ok := rec.Metadata["cloudflare_tags"]; ok && tags != "" {
+				if tags := rec.Metadata["cloudflare_tags"]; tags != "" {
 					cfmeta += ",cloudflare_tags=" + tags
 				}
 				if cfmeta != "" {
@@ -404,12 +404,12 @@ func formatDsl(rec *models.RecordConfig, defaultTTL uint32) string {
 	}
 
 	cfcomment := ""
-	if comment, ok := rec.Metadata["cloudflare_comment"]; ok && comment != "" {
+	if comment := rec.Metadata["cloudflare_comment"]; comment != "" {
 		cfcomment = fmt.Sprintf(", CF_COMMENT(%s)", jsonQuoted(comment))
 	}
 
 	cftags := ""
-	if tags, ok := rec.Metadata["cloudflare_tags"]; ok && tags != "" {
+	if tags := rec.Metadata["cloudflare_tags"]; tags != "" {
 		// Convert comma-separated tags to CF_TAGS("tag1", "tag2", ...)
 		tagList := strings.Split(tags, ",")
 		quotedTags := make([]string, len(tagList))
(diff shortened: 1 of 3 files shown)


integrationTest/integration_test.go
```diff
@@ -1248,6 +1248,49 @@ func makeTests() []*TestGroup {
...
+			only("CLOUDFLAREAPI"),
+			domainMeta(map[string]string{"cloudflare_manage_comments": "true"}),
+			CfCommentCreate(), CfCommentChange(),
+		),
+
+		testgroup("CF_COMMENT remove",
+			only("CLOUDFLAREAPI"),
+			domainMeta(map[string]string{"cloudflare_manage_comments": "true"}),
+			CfCommentCreate(), CfCommentRemove(),
+		),
```
tlimoncelli, line 1259: These can be combined. No need to clear the zone between tests. (for example, `testgroup("MX"` does it all in one group)

```suggestion
		testgroup("CF_COMMENT create",
			only("CLOUDFLAREAPI"),
			domainMeta(map[string]string{"cloudflare_manage_comments": "true"}),
			CfCommentCreate(),
			CfCommentChange(),
			CfCommentRemove(),
		),
```

integrationTest/integration_test.go
```diff
@@ -1248,6 +1248,49 @@ func makeTests() []*TestGroup {
...
+			CfTagsCreate(), CfTagsChange(),
+		),
+
+		testgroup("CF_TAGS remove",
+			only("CLOUDFLAREAPI"),
+			alltrue(*enableCFTags),
+			domainMeta(map[string]string{"cloudflare_manage_tags": "true"}),
+			CfTagsCreate(), CfTagsRemove(),
+		),
+
```
tlimoncelli, line 1271: Consolidate these:

```suggestion
		testgroup("CF_TAGS create",
			only("CLOUDFLAREAPI"),
			alltrue(*enableCFTags),
			domainMeta(map[string]string{"cloudflare_manage_tags": "true"}),
			CfTagsCreate(),
			CfTagsChange(),
			CfTagsRemove(),
		),

```

integrationTest/helpers_integration_test.go
```diff
@@ -51,6 +51,32 @@ func CfFlattenOn() *TestCase {
 	return tc("flattenon", cfFlattenCNAME("cflatten", "example.com.", "on"))
 }
 
+// Helper constants/funcs for the CLOUDFLARE comments testing:
+
+// A-record with comment
+func CfCommentCreate() *TestCase {
```
tlimoncelli: Now that CfCommentCreate() and friends are only used once, please inline them. That has the benefit of making the test data more visible.

commands/getZones.go
```diff
@@ -280,6 +280,28 @@ func GetZone(args GetZoneArgs) error {
...
 			}
+
+			// Check if any records have comments or tags, and add management flags if so
+			hasComments := false
+			hasTags := false
+			for _, rec := range recs {
+				if comment, ok := rec.Metadata["cloudflare_comment"]; ok && comment != "" {
+					hasComments = true
+				}
+				if tags, ok := rec.Metadata["cloudflare_tags"]; ok && tags != "" {
```
tlimoncelli: rec.Metadata["cloudflare_tags"] returns "" if there is no "cloudflare_tags" key.  It seems like the "ok" checking is overkill.

This would make CF_TAGS("") and not having any tags be equivalent. (which is a good thing. I think?)

documentation/provider/cloudflareapi.md
cafferata, line 1: I have reviewed the [documentation changes](https://docs.dnscontrol.org/~/revisions/cyeS6bhK0FAbBxJ1txuB/provider/cloudflareapi) and they look very good and complete. Thank you @allixsenos for the clear improvements and detailed explanation around using comments/tags.

tlimoncelli:
I can't wait to use this feature!!!

PR #4062, opened 2026-02-10
BUGFIX: Fix SPF parsing of qualified mechanisms like +mx

### LINKED ISSUE #4042: Regression in `SPF_BUILDER` using `+mx`
Is it expected for validation to fail in [`SPF_BUILDER`](https://docs.dnscontrol.org/language-reference/domain-modifiers/spf_builder) [.js](https://github.com/StackExchange/dnscontrol/blob/e5e79346785c2aae0bd5956c4363a5f7a61d1712/pkg/js/helpers.js#L1696) when using [`+mx`](https://www.rfc-editor.org/rfc/rfc7208#section-4.7) and `overflow`?

``` js
var DNS_BIND = NewDnsProvider("bind");

D("example.com",
    REG_NONE,
    DnsProvider(DNS_BIND),

    SPF_BUILDER({
        overflow: "_spf%d", // REMOVE TO FIX
        parts: [
            "v=spf1",
            "+mx", // accept from this domain's mx records
            "-all", // reject all other mail servers
        ],
    }),
)
```

dnscontrol v4.32.0 returns the error:

```
2026/02/02 11:22:09 1 Validation errors:
2026/02/02 11:22:09 ERROR: unsupported SPF part mx
```

If the `overflow` field is removed, the validation passes without error.

This configuration worked previously.

cafferata:
Fixes a regression where `SPF_BUILDER` with `overflow` rejected valid RFC 7208 qualified mechanisms like `+mx`, `~mx`, `?a`, etc. with the error `unsupported SPF part mx`. The root cause was that `lcpart` (the lowercase comparison variable) was created before stripping the qualifier prefix (`+`,`-`,`~`,`?`), so `HasPrefix` checks against `"mx"` or `"a"` failed when the string was still `"+mx"` or `"+a"`. Qualifiers on mechanisms are valid per [RFC 7208 Section 4.7](https://www.rfc-editor.org/rfc/rfc7208#section-4.7). A regression test covering all qualifier/mechanism combinations has been added.

Fixes #4042

### DIFF
--- a/pkg/spflib/parse.go
+++ b/pkg/spflib/parse.go
@@ -52,11 +52,11 @@ func Parse(text string, dnsres Resolver) (*SPFRecord, error) {
 		if part == "" {
 			continue
 		}
-		lcpart := strings.ToLower(part) // We have seen "Ip4" instead of "ip4".  Let's be gracious and allow it.
 		p := &SPFPart{Text: part}
 		if qualifiers[part[0]] {
 			part = part[1:]
 		}
+		lcpart := strings.ToLower(part) // We have seen "Ip4" instead of "ip4".  Let's be gracious and allow it.
 		rec.Parts = append(rec.Parts, p)
 		if part == "all" {
 			// all. nothing else matters.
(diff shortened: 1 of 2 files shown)


cafferata:
Ping @grahammiln, @tlimoncelli.

PR #4077, opened 2026-02-12
CNR: add support for DNAME, DHCID, and SVCB record types

AsifNawaz-cnic:


tlimoncelli (approved):
Thanks!

P.S. It's great to have a vendor maintain their own provider. It is rare and very appreciated!

PR #4139, opened 2026-03-08
BUGFIX: Make `OPENPGPKEY` records functional.

gucci-on-fleek:
I'm unable to get DNSControl to create an `OPENPGPKEY` record that `gpg` or [`openpgpkey`](https://github.com/letoams/hash-slinger/) will accept. From what I can tell, it looks like DNSControl always corrupts or misinterprets the input, meaning that there is no possible input that will produce a valid result. But if I build DNSControl with this PR applied, both `gpg` and `openpgpkey` accept the `OPENPGPKEY` record without any complaints.

I'm probably just misunderstanding something though, since I wouldn't expect for the current code to not work at all. 

I've also expanded the `OPENPGPKEY` documentation, and fixed a small bug with the `--debug` argument. Full commit messages:

---

### BUGFIX: Make `OPENPGPKEY` records functional.

Previously, #3718 introduced support for `OPENPGPKEY` records. However, that PR assumed that the `OPENPGPKEY` key data was hex-encoded. [RFC 7929 §2.3](https://datatracker.ietf.org/doc/html/rfc7929#section-2.3) states that when in "Presentation Format", `OPENPGPKEY` record data must be base64-encoded instead. Since most hex-encoded data can be parsed as base64, DNS providers would not reject a hex-encoded key, but they would instead parse it as base64, leading to a corrupt result.

Additionally, the previous PR also unconditionally lower-cased the key data, making it impossible to use base64-encoded keys. Once again, lowercased base64-encoded data can usually be parsed as base64, but the result will be corrupted.

This current PR fixes both of these issues by instead treating the key as base64-encoded data, and by no longer lower-casing the key data. To maintain backwards compatibility with existing hex-encoded keys, the code attempts to detect keys that appear to be hex-encoded, and decodes them as hex before re-encoding them as base64.

---

### BUGFIX: Make `--debug` not print the version.

Currently, using the `--debug` command-line argument just prints the DNSControl version and exits. This is because the long option `--debug` is mapped to the short option `-v`, but `-v` is used for `--version`, so the parser gets confused. This commit remaps the long option `--version` to the short option `-V`, which prevents the parser from getting confused and makes `--debug` work again.

### DIFF
--- a/documentation/provider/cloudflareapi.md
+++ b/documentation/provider/cloudflareapi.md
@@ -4,6 +4,7 @@ This is the provider for [Cloudflare](https://www.cloudflare.com/).
 
 * SPF records are silently converted to RecordType `TXT` as Cloudflare API fails otherwise. See [StackExchange/dnscontrol#446](https://github.com/StackExchange/dnscontrol/issues/446).
 * This provider currently fails if there are more than 1000 corrections on one domain. This only affects "push". This usually when moving a domain with many records to Cloudflare.  Try commenting out most records, then uncomment groups of 999. Typical updates are less than 1000 corrections and will not trigger this bug. See [StackExchange/dnscontrol#1440](https://github.com/StackExchange/dnscontrol/issues/1440).
+* DNS records that Cloudflare injects and maintains are ignored. That includes SOA records, NS records at the domain's apex, and the MX/DKIM records created as part of Cloudflare mail routing.
 
 ## Configuration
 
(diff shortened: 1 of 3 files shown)


tlimoncelli (approved):

gucci-on-fleek:
@tlimoncelli 

(Sorry for the delay, I was travelling last week)

> @gucci-on-fleek I was trying to make a PR based on your PR. I did the usual "git push" which gives me a URL to click to make the PR, but it actually accepted the push. oops! i guess the permissions were wide open. Sorry.

Fine by me, thanks for pushing the fixes! That worked because I had the “Allow edits and access to secrets by maintainers” box checked (which I think is the default):

image

> I was able to make the change to `parse_tests/056-openpgpkey.js` and `parse_tests/056-openpgpkey.json` to test both base64 and hex. I hope that's ok. Please verify that `parse_tests/056-openpgpkey.json` is what you'd expect.

Yes, that looks like the correct output to me. Thanks again!

PR #4172, opened 2026-03-17
CICD: only run GHA release_draft on the main repo

chicks-net:
## Context

When I sync'd the dnscontrol repo into my forked repo it reran these jobs.  The jobs all fail since the credentials are not available.

This also ended up reformatting the yaml.  Let me know if that is an issue.

## Done

- 🛶 only run GHA release_draft on the main repo

## Meta

(Automated in `.just/gh-process.just`.)

### DIFF
--- a/.github/workflows/release_draft.yml
+++ b/.github/workflows/release_draft.yml
@@ -1,70 +1,71 @@
-name: "RELEASE: Make release candidate"
+name: 'RELEASE: Make release candidate'
 
 on:
-  push:
-    tags:
-      - v[0-9]+.[0-9]+.[0-9]+
-      - v[0-9]+.[0-9]+.[0-9]+-*
+    push:
+        tags:
+            - v[0-9]+.[0-9]+.[0-9]+
+            - v[0-9]+.[0-9]+.[0-9]+-*
 
 jobs:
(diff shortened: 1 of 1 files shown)


cafferata (changes requested):
Thanks for the fix! The `if: github.repository == 'StackExchange/dnscontrol'` guard makes sense.

Could you resubmit without the indentation/quote reformatting? The rest of the workflow files in this repo use 2-space indentation and double quotes. Keeping just the one-line `if:` addition would make the diff much easier to review and stay consistent with the other files.

chicks-net:
> Thanks for the fix! The `if: github.repository == 'StackExchange/dnscontrol'` guard makes sense.
> 
> Could you resubmit without the indentation/quote reformatting? The rest of the workflow files in this repo use 2-space indentation and double quotes. Keeping just the one-line `if:` addition would make the diff much easier to review and stay consistent with the other files.

Totally makes sense.  Will do.

chicks-net:
Given https://github.com/StackExchange/dnscontrol/discussions/4174 this is going to need to be redone once we move the repo so I'm closing this for now and I'll make a new PR for the new home.

PR #4176, opened 2026-03-23
DNSCALE: allow third-party NS records at zone apex

dnscale-ops:
## Summary

- Stop filtering all apex NS records; only skip DNScale's own nameservers
  (*.dnscale.eu, *.dnscale.com) which are system-managed
- Remove `NsAtApex` audit rejection so users can declare third-party NS records
- Document apex NS behavior and multi-provider DNS setup with `NAMESERVER()` example

## Motivation

DNScale now supports multi-provider DNS setups where users add third-party
nameserver records at the zone apex. The provider was previously blocking all
apex NS records, which prevented this use case.

## Changes

| File | Change |
|------|--------|
| `providers/dnscale/dnscaleProvider.go` | Filter only DNScale's own NS records at apex instead of all NS records; add multi-provider documentation to `GetNameservers` |
| `providers/dnscale/auditrecords.go` | Remove `NsAtApex` rejection |
| `documentation/provider/dnscale.md` | Add "Nameservers and apex NS records" section with multi-provider example |

### DIFF
--- a/documentation/provider/dnscale.md
+++ b/documentation/provider/dnscale.md
@@ -64,6 +64,33 @@ DNScale supports the following record types:
 - TLSA
 - TXT
 
+## Nameservers and apex NS records
+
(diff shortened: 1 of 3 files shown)


TomOnTime (approved):

dnscale-ops:
Thank you @TomOnTime!

PR #4180, opened 2026-03-26
RWTH: Add Support for Domains with DNSSEC Disabled and Migrate to diff2

psuet:
Right now the RWTH Provider only supports domains with DNSSEC enabled, since the return values of the zone-lookup call are different for the different DNSSEC states.

This PR fixes this problem.

Additionally the provider now uses the diff2 instead of the old diff1.

@MisterErwin :)

### DIFF
--- a/providers/rwth/api.go
+++ b/providers/rwth/api.go
@@ -34,20 +34,50 @@ type RecordReply struct {
 	Editable  bool      `json:"editable"`
 }
 
+type dnssecKey struct {
+	CreatedAt time.Time `json:"created_at"`
+}
+
+// dnssecStatus supports both API variants:
+// - dnssec: false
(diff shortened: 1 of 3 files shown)


MisterErwin (approved):
looks good to me (and was tested against the live system)

TomOnTime:
FYI: Things are stalled right now. I hope I can get back to DNSControl next week.

TomOnTime:
Please rebase

TomOnTime:
Greetings!  Thank you for your patience while we migrated to the new GitHub org.

Now that the migration is complete, please rebase.  Thank you.

TomOnTime (approved):
Thank you!  I'm sure users will appreciate the reliability gains from using Diff2!

PR #4188, opened 2026-04-05
CLOUDNS: map CLOUD_WR to CLOUDNS_WR so redirect records are not recreated every push

### LINKED ISSUE #3972: CLOUDNS_WR (redirect) recreated on every push
**Describe the bug**
When using the CLOUDNS_WR record, the record is being deleted and recreated on every push.
```
#1: - DELETE WR example.org https://example2.org ttl=3600, ClouDNS ID: 1234
#2: + CREATE CLOUDNS_WR example.org https://example2.org ttl=3600
```

I tried to create a patch:

```diff
diff --git a/providers/cloudns/cloudnsProvider.go b/providers/cloudns/cloudnsProvider.go
index d5f85fdc48..0721a26a0a 100644
--- a/providers/cloudns/cloudnsProvider.go
+++ b/providers/cloudns/cloudnsProvider.go
@@ -426,8 +426,8 @@
                rc.DsDigestType = uint8(dsDigestType)
                rc.DsDigest = r.Target
                err = rc.SetTarget(r.Target)
-       case "CLOUD_WR":
-               rc.Type = "WR"
+       case "WR":
+               rc.Type = "CLOUDNS_WR"
                err = rc.SetTarget(r.Target)
        case "LOC":
                loc := fmt.Sprintf("%s %s %s %s %s %s %s %s %s %s %s %s",
```

Which kinda fixes the issue, but instead it now only modifies the record all the time:
```
#1: ± MODIFY CLOUDNS_WR example.org: (https://example2.org ttl=3600) -> (https://example2.org ttl=3600), (location=DEFAULT) -> (location=DEFAULT), ClouDNS ID: 1234:
```

I'm not sure how to proceed with the patch, so I hope to get some help here!

**To Reproduce**
Explained above here 

**Expected behavior**
No change when pushing the records

**DNS Provider**
 - ClouDNS

raman1236:
## Summary

Fixes #3972

CLOUDNS_WR (web redirect) records are destroyed and recreated on every `dnscontrol push` even when nothing has changed.

## Root Cause

When records are **fetched** from the ClouDNS API, the `toRc` function converts the API type `CLOUD_WR` to `WR`. However, dnscontrol registers the custom record type as `CLOUDNS_WR` (line 85). This means:

- Fetched records have type `WR`
- Desired records (from user config) have type `CLOUDNS_WR`

The diff algorithm sees them as different record types and generates a DELETE + CREATE on every push.

## Fix

1. In `toRc()`: Map `CLOUD_WR` → `CLOUDNS_WR` (instead of → `WR`) so fetched records match the registered custom type
2. In the request builder: Remove `WR` from the "nothing special" case since the type is now always `CLOUDNS_WR` (which was already handled in the case below)

## Testing

- Added `TestToRcConvertsCloudWRToCloudnsWR` unit test that verifies the type mapping
- Build passes cleanly

### DIFF
--- a/providers/cloudns/cloudnsProvider.go
+++ b/providers/cloudns/cloudnsProvider.go
@@ -406,7 +406,7 @@ func toRc(domain string, r *domainRecord) (*models.RecordConfig, error) {
 		rc.DsDigest = r.Target
 		err = rc.SetTarget(r.Target)
 	case "CLOUD_WR":
-		rc.Type = "WR"
+		rc.Type = "CLOUDNS_WR"
 		err = rc.SetTarget(r.Target)
 	case "LOC":
 		loc := fmt.Sprintf("%s %s %s %s %s %s %s %s %s %s %s %s",
@@ -459,7 +459,7 @@ func toReq(rc *models.RecordConfig) (requestParams, error) {
 	}
 
 	switch rc.Type { // #rtype_variations
-	case "A", "AAAA", "NS", "PTR", "TXT", "SOA", "ALIAS", "CNAME", "WR", "DNAME":
+	case "A", "AAAA", "NS", "PTR", "TXT", "SOA", "ALIAS", "CNAME", "DNAME":
 		// Nothing special.
 	case "CLOUDNS_WR":
 		req["record-type"] = "WR"
--- a/providers/cloudns/cloudnsProvider_test.go
+++ b/providers/cloudns/cloudnsProvider_test.go
@@ -1,0 +1,27 @@
+package cloudns
+
+import (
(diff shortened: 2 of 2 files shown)


raman1236:
Friendly ping — this PR fixes the `CLOUD_WR` → `CLOUDNS_WR` record type mapping so redirect records aren't recreated on every push. Would appreciate a review. Thanks!

TomOnTime:
Thank you for your patience while the repo was moving to a new GitHub org.

Thanks for fixing this!  I'm sure a lot of people found that bug to be very annoying.

PR #4189, opened 2026-04-05
CLOUDFLAREAPI: allow CNAME + other records when Cloudflare proxy is enabled

### LINKED ISSUE #4181: CNAME + other record validation should account for Cloudflare proxied (flattened) CNAMEs
## Problem

When using the Cloudflare provider with `CF_PROXY_ON`, dnscontrol rejects configurations that have a CNAME and another record type (e.g. MX) on the same name:

```
cannot have CNAME and MX record with same name: mailscanner.example.com
```

## Why this is incorrect for proxied Cloudflare records

When a CNAME is proxied (`CF_PROXY_ON`) in Cloudflare, the CNAME is **flattened** — Cloudflare resolves it internally and returns A/AAAA records to external queries. From a DNS protocol perspective, no CNAME record is ever visible to resolvers. This means the RFC 1034 §3.6.2 restriction (CNAME must be the only record at a name) does not apply, because the CNAME is never actually served.

Cloudflare itself allows this configuration and it works correctly. The validation in dnscontrol (`checkCNAMEs` in `pkg/normalize/validate.go`) does not account for this provider-specific behavior.

## Example

```js
CNAME("mailscanner", "d2mcmwo95c48hp.cloudfront.net.", CF_PROXY_ON),
// ... elsewhere ...
MX("mailscanner", 10, "inbound-smtp.eu-central-1.amazonaws.com."),
```

Both records exist and function correctly in Cloudflare, but `dnscontrol check` rejects this configuration.

## Expected behavior

The CNAME + other record type validation should be skipped (or downgraded to a warning) when the CNAME is proxied through Cloudflare (`CF_PROXY_ON`), since the CNAME is flattened and never served as an actual CNAME to resolvers.

## Current workaround

Use `IGNORE` for one of the conflicting records, which means dnscontrol can no longer manage it.

raman1236:
### Description

Fixes the CNAME + other record type validation to allow coexistence when the CNAME is proxied through Cloudflare (`CF_PROXY_ON` or `CF_PROXY_FULL`).

### Problem

When using the Cloudflare provider with `CF_PROXY_ON`, dnscontrol rejects configurations that have a CNAME and another record type (e.g., MX) on the same name:

```
cannot have CNAME and MX record with same name: mailscanner.example.com
```

However, when a CNAME is proxied, Cloudflare flattens it internally and returns A/AAAA records to resolvers. The CNAME is never served to the public, so the RFC 1034 §3.6.2 restriction does not apply. Cloudflare itself allows this configuration and it works correctly.

### Root Cause

The `checkCNAMEs` function in `pkg/normalize/validate.go` unconditionally rejects any record type (except AKAMAICDN) that shares a label with a CNAME, without checking whether the CNAME uses Cloudflare proxy flattening.

### Fix

When building the CNAME label map, also track which CNAMEs have `cloudflare_proxy` metadata set to `"on"` or `"full"`. Skip the conflict error for those labels, following the same pattern as the existing AKAMAICDN exception.

### Tests

Added `TestCNAMECloudflareProxied` that verifies:
- Proxied CNAME + MX on the same label produces no error
- Non-proxied CNAME + MX on the same label still produces an error

Fixes #4181

### DIFF
--- a/pkg/normalize/validate.go
+++ b/pkg/normalize/validate.go
@@ -636,18 +636,28 @@ func checkAutoDNSSEC(dc *models.DomainConfig) (errs []error) {
 
 func checkCNAMEs(dc *models.DomainConfig) (errs []error) {
 	cnames := map[string]bool{}
+	proxiedCnames := map[string]bool{}
 	for _, r := range dc.Records {
 		if r.Type == "CNAME" {
 			if cnames[r.GetLabel()] {
 				errs = append(errs, fmt.Errorf("%s: cannot have multiple CNAMEs with same name: %s", r.FilePos, r.GetLabelFQDN()))
 			}
 			cnames[r.GetLabel()] = true
+			if p, ok := r.Metadata["cloudflare_proxy"]; ok && (p == "on" || p == "full") {
+				proxiedCnames[r.GetLabel()] = true
+			}
 		}
 	}
 	for _, r := range dc.Records {
 		if cnames[r.GetLabel()] && r.Type != "CNAME" {
 			// Allow AKAMAICDN and CNAME to have same name
 			if r.Type == "AKAMAICDN" {
+				continue
+			}
+			// Cloudflare proxied (flattened) CNAMEs are resolved internally
+			// and never served as actual CNAME records, so the RFC 1034 §3.6.2
+			// restriction does not apply.
+			if proxiedCnames[r.GetLabel()] {
 				continue
 			}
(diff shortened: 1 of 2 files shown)


raman1236:
Friendly ping — this PR allows CNAME + other records to coexist when Cloudflare proxy is enabled (since proxied CNAMEs are flattened at the edge). Would appreciate a review. Thanks!

TomOnTime:
Thank you for sending this!

TomOnTime:
Looks good! Thanks for the fix!

PR #4190, opened 2026-04-05
BUGFIX: preview should display long TXT records as one string

### LINKED ISSUE #2834: Display of long TXT record changes is confusing
**Describe the bug**
When a TXT record is changed, the preview shows the record as `"255 chars" "255 chars" "remaining chars"` instead of `"all the chars"`. This is visually confusing to users. It also violates [Opinion #8](https://docs.dnscontrol.org/developer-info/opinions#opinion-8-txt-records-are-one-long-string).

**To Reproduce**
Change a TXT record that is longer than 255-octets.

**Expected behavior**
The "preview" display should show one long string, not the 255-octet chunks.

**DNS Provider**
 - all

**Additional context**
n/a

raman1236:
### Description

Fixes the confusing display of long TXT records in the preview/diff output. Previously, TXT records longer than 255 octets were shown as chunked segments:

```
"v=DKIM1; k=rsa; p=MIIB...long..." "remaining..."
```

Now they are displayed as a single string per [Opinion #8](https://docs.dnscontrol.org/developer-info/opinions#opinion-8-txt-records-are-one-long-string):

```
"v=DKIM1; k=rsa; p=MIIB...long...remaining..."
```

### Root Cause

`ToComparableNoTTL()` (used by the diff engine for both comparison and display) called `txtutil.EncodeQuoted()`, which splits TXT values into 255-octet chunks before quoting.

### Fix

- Added `txtutil.EncodeSingle()` that quotes the TXT value as a single string without chunking
- Changed `ToComparableNoTTL()` to use `EncodeSingle()` instead of `EncodeQuoted()`
- Zone file output (`prettyzone`) continues to use `EncodeQuoted()` with chunking, as that format is required for DNS wire compatibility

### Tests

All existing txtutil, diff2, and models tests pass.

Fixes #2834

### DIFF
--- a/models/record.go
+++ b/models/record.go
@@ -344,7 +344,7 @@ func (rc *RecordConfig) ToComparableNoTTL() string {
 		// SoaSerial is not included because it isn't used in comparisons.
 	case "TXT":
 		// fmt.Fprintf(os.Stdout, "DEBUG: ToComNoTTL raw txts=%s q=%q\n", rc.target, rc.target)
-		r := txtutil.EncodeQuoted(rc.target)
+		r := txtutil.EncodeSingle(rc.target)
 		// fmt.Fprintf(os.Stdout, "DEBUG: ToComNoTTL cmp txts=%s q=%q\n", r, r)
 		return r
 	case "LUA":
--- a/pkg/txtutil/txtcode.go
+++ b/pkg/txtutil/txtcode.go
@@ -36,6 +36,13 @@ func EncodeQuoted(t string) string {
 	return txtEncode(ToChunks(t))
 }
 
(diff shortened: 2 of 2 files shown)


raman1236:
Friendly ping — this PR fixes the preview display for long TXT records to show them as a single string rather than split into 255-byte chunks. Would appreciate a review. Thanks!

KaiSchwarz-cnic:
Hey Raman!

We had a public holiday on Friday and today I am on sick leave. Hopefully back tomorrow.

Kai

TomOnTime:
Hello!  Thank you for your patience while we moved DNSControl to our new GitHub org.

Thank you for the PR.  The output is definitely more readable this way.  Great idea!

PR #4194, opened 2026-04-08
BUGFIX: LOC record latitude/longitude overflow on 32 bit archs

al20ov:
The 045-loc.js test fails for me on i686:
```
--- FAIL: TestParsedFiles (0.55s)
    --- FAIL: TestParsedFiles/045-loc.js (0.02s)
        js_test.go:51: json: cannot unmarshal number -2403461648 into Go struct field DomainConfig.domains.records.loclongitude of type int
FAIL
FAIL	github.com/StackExchange/dnscontrol/v4/pkg/js	0.553s
FAIL
```

So in pkg/js/helpers.js, I found out that the bit shift operator always returns a signed int32 which means `1 << 31` evaluates to -2147483648 whereas `Math.pow(2, 31)` gives the correct value of +2147483648 for the equator and prime meridian.

And in models/record.go, `LocLatitude` and `LocLongitude` in `UnmarshalJSON` were typed as `int` which didn't match the main struct or RFC 1876 apparently and on 32 bit architectures, `int` ends up as a signed 32 bit integer and any lat/lon coordinates north of the equator or east of the prime meridian can overflow that which causes `json.Unmarshal` to fail.

### DIFF
--- a/models/record.go
+++ b/models/record.go
@@ -172,8 +172,8 @@ func (rc *RecordConfig) UnmarshalJSON(b []byte) error {
 		LocSize            uint8             `json:"locsize,omitempty"`
 		LocHorizPre        uint8             `json:"lochorizpre,omitempty"`
 		LocVertPre         uint8             `json:"locvertpre,omitempty"`
-		LocLatitude        int               `json:"loclatitude,omitempty"`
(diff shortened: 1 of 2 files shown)


TomOnTime:
Thanks for finding this bug and fixing it!

PR #4199, opened 2026-04-14
BUG: TypeScript definition of TXT is incorrect

### LINKED ISSUE #4198: `dnscontrol write-types` generates incorrect type for `contents` in TXT

**Describe the bug**
The [documentation for TXT](https://docs.dnscontrol.org/language-reference/domain-modifiers/txt) states that "The contents is either a single or multiple strings. To specify multiple strings, specify them as an array." The example also already includes the example: `TXT("multiple", ["one", "two", "three"]),  // Multiple strings`.

However, the generated TypeScript file declares the function `TXT` to only accept a string type.

```ts
declare function TXT(name: string, contents: string, ...modifiers: RecordModifier[]): DomainModifier;
```

**To Reproduce**
1. Run `dnscontrol write-types`.
2. See `contents` parameter to be only of type `string`.

**Expected behavior**
`contents` parameter for `TXT` should also accept `string[]`, similar to `LUA`.

**DNS Provider**
 - FILL IN

**Additional context**

dasyad00:
Resolves #4198

### DIFF
--- a/commands/types/dnscontrol.d.ts
+++ b/commands/types/dnscontrol.d.ts
@@ -4087,7 +4087,7 @@ declare function TTL(ttl: Duration): RecordModifier;
  *
  * @see https://docs.dnscontrol.org/language-reference/domain-modifiers/txt
  */
-declare function TXT(name: string, contents: string, ...modifiers: RecordModifier[]): DomainModifier;
+declare function TXT(name: string, contents: string | string[], ...modifiers: RecordModifier[]): DomainModifier;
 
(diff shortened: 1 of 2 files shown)


cafferata (approved):

TomOnTime (approved):
Approved

TomOnTime:
FYI: Right now the project is in transition but I hope to be able to address this once we've moved to our new git org.

TomOnTime (approved):

PR #4203, opened 2026-04-22
GIDINET: Auto-suggest nameservers and add premium DNS helper

zupolgec:
## Summary

- **`GetNameservers` now returns the static free-tier Gidinet nameservers** (`dnsl1/dnsl2.gidinet.com`). Previously it returned `nil, nil`, which meant DNSControl could not derive registrar delegation from the DNS provider for Gidinet-hosted zones. With this change, zones declared as `D(..., REG_GIDINET, DnsProvider(DSP_GIDINET), ...)` get correct NS suggestions automatically, matching the behavior of providers like Cloudflare or netcup.
- **New `GIDINET_PREMIUM_NS()` JS helper** that emits `NAMESERVER("dns1..dns5.gidinet.com.")` for zones on Gidinet's premium DNS tier. Intended to be paired with `DnsProvider(DSP_GIDINET, 0)` so the free-tier defaults from `GetNameservers` are skipped and only the five premium nameservers drive the delegation.
- **`filterApexNS` now drops apex NS records silently when they were synthesized from `dc.Nameservers`** (either auto-injected by `GetNameservers` or declared via `NAMESERVER()`), since `AddNSRecords` adds those to `dc.Records` automatically. Previously it warned on every preview once `GetNameservers` began returning values. Truly unexpected apex NS targets (e.g. a stray `NS("@", "something-else.")`) still emit a warning because the Gidinet DNS API cannot honor them.
- **Documentation updated** to describe both tiers and the `GIDINET_PREMIUM_NS()` usage pattern.

## Motivation

Before this change, a zone served by Gidinet DNS required a redundant `NS_GIDINET()`-style block of explicit `NAMESERVER()` calls just so `REG_GIDINET` could propose the correct delegation. Users hit a no-op/warning loop where the registrar had nothing to compare against and/or where apex NS records raised spurious warnings. Gidinet's nameservers are static per-tier, so returning them from `GetNameservers` is safe and matches the convention used by other providers.

## Test plan

- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `gofmt -l providers/gidinet/ pkg/js/`
- [x] `dnscontrol check` on a real Gidinet-hosted configuration
- [x] `dnscontrol preview` on a free-tier zone: registrar correctly proposes `dnsl1/dnsl2.gidinet.com` delegation without explicit `NAMESERVER()` in the zone file; no spurious apex-NS warnings.
- [ ] `dnscontrol preview` on a premium-tier zone using `DnsProvider(DSP_GIDINET, 0) + GIDINET_PREMIUM_NS()` (pending availability of a premium zone for the author to validate end-to-end).
- [x] IR inspection (`print-ir`) on a synthetic premium-tier config confirms the helper expands to the five `NAMESERVER("dns1..5.gidinet.com.")` entries.

## Notes

- No interface or API changes.
- Backwards compatible: zones that already include explicit `NAMESERVER("dnsl1/dnsl2…")` records still work (the registrar-side dedup already handles duplicates).

### DIFF
--- a/documentation/provider/gidinet.md
+++ b/documentation/provider/gidinet.md
@@ -123,13 +123,20 @@ Allowed TTL values (in seconds):
 
 ### Nameservers
 
-Gidinet's default nameservers are:
+Gidinet offers two DNS tiers with different nameserver sets.
+
+**Free tier (default):**
 - `dnsl1.gidinet.com`
 - `dnsl2.gidinet.com`
 
-**Apex NS records are automatically filtered** by the DNS provider with a warning message. Gidinet does not support modifying NS records at the zone apex via the DNS API - they are managed by the registrar.
+**Premium DNS:**
+- `dns1.gidinet.com`
+- `dns2.gidinet.com`
+- `dns3.gidinet.com`
+- `dns4.gidinet.com`
+- `dns5.gidinet.com`
 
-To manage nameserver delegation, use Gidinet as a **registrar** with the `NAMESERVER()` function:
+The DNS provider returns the free-tier nameservers via `GetNameservers`, so free-tier zones need no explicit `NAMESERVER(...)` — DNSControl will suggest the correct delegation to the registrar automatically:
 
 {% code title="dnsconfig.js" %}
 ```javascript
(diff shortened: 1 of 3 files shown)


TomOnTime:
Thank you for sending this!

TomOnTime (approved):

TomOnTime:
I like that you added  a helper to make it easier to use!

PR #4204, opened 2026-04-24
CHORE: DNSControl.org zone/zone transfer to TransIP

cafferata:
Initial zone configuration for `dnscontrol.org`, based on the records provided by @TomOnTime.

## Records

- `ALIAS @` → `stackexchange.github.io.`
- `CNAME docs` → `db3053e25d-hosting.gitbook.io.`
- `CNAME www` → `dnscontrol.org.`
- `TXT @` Google Search Console verification
- `SPF_NONE` (domain does not send email)

**Notes**: The zone is staged ahead of the domain transfer. Nameservers will be pointed to the new registrar (TransIP) once Stack completes the transfer.

### DIFF
--- a/.dns/dnsconfig.js
+++ b/.dns/dnsconfig.js
@@ -1,0 +1,16 @@
+var PROVIDER_NONE = NewRegistrar('none');
+var PROVIDER_TRANSIP = NewDnsProvider('transip', '-');
+
+DEFAULTS(
+  DefaultTTL('1d')
+)
+
+require_glob(
+  './domains/',
+  false
+);
+
+require_glob(
+  './domains/dnscontrol.org/',
+  false
+);
--- a/.dns/domains/dnscontrol.org.js
+++ b/.dns/domains/dnscontrol.org.js
@@ -1,0 +1,8 @@
+D(
+  'dnscontrol.org',
+  PROVIDER_NONE,
+  DnsProvider(PROVIDER_TRANSIP),
+  ALIAS('@', 'stackexchange.github.io.'),
+  CNAME('www', 'dnscontrol.org.'),
+  END,
+);
--- a/.dns/domains/dnscontrol.org/gitbook.js
+++ b/.dns/domains/dnscontrol.org/gitbook.js
@@ -1,0 +1,3 @@
+D_EXTEND('dnscontrol.org',
+  CNAME('docs', 'db3053e25d-hosting.gitbook.io.'),
+);
--- a/.dns/domains/dnscontrol.org/google-site-verification.js
+++ b/.dns/domains/dnscontrol.org/google-site-verification.js
@@ -1,0 +1,3 @@
+D_EXTEND('dnscontrol.org',
+  TXT('@', 'google-site-verification=iItI_5CTvxVLRyanzMjE9xYi_E2IEotex05tH2AgrVw'),
+);
--- a/.dns/domains/dnscontrol.org/spf.js
+++ b/.dns/domains/dnscontrol.org/spf.js
@@ -1,0 +1,9 @@
+D_EXTEND('dnscontrol.org',
+  SPF_BUILDER({
+    label: '@',
+    parts: [
+      'v=spf1',
+      '-all',
+    ],
+  }),
+);


TomOnTime (commented):
I like the general idea of this but not the implementation.

* nitpick: `D_EXTEND` is very slow. I don't want to encourage its use.
* blocker: One domain per file is good.  Individual files for spf, google site verification, etc. is better done as macros.
* needs more discussion: I strongly feel this should be in a separate repo.  The domain is not part of the project.  For example, if someone were to fork the project, I wouldn't want them to have to delete this and replace it with their own domain.

I think this needs a lot more discussion: I think what we're doing here is trying to decide what "best practice" we want to normalize for our users.  I could be talked out of it, but my vision is two parts: (1) provide `dnscontrol init` (which I haven't tried yet) to help people get started, (2) provide a repo that shows the result of `dnscontrol init` so that (a) people can see what the result looks like, (b) people can see a working example that is exactly what we recommend, (c) people that don't want to or can't use `init` can clone this repo as an alternative way of getting started.

TomOnTime:
@cafferata I created https://github.com/dnscontrol/dns-config as an very simple example for people just getting started. It actually updates a real domain.

cafferata:
Thanks for creating https://github.com/dnscontrol/dns-config, that covers the example for people who are just getting started.

This pull request also addresses a second problem: managing the `dnscontrol.org` zone itself as code. The zone is already live on the TransIP nameservers with the records from this pull request, but as far as I can see it isn't managed as code yet. dns-config uses `best-spaghetti-sauce-ever.com` as its example domain, so it doesn't cover `dnscontrol.org`.

Where would you like the `dnscontrol.org` configuration to live: in a separate repo, or in dns-config? I'll keep this pull request open until we've decided. After that I'll move or rework it into one file for the domain, with macros for the SPF and site verification records and without `D_EXTEND`, as suggested in your review.

TomOnTime:
Hello friend!

I think it would be great to have dnscontrol.org in this repo and best-spaghetti-sauce-ever.com in dns-config.  Maybe we rename them -simple and -fancy?  That way we satisfy both audiences.

Tom

cafferata:
Following up on your review (https://github.com/DNSControl/dnscontrol/pull/4204#pullrequestreview-4215149050) and on your suggestion to rename them `-simple` and `-fancy`.

Since https://github.com/DNSControl/dns-config already exists and already pushes to a real domain, I would rather add `dnscontrol.org` to it as a second domain than create a second repo. That keeps your point about the main repo https://github.com/DNSControl/dnscontrol intact, because nobody who forks DNSControl inherits our domain, and one repo can serve both audiences: `best-spaghetti-sauce-ever.com` as the fictional starter and `dnscontrol.org` as a real zone that is actually pushed. It also avoids maintaining two sets of workflows, secrets and Dependabot configuration.

Your review points are covered in that move:

* No `D_EXTEND`.
* One file per domain: `domains/best-spaghetti-sauce-ever.com.js` and `domains/dnscontrol.org.js`, loaded with `require_glob("./domains/")`.
* The SPF and Google site verification records go inline in the domain file instead of into separate files. The one macro I would add is `NO_EMAIL` (SPF `-all`, null MX, DMARC `p=reject`), because both domains use it.

One thing to decide before I open the pull request: https://github.com/DNSControl/dns-config/blob/main/.github/workflows/pr_preview.yml and https://github.com/DNSControl/dns-config/blob/main/.github/workflows/pr_push.yml both filter on `'*dnsconfig.js'`, and a GitHub path filter does not match across `/`, so changes under `domains/` would trigger neither a preview nor a push. Moving to one file per domain means adding `domains/**` to the `paths:` of both workflows. If you would rather leave the workflows alone, I can keep both domains in `dnsconfig.js`.

If you agree on the location, I will close this pull request and open one against https://github.com/DNSControl/dns-config.

cafferata:
> nitpick: `D_EXTEND` is very slow. I don't want to encourage its use.

Splitting this off, because it reaches well beyond this pull request: I did not know `D_EXTEND()` was slow, and every customer setup I maintain uses it.

The documentation does not mention performance anywhere. The ProTips on https://docs.dnscontrol.org/language-reference/top-level-functions/d_extend only warn about readability, and I could not find an issue or pull request that records it either.

So I asked Claude Opus 5 to work out the performance impact:

> `_getDomainObject()` in `pkg/js/helpers.js` walks every domain defined so far on each `D_EXTEND()` call, which makes the total work quadratic in the number of domains. Generated configurations with 10 records per domain, declared once inside `D()` and once through one `D_EXTEND()` per record, confirm this (`dnscontrol check`, three runs each):
>
> | Domains | In `D()` | Via `D_EXTEND()` |
> | ------- | -------- | ---------------- |
> | 25      | 0.02s    | 0.04s            |
> | 100     | 0.05s    | 0.28s            |
> | 400     | 0.15s    | 3.85s            |

At a few hundred domains that clearly hurts, while at the scale I run (below 30 domains per configuration) it costs hundredths of a second.

Is parse time what you have in mind, or does `D_EXTEND()` cost something later in the run as well? Either way I think it deserves a line in the documentation, because the readability ProTip is the only hint there is today. Happy to send that pull request once I know what it should say.

TomOnTime:
> Following up on your review ([#4204 (review)](https://github.com/DNSControl/dnscontrol/pull/4204#pullrequestreview-4215149050)) and on your suggestion to rename them `-simple` and `-fancy`.
> 
> Since https://github.com/DNSControl/dns-config already exists and already pushes to a real domain, I would rather add `dnscontrol.org` to it as a second domain than create a second repo. That keeps your point about the main repo https://github.com/DNSControl/dnscontrol intact, because nobody who forks DNSControl inherits our domain, and one repo can serve both audiences: `best-spaghetti-sauce-ever.com` as the fictional starter and `dnscontrol.org` as a real zone that is actually pushed. It also avoids maintaining two sets of workflows, secrets and Dependabot configuration.

`best-spaghetti-sauce-ever.com` is not fictional. Last year the ad revenue made nearly 3 USD!

However, I understand your point.

> Your review points are covered in that move:
> 
> * No `D_EXTEND`.
> * One file per domain: `domains/best-spaghetti-sauce-ever.com.js` and `domains/dnscontrol.org.js`, loaded with `require_glob("./domains/")`.
> * The SPF and Google site verification records go inline in the domain file instead of into separate files. The one macro I would add is `NO_EMAIL` (SPF `-all`, null MX, DMARC `p=reject`), because both domains use it.
> 
> One thing to decide before I open the pull request: https://github.com/DNSControl/dns-config/blob/main/.github/workflows/pr_preview.yml and https://github.com/DNSControl/dns-config/blob/main/.github/workflows/pr_push.yml both filter on `'*dnsconfig.js'`, and a GitHub path filter does not match across `/`, so changes under `domains/` would trigger neither a preview nor a push. Moving to one file per domain means adding `domains/**` to the `paths:` of both workflows. If you would rather leave the workflows alone, I can keep both domains in `dnsconfig.js`.
> 
> If you agree on the location, I will close this pull request and open one against https://github.com/DNSControl/dns-config.

I don't understand why people are so enamored with one file per domain.  Having all the domains in one dnsconfig.js makes it easier to search, easier to do global updates, etc.  If I had hundreds of domains it would be different.  It's also easier to explain to new people.

Plus, doesn't "init" add all new domains to dnsconfig.js?

I still prefer two repos: -simple and -fancy.    If we're going to have 1 repo, `best-spaghetti-sauce-ever.com` should be in the dnsconfig.js and `dnscontrol.org` could be in its own file.  That way people see two options and can make their own decision.

Tom

PR #4207, opened 2026-04-25
AXFRDDNS: Add missing ALPN token 'dot'

Mr-Foo:
AXFRDDNS does not comply with RFC 9103 "DNS Zone Transfer over TLS " section 7.1 "Connection Establishment", which states:
> During connection establishment, the Application-Layer Protocol
> Negotiation (ALPN) token "dot" [DoT-ALPN] MUST be selected in the TLS
> handshake. [1]

Without ALPN `bind` (tested 9.20.22) refuses to return the zone:
```raw
(16.172.in-addr.arpa): view default: query failed (no ALPN) for 16.172.in-addr.arpa/IN/AXFR at query.c:12143
```

While `dnscontrol` logs:
```raw
INFO#1: Domain "16.172.in-addr.arpa" provider axfrddns Error: [Error] AXFRDDNS: nameserver refused to transfer the zone 16.172.in-addr.arpa: dns: bad xfr rcode: 2
```

Since the requirement in RFC 9103 is strict, I propose to unconditionally add `NextProtos: dot`.

[1] [RFC 9103](https://datatracker.ietf.org/doc/rfc9103/)


Mr-Foo:
@hnrgrgr please review.

TomOnTime:
Thank you for sending this!

Mr-Foo:
Thank you for your consideration.

I will be happy to revisit this PR once you are ready.

TomOnTime (approved):

TomOnTime:
CC @hnrgrgr for confirmation

Mr-Foo:
Thank you @TomOnTime!
Just upgraded to the new version :heart:

PR #4223, opened 2026-05-03
BUILD: Fix typo in GoReleaser GHCR image path

cafferata:
The GHCR image path contained 'dnscontrodnscontrol' instead of 'dnscontrol', causing pushes to be written under the wrong namespace. The same typo also appeared in the release notes Docker example.


TomOnTime (approved):
Approved

PR #4230, opened 2026-05-07
CICD: Integration test output not visible

### LINKED ISSUE #4220: Fix dockerhub
As @cafferata pointed out in https://github.com/DNSControl/dnscontrol/pull/4214:

> One operational follow-up that's not in the diff but blocks the rename being effective: Docker Hub for the new org still needs to be set up. The current `.goreleaser.yml` publishes to [`hub.docker.com/r/stackexchange/dnscontrol`](https://hub.docker.com/r/stackexchange/dnscontrol), and once that's pointed at a new namespace (presumably `dnscontrol/dnscontrol` or similar), the org has to exist on Docker Hub with the right credentials wired into the release workflow. Worth flagging so it doesn't get forgotten between the GitHub rename and the next release.

DockerHub orgs are $15/seat/month.  Before we create this org, let's make sure its absolutely needed.

* Is ghcr.io sufficient?

### LINKED ISSUE #4227: Create first release now that we're independent
CC @cafferata @chicks-net for feedback and help.

# Issue

A lot broke when the repos moved to the DNSControl org.  We are focusing on getting back to parity with the old system.  Parity includes:

* Binaries created
* Containers sent to Docker Hub and GHCR
* Homebrew gets the latest binary (under the new name)
* GitBook integration working
* (maybe other things too)

Anti-goals include:

* New features, bug fixes, new providers.  Let's avoid scope creep.

# Resolution

* We will do a series of test releases under v4.36.x.  They will not be released to the world. (they remain private)
* When the process works fully, we'll do a release v4.37.0

### LINKED ISSUE #4233: Fix GHCR container access
# Issue

Container access gets an unauthorized error.

```
[09:59:30 PM]$ docker pull ghcr.io/dnscontrol/dnscontrol:4.37.0
Error response from daemon: error from registry: unauthorized
unauthorized

[10:00:45 PM]$ docker pull ghcr.io/dnscontrol/dnscontrol:latest
Error response from daemon: error from registry: unauthorized
unauthorized
```

# Resolution

I fixed the setting in the GH console.  We need to push a new release to see if it worked.

TomOnTime:
Fixes https://github.com/DNSControl/dnscontrol/issues/4233
Fixes https://github.com/DNSControl/dnscontrol/issues/4227
Fixes https://github.com/DNSControl/dnscontrol/issues/4220

# Issue

Issue 1: The output of the integration tests isn't displayed anywhere. It seems that the output is generated (provider tests take the right amount of time) but I don't see anyplace in the UI that I can view the results.
- Example: 
https://github.com/DNSControl/dnscontrol/actions/runs/25500999681/job/74833580403

Issue 2: The code that sets "END" is failing with a `if [[ not found` error.

# Resolution

- Issue 2:  It turns out the code to generate `$END` is using `if [[` which is a bash-ism, and this GHA defaults to `sh`.  Change to the `bash` shell for that step.

### DIFF
--- a/.github/workflows/pr_integration_tests.yml
+++ b/.github/workflows/pr_integration_tests.yml
@@ -271,6 +271,7 @@ jobs:
           echo "END: Using: END=$END"
           go install gotest.tools/gotestsum@latest
           if [ -n "$${{ matrix.provider }}_DOMAIN" ] ; then
+            echo "DEBUG: matrix.provider=$${{ matrix.provider }}_DOMAIN"
             gotestsum --junitfile ${TEST_RESULTS}/gotestsum-report.xml -- \
                 -timeout 30m -v \
                 -verbose \
@@ -279,7 +280,9 @@ jobs:
                 -cfredirect=false \
                 -cfflatten=false \
                 -cftags=false \
-                -end $END ./... > ${TEST_RESULTS}/gotestsum-output.txt
+                -end $END \
+                ./... \
+                > ${TEST_RESULTS}/gotestsum-output.txt
           else
             echo "Skip test for ${{ matrix.provider }} provider"
           fi
--- a/.github/workflows/pr_integration_tests.yml
+++ b/.github/workflows/pr_integration_tests.yml
@@ -246,10 +246,8 @@ jobs:
           DO_WORKERS=false
 
           # Only for PRs
-          echo "DEBUG: ONLY FOR PRs: ${{ github.event_name }} == pull_request"
           if [[ "${{ github.event_name }}" == "pull_request" ]]; then
             # Check for 'longtest' label
-            echo "DEBUG: check label: ${{ github.event.pull_request.labels.*.name }}"
             #if [[ "${{ github.event.pull_request.labels.*.name }}" =~ "longtest" ]]; then
             if [[ " $LABELS " =~ " longtest " ]]; then
               END=999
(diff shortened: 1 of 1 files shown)


TomOnTime:
@chicks-net @cafferata:  I'm stumped.  It's probably something obvious.  I see the output collected in `${TEST_RESULTS}/gotestsum-output.txt` but I don't see any code that gives it to GitHub to display it.  Any thoughts?

cafferata:
@TomOnTime The output was visible before #4153 (March 11). That PR added the `> ${TEST_RESULTS}/gotestsum-output.txt` redirect to the gotestsum command, which sends all stdout to the file instead of the Actions log. The file gets uploaded as an artifact, but there's nothing that prints it back to the log.

I verified this by downloading the full job log archive for this PR's run:
https://github.com/DNSControl/dnscontrol/suites/67915295585/logs?attempt=1

Looking at the TRANSIP job log, there's a 10 minute gap between `END: Using: END=999` and the upload-artifact step. The tests ran, but all output went to the file.

The fix for Issue 1 would be to replace `>` with `| tee`:

```bash
gotestsum ... ./... 2>&1 | tee ${TEST_RESULTS}/gotestsum-output.txt
```

This writes to both the file (for the artifact) and stdout (for the Actions log).

TomOnTime:
Ah!!!  Thank you!!

PR #4208, opened 2026-04-25
FEATURE: Init command

cafferata:
# Add interactive `dnscontrol init` onboarding wizard

Closes #4205

## Summary

A new user can run `dnscontrol init` and have a working `creds.json` plus a starter `dnsconfig.js` within a minute. The wizard asks the right questions for the chosen provider and writes both files for them. Existing workflows are unaffected; users who prefer to hand edit `creds.json` can keep doing exactly that.

## Why

DNSControl's biggest entry barrier is the very first ten minutes. Newcomers have to discover, on their own, what fields belong in `creds.json` for their provider, where to create the API token, whether a value should be a secret, whether it spans multiple lines. That information lives in three places (the provider documentation, the provider's source, and the integration tests workflow) and rarely all in the same form. After helping multiple developers through that ramp myself, the friction is recognisable every time.

The wizard collapses that into a single guided session. A maintainer describes their provider's credential fields once via `RegisterCredsMetadata`, and from then on every newcomer reuses that knowledge.

## What this gives users

- **A guided start.** Pick a DNS provider, optionally a registrar, type or paste your credentials, name a domain. Done.
- **Sensible defaults.** Where the DNS provider can also act as a registrar, the wizard offers to reuse the same credentials. Pressing Enter accepts the suggested defaults where they exist. PEM blocks open the user's `$EDITOR` so multiline values stay intact.
- **Immediate feedback.** Right after writing, the wizard offers to compare the user's `dnsconfig.js` with the live zones at the provider, and to run `dnscontrol preview` as a sanity check.
- **A community welcome.** The flow closes with pointers to the documentation, GitHub Discussions, and the monthly community video call.

## What this gives provider maintainers

A small registration block next to the existing `RegisterMaintainer` call:

```go
providers.RegisterCredsMetadata("MYPROVIDER", providers.CredsMetadata{
    DisplayName: "My Provider",
    Kind:        providers.KindDNS,
    PortalURL:   "https://portal.example.com/api-tokens",
    Fields: []providers.CredsField{
        {Key: "apitoken", Label: "API Token", Required: true, Secret: true},
    },
})
```

That is enough for the provider to appear in the wizard. Providers with multiple auth methods (TransIP access token versus account name plus private key, or Route53 static keys versus assume role) can branch with `Internal` and `ShowIf`.

Three providers ship with metadata in this PR as a starting point and a worked example: NONE, BIND, and TransIP. The remaining providers can land incrementally in follow up PRs, one per maintainer.

## Scope and follow ups

[...]

PR #4240, opened 2026-05-10
BUG: Fix get-zones argument parsing to match documentation

### LINKED ISSUE #4235: `get-zones` argument parsing VS documentation mismatch
**Describe the bug**
The usage of the `get-zones` subcommand does not seem to work like described in https://docs.dnscontrol.org/commands/get-zones

**To Reproduce**
1. According to docs `dnscontrol get-zones --format=nameonly desec all` should work as of v4.0 (note: "desec" does exist as entry in my creds.json)
2. I get a return code of 1, and the output `Arguments should be: credskey providername zone(s) (Ex: r53 ROUTE53 example.com)`

**Expected behavior**
- the command to return a list of domains registered with that provider

**DNS Provider**
irrelevant

**Additional context**
- DNSControl version 4.37.1
- I was previously successfully using `dnscontrol get-zones --format=nameonly desec - all`, this does also not work anymore, and returns with the same error message. Quoting the `-` or bash instead of zsh do not make a difference.
- Only the format `dnscontrol get-zones --format=nameonly desec DESEC all` gives expected output, but also gives the hint: `WARNING: To retain compatibility in future versions, please change "DESEC" to "-". See "https://docs.dnscontrol.org/commands/get-zones"` - which as I said is not working for me
- `dnscontrol check-creds desec` works as expected

cafferata:
I ran into this same issue while working on #4208 and can confirm the behavior @stefanwascoding describes. The only form that works requires the provider name explicitly:

```shell
dnscontrol get-zones transip TRANSIP all
```

```text
WARNING: To retain compatibility in future versions, please change "TRANSIP" to "-". See "https://docs.dnscontrol.org/commands/get-zones"
```

But following that suggestion also fails:

```shell
dnscontrol get-zones transip - all
```

```text
Arguments should be: credskey providername zone(s) (Ex: r53 ROUTE53 example.com)
```

So neither the documented 2 argument form (`transip all`) nor the suggested dash form (`transip - all`) works.

This pull request fixes the argument parsing so that `get-zones` accepts 2 arguments (`credkey zone`), resolving the provider type from the `TYPE` field in `creds.json`. The old 3 argument form still works but now shows a deprecation warning with the correct command:

```shell
dnscontrol get-zones transip TRANSIP all
```

```text
WARNING: The provider name argument is deprecated. Please use "dnscontrol get-zones transip all" instead. See "https://docs.dnscontrol.org/commands/get-zones"
```

With 3+ arguments, the code now detects whether the second argument is a registered provider type to distinguish the deprecated form (`credkey provider zone...`) from the new multi-zone form (`credkey zone1 zone2...`):

```shell
dnscontrol get-zones --format=nameonly transip dnscontrol.org dnscontrol.nl
```

```text
dnscontrol.org
dnscontrol.nl
```

Additionally:

- CLI help is shortened and links to online documentation
- Examples use descriptive `credkey` names (`my_route53` instead of `myr53`) so they are self-explanatory without the provider name
- `Documentation:` links are added to `get-zones` and `check-creds`
- `documentation/commands/get-zones.md` is updated to match

Tested against the TransIP provider with the `dnscontrol.org` zone from #4204.

- [x] `go test ./commands/`
- [x] `dnscontrol get-zones transip all` (2 args, new style)
- [x] `dnscontrol get-zones transip dnscontrol.org` (specific zone)
- [x] `dnscontrol get-zones transip dnscontrol.org dnscontrol.nl` (multiple zones, new style)
- [x] `dnscontrol get-zones transip TRANSIP all` (3 args, deprecation warning)
- [x] `dnscontrol get-zones transip` (too few args, error message)
- [x] `dnscontrol get-zones --format=zone transip dnscontrol.org`
- [x] `dnscontrol get-zones --format=tsv transip dnscontrol.org`
- [x] `dnscontrol get-zones --format=nameonly transip all`

<details>
    <summary>CLI diff</summary>

```diff
--- before (v4.37.1)
+++ after
@@ -2,7 +2,7 @@
    dnscontrol get-zones - gets a zone from a provider (stand-alone)
 
 USAGE:
-   dnscontrol get-zones [command options] credkey provider zone [...]
+   dnscontrol get-zones [command options] credkey zone [...]
 
 CATEGORY:
    utility
@@ -11,34 +11,18 @@
    Download a zone from a provider.  This is a stand-alone utility.
 
    ARGUMENTS:
-      credkey:  The name used in creds.json (first parameter to NewDnsProvider() in dnsconfig.js)
+      credkey:  The name used in creds.json
-      provider: The name of the provider (second parameter to NewDnsProvider() in dnsconfig.js)
       zone:     One or more zones (domains) to download; or "all".
 
-   FORMATS:
-      --format=js        dnsconfig.js format (not perfect, just a decent first draft)
-      --format=djs       js with disco commas (leading commas)
-      --format=zone      BIND zonefile format
-      --format=tsv       TAB separated value (useful for AWK)
-      --format=nameonly  Just print the zone names
-
-   The columns in --format=tsv are:
-      FQDN (the label with the domain)
-      ShortName (just the label, "@" if it is the naked domain)
-      TTL
-      Record Type (A, AAAA, CNAME, etc.)
-      Target and arguments (quoted like in a zonefile)
-      Either empty or a comma-separated list of properties like "cloudflare_proxy=true"
-
-   The --ttl flag only applies to zone/js/djs formats.
-
    EXAMPLES:
-      dnscontrol get-zones myr53 ROUTE53 example.com
-      dnscontrol get-zones gmain GANDI_V5 example.com other.com
-      dnscontrol get-zones cfmain CLOUDFLAREAPI all
-      dnscontrol get-zones --format=tsv bind BIND example.com
-      dnscontrol get-zones --format=djs --out=draft.js gcloud GCLOUD example.com
+      dnscontrol get-zones my_route53 example.com
+      dnscontrol get-zones my_gandi example.com other.com
+      dnscontrol get-zones my_cloudflare all
+      dnscontrol get-zones --format=tsv my_bind example.com
+      dnscontrol get-zones --format=djs --out=draft.js my_gcloud example.com
 
+   Documentation: https://docs.dnscontrol.org/commands/get-zones
+
 OPTIONS:
    --creds string   Provider credentials JSON file (or !program to execute program that outputs json) (default: "creds.json")
    --format string  Output format: js djs zone tsv nameonly (default: "zone")
```
</details>

Fixes #4235


TomOnTime (approved):
Thanks for fixing this and cleaning how it works!

TomOnTime (commented):
nitpick: Previously the code allowed a list of zone names.  That's why the `-h` listing includes `[...]` at the end of the usage spec:

```
USAGE:
   dnscontrol get-zones [command options] credkey zone [...]
```

I'm ok with fixing the code or simply updating the docs. I don't know many times when listing multiple zones is needed.

Related: (mostly thinking out loud. These don't have to be part of this PR).  I wonder if we should drop the `all` and just allow an empty list to mean all zones.  I've also thought that we should create `list-zones` as an alias for `get-zones --format=nameonly credkey all`

cafferata:
Thanks for the approval and feedback! I'll address each point:

**Multi-zone support:** Good catch. I hadn't thought about this scenario (I've never used the multi-zone form myself). Since it's a small fix, I went ahead and implemented it right away by checking whether the second argument is a registered provider type. If it is: deprecated path with warning. If not: new format with multiple zones. This should cover all remaining edge cases around `get-zones` argument parsing.

**`list-zones` alias:** Great idea! I ran into the same need in https://github.com/DNSControl/dnscontrol/pull/4208 when wanting to list domains during `init`. A dedicated `list-zones` command would have been a much cleaner fit there. Happy to explore that in a follow-up.

**Removing the [v3.16](https://docs.dnscontrol.org/release/v316) migration notes from the docs:** While cleaning up the documentation I kept the "As of v3.16" section as-is, but I think we can safely remove it entirely. [v3.16.0](https://github.com/DNSControl/dnscontrol/releases/tag/v3.16.0) was released over 4 years ago (May 2022). Anyone still on v3.15 or older is unlikely to consult the current docs for migration guidance. What do you think?

TomOnTime:
> Thanks for the approval and feedback! I'll address each point:
> 
> **Multi-zone support:** Good catch. I hadn't thought about this scenario (I've never used the multi-zone form myself). Since it's a small fix, I went ahead and implemented it right away by checking whether the second argument is a registered provider type. If it is: deprecated path with warning. If not: new format with multiple zones. This should cover all remaining edge cases around `get-zones` argument parsing.

👍 

> **`list-zones` alias:** Great idea! I ran into the same need in #4208 when wanting to list domains during `init`. A dedicated `list-zones` command would have been a much cleaner fit there. Happy to explore that in a follow-up.

👍 

> **Removing the [v3.16](https://docs.dnscontrol.org/release/v316) migration notes from the docs:** While cleaning up the documentation I kept the "As of v3.16" section as-is, but I think we can safely remove it entirely. [v3.16.0](https://github.com/DNSControl/dnscontrol/releases/tag/v3.16.0) was released over 4 years ago (May 2022). Anyone still on v3.15 or older is unlikely to consult the current docs for migration guidance. What do you think?

Agreed.

cafferata:
Thanks! I've pushed the v3.16 migration guide removal. This cleans up references across `check-creds.md`, `creds-json.md`, `NewRegistrar.md`, `NewDnsProvider.md`, `SUMMARY.md`, and removes `documentation/release/v316.md` entirely.

Ready for a final review whenever you have a moment.

TomOnTime (approved):
Awesome!

PR #4249, opened 2026-05-13
LUADNS: support OPENPGPKEY record type

riku22:
Adds support for OPENPGPKEY record type in LUADNS.


TomOnTime:
Thanks for updating this!

Issue #4088, opened 2026-02-20
Suggested changes to the README.md


tlimoncelli:
1. Reorg the list of providers

Now that the list is so long, we probably want to list them in a more compact way, and group them.

Some thoughts:
* Instead of 2 lists (DSPs and Registrars), one lists with an "*" to mean "registrar only".
* Instead of a long list, use be a table with 5 names across.
* Group them by category: "Cloud Providers", "DNS SaaS providers", "Self-Hosted" (BIND, PowerDNS), "Network Hardware"

2. The opening paragraph should use more updated terminology. IaC instead of DSL.

DNSControl is Infrastructure as Code for DNS. It includes a full-featured configuration language (Javascript-compatible), plus "plug-ins" that speak to DNS provider APIs. It can upload the same DNS records to multiple providers to improve reliability.  It even generates the most beautiful BIND zone files ever. It runs anywhere Go runs (Linux, macOS, Windows, FreeBSD). It is distributed as source, binaries, and Docker containers.  Get started easily with the import feature!

3.  Move the example up higher

The example really sells it.  Move that up.

cafferata:
@TomOnTime I've put together an initial implementation in https://github.com/DNSControl/dnscontrol/pull/4263. The three changes from your suggestions: "Infrastructure as Code" framing in the opening paragraph, the code example moved above the provider list, and the two provider lists combined into a single table. Let me know if this is what you had in mind.

PR #4263, opened 2026-05-14
DOCS: Reorganize README.md

cafferata:
The opening paragraph now uses "Infrastructure as Code" framing instead of "domain specific language (DSL)". The code example is moved up so readers see working code before the provider list. The two separate provider lists (57 DNS providers and 19 registrars) are combined into a single 5-column table with superscript markers: ¹ for providers that also support registrar functions and ² for registrar-only providers. Each provider links to its documentation page on https://docs.dnscontrol.org.

Addresses #4088.


cafferata:
Preview of the [rendered README](https://github.com/cafferata/dnscontrol/blob/docs/readme-reorganization/README.md) 📚

TomOnTime (approved):
Wow! That looks so much better!  Thanks!

PR #4264, opened 2026-05-14
LUADNS: changed to use record set to perform bulk operations

riku22:
Changed LUADNS to use record sets to perform bulk operations on records.

I think this will reduce the number of requests and the time it takes a bit.


TomOnTime (approved):

TomOnTime:
Excellent!  This should speed up big changes.

Issue #4210, opened 2026-04-27
get-zones skips children on AutoDNS


dkorunic:
**Describe the bug**
get-zones skips over children zones on AutoDNS. There are [multiple levels](https://help.internetx.com/display/APIXMLEN/User+status) of users on InternetX AutoDNS and sub-users which can be owners of domains. Due to `Children: false` in `getZones()` we are skipping all domains that typically AutoDNS interface shows for master/admin users.

**To Reproduce**
1. Create sub-users on AutoDNS.
2. Assign a domain to sub-user.
3. Try to fetch domain with "get-zones" either directly or with "all" keyword from a top-level user.

**Expected behavior**
We should always default to safe default with `Children: true` the same way AutoDNS UI does. User can filter out unneeded domains.

**DNS Provider**
 - AUTODNS

**Additional context**
[User levels documentation on AutoDNS](https://help.internetx.com/display/APIXMLEN/User+status)

TomOnTime:
CC arnoschoon

arnoschoon:
Hi @dkorunic ,

let me see if I understand your case correctly.

You have a single overarching API user/serviceaccount to manage zones in AutoDNS but need to split those zones into smaller chunks to divide them into teams or something?

As far as my experience with the UI goes this checkbox is not on by default, I think we might be better of if we find a way to make it optional for this tool.

Currently we have a pipeline per team (set of zones) in order to allow team members from managing their own DNS, my guess is that we never ran into this issue you're describing.

PR #4272, opened 2026-05-16
AUTODNS: support init command

TomOnTime:
## Summary

Register `CredsMetadata` for AUTODNS so the provider is offered by the `dnscontrol init` wizard.

CC: @arnoschoon

## Test plan

- [ ] `go build ./...` passes
- [ ] `dnscontrol init` lists AUTODNS as an option


providers/autodns/autoDnsProvider.go
```diff
@@ -54,6 +54,33 @@ func init() {
 	}, features)
 	providers.RegisterDomainServiceProviderType(providerName, fns, features)
 	providers.RegisterMaintainer(providerName, providerMaintainer)
+	providers.RegisterCredsMetadata(providerName, providers.CredsMetadata{
+		DisplayName: "AutoDNS",
+		Kind:        providers.KindDNS | providers.KindRegistrar,
+		DocsURL:     "https://docs.dnscontrol.org/provider/autodns",
+		PortalURL:   "https://login.autodns.com/", // TODO: Verify
```
arnoschoon, line 62: To be honest I've never used this, the reseller that manages AutoDNS is based in the Netherlands so I always used https://login.autodns.nl/.

The API is on api.autodns.com (the .com-tld) so I guess this is correct.

arnoschoon (approved):

arnoschoon:
@TomOnTime — heads up that #4391 adds a new optional `children` creds field to AUTODNS (opt-in, default off; when `"true"`, `get-zones` also returns zones owned by sub-users, fixing #4210).

Since this PR registers `CredsMetadata` for the `init` wizard, should `children` be included there too? It'd be an optional, non-secret field, e.g.:

```go
{
    Key:   "children",
    Label: "Include sub-user zones",
    Help:  "Set to \"true\" so get-zones also lists zones owned by sub-users (master/admin accounts). Optional; defaults to off.",
},
```

Happy to either add it here once this merges, or fold it into #4391 — whichever you prefer. Just didn't want the two to drift out of sync.

TomOnTime:
Please fold it into https://github.com/DNSControl/dnscontrol/pull/4391

Thanks!

TomOnTime:
P.S. I added the "children" setting.  CC: @arnoschoon

PR #4275, opened 2026-05-16
BUNNY_DNS: support init command

TomOnTime:
## Summary

Register `CredsMetadata` for BUNNY_DNS so the provider is offered by the `dnscontrol init` wizard.

CC: @ppmathis

## Test plan

- [ ] `go build ./...` passes
- [ ] `dnscontrol init` lists BUNNY_DNS as an option

### DIFF
--- a/providers/bunnydns/bunnydnsProvider.go
+++ b/providers/bunnydns/bunnydnsProvider.go
@@ -55,7 +55,7 @@ func init() {
 		DisplayName: "Bunny DNS",
 		Kind:        providers.KindDNS,
 		DocsURL:     "https://docs.dnscontrol.org/provider/bunnydns",
-		PortalURL:   "https://dash.bunny.net/account/api-key", // TODO: Verify
+		PortalURL:   "https://dash.bunny.net/account/api-key",
 		Fields: []providers.CredsField{
 			{
 				Key:      "api_key",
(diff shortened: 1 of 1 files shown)


providers/bunnydns/bunnydnsProvider.go
```diff
@@ -50,6 +50,22 @@ func init() {
 
 	providers.RegisterCustomRecordType("BUNNY_DNS_RDR", providerName, "")
 	providers.RegisterCustomRecordType("BUNNY_DNS_PZ", providerName, "")
+
+	providers.RegisterCredsMetadata(providerName, providers.CredsMetadata{
+		DisplayName: "Bunny DNS",
+		Kind:        providers.KindDNS,
+		DocsURL:     "https://docs.dnscontrol.org/provider/bunnydns",
+		PortalURL:   "https://dash.bunny.net/account/api-key", // TODO: Verify
```
ppmathis: This URL is indeed correct and there is also only a single type of API key, so this verify todo can be dropped.

ppmathis (changes requested):
Thanks for the PR! This is indeed all that is needed to be configured to get `dnscontrol init` working for bunny.net, so once that todo comment is removed I personally see no reason not to merge this.

TomOnTime:
Thanks for the confirmation!

PR #4288, opened 2026-05-16
GIDINET: support init command

TomOnTime:
## Summary

Register `CredsMetadata` for GIDINET so the provider is offered by the `dnscontrol init` wizard.

CC: @zupolgec

## Test plan

- [ ] `go build ./...` passes
- [ ] `dnscontrol init` lists GIDINET as an option


providers/gidinet/gidinetProvider.go
```diff
@@ -71,6 +71,27 @@ func init() {
...
+				Help:     "Your Gidinet account username.",
+				Required: true,
+			},
+			{
+				Key:      "password",
+				Label:    "Password",
+				Help:     "Your Gidinet account password.",
+				Secret:   true,
+				Required: true,
+			},
```
zupolgec: The `PortalURL` had a `// TODO: Verify` placeholder. Gidinet's API auth isn't the control-panel login: the **username** is the normal account username, but the **password** must be the dedicated "API password" generated at https://www.gidinet.com/modules/private/account_password/. This suggestion points `PortalURL` straight at that page, adds a `Notes` line, and clarifies the field labels/help.

```suggestion
		PortalURL:   "https://www.gidinet.com/modules/private/account_password/",
		Notes:       "The username is your usual Gidinet account username, but the password must be the dedicated \"API password\" generated in the Gidinet control panel — not your normal login password.",
		Fields: []providers.CredsField{
			{
				Key:      "username",
				Label:    "Username",
				Help:     "Your Gidinet account username (same as the control-panel login).",
				Required: true,
			},
			{
				Key:      "password",
				Label:    "API password",
				Help:     "The \"API password\" generated at https://www.gidinet.com/modules/private/account_password/ — not your normal login password.",
				Secret:   true,
				Required: true,
			},
```

zupolgec (commented):
Thanks @TomOnTime ! Builds clean, tests pass, and `dnscontrol init` lists GIDINET. One correction to the creds metadata, plus a related docs fix.

**Creds metadata** — see the inline suggestion. Gidinet doesn't authenticate with the control-panel login: the username is the normal account username, but the password must be the dedicated **"API password"** generated at https://www.gidinet.com/modules/private/account_password/. The suggestion resolves the `// TODO: Verify` on `PortalURL` (it's `.com`, and points right at the API-password page), adds a `Notes` line, and clarifies the field labels/help.

**Docs** — `documentation/provider/gidinet.md` isn't in this PR, but its Activation section currently claims the credentials "are the same ones you use to log in to the control panel," which is wrong for the password. Worth fixing in the same PR. Suggested replacement for that section:

```markdown
## Activation

1. Log in to the [Gidinet Control Panel](https://www.gidinet.com/)
2. Generate an **API password** at [https://www.gidinet.com/modules/private/account_password/](https://www.gidinet.com/modules/private/account_password/)
3. In `creds.json`, use your account **username** together with that **API password** — the API password is distinct from your normal login password
```

TomOnTime:
Thanks!  I've updated the documentation.

TomOnTime:
Imma merge this so that it gets into the new release.  We can fix any docs in an separate PR.

PR #4289, opened 2026-05-16
HEDNS: support init command

TomOnTime:
## Summary

Register `CredsMetadata` for HEDNS so the provider is offered by the `dnscontrol init` wizard.

CC: @rblenkinsopp

## Test plan

- [ ] `go build ./...` passes
- [ ] `dnscontrol init` lists HEDNS as an option


providers/hedns/hednsProvider.go
```diff
@@ -77,6 +77,38 @@ func init() {
 	}
 	providers.RegisterDomainServiceProviderType(providerName, fns, features)
 	providers.RegisterMaintainer(providerName, providerMaintainer)
+	providers.RegisterCredsMetadata(providerName, providers.CredsMetadata{
+		DisplayName: "Hurricane Electric DNS",
+		Kind:        providers.KindDNS,
+		DocsURL:     "https://docs.dnscontrol.org/provider/hedns",
+		PortalURL:   "https://dns.he.net/", // TODO: Verify
```
rblenkinsopp: Yes, this is the correct URL, I'm not sure why this is TODO as this is very easy to verify, was this Claude submission reviewed?

providers/hedns/hednsProvider.go
```diff
@@ -77,6 +77,38 @@ func init() {
...
+		Fields: []providers.CredsField{
+			{
+				Key:      "username",
+				Label:    "Username",
+				Help:     "Your dns.he.net account username.",
```
rblenkinsopp: I would change this (as the account is broader than just DNS):

```suggestion
				Help:     "Your Hurricane Electric account username.",
```

providers/hedns/hednsProvider.go
```diff
@@ -77,6 +77,38 @@ func init() {
...
+				Required: true,
+			},
+			{
+				Key:      "password",
+				Label:    "Password",
+				Help:     "Your dns.he.net account password.",
```
rblenkinsopp:
```suggestion
				Help:     "Your Hurricane Electric account password.",
```

providers/hedns/hednsProvider.go
```diff
@@ -77,6 +77,38 @@ func init() {
...
+				Key:    "totp-key",
+				Label:  "TOTP shared secret (optional)",
+				Help:   "Shared TOTP secret used to generate a valid TOTP code. Leave blank if you do not use TOTP.",
+				Secret: true,
+			},
+			{
+				Key:   "session-file-path",
+				Label: "Session file path (optional)",
+				Help:  "Path to a directory where the .hedns-session file is stored to reuse the authenticated session. Leave blank to log in each run.",
+			},
```
rblenkinsopp, line 110: For this, I'd be inclined to set the default to `.`, as without this the user will get the suprising behaviour only being able to issue one command every 30 seconds if TOTP is enabled.

```suggestion
			{
				Key:     "session-file-path",
				Label:   "Session file path (optional)",
				Help:    "Path to a directory where the .hedns-session file is stored to reuse the authenticated session. Leave blank to log in each run.",
				Default: ".",
			},
```

While this is slightly less secure, it is documented in the associated documentation, and since this is for the `init` getting started workflow, I would favour useability in this instance.

rblenkinsopp (changes requested):

rblenkinsopp:
Thanks for the submission, mostly looks good with a some minor suggestions/alterations.

providers/hedns/hednsProvider.go
TomOnTime: Yes, Claude created the PR.  Even if I had manually made the PR I would have asked ppl to verify I guessed the right URLs :)

TomOnTime:
Thanks for the reviews and suggestions!

PR #4301, opened 2026-05-17
FEATURE: `init` - verify credentials, select zones and import records

### LINKED ISSUE #4300: `dnscontrol init` feature ideas
Some additional features I'd like to see in init:

If `ListZones()` is implemented on a provider, there are a lot of features we can add.  `ListZones()` is the only way (currently) to test credentials. There's no "no-op" or "Test Credentials" call, though maybe we should add one?

1. Verify the credentials

If they fail, offer the option of correcting them or just leave them as-is. (users may have reasons the secrets might not work yet)

2. If `ListZones()` is supported, list the domains available.

Users should still be able to manually enter a zone name.

3. After each zone name is entered, offer to get the records and add them to dnsconfig.js.

The equivalent of `dnscontrol get-zones --format=js credkey zonename`

Sadly, its quite difficult to update dnsconfig.js using automated methods.  A good first step would be to write the records to a file and tell the human to add it to dnsconfig.js.  Or blindly append it to the end of dnsconfig.js and warn the user this may result in a duplicate?

cafferata:
I am really excited about this one. Before `init`, onboarding a new provider meant reading the documentation and taking the first hurdle manually. The `init` command removed that barrier, but it still generated placeholder configs that required manual editing before they did anything useful. With this change, `init` produces a working `dnsconfig.js` from the start: real credentials, real zones, real records. The onboarding experience goes from "here is a template, now figure it out" to "here is your actual DNS, ready for `dnscontrol preview`".

This also opens the door to a follow-up feature I have been thinking about: syncing from a remote provider back into DNSControl. A diff between the live zone and the local config, with the option to pull changes back. But that is for another PR.

Builds on the existing `init` wizard with three capabilities: credential verification, zone-based domain selection, and automatic record import.

Credential verification instantiates the provider and calls `ListZones` immediately after the user enters credentials. When verification fails (for example a typo in the API token), `init` prints the error with visual separation and offers retry or abort. The provider selection and entry name are kept across retries. Registrar credentials are verified the same way, except when the registrar reuses the DNS provider account (already verified).

When `ListZones` returns zones, `init` shows a multi-select picker so the user can choose which domains to manage. Extra domains can still be added manually. When no zones are available (DNS provider is NONE, or the provider does not support `ListZones`), `init` falls back to the existing free-form prompt.

After zone selection, `init` fetches the existing DNS records for each selected domain via `GetZoneRecords`. Records are sorted with `prettyzone.PrettySort`, formatted with `formatDsl` (reused from `get-zones`), and written into the generated `dnsconfig.js` with a `DefaultTTL()` directive when the most common TTL differs from the default 300. SOA records and apex NS records are filtered out. When the fetch fails for a domain, `init` falls back to the placeholder `A("@", "1.2.3.4")` and prints a warning.

Six flow tests cover the main paths using stubbed provider calls and a `stubAsker`.

A screen recording of the happy path is attached, demonstrating the full flow from provider selection through credential verification, zone picking, record import, and the final generated files.

https://github.com/user-attachments/assets/720b7444-ef4a-456e-ac19-52a4959f7196

The additional steps are described in the updated documentation: https://docs.dnscontrol.org/~/revisions/X3HqhZ6QMtiZKZ5xzBUq/commands/init#what-happens

An example of retrying after a credential error: https://docs.dnscontrol.org/~/revisions/X3HqhZ6QMtiZKZ5xzBUq/commands/init#example-retrying-after-a-credential-error


commands/init.go
```diff
@@ -218,6 +235,192 @@ func confirmAndWrite(asker Asker, args InitArgs, existingCreds map[string]map[st
...
+	return nil, nil
+}
+
+func askDomainsWithZones(asker Asker, availableZones []string) ([]string, error) {
+	if len(availableZones) == 0 {
+		return askDomains(asker)
+	}
+
+	fmt.Println()
+	useList, err := asker.Confirm("Pick domains from the zone list?", true)
```
TomOnTime: A new user might not understand this.  How about asking, "I can query $PROVIDER to see what domains/zones exist in the account. Would you like me to do that and allow you to pick from the list?"

commands/init.go
```diff
@@ -218,6 +235,192 @@ func confirmAndWrite(asker Asker, args InitArgs, existingCreds map[string]map[st
...
+
+	for {
+		more, err := asker.Confirm("Add another domain manually?", false)
+		if err != nil {
+			return nil, err
+		}
+		if !more {
+			break
+		}
+		next, err := askRequiredDomain(asker, "Domain name", "")
```
TomOnTime: If I change my mind, this becomes a dead end.  Suggestion: Allow the user to give an empty response to indicate they don't want to do this.

```
? Add another domain manually? Yes
? Domain name 
A domain name is required.
? Domain name 
A domain name is required.
? Domain name 
```

TomOnTime (commented):
Suggestion:  When prompted for a cred value, indicate the name used in creds.json:

For example, in this situation it would be nice to know that `apiuser`API user (email)` is input for `apiuser` (mostly obvious) and `API key` is `apikey` (could be confused with apitoken).

```
== DNS provider: Cloudflare ==

API settings for Cloudflare: https://dash.cloudflare.com/profile/api-tokens
Cloudflare supports two auth methods: a scoped API token (recommended) or the legacy global API key paired with the account email.
? Which authentication method do you want to use? (required) Global API key + email
? API user (email) (required) tal@whatexit.org
? API key (required) [? for help] *************************************
```

TomOnTime:
```
Compare domains in dnsconfig.js with zones at Cloudflare?
```

I wasn't sure if this was going to compare the list of domains or what is in the domains (the records).  Perhaps be more explicit?  "Would you like to see a comparison of which domains are at $PROVIDER to see if we missed anything?"

Or, why ask at all?  If we have the creds and know that ListZones() works, why not just say, "I will now compare the list of domains at the server against dnsconfig.js so you can see if we are missing any." and do it?

commands/init.go
cafferata: Rephrased to `Select from the 2 zone(s) found at Cloudflare?` which makes clear what is being offered. The zone list was already retrieved during credential verification so the user has context; the prompt now reflects the count and provider name.

commands/init.go
cafferata: Fixed. Replaced `askRequiredDomain` with a plain `Input` that accepts empty values. An empty domain name now breaks out of the loop gracefully.

cafferata:
> Or, why ask at all? If we have the creds and know that ListZones() works, why not just say, "I will now compare the list of domains at the server against dnsconfig.js so you can see if we are missing any." and do it?

Good call. Removed the confirm entirely. It now just prints "Comparing domains in dnsconfig.js with zones at Cloudflare..." and runs the comparison automatically.

cafferata:
> Suggestion: When prompted for a cred value, indicate the name used in creds.json

Done! The prompt now shows the key in brackets, e.g. `API user (email) [apiuser] (required)`. When the label falls back to the key itself (because no label was registered), the bracket is skipped to avoid duplication.

TomOnTime:
I found another dead end:

```
== Registrar: No registrar ==

Use NONE when you do not want DNSControl to manage nameserver delegation (that is, the nameservers for this domain will be managed manually).
? creds.json entry name for this provider none_primary

? Select from the 15 zone(s) found at Cloudflare? No
? First domain name for dnsconfig.js 
A domain name is required.
? First domain name for dnsconfig.js 
A domain name is required.
? First domain name for dnsconfig.js 
A domain name is required.
? First domain name for dnsconfig.js [? for help] 
```

TomOnTime:
I think it would be more natural to name the creds entry (`? creds.json entry name for this provider...`) earlier.  In fact, we might want to include some introductory text such as:

> The next step is to create an entry in creds.json. Each entry stores a set of credentials (usually a API key, token, or PAT) and other information required to authenticate API calls.
>
> The first thing you'll be asked is for the "credkey", which is the name given to each set of credentials. For example, if you have 2 accounts at gandi the two credkeys might be gandi_primary and gandi_testacct.

cafferata:
> I found another dead end

Fixed. Leaving the first domain name empty now returns to the zone selection prompt instead of looping on "A domain name is required." When no zones are available (DNS = NONE), it re-prompts with "At least one domain is required."

cafferata:
> I think it would be more natural to name the creds entry earlier.

Done. The entry name ("credkey") is now the first thing asked after the provider header, with introductory text explaining what a `creds.json` entry is and what the credkey means. The portal URL and credential fields follow after.

Screenshot 2026-05-18 at 16 21 53

TomOnTime:
This is so awesome!  I expect this to really help our adoption rate!

PR #4304, opened 2026-05-18
AXFRDDNS: support unix domain sockets in update-mode/transfer-mode

tomfitzhenry:
My usecase is doing AXFR/DDNS via DNS-over-TLS with PKCS11, and to a server that has a self-signed cert whose [pubkey pin](https://datatracker.ietf.org/doc/html/rfc7858#section-4.2) I want to verify. This could plausibly be natively added to dnscontrol, but it would be a fiddly amount of config in the AXFRDDNS provider.

Unix domain sockets allow connecting via another program, that already exposes lots of the above configuration (stunnel/ghostunnel).

Why not just have that other program listen on a localhost port? That's then exposed to all processes on the host. Unix domain sockets are restricted by file ACLs.

### DIFF
--- a/documentation/provider/axfrddns.md
+++ b/documentation/provider/axfrddns.md
@@ -10,10 +10,12 @@ To use this provider, add an entry to `creds.json` with `TYPE` set to `AXFRDDNS`
 
 Zone transfers and DDNS updates default to TCP when using this provider.
 
-The following two parameters in `creds.json` allow switching to TCP or TCP over TLS.
+The following two parameters in `creds.json` allow switching to TCP, TCP over TLS, or Unix domain sockets.
 
-* `update-mode`: May contain `tcp` (the default), `udp`, or `tcp-tls`.
-* `transfer-mode`: May contain `tcp` (the default), or `tcp-tls`.
+* `update-mode`: May contain `tcp` (the default), `udp`, `tcp-tls`, or `unix`.
+* `transfer-mode`: May contain `tcp` (the default), `tcp-tls`, or `unix`.
+
+When using `unix` mode, `master` and/or `transfer-server` should be set to the path of the Unix domain socket.
 
 ### Authentication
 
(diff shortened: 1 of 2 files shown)


TomOnTime (commented):
Looks good!

Would you please also add a call to providers.RegisterCredsMetadata()?  

This is a new feature, which is covered in Step 4 of https://docs.dnscontrol.org/developer-info/writing-providers

You can see an example in https://github.com/DNSControl/dnscontrol/pull/4297

Thanks!

P.S.  If you wish to do that in a separate PR, please let me know and I'll approve this as-is.

TomOnTime:
We can do the providers.RegisterCredsMetadata part in another PR.  merging.

tomfitzhenry:
Thank you! I looked at providers.RegisterCredsMetadata for AXFRDDNS, and found a surprising amount of flexibility in the provider, so it's not clear what the init wizard should guide the users towards.

I had hoped to find time to get more familiarity with this provider, but that's not yet happened. :/

TomOnTime:
> Thank you! I looked at providers.RegisterCredsMetadata for AXFRDDNS, and found a surprising amount of flexibility in the provider, so it's not clear what the init wizard should guide the users towards.
> 
> I had hoped to find time to get more familiarity with this provider, but that's not yet happened. :/

The wizard only has to set the variables in profiles.json:

```
  "AXFRDDNS": {
    "TYPE": "AXFRDDNS",
    "domain": "$AXFRDDNS_DOMAIN",
    "master": "$AXFRDDNS_MASTER",
    "nameservers": "ns.example.com",
    "transfer-key": "$AXFRDDNS_TRANSFER_KEY",
    "transfer-mode": "$AXFRDDNS_TRANSFER_MODE",
    "update-key": "$AXFRDDNS_UPDATE_KEY",
    "update-mode": "$AXFRDDNS_UPDATE_MODE"
  },
  "AXFRDDNS_DNSSEC": {
    "TYPE": "AXFRDDNS",
    "domain": "$AXFRDDNS_DNSSEC_DOMAIN",
    "master": "$AXFRDDNS_DNSSEC_MASTER",
    "nameservers": "ns.example.com",
    "transfer-key": "$AXFRDDNS_DNSSEC_TRANSFER_KEY",
    "transfer-mode": "$AXFRDDNS_DNSSEC_TRANSFER_MODE",
    "update-key": "$AXFRDDNS_DNSSEC_UPDATE_KEY",
    "update-mode": "$AXFRDDNS_DNSSEC_UPDATE_MODE"
  },

```

PR #4315, opened 2026-05-20
TXTUTIL: accept adjacent quoted chunks without a separator

tresni:
## Summary

Relax `txtDecode` to accept two character-strings appearing back-to-back without an intervening space (e.g. `"foo""bar"`).

## Motivation

Route 53 has been observed returning long TXT records (e.g. DKIM keys) in a form like:

```
"v=DKIM1; k=rsa; p=...Be/""cXaN28e...IDAQAB"
```

with no whitespace between adjacent character-strings. The current parser is strict about RFC1035 zonefile syntax — after a closing quote it requires a space — and returns `txtDecode expected whitespace after close quote`, which the caller in `models/t_parse.go` surfaces as the opaque `invalid TXT record: <contents>`. Every `dnscontrol preview` / `push` against the affected zone then fails on read, before any change can be planned.

This bug surfaces in Route 53 today but the fix lives in shared parser code used by ~a dozen providers (gcloud, cloudflare, hetzner, hetznerv2, gandiv5, dnsimple, hedns, cnr, route53, …), so any provider exhibiting the same quirk benefits.

## Behavior

In `StateWantSpace` (just consumed a closing `"`), a `"` now transitions to `StateQuoted` to start the next chunk, rather than erroring. Any other non-space character still errors. Everything else is unchanged.

## Tests

Added three cases to `TestTxtDecode`:

- `"foo""bar"` → `foobar`
- `"a""b""c"` → `abc`
- `"escaped\"quote""next"` → `escaped"quotenext`

Existing tests continue to pass.

## Test plan

- [x] `go test ./pkg/txtutil/...` passes
- [x] `go test ./models/...` passes
- [x] `go build ./...` succeeds
- [ ] Verified locally against a real Route 53 zone containing a DKIM TXT record that previously failed to parse

### DIFF
--- a/pkg/txtutil/txtcode.go
+++ b/pkg/txtutil/txtcode.go
@@ -134,9 +134,15 @@ func txtDecode(s string) (string, error) {
 			state = StateQuoted
 
 		case StateWantSpace:
-			if c == ' ' {
+			switch c {
+			case ' ':
 				state = StateStart
-			} else {
+			case '"':
+				// Tolerate adjacent quoted character-strings without a
+				// separating space (e.g. `"foo""bar"`). Route 53 has been
+				// observed to return long TXT records in this form.
+				state = StateQuoted
+			default:
 				return "", fmt.Errorf("txtDecode expected whitespace after close quote q(%q)", s)
 			}
 		}
--- a/pkg/txtutil/txtcode.go
+++ b/pkg/txtutil/txtcode.go
@@ -141,6 +141,7 @@ func txtDecode(s string) (string, error) {
 				// Tolerate adjacent quoted character-strings without a
 				// separating space (e.g. `"foo""bar"`). Route 53 has been
 				// observed to return long TXT records in this form.
+				// Whether or not this is valid is questionable but we'll accept it because... Amazon.
 				state = StateQuoted
 			default:
 				return "", fmt.Errorf("txtDecode expected whitespace after close quote q(%q)", s)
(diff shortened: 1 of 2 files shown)


pkg/txtutil/txtcode.go
```diff
@@ -134,9 +134,15 @@ func txtDecode(s string) (string, error) {
...
 		case StateWantSpace:
-			if c == ' ' {
+			switch c {
+			case ' ':
 				state = StateStart
-			} else {
+			case '"':
+				// Tolerate adjacent quoted character-strings without a
+				// separating space (e.g. `"foo""bar"`). Route 53 has been
+				// observed to return long TXT records in this form.
```
TomOnTime, line 143:
```suggestion
				// observed to return long TXT records in this form.
				// Whether or not this is valid is questionable but we'll accept it because... Amazon.
```

pkg/txtutil/txtcode_test.go
```diff
@@ -67,6 +67,11 @@ func TestTxtDecode(t *testing.T) {
 		{`"one" "more" `, []string{`one`, `more`}},
 		// Edge case: unquoted strings are treated as literals to be joined with no space!
 		{`v=spf1 -all`, []string{`v=spf1-all`}},
+		// ROUTE53 has been observed returning long TXT records with adjacent
+		// quoted character-strings and no separator between them.
```
TomOnTime, line 71:
```suggestion
		// quoted character-strings and no separator between them.
		// Whether or not this is valid is questionable but we'll accept it because... Amazon.
```

TomOnTime:
Looks good!  I'm glad someone besides myself was able to understand my weird parser.

Commit those 2 comments and I'll merge.

tresni:
@TomOnTime Committed. If you get a chance to look at #4312 as well I'd much appreciate it :-)

PR #4310, opened 2026-05-19
ORACLE: Enable NO_PURGE and IGNORE*() by upgrading to diff2

fabienmazieres:
I want Oracle provider to use diff2 package,
So that Oracle can reliably support NO_PURGE and IGNORE()

Related to #4309 

I also improved the batch handling of zone updates to be more inline with what I found in other providers, the current method is similar to what is done for Route53

PR #4316, opened 2026-05-20
ORACLE: Upgrade to diff2 and change to RRSet API

fabienmazieres:
I want Oracle provider to use diff2 package,
So that Oracle can reliably support NO_PURGE and IGNORE()

Related to https://github.com/DNSControl/dnscontrol/issues/4309

I also changed the API calls to use the RRSet API instead of Zone.

There does seem to be quite a difference (~40%)  between both methods when running the current test suite :
Tests with RRSet API = ~415-425 sec
Tests with batched zone API (see #4310 ) = ~300-305 sec

I will improve on the zone update method ( #4310 ) with a couple of things I learned when working on this PR, I keep this PR as a record of testing with RRSet and its impact on performance.

### DIFF
--- a/providers/oracle/oracleProvider.go
+++ b/providers/oracle/oracleProvider.go
@@ -4,12 +4,13 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"strings"
 	"time"
 
 	"github.com/DNSControl/dnscontrol/v4/models"
-	"github.com/DNSControl/dnscontrol/v4/pkg/diff"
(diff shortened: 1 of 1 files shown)


fabienmazieres:
There does seem to be quite a difference (~40%) between both methods when running the current test suite :
Tests with RRSet API = ~415-425 sec
Tests with batched zone API (see https://github.com/DNSControl/dnscontrol/pull/4310 ) = ~300-305 sec

I will improve on the zone update method ( https://github.com/DNSControl/dnscontrol/pull/4310 ) with a couple of things I learned when working on this PR, I keep this PR as a record of testing with RRSet and its impact on performance.

PR #4325, opened 2026-05-24
CLOUDFLAREAPI: Reduce integration tests to match the free tier

TomOnTime:
CC @tresni @allixsenos @jzhang-sre 

# Issue

Some tests fail because they use features that are not available in the "free tier" account we use for testing.

# Resolution

Reduce the tests to avoid them.

The following features will no longer be regularly tested:

  * CNAME flattening
  * Tags
  * "Single redirects" that use the "Matches" subcommand

NOTE: Would you like these features to be tested regularly?  We'd gladly accept the donation to enable us to buy a "business plan"-level account.  (Cloudflare, if you're listening, we'd love a partnership!)

### DIFF
--- a/.github/workflows/pr_integration_tests.yml
+++ b/.github/workflows/pr_integration_tests.yml
@@ -259,7 +259,6 @@ jobs:
         run: |
           END=3
           echo "END: Default set to $END"
-          DO_WORKERS=false
 
           # Only for PRs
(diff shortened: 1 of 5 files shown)


allixsenos:
I will reach out to Cloudflare to see about sponsored zones for testing.

PR #4330, opened 2026-05-25
NAMECHEAP: Fix paginate zone listing to work with more than 20 items

### LINKED ISSUE #3824: Possible bug introduced by #3775 with Namecheap provider; Zone "xxx" does not exist ...
**Describe the bug**

I *think* that #3775 may have introduced a bug in the namecheap provider.  With 4.25.0 it works correctly, but with 4.26.0 it says:

*Note I've switch my domain name for 'example.com' in this bug, but I was using my domain name*

```text
 $ dnscontrol preview --creds "./creds.sh" --config "./dnsconfig.js" --domains example.com
CONCURRENTLY checking for 1 zone(s)
SERIALLY checking for 0 zone(s)
Waiting for concurrent checking(s) to complete...DONE
******************** Domain: example.com
1 correction (namecheap)
INFO#1: Zone "example.com" does not exist. Can not create because "namecheap" does not implement ZoneCreator
CONCURRENTLY gathering records of 1 zone(s)
SERIALLY gathering records of 0 zone(s)
Waiting for concurrent gathering(s) to complete...DONE
******************** Domain: tulipgirl.uk
Done. 1 corrections.
completed with errors
```

Whereas with 4.25.0:

```text
# dnscontrol preview --creds "./creds.sh" --config "./dnsconfig.js" --domains example.com
CONCURRENTLY checking for 1 zone(s)
SERIALLY checking for 0 zone(s)
Waiting for concurrent checking(s) to complete...DONE
CONCURRENTLY gathering records of 1 zone(s)
SERIALLY gathering records of 0 zone(s)
Waiting for concurrent gathering(s) to complete...DONE
******************** Domain: example.com
Done. 0 corrections.
```

*Note I've switch my domain name for 'example.com'.*

**Expected behavior**

I expected that 4.26.0 would work like 4.25.0.  The domain 

**DNS Provider**
 
Namecheap

ajkavanagh:
Provides a fix for the issue #3824. If the user has more than 20 domains at Namecheap then dnscontrol only gets the first 20. This patch performs paging such that all the domains can be retrieved.

* Fetch all the pages from Namecheap when listing zones so get-zones and zone existence checks do not miss domains beyond the default first page of 20 results.
* Add regression test covering multi-page domains.getList responses.

Assisted-by: OpenCode gpt-5.4 via OpenCode

Fixes: #3824

### DIFF
--- a/providers/namecheap/namecheapProvider.go
+++ b/providers/namecheap/namecheapProvider.go
@@ -1,10 +1,16 @@
 package namecheap
 
 import (
+	"bytes"
 	"encoding/json"
+	"encoding/xml"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


TomOnTime (approved):

TomOnTime:
Thanks for fixing that!

PR #4339, opened 2026-05-30
BUG: Fix CAA_BUILDER type declarations

thelegy:
# Summary
The type declarations for the `CAA_BUILDER` currently require several fields, that the underlying function does not actually require. In addition to that, the `issue*` fields do not actually take any string, like the type currently describes, but only either a list of strings or `"none"`. This PR updates the type declarations and documentation in accordance with the current implementation.

# Changes
I updated the `documentation/language-reference/domain-modifiers/CAA_BUILDER.md` to include the correct types of the fields. Also updated the description of the parameters to document, that they are optional and what their default values are.

`./bin/generate-all.sh` was used to regenerate the `commands/types/dnscontrol.d.ts`.

# Testing
I ran `go run main.go write-types` to generate the new `types-dnscontrol.d.ts`. I then tested the generated declaration file in my own DNSControl configuration repository and verified that valid `CAA_BUILDER` invocations, such as the following, now type-check correctly:
```ts
CAA_BUILDER({
  issue: ["letsencrypt.org;validationmethods=dns-01"],
  issue_critical: true,
})
```

### DIFF
--- a/commands/types/dnscontrol.d.ts
+++ b/commands/types/dnscontrol.d.ts
@@ -541,21 +541,21 @@ declare function CAA(name: string, tag: "issue" | "issuewild" | "iodef" | "conta
  * ### Parameters
  *
  * * `label:` The label of the CAA record. (Optional. Default: `"@"`)
- * * `iodef:` Report all violation to configured mail address.
+ * * `iodef:` Report all violation to configured mail address. (Optional. Default: `""`)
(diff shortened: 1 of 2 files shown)


TomOnTime (approved):

TomOnTime:
Thanks for fixing this!

PR #4350, opened 2026-06-02
DNSIMPLE: Update to new NS1 & NS3 nameservers

onlyhavecans:
This updates the nameservers for the DNSimple dns provider per the migration documented here.

https://support.dnsimple.com/articles/announcement-ns1-ns3-ip-addresses/

All domains that use the DNSimple DnsProvider will be switched to the up-to-date list of nameservers here, there should be no disruptions for this change.

https://support.dnsimple.com/articles/dnsimple-nameservers/

> The legacy `ns1.dnsimple.com` and `ns3.dnsimple.com` hostnames will continue to resolve after the migration but they are no longer recommended.

### DIFF
--- a/providers/dnsimple/dnsimpleProvider.go
+++ b/providers/dnsimple/dnsimpleProvider.go
@@ -58,9 +58,9 @@ func init() {
 const stateRegistered = "registered"
 
 var defaultNameServerNames = []string{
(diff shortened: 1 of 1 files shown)


TomOnTime (approved):

TomOnTime:
Thanks for the update!

PR #4359, opened 2026-06-03
HEDNS: Further improve authentication performance

rblenkinsopp:
## Summary

Builds on the caching/authentication work in #4348.

When the provider authenticates (or resumes a session), the page HE returns is already the account dashboard, which contains the domain listing. Previously the provider discarded that response and issued a second `GET /` to read the domain table.

This reuses the authentication response directly, removing one request.

## Testing

Tested against a live HE account; both a cold login (no session) and warm session reuse pass the HEDNS integration suite.


TomOnTime (approved):
Thanks!

PR #4225, opened 2026-05-05
DOCS: Unwrap hard-wrapped paragraphs in GitBook documentation

cafferata:
Reflow prose in 15 documentation pages so each paragraph is a single line instead of being hard-wrapped at ~70 characters. Hard-wrapped Markdown caused awkward soft-wrap rendering on docs.dnscontrol.org. Code blocks, front matter, lists, blockquotes, tables, HTML, and {% hint %} / {% code %} blocks are left untouched. No content changes.

**Before**

Screenshot 2026-05-05 at 08 31 55

<https://docs.dnscontrol.org/getting-started/migrating>

**After**

Screenshot 2026-05-05 at 08 43 08

<https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/getting-started/migrating>

| Before (live) | After (revisie) |
| --- | --- |
| https://docs.dnscontrol.org/getting-started/migrating | https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/getting-started/migrating |
| https://docs.dnscontrol.org/provider/hetzner | https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/provider/hetzner |
| https://docs.dnscontrol.org/advanced-features/code-tricks | https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/advanced-features/code-tricks |
| https://docs.dnscontrol.org/advanced-features/concurrency-verified | https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/advanced-features/concurrency-verified |
| https://docs.dnscontrol.org/advanced-features/opinions | https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/advanced-features/opinions |
| https://docs.dnscontrol.org/advanced-features/testing-txt-records | https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/advanced-features/testing-txt-records |
| https://docs.dnscontrol.org/advanced-features/json-reports | https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/advanced-features/json-reports |
| https://docs.dnscontrol.org/commands/colors | https://docs.dnscontrol.org/~/revisions/Uy3v06HuP9Hvf4EPf9c7/commands/colors |
[...]

PR #4368, opened 2026-06-07
DOCS: Explain Bunny CNAME/ALIAS behavior.

jfexyz:
Just an addition to explain that while Bunny DNS allows root CNAMEs, here in DNSControl, you should use `ALIAS` and the CNAME will result. This was initially unclear to me when setting up DNSControl.

### DIFF
--- a/documentation/provider/bunnydns.md
+++ b/documentation/provider/bunnydns.md
@@ -37,6 +37,19 @@ export BUNNY_DNS_API_KEY=XXXXXXXXX
 
 This provider does not recognize any special metadata fields unique to Bunny DNS.
 
+## Limitations
+
+### Records
+
+#### CNAME / ALIAS
+
+Bunny supports adding CNAME records for the root domain (`@`), but within
+DNSControl the `ALIAS` record should be used.
+
+```javascript
+D('example.com', REG_NONE, DnsProvider(DSP_BUNNY_DNS), ALIAS('@', 'example2.com'));
+```
+
 ## Usage
 
 An example configuration:
--- a/documentation/provider/bunnydns.md
+++ b/documentation/provider/bunnydns.md
@@ -1,19 +1,22 @@
 # Configuration
 
-To use this provider, add an entry to `creds.json` with `TYPE` set to `BUNNY_DNS` along with
-your [Bunny API Key](https://dash.bunny.net/account/settings).
+To use this provider, add an entry to `creds.json` with `TYPE` set to
+`BUNNY_DNS` along with your
+[Bunny API Key](https://dash.bunny.net/account/settings).
 
 Example:
 
 {% code title="creds.json" %}
+
 ```json
 {
-  "bunny_dns": {
-    "TYPE": "BUNNY_DNS",
-    "api_key": "your-bunny-api-key"
-  }
+    "bunny_dns": {
+        "TYPE": "BUNNY_DNS",
+        "api_key": "your-bunny-api-key"
+    }
 }
 ```
+
 {% endcode %}
 
 You can also use environment variables:
@@ -23,19 +26,22 @@ export BUNNY_DNS_API_KEY=XXXXXXXXX
 ```
 
 {% code title="creds.json" %}
+
 ```json
 {
-  "bunny_dns": {
-    "TYPE": "BUNNY_DNS",
-    "api_key": "$BUNNY_DNS_API_KEY"
-  }
+    "bunny_dns": {
+        "TYPE": "BUNNY_DNS",
+        "api_key": "$BUNNY_DNS_API_KEY"
+    }
 }
 ```
+
 {% endcode %}
 
 ## Metadata
 
-This provider does not recognize any special metadata fields unique to Bunny DNS.
+This provider does not recognize any special metadata fields unique to Bunny
+DNS.
 
 ## Limitations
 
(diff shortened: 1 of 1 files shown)


documentation/provider/bunnydns.md
```diff
@@ -37,6 +37,19 @@ export BUNNY_DNS_API_KEY=XXXXXXXXX
...
+
+### Records
+
+#### CNAME / ALIAS
+
+Bunny supports adding CNAME records for the root domain (`@`), but within
+DNSControl the `ALIAS` record should be used.
+
+```javascript
+D('example.com', REG_NONE, DnsProvider(DSP_BUNNY_DNS), ALIAS('@', 'example2.com'));
```
TomOnTime:
```suggestion
D('example.com', REG_NONE, DnsProvider(DSP_BUNNY_DNS),
    ALIAS('@', 'example2.com'),
);
```

TomOnTime (changes requested):

documentation/provider/bunnydns.md
jfexyz: The change here is just formatting? Sure, can do, and I see you supposedly use prettier in the repo, but when I initially created the PR so much of the file was reformatted that I undid all the formatting changes and disabled prettier...

documentation/provider/bunnydns.md
```diff
@@ -23,70 +26,99 @@ export BUNNY_DNS_API_KEY=XXXXXXXXX
...
+DNS.
+
+## Limitations
+
+### Records
+
+#### CNAME / ALIAS
+
+Bunny supports adding CNAME records for the root domain (`@`), but within
+DNSControl the `ALIAS` record should be used.
```
TomOnTime, line 53: I think this needs a little more explanation.  Something like:

Bunny supports adding CNAME records for the root domain (`@`). Use the `ALIAS` record type and DNSControl will generate a CNAME as appropriate.

documentation/provider/bunnydns.md
```diff
@@ -23,70 +26,99 @@ export BUNNY_DNS_API_KEY=XXXXXXXXX
...
 ```javascript
-var REG_NONE = NewRegistrar("none");
-var DSP_BUNNY_DNS = NewDnsProvider("bunny_dns");
+var REG_NONE = NewRegistrar('none');
+var DSP_BUNNY_DNS = NewDnsProvider('bunny_dns');
 
-D("example.com", REG_NONE, DnsProvider(DSP_BUNNY_DNS),
-    A("test", "1.2.3.4"),
-);
+D('example.com', REG_NONE, DnsProvider(DSP_BUNNY_DNS), A('test', '1.2.3.4'));
```
TomOnTime, line 71: JavaScript code should use double quotes (") for strings, not single quotes ('). 

See https://docs.dnscontrol.org/developer-info/styleguide-doc

documentation/provider/bunnydns.md
```diff
@@ -23,70 +26,99 @@ export BUNNY_DNS_API_KEY=XXXXXXXXX
...
     BUNNY_DNS_RDR("@", "https://foo.bar"),
 ```
+
 {% endcode %}
 
 ### Pull Zone (PZ)
 
-You can configure Bunny's Pull Zone type with `BUNNY_DNS_PZ`. The target is the Pull Zone ID:
+You can configure Bunny's Pull Zone type with `BUNNY_DNS_PZ`. The target is the
+Pull Zone ID:
```
TomOnTime, line 109:
```suggestion
You can configure Bunny's Pull Zone type with `BUNNY_DNS_PZ`. The target is the Pull Zone ID:
```

CC @cafferata Is it ok that these lines are being broken?  I believe you recently unbroke lines.

TomOnTime (changes requested):

documentation/provider/bunnydns.md
jfexyz: Ok, y'all have a `.prettierrc` file that says otherwise. Which is why this PR is so huge... you have your repo set up to use Prettier, but don't actually seem to be using it properly. Hence me having to spend time disabling and working around this. I'll probably just drop wanting to get this merged, tbh. When y'all figure out your systems set up, I'm happy to take another crack at it.

TomOnTime: Ah, that's certainly a contradiction!  We use prettier for pkg/js/helpers.js and nothing else.  We should update the style guide (or maybe change pkg/js/helpers.js to be the same as the docs?). CC @cafferata for his opinion.

jfexyz: Or use a .prettierignore file to ignore the docs. Or ideally just use Prettier for the docs (you can have a separate .prettierrc in the docs dir with different settings) and eliminate all the bikeshedding...

To be clear, the huge problem right now is that because you have that .prettierrc file, Zed and VS Code and any other editor set up to respect them will reformat everything in a file when you make edits.

TomOnTime: (thank you for your patience. I was traveling.)

Ah, that's an interesting point about Zed and VS Code.  I thought it wasn't a big issue because bin/generate-all.sh is specific to which file gets run.

I like your proposed solution of using .prettierignore.  Would you add that to this PR?

TomOnTime:
cafferata let me know that he's busy so I won't wait for his reply.  Please resolve conflicts and I'll merge this. We can fix any issues later. Thank you for your patience!

documentation/provider/bunnydns.md
cafferata: Sorry for the late reply, Tom. Yes, please keep paragraphs on a single line. In #4225 I unwrapped hard-wrapped paragraphs because GitBook showed the line breaks on docs.dnscontrol.org. Pages that still have wrapped lines, such as https://docs.dnscontrol.org/provider/scaleway, now render fine, so GitBook seems to handle it today. Single-line paragraphs keep the documentation consistent and avoid the whole-file reformatting @jfexyz ran into.

cafferata:
@jfexyz sorry for not replying when Tom asked for my opinion here. I was very busy at the time, so unfortunately I couldn't spend time on open source work. You were right: the root `.prettierrc` (`singleQuote: true`) is only meant for `pkg/js/helpers.js`, but editors that respect it reformat every Markdown file, while the documentation style guide asks for double quotes. Since the fork of this pull request is gone, it can't be reopened. Would you like to open a new one with just the CNAME/ALIAS section? The explanation is still missing from https://docs.dnscontrol.org/provider/bunnydns. If you'd rather not, I'm happy to pick it up and credit you.

Issue #4400, opened 2026-06-28
BUG: CNR integration tests failing due to ALIAS/DNSSEC incompatibility


TomOnTime:
**Describe the bug**

The integration tests are failing:

https://github.com/DNSControl/dnscontrol/actions/runs/28322093234/job/83905681568

```
=== Failed
=== FAIL: integrationTest TestDNSProviders/***-cnr.com/52:ALIAS_on_apex:ALIAS_at_root (0.90s)
    helpers_integration_test.go:236: 
        GENERATE_ZONE: ***-cnr.com
        + CREATE ALIAS ***-cnr.com foo.com. ttl=300
        
    helpers_integration_test.go:241: Error while updating zone "***-cnr.com". [541 Invalid attribute value; ALIAS record cannot be added to a signed zone. Please disable signing first.]
        --- FAIL: TestDNSProviders/***-cnr.com/52:ALIAS_on_apex:ALIAS_at_root (0.90s)

=== FAIL: integrationTest TestDNSProviders/***-cnr.com/53:ALIAS_to_nonfqdn:ALIAS_at_root (0.70s)
    helpers_integration_test.go:236: 
        GENERATE_ZONE: ***-cnr.com
        + CREATE A foo.***-cnr.com 1.2.3.4 ttl=300
        + CREATE ALIAS ***-cnr.com foo.***-cnr.com. ttl=300
        
    helpers_integration_test.go:241: Error while updating zone "***-cnr.com". [541 Invalid attribute value; ALIAS record cannot be added to a signed zone. Please disable signing first.]
        --- FAIL: TestDNSProviders/***-cnr.com/53:ALIAS_to_nonfqdn:ALIAS_at_root (0.70s)

=== FAIL: integrationTest TestDNSProviders/***-cnr.com/54:ALIAS_on_subdomain:ALIAS_at_subdomain (0.69s)
    helpers_integration_test.go:236: 
        GENERATE_ZONE: ***-cnr.com
        + CREATE ALIAS test.***-cnr.com foo.com. ttl=300
        
    helpers_integration_test.go:241: Error while updating zone "***-cnr.com". [541 Invalid attribute value; ALIAS record cannot be added to a signed zone. Please disable signing first.]
        --- FAIL: TestDNSProviders/***-cnr.com/54:ALIAS_on_subdomain:ALIAS_at_subdomain (0.69s)
```

**To Reproduce**
See above

**Expected behavior**
Tests should pass.

**DNS Provider**
[...]

PR #4401, opened 2026-06-29
CNR: Skip ALIAS on DNSSEC-signed zones, add SMIMEA + AUTODNSSEC support, and enhance SVCB handling

AsifNawaz-cnic:
Thank you @TomOnTime  for bringing this into our attention.

CNR API does not accept ALIAS records when a dns zone is DNSSEC-signed.

Fixes the issue #4400


TomOnTime (approved):
Great work!  Thanks for the fast PR!

PR #4417, opened 2026-07-03
TEST: Add validations and tests for SOA records to reject multiple SOAs

istenrot:
This PR adds validations to reject multipel SOA records. Also unit tests for new function checkMultipleSOAs added.

I did run integration tests only for Route 53 but there isn't anything Route 53 provider spesific.


TomOnTime (approved):
Perfect!

PR #4419, opened 2026-07-04
BUGFIX: Remove PORKBUN_URLFWD since it's an alias to URL and URL301

TomOnTime:
# Issue

According to https://github.com/DNSControl/dnscontrol/pull/4414#issuecomment-4880495022 

We can remove PORKBUN_URLFWD since it's an alias to URL and URL301 (as of https://github.com/DNSControl/dnscontrol/pull/3951)

# Resolution

Remove PORKBUN_URLFWD and related tests.

CC [imlonghao](https://github.com/imlonghao) (Please let me know if this works. Also... I've removed everything related to urlfwd.  Is that what you were suggesting or did I go too far?)

By the way... we could avoid this kind of breakage if I had a test account for Porkbun.  That way the GHA tests would test porkbun with every PR.  Is that something you can arrange?  It can be a free/test account, or just an API key locked to a domain/zone that you've set up.  Details are here: https://docs.dnscontrol.org/developer-info/byo-secrets  Thanks for considering that!


imlonghao:
> Is that what you were suggesting or did I go too far?

Yes, this is what I want to do **yesterday**.

But today I noticed porkbun has third type of url forwarding named `masked`, this type is basically a `<iframe>`, not 301 nor 302 forwarding.

This type can be only created in web dashboard, no description in API docs at all, but listing in getUrlForwarding API response. I'm asking support about it.

If we want to support this type I think we need to keep the PORKBUN_URLFWD. Sorry for the back and forth.

> GHA tests would test porkbun with every PR

It will need a domain register with them, so not now, just ping me when test needed.

TomOnTime:
I'm not sure I understand.  Would you like to make a PR instead?

imlonghao:
I will make a PR in a few days.

PR #4424, opened 2026-07-05
ALIDNS: audit punycoded CNAME targets

bytemain:
## Summary

Fixes the AliDNS audit path for internationalized CNAME targets that have already been converted to ACE/punycode before `AuditRecords` runs.

AliDNS supports ASCII and Chinese IDN values, but rejects non-Chinese IDN targets such as `ööö.com.`. During integration testing, that target reaches the provider audit as `xn--ndaaa.com.`, so the existing ASCII/Chinese check sees only ASCII and allows it through. The AliDNS API then rejects the create request with `SubDomainInvalid.Value`.

This changes `targetConstraint` to decode the target with `idna.ToUnicode` before applying the existing AliDNS string validation. That preserves support for ASCII and Chinese IDN targets while rejecting non-Chinese IDN targets during audit, so the integration test is skipped as provider-unsupported instead of failing at the API call.

related to: https://github.com/DNSControl/dnscontrol/issues/4421

## Tests

```bash
go test ./providers/alidns
go test ./integrationTest -run TestNonExistent -count=0
```

### DIFF
--- a/providers/alidns/auditrecords.go
+++ b/providers/alidns/auditrecords.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/DNSControl/dnscontrol/v4/models"
 	"github.com/DNSControl/dnscontrol/v4/pkg/rejectif"
+	"golang.org/x/net/idna"
 )
 
 // isValidAliDNSString checks if a string contains only ASCII or Chinese characters.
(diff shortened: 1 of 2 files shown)


TomOnTime:
Thanks for fixing this!

PR #4427, opened 2026-07-06
v5 - add non-porkbun compatibility to url301

willpower232:
https://github.com/DNSControl/dnscontrol/issues/4421#issuecomment-4887078054

https://github.com/DNSControl/dnscontrol/issues/4421#issuecomment-4890438490

not sure if there is a neat way of detecting the provider or not, probably preferable to making a second kind of URL301 type


willpower232:
I confirm this change makes v5 great from a namecheap point of view

TomOnTime:
Thanks for the quick response!

willpower232:
@TomOnTime fyi I don't think I can check off namecheap on the list so presumably you'll catch up on that with all the others saying it is okay at some point :sweat_smile:

Issue #4421, opened 2026-07-05
Provider maintainers: Please test this release!


TomOnTime:
# Issue

See https://github.com/orgs/DNSControl/discussions/4415 for more details.

https://github.com/DNSControl/dnscontrol/pull/4414 (branch `release_candidate_v5`) needs extra testing because it changes a lot. There should be no user-visible changes. However a lot of code was refactored.

Therefore, I'm requesting that each provider maintainer test this branch by July 18, 2026 (that's 2 weeks). If all goes well, this will become DNSControl v5.0 shortly after.

- [ ] providers/adguardhome (@ishanjain28)
- [x] providers/akamaiedgedns (@edglynes)
- [x] providers/alidns (@bytemain)
- [ ] providers/autodns (@arnoschoon)
- [ ] providers/axfrddns (@hnrgrgr)
- [x] providers/azuredns (@vatsalyagoel)
- [x] providers/azureprivatedns (@matthewmgamble)
- [x] providers/bind (@TomOnTime)
- [ ] providers/bunnydns (@ppmathis)
- [x] providers/cloudflare (@tresni)
- [ ] providers/cloudns (@pragmaton)
- [x] providers/cnr (@AsifNawaz-cnic)
- [ ] providers/cscglobal (@mikenz)
- [ ] providers/desec (@D3luxee)
- [x] providers/digitalocean (@chicks-net)
- [ ] providers/dnscale (@dnscale-ops)
- [x] providers/dnsimple (@onlyhavecans)
- [ ] providers/dnsmadeeasy (@vojtad)
- [ ] providers/domainnameshop (@SimenBai)
- [ ] providers/exoscale (@Giza)
- [x] providers/fortigate (@KlettIT)
- [x] providers/gandiv5 (@TomOnTime)
- [x] providers/gcloud (@riyadhalnur)
- [x] providers/gcore (@xddxdd)
- [ ] providers/gidinet (@zupolgec)
- [x] providers/hedns (@rblenkinsopp)
- [x] providers/hetzner (@das7pad)
- [ ] providers/hetznerv2 (@das7pad)
- [ ] providers/hostingde (@juliusrickert)
- [x] providers/huaweicloud (@huihuimoe)
- [x] providers/infomaniak (@jbelien)
- [x] providers/inwx (@patschi)
- [ ] providers/joker (@atrull)
- [ ] providers/linode (@koesie10)
- [ ] providers/loopia (@systemcrash)
[...]

PR #4451, opened 2026-07-08
GIDINET: v5 test fixes and modernize record conversion

zupolgec:
Tested GIDINET against `release_candidate_v5` (#4421). Two fixes so the
integration test passes, plus a small modernization.

- Reject null SRV targets in `AuditRecords` — the Gidinet API errors on them
  (code 4), so the `43:SRV Null Target` case now skips instead of failing.
- Exclude GIDINET from the `NS only APEX` test — the provider already refuses to
  modify apex NS records, so the test expected changes that never come.
- `toRecordConfig`: use `SetTargetMX`/`SetTargetSRV` instead of setting
  the fields by hand (v5 cookbook). No behaviour change.

Integration test green; unit tests pass.


TomOnTime:
Thanks!

PR #4452, opened 2026-07-08
UNIFI: new API compatibility fixes + tests

zupolgec:
Tested UNIFI against `release_candidate_v5` (#4421) on a controller running the
new API (Network 10.x). The new API rejects several things the provider was
sending; fixes below. Validated with the integration test locally and against
the same controller via UniFi cloud (console proxy).

- `ttlSeconds` is only accepted for A/AAAA/CNAME — omit it for MX/TXT/SRV (the
  API returns `Unknown request body property '$.ttlSeconds'`).
- SRV: the new API splits the label into `service`/`protocol`/`domain`
  (e.g. `_sip`/`_tcp`/`example.com`). Added those fields, split on write and
  rebuild on read.
- MX/TXT/SRV have no per-record TTL, so normalize them to a fixed value to avoid
  a perpetual diff. UNIFI is excluded from the SRV-TTL test group.
- Reject TXT with interior double quotes (API returns `incorrectly quoted
  value`).
- API auto-detection cached negative results, so one transient probe failure
  marked both APIs unavailable for the rest of the session. Cache only positive
  results.
- Added unit tests for the record converters and switched to
  `SetTargetMX`/`SetTargetSRV` (cookbook).

Integration test green on local and cloud.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -892,6 +892,7 @@ func makeTests() []*TestGroup {
 		// https://github.com/DNSControl/dnscontrol/issues/2066
 		testgroup("SRV",
 			requires(providers.CanUseSRV),
+			not("UNIFI"), // UniFi has no per-record TTL for SRV records.
 			tc("Create SRV333", ttl(srv("_sip._tcp", 5, 6, 7, "foo.com."), 333)),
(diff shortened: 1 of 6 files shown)


TomOnTime (approved):
Thanks for the update!

PR #4453, opened 2026-07-08
OVH: fix ovh integration tests for release_candidate_v5

masterzen:
## What?

This makes the integration tests pass for the OVH provider on top of the release_candidate_v5 work.

## Why?

TXT record parsing seems to have changed, and TXT records where quoted again when parsed.


masterzen:
While this fixes some of the integration tests, some are still failing:
```
=== RUN   TestDNSProviders/dnscontroltest.ovh/27:complex_TXT:a_0-byte_TXT
    helpers_integration_test.go:255: OVHcloud API error (status code 400): Client::BadRequest: "{'target': ['Invalid input.']}" (X-OVH-Query-Id: EU.ext-4.6a4ea284.3587060.75b31d3c7614bc9e33a63aea301fdb8e)
=== RUN   TestDNSProviders/dnscontroltest.ovh/97:structured_TXT_as_native_records:Create_native_OVH_records
    helpers_integration_test.go:255: OVHcloud API error (status code 400): Client::BadRequest: "Invalid subdomain: dkim" (X-OVH-Query-Id: EU.ext-3.6a4ea343.1196486.1857a3461e2a5378c77f82354f6a3541)
```

I need to have a deeper look.

TomOnTime:
FYI: If miekg accepts https://codeberg.org/miekg/dns/pulls/916, the way TXTs are handled might change for the better.

TomOnTime (approved):

TomOnTime:
Looks good!

It's marked draft.  Ready for merge?

masterzen:
> Looks good!
> 
> It's marked draft. Ready for merge?

Not yet, I still need to find a way to fix the round-trip problem outlined [here](https://github.com/DNSControl/dnscontrol/issues/4421#issuecomment-4924340708).

Basically, the integration test 97 writes `v=DMARC1; p=none; rua=mailto:dmarc@yourdomain.com` but when read again and passed into `PopulateFromStringFunc` with `txtutil.ParseQuoted`, removes all spaces.

I'm still not sure how to properly handle this, and have all TXT tests to pass.

OVH treats TXT as if they were double-quoted strings in a zone file, that's why I'm surprised txtutil.ParseQuoted removes all spaces. I don't think whatever we were doing in `main` was doing this.

TomOnTime:
Try this:

Add this to `providers/ovh/ovhProvider.go`:

```
// ovhParseTXT decodes a TXT target as received from the OVH API.
// We must join the strings ourselves. OVH is "smart" and joins with spaces for DMARC/DKIM/SPF,
// and joins with "" (null string) for TXT records. We don't want it to be so smart, so we do the join ourselves.
func ovhParseTXT(s string) (string, error) {
      return strings.Join(models.ParseQuotedTxt(s), ""), nil
}
```

and the call to PopulateFromStringFunc becomes:

```
if err := rec.PopulateFromStringFunc(rtype, r.Target, origin, ovhParseTXT); err != nil {
```

TomOnTime:
Please update the branch to  to release_candidate_v5 and re-run the tests. There's a chance something I changed fixed this.

masterzen:
> Please update the branch to to release_candidate_v5 and re-run the tests. There's a chance something I changed fixed this.

Sorry, couldn't come back to the problem before today.

I'm still seeing test 97 TXT records failures (ie removal of spaces inside a TXT target).
I'm wondering if the provider shouldn't pretend to dnscontrol that the string is indeed quoted, because `txtDecode` assumes `foo bar` is not the same as `"foo bar"`.

> Try this:
> 
> Add this to `providers/ovh/ovhProvider.go`:
> 
> ```
> // ovhParseTXT decodes a TXT target as received from the OVH API.
> // We must join the strings ourselves. OVH is "smart" and joins with spaces for DMARC/DKIM/SPF,
> // and joins with "" (null string) for TXT records. We don't want it to be so smart, so we do the join ourselves.
> func ovhParseTXT(s string) (string, error) {
>       return strings.Join(models.ParseQuotedTxt(s), ""), nil
> }
> ```
> 
> and the call to PopulateFromStringFunc becomes:
> 
> ```
> if err := rec.PopulateFromStringFunc(rtype, r.Target, origin, ovhParseTXT); err != nil {
> ```

Well, this solves 97, but, now 27 and 28 fail:
```
        --- FAIL: TestDNSProviders/dnscontroltest.ovh/27:complex_TXT:TXT_with_1_dq-1interior (0.70s)
        --- FAIL: TestDNSProviders/dnscontroltest.ovh/28:TXT_backslashes:TXT_with_backslashs (0.66s)
```

TomOnTime:
@masterzen

Eventually we'll be upgrading to the new NewRecordConfig factories... we might as well do it now. Those new functions handle all this stuff a lot better.

I did the work (but can't test it) in `release_candidate_v5_ovh_try2` https://github.com/DNSControl/dnscontrol/pull/4503

Feel free to clone that and work with it as a base.

It should work for most tests.  Because test 27/28 fail, I updated auditrecords.go to skip those.

Once all that works, I recommend a few experiments:

1. Try removing the case for TXT in nativeToRecord. Maybe the default case is sufficient.
2. Try removing (one at a time) the rejectif statements in auditrecords.go.  Maybe the new code handles them better now.

Hope that helps!

masterzen:
Closing in favor of https://github.com/DNSControl/dnscontrol/pull/4636

PR #4456, opened 2026-07-10
HETZNER: remove defunct provider

das7pad:
The legacy Hetzner provider is now officially defunct and we can remove it.

``` 
$ curl -i https://dns.hetzner.com/api/v1/zones
HTTP/2 301 
location: https://console.hetzner.com/
```


TomOnTime:
yay! thanks for cleaning this up!

PR #4469, opened 2026-07-13
DNSMADEEASY: fix TXT ≥256-byte round-trip and ALIAS/ANAME diff loop; document sandbox testing

vojtad:
Fixes three DNS Made Easy provider bugs surfaced by the integration suite, plus
docs for sandbox/integration testing.

Fixes:
- TXT values ≥256 bytes: send a single unchunked quoted value (EncodeSingle)
  instead of RFC1035 chunks, which DME stores with a literal joining space and
  never round-trips → perpetual spurious MODIFY.
- ALIAS records: move the ALIAS→ANAME rename into the write serializer instead
  of mutating desired records pre-diff, so the diff compares ALIAS vs ALIAS and
  converges (previously looped DELETE ALIAS + CREATE ANAME).
- ALIAS-to-nonfqdn integration test: exclude DNSMADEEASY, since DME validates
  ANAME target resolvability at create time and an unpublished in-zone target
  can't resolve.

Docs:
- Document the sandbox environment (control panel, credentials, `sandbox` creds
  key), add an integration-testing section with command and raised timeout, and
  clarify DNSMADEEASY_DEBUG_HTTP usage.

Verified against the DME sandbox: full integration suite passes. **Not tested in production since I am not using DNS Made Easy anymore.**

### DIFF
--- a/documentation/provider/dnsmadeeasy.md
+++ b/documentation/provider/dnsmadeeasy.md
@@ -6,6 +6,7 @@ along with your `api_key` and `secret_key`. More info about authentication can b
 Example:
 
 {% code title="creds.json" %}
+
 ```json
 {
   "dnsmadeeasy": {
@@ -15,6 +16,7 @@ Example:
   }
 }
 ```
+
 {% endcode %}
 
 ## Records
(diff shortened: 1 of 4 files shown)


TomOnTime (approved):

TomOnTime:
Thanks for fixing those!

Do me a favor and re-run the tests on the release_candidate_v5 branch. I've updated code related to txt records and I want to make sure I haven't broken your txt fix.  Thanks!

vojtad:
All passing on the release_candidate_v5 branch.

PR #4501, opened 2026-07-18
Modernize PowerDNS provider

TomOnTime:
## What changed

- pass `DomainConfig` into PowerDNS record conversion
- replace legacy record construction and parsing with v3 factories
- preserve provider-specific SVCB/HTTPS auto hints while using a v3 base record
- serialize standard outbound records from v3 RDATA
- modernize the existing TXT, LUA, auto-hint, and audit tests

## Validation

- `../../bin/is_modern.sh` from `providers/powerdns` (clean)
- `go test ./providers/powerdns`
- `env GOCACHE=/tmp/dnscontrol-go-cache go test ./...`
- `env GOCACHE=/tmp/dnscontrol-go-cache GOLANGCI_LINT_CACHE=/tmp/dnscontrol-golangci-cache bin/generate-all.sh`
- `git diff --check`


providers/powerdns/convert.go
```diff
@@ -5,63 +5,67 @@ import (
...
-
-	rc := &models.RecordConfig{
-		TTL:      uint32(ttl),
-		Original: r,
-		Type:     rtype,
+func toRecordConfig(dc *models.DomainConfig, r zones.Record, ttl int, name string, rtype string) (*models.RecordConfig, error) {
+	label := dc.LabelFromShort(name)
+	if strings.HasSuffix(name, ".") {
+		label = dc.LabelFromFQDNWithDot(name)
 	}
```
jpbede: Shouldn't this be something like the bellow otherwise this would always print a error?
At least in my tests this always printed a error.


```suggestion
    var label string 
	if strings.HasSuffix(name, ".") {
		label = dc.LabelFromFQDNWithDot(name)
     } else {
     	label = dc.LabelFromShort(name)
     }
```

TomOnTime: Good catch!

Actually, since the API is always going to give a string that is "example.com." (FQDN + "."), we can just do:

```
label := dc.LabelFromFQDNWithDot(name)
```

TomOnTime: I've pushed a new commit. Let me know if that works.

jpbede (approved):
Thanks a lot 👍 Saw some failing tests that I'm going to fix soon

TomOnTime:
I merged everything to release_candidate_v5 (because I'm an optimist).  Please pull that and re-run the integration tests to be sure the merge was successful.  Thanks!

PR #4531, opened 2026-07-21
NS1: enable init support

costasd:


TomOnTime (approved):
w00t!  thanks for adding this!

PR #4534, opened 2026-07-21
CHORE: fix a tiny bashism

costasd:
/bin/sh isn't symlinked to /bin/bash anymore in various distributions, ensure we don't use any bash-specific syntax in this script.


TomOnTime (approved):

TomOnTime:
Thanks!

PR #4562, opened 2026-07-22
luadns: Modernize record conversion

TomOnTime:
Dear @riku22. I could use your help. I'm doing upgrades to the code base and I have no way to test your provider. Please run the integration test and let me know the result. How to run integration tests is here: https://docs.dnscontrol.org/developer-info/integration-tests

## Summary

- replace legacy record construction with current domain-aware factories
- replace combined legacy getters with typed RDATA output
- preserve raw TXT and default nameserver behavior
- add focused conversion regression tests

## Validation

- `../../bin/is_modern.sh` (Steps 1-7 clean)
- `go test ./providers/luadns`
- `bin/generate-all.sh`
- `go test ./...`


riku22 (approved):
I verified that the integration tests pass.

I think there is no problem with this.

TomOnTime:
Thanks for running the tests!

PR #4566, opened 2026-07-22
namecheap: Modernize record conversion

TomOnTime:
Dear @willpower232. I could use your help. I'm doing upgrades to the code base and I have no way to test your provider. Please run the integration test and let me know the result. How to run integration tests is here: https://docs.dnscontrol.org/developer-info/integration-tests

## Summary

- build downloaded DNS and SRV records with modern record factories
- use the domain configuration for label normalization
- replace the obsolete combined-target getter

## Validation

- `../../bin/is_modern.sh` (Steps 1-7 clean)
- `go test ./providers/namecheap`
- `bin/generate-all.sh`
- `go test ./...`


willpower232:
initially looks good but now I'm smashing into their rate ilmits, I'll try again tomorrow :pray:

willpower232:
Okay most of the tests pass but we do have a few random fails which have this message in the output

```
helpers_integration_test.go:272: dns: bad TXT Txt: "TestDNSProviders" at line: 1:17
helpers_integration_test.go:235: runTests: dns: bad TXT Txt: "TestDNSProviders" at line: 1:17
helpers_integration_test.go:235: runTests: dns: bad TXT Txt: "TestDNSProviders" at line: 1:17
```

I presume that is a line reference to a different file but I don't know which one.

I added an extra sleep to try and avoid a minutely rate limit which might have helped me get to the end

image

Attached full outputs below, I did them in three parts spread out. I think the panic is related to sleeping for the rate limits and I gave up on one of them because it was just rate limited and then set a timer for an hour to try again.

[output.txt](https://github.com/user-attachments/files/30342905/output.txt)
[output copy.txt](https://github.com/user-attachments/files/30342906/output.copy.txt)
[output copy 2.txt](https://github.com/user-attachments/files/30342907/output.copy.2.txt)

Let me know when to have another look!

TomOnTime:
> Okay most of the tests pass but we do have a few random fails which have this message in the output
> 
> ```
> helpers_integration_test.go:272: dns: bad TXT Txt: "TestDNSProviders" at line: 1:17
> helpers_integration_test.go:235: runTests: dns: bad TXT Txt: "TestDNSProviders" at line: 1:17
> helpers_integration_test.go:235: runTests: dns: bad TXT Txt: "TestDNSProviders" at line: 1:17
> ```

It's converting a string so "line 1:17" means 17 bytes into the string. Not too useful.

I've updated the code to provide a more useful error message.

> I added an extra sleep to try and avoid a minutely rate limit which might have helped me get to the end

"go test" times out after 10 minutes.  The panic is indicating that the test framework is killing the process.  Add `-timeout 1h`

```
cd integrationTest
go test -timeout 1h -v -args -verbose -profile NAMECHEAP
```

I've improved TXT handling in general and made some other improvements.  I'll have one more commit in an hour or so.

willpower232:
every test now fails with 
```
helpers_integration_test.go:235: runTests: NewData(16, "TestDNSProviders was successful!", "willpower232testsdnscontrol.com.") failed: dns: bad TXT Txt: "TestDNSProviders" at line: 1:17
```

TomOnTime:
> every test now fails with
> 
> ```
> helpers_integration_test.go:235: runTests: NewData(16, "TestDNSProviders was successful!", "willpower232testsdnscontrol.com.") failed: dns: bad TXT Txt: "TestDNSProviders" at line: 1:17
> ```

At least I'm consistent! LOL

I'll take a look.

TomOnTime:
I think I fixed it. Please try again.

willpower232:
would you believe I managed to make the tests last over an hour? haha, I'll try again tomorrow when my daily rate limit has reset

willpower232 (approved):
I confirm that the 4 tests which fail on the main branch are the only tests failing on this branch so this is a go from me 🙏

TomOnTime:
Fantastic!  Thanks for your help!

Would you mind submitting a PR that marks those 4 tests as "skip"?

Find the tests in `integrationTest/integration_test.go` and mark them with not(). Here's an example:

```
                testgroup("Attl",
                        not("LINODE"),  // Linode does not support arbitrary TTLs: both are rounded up to 3600.
                        not("OPENWRT"), // OpenWRT does not support per record TTL
                        tc("Create Arc", ttl(a("testa", "1.1.1.1"), 333)),
                        tc("Change TTL", ttl(a("testa", "1.1.1.1"), 999)),
                ),
```

Thanks!
Tom

willpower232:
I would but the other cases in the group do pass, unless there is a way to exclude a single test case without moving it to its own group?

TomOnTime:
> I would but the other cases in the group do pass, unless there is a way to exclude a single test case without moving it to its own group?

@willpower232 : auditrecords.go ?

willpower232:
I tried but I don't think auditrecords.go can tell if a record is being changed (which is two of them) and it also seems to be checking existing records instead of new records so I can't tell that its a single apex NS

image

TomOnTime:
> I tried but I don't think auditrecords.go can tell if a record is being changed (which is two of them) and it also seems to be checking existing records instead of new records so I can't tell that its a single apex NS

Ah, good point.  You'll need to split out that test.

PR #4576, opened 2026-07-22
CICD: Translate PowerShell code to jq

TomOnTime:


TomOnTime:
CC: @chicks-net will be happy to see this!

PR #4593, opened 2026-07-22
NS1: Modernization fixes

costasd:
This changeset provides fixes for TXT and NAPTR handling under the new record converter.

* For TXT, it ensures that we parse it as a simple string and not an RFC1035 quoted representation.
* For NAPTR, it restores the usage of naptrAssureQuotedFields, but in a provider-private way, ensuring no other providers will be impacted now or in the future.

### DIFF
--- a/providers/ns1/convert_test.go
+++ b/providers/ns1/convert_test.go
@@ -1,0 +1,57 @@
+package ns1
+
+import (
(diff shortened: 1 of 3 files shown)


TomOnTime:
Thanks!

PR #4582, opened 2026-07-22
unifi: Modernize record conversion

TomOnTime:
Dear @zupolgec. I could use your help. I'm doing upgrades to the code base and I have no way to test your provider. Please run the integration test and let me know the result. How to run integration tests is here: https://docs.dnscontrol.org/developer-info/integration-tests

## Summary

- build legacy and current UniFi API records with modern factories
- precompute labels and TTLs shared across record-type branches
- modernize unit-test fixtures while preserving provider ID tests

## Validation

- `../../bin/is_modern.sh` (Steps 1-7 clean)
- `go test ./providers/unifi`
- `bin/generate-all.sh`
- `go test ./...`

Issue #2066, opened 2023-02-12
BUG: hosting_de SRV TTL change crash


tmkis2:
Changing just a TTL for a DNS record results in

FAILURE! [{Code:40010 ContextObject: ContextPath: Text:Record with name 'xxxxx', type 'SRV' and content 'xxxx' is a duplicate. Value:}]

Nothing is changed and preview always displays this as to be changed (TTL)

tlimoncelli:
CC: @juliusrickert (owner of the hostingde provider)

juliusrickert:
@tmkis2 could you provide a minimal reproducible example, please?

I won't get to fixing this until the end of next week.

tlimoncelli:
FYI: In case you need it... here's an example of how to add an integration test that covers this situation: https://github.com/StackExchange/dnscontrol/pull/2067/files

tmkis2:
@juliusrickert any existing entry will cause this if you change it to another TTL value like adding , TTL(600) to the end of this entry. If the entry doesn't change anything other than the TTL it will fail.
Just now I have to change the entry in hosting.de web panel or change the entry to some (invalid) value and then back to the right value with new TTL

tmkis2:
Okay now I am really sure that after the latest dnscontrol update something major happened with at least the hosting_de provider. Every push forces changed lines to be created first (which results in above error for TXT and CNAME records because they can't be full duplicates) and "works" for A records resulting in TWO a records (one with the old value and the second with the new value). Another run from dnscontrol push results in removing the old A record. So for now to change TXT or CNAME records you have to remove them first, run dnscontrol push and then add them with the new values again and run dnscontrol push again.

tlimoncelli:
Hey @Yannik could you take a look?

@juliusrickert 

These are the changes since the last version: (if I got the git command correct... which isn't always true!)

```
$ git log ...v3.25.0 providers/hostingde/
```

juliusrickert:
@tmkis2 @tlimoncelli I'll be looking into this this weekend.

Yannik:
@tmkis2 
> any existing entry will cause this if you change it to another TTL value like adding , TTL(600) to the end of this entry. If the entry doesn't change anything other than the TTL it will fail.

I'm having trouble reproducing your issue. I tried changing TTLs both on TXT records and CNAMEs, but didn't get any errors. I am using the latest master. Could you post both `dnsconfig.js` snippets and the commands you run to reproduce the problem?

[...]

PR #4594, opened 2026-07-22
UNIFI: Skip SRV TTL integration test, add console_id to test profile

zupolgec:
## Summary

- Add `UNIFI` to the `not()` list of the SRV TTL testgroup (issue #2066): the UniFi API does not support TTL on SRV records (`ttlSeconds` is rejected on write with `Unknown request body property '$.ttlSeconds'`), so this test can never pass. Same approach already used for OPENWRT and NAMECHEAP, which lack per-record TTL support.
- Add `console_id` to the UNIFI profile in `integrationTest/profiles.json` so tests can also run via UniFi cloud access (api.ui.com), not only against a local controller.

## Context

Found while running integration tests for #4582: `44:SRV:Change_TTL999` failed with "Expected changes, but got none". The failure is pre-existing on `release_candidate_v5` and is by design: the provider normalizes TTL to 300 for MX/TXT/SRV in `GetZoneRecordsCorrections` because the API cannot store it.

## Validation

- `go test -timeout 20m -run TestDNSProviders -v -args -verbose -profile UNIFI` against a local controller: **196 PASS, 0 FAIL** (previously 1 FAIL: `44:SRV:Change_TTL999`, now correctly skipped)
- Tested with go1.26.5 darwin/arm64

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -898,6 +898,7 @@ func makeTests() []*TestGroup {
 			not(
 				"OPENWRT",   // OpenWRT does not support per record TTL
 				"NAMECHEAP", // Namecheap does not support per record TTL
+				"UNIFI",     // UniFi API does not support TTL on SRV records
 			),
 			tc("Create SRV333", ttl(srv("_sip._tcp", 5, 6, 7, "foo.com."), 333)),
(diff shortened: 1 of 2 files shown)


zupolgec:
Closing: TomOnTime applied the same two changes directly on #4582 (commits 8c5af954 and 9ab148bb).

PR #4605, opened 2026-07-24
CSCGLOBAL: Use diff2.ByRecord instead of diff.NewCompat

TomOnTime:
Converts providers/cscglobal from the legacy diff.NewCompat() engine to diff2.ByRecord(), per the migration in pkg/diff2/diff2.go. The makePurge/makeAdd/makeEdit helpers now take *models.RecordConfig; the single batched sendZoneEditRequest (delete→create→modify order) is unchanged.


TomOnTime:
Hi @mikenz 

I no longer have a test environment for CSCGlobal. Could you do me a favor and run the integration tests?  It's going to take 9+ hours to run because of how their API works (sorry!) but it will eventually complete.

Instructions are here: https://docs.dnscontrol.org/developer-info/integration-tests

All the changes were "mechanical" and therefore should "just work".

Tom

TomOnTime:
CC @jzhang-sre I'm pretty sure SO delegates all their domains to a different provider and only uses CSCGlobal for registration. That said, since I made these changes "blind", you might want to do some extra testing here.

TomOnTime:
FYI: if I don't hear back by July 31, I'll merge this and pray it works.

PR #4606, opened 2026-07-24
SOFTLAYER: Use diff2.ByRecord instead of diff.NewCompat

TomOnTime:
Converts providers/softlayer from the legacy diff.NewCompat() engine to diff2.ByRecord(), per the migration in pkg/diff2/diff2.go.


TomOnTime:
CC @saurabhguptag   @jamielennox for visibility

PR #4583, opened 2026-07-22
vultr: Modernize record conversion

TomOnTime:
Dear @pgaskin. I could use your help. I'm doing upgrades to the code base and I have no way to test your provider. Please run the integration test and let me know the result. How to run integration tests is here: https://docs.dnscontrol.org/developer-info/integration-tests

## Summary

- replace legacy `RecordConfig` construction and target setters with record factories
- pass the domain configuration into Vultr record conversion
- preserve the existing IDNA, quoting, and round-trip behavior

## Testing

- `../../bin/is_modern.sh` (Steps 1-7 clean)
- `go test ./providers/vultr`
- `bin/generate-all.sh`
- `go test ./...`

PR #4609, opened 2026-07-26
VULTR: Changes for #4583 (and some other stuff)

pgaskin:
- Fixes for #4583
- Fix null MX handling (it previously didn't round-trip properly)
- Update audit (adding `rejectif.CaaHasEmptyTarget`)
- Simplify GetZoneRecords (and fix some theoretical edge cases)
- Implement ListZones
- Implement AutoDNSSEC
- Use govultr v3
- Verify concurrency
- Cleanup


providers/vultr/vultrProvider.go
```diff
@@ -118,61 +119,62 @@ func (api *vultrProvider) GetZoneRecords(dc *models.DomainConfig) (models.Record
...
+			printer.Warnf("All TTLs for a rrset (%v) must be the same. Using smaller of %v and %v.\n", r.Key(), r.TTL, ttl)
+			r.TTL = ttl
+		}
+	}
+
+	// Vultr rejects any operation which would result in mixed recordset TTLs,
+	// so delete/re-create entire recordsets at a time (this is a bit
+	// heavy-handed, but it's simpler than trying to be smart about it, and
+	// matches what other provders like akamaiedgedns and alidns do).
+	changes, actualChangeCount, err := diff2.ByRecordSet(curRecords, dc, nil)
```
pgaskin, line 137: @TomOnTime thoughts?

The only possible improvement (depending on what you consider better) would be to only delete the recordset if the TTLs actually differ from existing ones, and to edit instead of deleting if there's only one. This would make edits more concise, but also more complex.

pgaskin (commented):
All integration tests pass since 34aa197125a008d52fd6a42a75dbdeae93524eec `vultr: Fix null MX handling`. The other changes can be merged along with it, or separately.

Personally, I'd squash the first one into the other PR, then rebase the rest.

providers/vultr/vultrProvider.go
TomOnTime: I'm ok with leaving it as-is.

"Simple and always works" is better than "smart and difficult to debug".

TomOnTime (approved):
Thanks for the upgrade to govultr/v3 and so many other simplifications and improvements!

providers/vultr/vultrProvider.go
pgaskin: Yeah, that's how I saw it, especially given the nature of this sort of tool.

PR #4611, opened 2026-07-26
DNSIMPLE: fix mod_dnsimple: SVCB & HTTPS are also unqualified at dnsimple

onlyhavecans:
This is a "quick" fix to the mod_dnsimple branch and PR https://github.com/DNSControl/dnscontrol/pull/4545 which fixes SVCB & HTTPS records.

DNSimple does not return qualified domains in these either so add a quick helper to manage all the various forms.


TomOnTime (approved):

TomOnTime:
Thanks for the fix!

Issue #4422, opened 2026-07-05
About `dnscontrol preview` when zone does not exist


riku22:
**Describe the bug**
If the zone does not exist, running `dnscontrol preview` will display an error and stop.
`dnscontrol push` automatically creates zones if they don't already exist.
Shouldn't preview mode either create them as well or display a message indicating that they will be created automatically, rather than treating this as an error?

**To Reproduce**
Create `dnsconfig.js` specifying the non-existent zone and run `dnscontrol preview`.

The following error is displayed.

```
CONCURRENTLY checking for 1 zone(s)
SERIALLY checking for 0 zone(s)
Waiting for concurrent checking(s) to complete...DONE
******************** Domain: example.com
1 correction (luadns)
#1: Ensuring zone "example.com" exists in "luadns"
CONCURRENTLY gathering records of 1 zone(s)
SERIALLY gathering records of 0 zone(s)
Waiting for concurrent gathering(s) to complete...DONE
******************** Domain: example.com
INFO#1: Domain "example.com" provider luadns Error: 'example.com' not a zone in luadns account
Done. 1 corrections.
completed with errors
```

**Expected behavior**
Since the zone is automatically created, display that fact and exit without an error.

**DNS Provider**
Check with LUADNS and CLOUDFLARE

**Additional context**
REF: #4322

TomOnTime:
Does adding `--no-populate` help?

riku22:
I tried running it with the `--no-populate` option, but the results don't seem to change.

PR #4612, opened 2026-07-26
BUG: preview no longer errors when a zone will be created by push

### LINKED ISSUE #4322: CLOUDFLARE: Github Actions reports error when zone doesn't exist
**Describe the bug**
GitHub Actions throws an error when we're adding new domains to our repository. I'm happy to try to fix this, but wanted to make sure this isn't expected behavior before doing so.

**To Reproduce**
1. Create a zone file that does not exist previously in your Cloudflare tenant.
2. Run the `stackexchange/dnscontrol` image in a GitHub workflow with a step for `dnscontrol preview` (Or use the provided workflow below)
3. Notice that it fails

**DNS Provider**
 - Cloudflare

```
name: DNS Preview

on:
  pull_request:
    branches:
      - main
    paths:
      - src/**
      - .github/workflows/dnsPreview.yml

jobs:
  dns-preview:
    runs-on: ["self-hosted", "rhel"]
    container:
      image: stackexchange/dnscontrol:latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3

      - name: Check version
        run: dnscontrol version

      - name: Preview changes
        run: dnscontrol preview
        working-directory: ${{ github.workspace }}/src
        env:
          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
          CLOUDFLARE_API_KEY: ${{ secrets.CLOUDFLARE_API_KEY }}
```

lopster568:
`dnscontrol preview` exits non-zero with a raw provider error whenever `dnsconfig.js` references a domain that does not yet exist at the provider, even though `push` would create that zone without error. Preview intentionally does not run the zone-creation correction, but Phase 2 still fetched records for the not-yet-created zone, turning a "will be created" case into a hard failure. This breaks CI that runs `dnscontrol preview` on pull requests adding new domains.

This threads the `push` flag into `oneZone` and, when previewing a zone that Phase 1 flagged for creation, reports that the zone will be created on push instead of fetching its records. The message is suppressed when this run actually creates the zone (push, or preview with `--populate-on-preview`), so those keep showing the real record diff. Push behavior is unchanged, and providers that cannot create the zone still surface a real error.

Fixes #4322
Related: #4422

### DIFF
--- a/commands/ppreviewPush.go
+++ b/commands/ppreviewPush.go
@@ -340,7 +340,7 @@ func prun(args PPreviewArgs, push bool, interactive bool, out printer.CLI, repor
 		out.PrintfIf(fullMode, "Concurrently gathering: %q\n", zone.UniqueName)
 		go func(zone *models.DomainConfig, args PPreviewArgs, zcache *CmdZoneCache) {
 			start := time.Now()
-			err := oneZone(zone, args)
+			err := oneZone(zone, args, push)
 			if err != nil {
 				concurrentErrors.Store(true)
 			}
@@ -358,7 +358,7 @@ func prun(args PPreviewArgs, push bool, interactive bool, out printer.CLI, repor
 	out.Printf("SERIALLY gathering records of %d zone(s)\n", len(zonesSerial))
 	for _, zone := range zonesSerial {
 		out.Printf("Serially Gathering: %q\n", zone.UniqueName)
-		if err := oneZone(zone, args); err != nil {
+		if err := oneZone(zone, args, push); err != nil {
 			anyErrors = true
 		}
 	}
@@ -567,7 +567,7 @@ func oneZonePopulate(zone *models.DomainConfig, zc *CmdZoneCache) error {
 	return errors.Join(errs...)
 }
 
-func oneZone(zone *models.DomainConfig, args PPreviewArgs) error {
+func oneZone(zone *models.DomainConfig, args PPreviewArgs, push bool) error {
 	var errs []error
 	// Fix the parent zone's delegation: (if able/needed)
 	delegationCorrections, dcCount, err := generateDelegationCorrections(zone, zone.DNSProviderInstances, zone.RegistrarInstance)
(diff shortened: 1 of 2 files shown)


TomOnTime:
@riku22  Does this solve the problem you described in https://github.com/DNSControl/dnscontrol/issues/4422?

riku22:
Thank you for working on this issue.

The problem was resolved with LuaDNS, but it seems that an error occurs with Cloudflare.

```
CONCURRENTLY checking for 1 zone(s)
SERIALLY checking for 0 zone(s)
Waiting for concurrent checking(s) to complete...DONE
******************** Domain: example.com
1 correction (cloudflare)
#1: Ensuring zone "example.com" exists in "cloudflare"
CONCURRENTLY gathering records of 1 zone(s)
SERIALLY gathering records of 0 zone(s)
Waiting for concurrent gathering(s) to complete...DONE
******************** Domain: example.com
INFO#1: Zone "example.com" does not yet exist at provider "cloudflare". It will be created when you run `dnscontrol push`.
INFO#1: DetermineNS: zone "example.com"; Error: error while getting Nameservers for zone="example.com" with provider="cloudflare": zone not found
Done. 1 corrections.
completed with errors
```

lopster568:
Thanks for testing, @riku22. The remaining Cloudflare error comes from a different step than this PR changes.

This PR fixes the record-gathering path: preview no longer fetches records for a zone that push will create, which is why LuaDNS is clean now and #4322 is resolved.

The Cloudflare error is raised earlier, in `generateDelegationCorrections` calling `DetermineNameserversForProviders`. That queries the provider for the zone's nameservers before the zone exists, so a to-be-created zone still errors there. It runs ahead of this PR's change and isn't touched by it. LuaDNS derives its nameservers without a provider call, so it does not hit this path.

@TomOnTime I'd suggest keeping this PR scoped to #4322 and opening a separate one for the delegation path, which could reuse the `zoneWillBeCreated` check added here. Happy to fold it in instead if you would prefer.

TomOnTime:
@lopster568 I agree with your suggestion (merge this, keep cloudflare separate).

Thanks for fixing this part!

PR #4614, opened 2026-07-27
POWERDNS: exclude from DS and DNSKEY record tests

jpbede:
Disable the `DS` test group for PowerDNS as DS records are not supported a apex. Disable `DNSKEY` test group as well as DNSKEY records are only supported at apex for PowerDNS


TomOnTime:
Thanks!

PR #4553, opened 2026-07-22
gigahost: Modernize record conversion

TomOnTime:
Dear @jochristian. I could use your help. I'm doing upgrades to the code base and I have no way to test your provider. Please run the integration test and let me know the result. How to run integration tests is here: https://docs.dnscontrol.org/developer-info/integration-tests

## Summary

- replace legacy record construction with current domain-aware factories
- replace the combined legacy getter with typed RDATA output
- preserve Gigahost's hostname and raw TXT normalization
- add focused conversion regression tests

## Validation

- `../../bin/is_modern.sh` (Steps 1-7 clean)
- `go test ./providers/gigahost`
- `bin/generate-all.sh`
- `go test ./...`

PR #4617, opened 2026-07-27
GIGAHOST: Fix API quirks surfaced by the integration tests

jochristian:
Fixes for #4553 — provider-level Gigahost API quirks the integration tests surfaced, all verified against the live API with a dedicated test zone. None are caused by the modernization changes; they were latent in the provider (or the API's behavior has changed since the provider was written).

## Fixes

- **MX DELETE value form**: the delete endpoint 404s when given the `record_value` the API itself returns for MX; it requires the internal RDATA presentation `"<priority> <target>."`. Added `deleteValue()`.
- **429 rate limiting**: the API throttles hard enough that a normal-sized push can trip it. `request()` now retries up to 5 times with 2s→32s backoff, honoring `Retry-After`.
- **Null MX**: the API rejects a `.` target ("MX record value must be a valid mail server hostname") → audit-reject with `rejectif.MxNull`.
- **TXT >255 octets**: must be sent as RFC1035 quoted 255-octet chunks; the API returns that form with the outer quotes stripped, so it is decoded on read (unambiguous: a single character-string can't exceed 255 octets).
- **TXT with double quotes**: cannot round-trip. Raw interior quotes are rejected by the API's zone parser (it blindly wraps values in quotes); the escaped form stores correctly (verified via `dig`) but reads back mangled (trailing `\"` → `\`), breaking diffing and deletes → `rejectif.TxtHasDoubleQuotes`.
- **TXT leading/trailing spaces**: trimmed by the API on store → `rejectif.TxtStartsOrEndsWithSpaces`.
- **NAPTR is single-record per label**: POSTing a second NAPTR at the same label silently replaces the entire NAPTR RRset (201 "created", first record gone — verified via API and `dig`). Audit-rejects >1 NAPTR per label, and marks the `NAPTR delete second record` testcase `.AllowNoChanges()` so providers that audit-skip the multi-NAPTR case don't fail on the already-converged state.

## Validation

```
cd integrationTest && go test -v -verbose -profile GIGAHOST -run TestDNSProviders
ok  github.com/DNSControl/dnscontrol/v5/integrationTest  87.9s
239 PASS / 77 SKIP / 0 FAIL
```

Also: `go vet`, `gofmt`, `go test ./providers/gigahost` clean.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -839,7 +839,10 @@ func makeTests() []*TestGroup {
 				naptr("test", 100, 10, "U", "E2U+sip", "!^.*$!sip:customer-service@example.com!", "."),
 				naptr("test", 102, 10, "U", "E2U+email", "!^.*$!mailto:information@example.com!", "."),
 			),
-			tc("NAPTR delete second record", naptr("test", 100, 10, "U", "E2U+sip", "!^.*$!sip:customer-service@example.com!", ".")),
+			// AllowNoChanges: providers that audit-reject multiple NAPTRs at
+			// one label (e.g. GIGAHOST) skip "NAPTR second record" above, so
+			// this state may already be converged.
+			tc("NAPTR delete second record", naptr("test", 100, 10, "U", "E2U+sip", "!^.*$!sip:customer-service@example.com!", ".")).AllowNoChanges(),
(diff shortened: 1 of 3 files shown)


TomOnTime (approved):
Looks good!  Thanks for the fast response!

PR #4619, opened 2026-07-28
VERCEL: Modernize fields

TomOnTime:
* Change to new way to access record fields

CC @SukkaW for visibility.


SukkaW (commented):
LGTM overall, as long as intergration tests pass there should be no problem here.

PR #126, opened 2017-05-26
WIP: spf flattening logic.

captncraig:
Currently implemented splitting TXT record at 255 boundaries.

- [x] Add metadata to flatten / split TXT records
- [x] Separate subcommand to update caches
- [x] document everything
- [x] Pretty up ui thing for docs.

PR #226, opened 2017-10-11
Improved BIND docs (#221)

geek1011:
- Fixed highlighting language
- Improved markdown formatting
- Improved grammar

See #221

PR #4630, opened 2026-07-31
BUG: SPF flattening no longer drops the last term of an include with no "all"

shuvamk:
`SPFRecord.Flatten()` removes the last term of every include it flattens, whether or not
that term is an `all` mechanism. RFC 7208 §5.1 makes `all` optional and §4.7 defines the
default result when it is absent, so an include target that ends in something else is
legal — and common in the wild.

The result is a syntactically valid SPF record that is missing a live sending range, with
no warning and no error.

| input (flattened with `*`) | result on `main` | expected |
|---|---|---|
| `v=spf1 include:aspmx.googlemail.com -all`<br>`aspmx.googlemail.com` → `v=spf1 redirect=_spf.google.com`<br>`_spf.google.com` → two `ip4` + six `ip6` netblocks + `~all` | `v=spf1 ip4:74.125.0.0/16 ip4:209.85.128.0/17 ip6:2001:4860:4864::/56 ip6:2404:6800:4864::/56 ip6:2607:f8b0:4864::/56 ip6:2800:3f0:4864::/56 ip6:2a00:1450:4864::/56 -all`<br>— `ip6:2c0f:fb50:4864::/56` silently deleted | all eight netblocks retained |
| `v=spf1 include:spf-00123c01.pphosted.com -all`<br>(target resolves to six `ip4` terms and no `all`) | `v=spf1 ip4:67.231.153.0/24 ip4:67.231.145.0/24 ip4:67.231.149.0/24 ip4:67.231.152.48 ip4:67.231.148.205 -all`<br>— `ip4:67.231.152.182` silently deleted | all six `ip4` terms retained |
| `v=spf1 include:a.example.net -all`<br>`a.example.net` → `v=spf1 include:b.example.net`<br>`b.example.net` → `v=spf1 -all` | `panic: runtime error: slice bounds out of range [:-1]` at `flatten.go:124` | `v=spf1 -all`, no panic |
| `v=spf1 include:example.net -all` where the target ends in `~all` | `all` dropped | unchanged — `all` dropped |

Flattening a Google include drops a Google netblock: `aspmx.googlemail.com` is
`v=spf1 redirect=_spf.google.com`, and once the redirect is expanded the flattened result
ends in an `ip6` netblock rather than an `all`, so that netblock is deleted.

`spf-00123c01.pphosted.com` is the include that citibank.com, citi.com and citigroup.com
publish. Surveying the `include:` targets reachable from the 6000 highest-ranked domains
of the Cisco Umbrella top-1m, of which 2664 publish an SPF record: 52 of 967 distinct
include targets (5.4%) end in something other than `all`, and 57 of those 2664 domains
(2.1%) include at least one such target. Proofpoint's `spf-XXXXXXXX.pphosted.com` records
dominate; `_spf.hp.com`, `_q.ironport.com` (which publishes only three ranges, so it loses
a third of them) and `_corporate.eloqua.com` are also affected.

The third row is not a contrived input: `v=spf1 -all` is what every no-mail domain
publishes, and one level of indirection above it is enough to flatten to zero terms.

## Cause

`pkg/spflib/flatten.go:124` — the comment states the intent, the code does not check it:

```go
// include their parts (skipping final all term)
newRec.Parts = append(newRec.Parts, flattenedChild.Parts[:len(flattenedChild.Parts)-1]...)
```

The slice expression assumes the flattened child always ends in `all`. When it does not,
a real mechanism is dropped; when the child has no parts at all, the expression panics.
This has been the behaviour since the flattener was added in #126 (2017). Nothing caught
it because no existing test asserts `Flatten`'s output — `TestFlatten` only logs it — and
all twelve records in `testdata-dns1.json` end in `all`.

## Fix

Strip the final part only when it actually is an `all` mechanism, with or without a
qualifier. The comparison is case-insensitive, which matches the leniency `Parse()`
already applies to mechanism names ("We have seen `Ip4` instead of `ip4`") and keeps
today's behaviour for a target ending in `-ALL`.

## Tests

`TestFlattenTrailingAll` in `pkg/spflib/flatten_test.go`, nine table cases: the
`aspmx.googlemail.com` redirect chain; the Proofpoint record with no `all`; targets ending
in `~all`, `-all`, `?all`, unqualified `all` and `-ALL` (pinning the existing behaviour in
both directions); a nested include whose inner target has no `all`; and the no-mail chain
that flattens to zero terms.

It uses a small in-test `Resolver` rather than `NewCache("testdata-dns1.json")`. That
fixture stores each entry under a `"txt"` key, which `cacheEntry{SPF string}` does not
unmarshal, so every entry loads empty and the existing `spflib` tests fall through to live
DNS — the new test is hermetic instead. Happy to file that separately if it is useful.

I verified the test fails without the fix. With `git checkout main -- pkg/spflib/flatten.go`
(test kept), `go test ./pkg/spflib/ -run TestFlattenTrailingAll -v` fails four subtests:

```
--- FAIL: TestFlattenTrailingAll/include_of_a_redirect-only_target_keeps_the_last_netblock
    flatten_test.go:124: got  v=spf1 ... ip6:2a00:1450:4864::/56 -all
                        want v=spf1 ... ip6:2a00:1450:4864::/56 ip6:2c0f:fb50:4864::/56 -all
--- FAIL: TestFlattenTrailingAll/include_with_no_all_keeps_its_last_term
    flatten_test.go:124: got  v=spf1 ... ip4:67.231.148.205 -all
                        want v=spf1 ... ip4:67.231.148.205 ip4:67.231.152.182 -all
--- FAIL: TestFlattenTrailingAll/nested_include_with_no_all_keeps_its_last_term
--- FAIL: TestFlattenTrailingAll/include_of_a_target_that_only_includes_a_no-mail_domain
panic: runtime error: slice bounds out of range [:-1]
    .../pkg/spflib/flatten.go:124
```

The five `all`-terminated cases pass in both states, so the test pins the existing
behaviour as well as the new. All nine pass with the fix restored.

## Local verification

Because this is my first PR here, the workflows sit at `action_required` until a
maintainer approves the run, so the following are the pre-merge evidence. All run on this
branch, macOS/arm64, go1.26.0 via `GOTOOLCHAIN=auto`:

| command | result |
|---|---|
| `go build .` | success |
| `go test ./...` | 47 ok, 0 FAIL (identical to `main` @ 637fbaf7) |
| `golangci-lint run ./...` | 0 issues |
| `staticcheck ./...` | 0 issues |
| `go generate ./... && go fmt ./... && go mod tidy && bin/fmtjson …` | no files changed |
| `BIND_DOMAIN=example.com go test ./integrationTest/ -args -provider BIND` | ok |

`main` was green before the change, so there is no pre-existing failure to discount.

## One interaction worth flagging

If a partial flatten spec is used (not `*`) and an include target's last term is an
unmatched `redirect=`, that term is now retained rather than dropped. Six of the 967
surveyed include targets end in `redirect=`: `aspmx.googlemail.com`,
`calendar-server.bounces.google.com`, `cisco.com`, `_spf.yandex.net`, `fusemail.net` and
`spf.smtpcorp.com`. Both corners, with
`child.example.net = v=spf1 ip4:1.2.3.4 redirect=other.example.org`:

| parent record | `main` | this PR |
|---|---|---|
| `v=spf1 include:child.example.net -all` | `v=spf1 ip4:1.2.3.4 -all` | `v=spf1 ip4:1.2.3.4 redirect=other.example.org -all` |
| `v=spf1 include:child.example.net` | `v=spf1 ip4:1.2.3.4` | `v=spf1 ip4:1.2.3.4 redirect=other.example.org` |

In the first row the retained `redirect=` is inert: RFC 7208 §5.1 says any `redirect`
modifier "MUST be ignored when there is an 'all' mechanism in the record, regardless of
the relative ordering of the terms", so the two outputs are semantically identical. In the
second row there is no `all`, the `redirect=` is last and valid, and it is honoured —
whereas `main` discards it and loses the redirect target's addresses entirely. The new
behaviour is therefore never semantically worse, which is why I did not add a guard.

`Parse()` does reject `redirect=` in a non-final position, so re-parsing the first row's
output errors. If you would rather I keep dropping a trailing unmatched `redirect=`, or
expand it regardless of the spec, say the word and I will amend.


TomOnTime:
Wow! That's an excellent find!  This bug has been in the code since 2017-09-29 (almost 9 years!)

Thank you for identifying this and submitting a PR.

As far as redirect: I'd like to merge this PR then handle that separately.  redirects should be removed when they are ignored. That's the kind of thing an optimizer should do. That said, there should be a way to disable this feature in case someone feels strongly about not wanting that particular optimization.

TomOnTime:
> the workflows sit at action_required until a maintainer approves the run

Due to GH's permissions system, I have to run it separately: Here are the results: https://github.com/DNSControl/dnscontrol/pull/4632

By the way... I'm curious what prompt you used to find this bug.

shuvamk:
Thanks for running the workflows — #4632 is green, so that clears the only gate I couldn't check myself.

On `redirect=`: merging this first and handling it separately works for me, and your framing actually resolves the residual I flagged. The one case where this PR changes `redirect=` output is a partial flatten spec where the target's last term is an unmatched `redirect=` — it's now retained instead of dropped. When the parent has an `all`, the retained term is semantically inert (§5.1 says it MUST be ignored), but the record no longer round-trips: `Parse()` rejects `redirect=` in a non-final position, so `v=spf1 ip4:1.2.3.4 redirect=other.example.org -all` parses back as an error. Stripping ignored redirects in the optimizer, as you describe, removes that case entirely. Happy to implement it behind a disable flag if you want it as a follow-up PR — just say which way you'd like the flag to default.

As for how I found it: I should be upfront that this was AI-assisted — there's a `Co-Authored-By: Claude Opus 5` trailer on the commit. It's an agent I run over a project's source looking for correctness bugs, so it's less one prompt than a standing set of rules, but the heuristic that actually caught this one is easy to state: **look for a comment that asserts an invariant the code immediately below it doesn't check.**

```go
// include their parts (skipping final all term)
newRec.Parts = append(newRec.Parts, flattenedChild.Parts[:len(flattenedChild.Parts)-1]...)
```

The comment says "final all term". The slice drops the final term whatever it is. Someone knew the intent well enough to write it down and the code just never enforced it — that gap is where this class of bug lives, and it survives review precisely because the comment reads as if the check exists.

That only makes it a suspicion. The rules that turn it into a PR are the expensive part, and they're mostly about not wasting your time:

- **Prove it's reachable with a realistic input, not a constructed one.** Hence the Umbrella top-1m survey rather than my asserting "this probably happens" — 52 of 967 distinct include targets don't end in `all`. That measurement is also what ranked this above another open finding I have in `spflib`.
- **Prove the test fails without the fix.** Stash the source change, watch the test go red, restore it, watch it go green. No "should fix".
- **Re-resolve every external fact at write time.** An earlier draft of this body asserted a DNS fact about a specific domain that was wrong and that you could have falsified with one `dig`. Caught before posting, but it's the failure mode I'd warn anyone about: one invented number and a maintainer rightly discounts every other figure in the body, including the measured ones.

Broadly, the agent is much better at the first step than the last, which is why the verification rules are so blunt.

The other `spflib` finding is #226 — `Parse()` recurses without a depth or cycle limit, so a cyclic third-party `include:` chain overflows the stack unrecoverably. I held it back because a fix threads a visited-set or depth parameter through a public recursive API, which felt like your call to make rather than something to drop on you unsolicited. Want an issue for it?

TomOnTime:
Yes, please submit a separate PR that handles redirect. The default should be to remove it.

I don't mind LLM-generated PRs as long as the code is good, and I can understand it :)

As far as the Parse() recursion: Please open a bug.

PR #4635, opened 2026-08-01
NAMECHEAP: just skip the failing tests

willpower232:
[As requested](https://github.com/DNSControl/dnscontrol/pull/4566#issuecomment-5079083487), Namecheap now has no test fails once you exclude any timeouts because of their rate limits


TomOnTime:
Thanks!

It looks like "Change MX p" could be fixed by deleting the entire record and reinserting it.  Namecheap wouldn't be the first provider that requires that.  Please consider fixing this in the future.

Merging in 3... 2... 1...

PR #4657, opened 2026-08-02
ALIDNS: Modernize field access

TomOnTime:
Hi!

These changes are low risk but please test and let me know if they pass.

This should be the last PR for ALIDNS for a while.

Thanks for helping out!
Tom


bytemain:
Tested mod_alidns (PR #4657) against the full AliDNS integration suite — all green. 217 PASS / 0 FAIL / 90 SKIP (skips are unrelated providers/unsupported record types). ok .../integrationTest 568.509s. Full log can see attachment 
[rc-integration-test-alidns.zip](https://github.com/user-attachments/files/30630045/rc-integration-test-alidns.zip). 

Only warnings are TTL-below-600 on the free (mianfei) test-domain plan, which is expected. LGTM. 👍

TomOnTime:
Thanks for the fast reply!

PR #4658, opened 2026-08-02
AXFRDDNS: convert to codeberg.org/miekg/dns (dnsv2)

### LINKED ISSUE #4358: Please convert AXFRDDNS to use dnsv2

shuvamk:
Fixes #4358. Based on `release_candidate_v5`, as you asked on the issue.

#4529 had already done the record-conversion side, so what was left was the network layer: `dnsv1.Conn`, `dnsv1.Transfer`, `dnsv1.Client` and the RFC 2136 message builders. `providers/axfrddns` no longer imports `github.com/miekg/dns`.

### Three places where the dnsv2 API forced a behaviour change

**1. Single-message AXFR would have taken 30 seconds.** This is the one worth reading.

`transferInAXFR` only checks for the terminating SOA when `options.TimersOnly` is true, and it sets `TimersOnly` at the *end* of the first loop iteration ([`transfer.go:174-186`](https://codeberg.org/miekg/dns)). v1's `inAxfr` flips `first` *before* the `isSOALast` check (`xfr.go:117-131`), so v1 returns on the closing SOA even when the whole zone arrives in one message. Under dnsv2 the read loop just keeps going until the server closes or `ReadTimeout` (30 s) expires.

Measured against BIND 9.20 in a container, `allow-transfer any`:

| | 8-record zone (1 message) | 1208-record zone (3 messages) |
|---|---|---|
| `release_candidate_v5` (dnsv1) | 4.4 ms | 17.1 ms |
| this branch, before the fix | **30.0 s** | 18 ms |
| this branch as it stands | 2.0–4.9 ms | 14.8 ms |

`transfer-message-size` defaults to 20480 bytes, so a typical AXFRDDNS zone fits in one message and this would have cost 30 s on every `preview` and every `push` — including the integration tests.

`FetchZoneRecords` now latches once the accumulated answer ends in a SOA and closes the connection. It drains the remaining envelopes rather than breaking, because `TransferInWithConn` hands back an unbuffered channel — with a bare `break`, a server that sends a spurious envelope after the closing SOA parks the library's reader goroutine forever on `ch <-` (16 of 20 transfers, measured). Pinned by `TestFetchZoneRecordsReturnsOnTheClosingSOA`.

**The real fix is one line upstream in your fork** — `transferInAXFR` should make the terminating-SOA check on the first iteration too, the way v1's `inAxfr` does. Every other dnsv2 AXFR caller has this stall today. The provider-side latch stays correct either way, so this doesn't block on it, but you may prefer to fix it there and simplify here.

**2. A connection dropped mid-AXFR used to be an error and silently became a success.** `dnsv1.Transfer` reported a read error on the envelope channel; dnsv2 closes the channel silently. Left alone, a truncated transfer would have been accepted as a complete zone and every record that didn't arrive would have been diffed as a **deletion**. `FetchZoneRecords` now requires the closing SOA of RFC 5936 §2.2 before trusting the answer, which restores the v1 failure. The guard counts records rather than messages, since a server may answer the opening SOA in a message of its own (v1 special-cased exactly that at `xfr.go:114-118`) — `TestFetchZoneRecordsDoesNotStopOnTheOpeningSOA` pins that boundary.

**3. Refused-transfer error text changed.** Now `refused to transfer the zone example.com: dns: bad rcode: NOTAUTH` rather than `...: NOT AUTH (9)`. The old text came from matching `msg.Error.Error()` against the literal `"dns: bad xfr rcode: 9"`, which dnsv2 no longer produces.

### TSIG

There's no `TsigSecret` map in dnsv2 — a `Client` carries a `TSIGSigner`. So `Key.secret` now holds the decoded secret rather than its base64 form (`readKey` already validated the encoding, it just threw the result away), and `Key.signer()` returns `dnsv2.HmacTSIG`, or `md5Provider` for HMAC-MD5, which `HmacTSIG` doesn't implement. `md5Provider` was already carrying HMAC-MD5 for the same reason under v1 and is reshaped to the `TSIGSigner` interface.

`Client.Exchange` neither signs nor verifies, unlike v1's, so the update path calls `dnsv2.TSIGSign` before the exchange and `dnsv2.TSIGVerify` on the reply. As in v1 (`client.go:267`, which verifies only when the reply carries a TSIG), an **unsigned** reply is still not rejected while a **wrongly-signed** one is — both directions are pinned by `TestBuildCorrection`.

### RFC 2136 builders

dnsv2 has no `Msg.SetUpdate/Insert/Remove/RemoveName`, so `newUpdate`, `insert`, `remove` and `removeName` reimplement them locally against RFC 2136 §2.3, §2.5.1, §2.5.3 and §2.5.4.

Verified by construction rather than by inspection: the same update built both ways and packed, for A, AAAA, CNAME, MX, NS, TXT, SRV, CAA, PTR, NAPTR and TLSA across all three operations, is **byte-for-byte identical**. A message holding all eleven is identical in length and decodes identically under the v1 parser, differing only in which of two equivalent name-compression offsets it picks.

### Tests

`providers/axfrddns/axfrddnsProvider_test.go` is new — the provider's first tests. TSIG key parsing and algorithm mapping; AXFR against an in-process nameserver unauthenticated and under HMAC-SHA256/SHA512/MD5; rejection of a wrong TSIG secret, a truncated transfer and a reply signed with the wrong key; acceptance of an unsigned reply; the closing-SOA return; a zone whose opening SOA arrives in a message of its own; the reported text of a refused transfer; the RFC 2136 encoding of each update operation; message chunking; and a DDNS update pushed to a server that verifies the TSIG and replies signed.

Each behaviour above was checked by mutation — perturb the source, confirm the matching test goes red, restore, confirm green.

`go test ./...` 71 ok / 0 fail. `go vet` clean, `staticcheck` clean, `golangci-lint` clean apart from the pre-existing `pkg/js/hash.go:21` finding that already carries a `//lint:ignore`. `-race` and `-count=5` clean on the package. `bin/generate-all.sh` leaves the tree empty. `gofmt -l` clean on `providers/axfrddns/`. BIND integration test passes.

### Two things you may want to know

- **A bug in your fork, not reachable from here.** `dnsv2.Msg.Pack()` fails with `dns packing: overflow RR header` on a zero-rdata `ANY` RR whose owner name doesn't compress against the question name — e.g. zone `example.com.` with name `x.net.`, or zone `example.org.` with name `a.example.com.`. `removeName` produces exactly that RR shape. It can't be hit from axfrddns because owner names always come from `dc` and are always in-zone, so I've left the code alone.
- **Sequencing with #4653.** That PR deliberately did *not* extract axfrddns's converter, because this branch rewrites the same function. Worth landing this one first.

`go.mod` and `go.sum` are untouched — both modules were already there.


TomOnTime:
Hey @hnrgrgr!  Are you ok with these changes?  They pass all tests.

TomOnTime:
All tests pass. I'm merging.  Sorry for the fast pace but this PR is blocking other work.

PR #4685, opened 2026-08-02
NEW PROVIDER: NexDNS

morozov-alexey:
- [x] Invite accepted
- [x] Email to Faisal received
- [x] Remaining PR notes resolved

Adds a DNS provider for NexDNS. I run the service and will maintain the provider.

The API addresses one record value at a time, so the provider uses diff2.ByRecord(). A record's id is derived from its name, type and content, so an id stops resolving as soon as the record it names changes; each correction therefore uses only the id it read from the zone it is about to modify.

Supported: A, AAAA, ALIAS, CAA, CNAME, DNAME, DS, MX, NS, PTR, SRV, TLSA, TXT. The SOA record and the apex NS records are maintained by the platform and rejected by the API, so both are left out of the zone contents and the desired state, and DocDualHost is Cannot.

The integration suite passes: 249 cases, 0 failures. It needed two changes in integrationTest/integration_test.go, and I would rather flag them than have you find them.

NEXDNS is added to the not(...) list of the apex-NS group: the apex NS records follow the zone's nameserver group and the API rejects an edit, which is what NAMEDOTCOM and NETCUP are excluded for.

In "IGNORE with modify", one address changes from 10.10.10.10 to 9.9.9.9. That value is only the "before" of the change the following cases make (10 -> 11 -> 12 -> 13), so the range looks incidental to what the test checks, while an RFC 1918 address is refused outright by any provider serving public zones only; 9.9.9.9 is already used elsewhere in the same file. Happy to exclude NEXDNS from the group instead if you prefer, but that would drop the whole of the IGNORE semantics from this provider's coverage over one value the test does not depend on.

Please create the GitHub label 'provider-NEXDNS'.

### DIFF
--- a/integrationTest/integration_test.go
+++ b/integrationTest/integration_test.go
@@ -500,6 +500,7 @@ func makeTests() []*TestGroup {
 				"NAMEDOTCOM",        // "Ignores @ for NS records"
 				"NAMECHEAP",         // cannot handle single NS at apex but does handle dual
 				"NETCUP",            // NS records not currently supported.
+				"NEXDNS",            // The apex NS records follow the zone's nameserver group and are not editable.
 				"PORKBUN",           // Record ignored.
 				"SAKURACLOUD",       // Silently ignores requests to remove NS at @.
 				"TRANSIP",           // "it is not allowed to have an NS for an @ record"
@@ -1992,7 +1993,7 @@ func makeTests() []*TestGroup {
 			),
 			tc("Create some records",
 				a("foo", "1.1.1.1"),
-				a("foo", "10.10.10.10"),
+				a("foo", "9.9.9.9"),
 				aaaa("foo", "2003:dd:d7ff::fe71:aaaa"),
 				mx("foo", 10, "aspmx.l.google.com."),
 				mx("foo", 20, "alt1.aspmx.l.google.com."),
--- a/integrationTest/profiles.json
+++ b/integrationTest/profiles.json
@@ -300,6 +300,11 @@
     "customer-number": "$NETCUP_CUSTOMER_NUMBER",
     "domain": "$NETCUP_DOMAIN"
   },
+  "NEXDNS": {
+    "TYPE": "NEXDNS",
+    "api_token": "$NEXDNS_API_TOKEN",
+    "domain": "$NEXDNS_DOMAIN"
+  },
   "NETLIFY": {
     "TYPE": "NETLIFY",
     "domain": "$NETLIFY_DOMAIN",
--- a/providers/nexdns/auditrecords.go
+++ b/providers/nexdns/auditrecords.go
@@ -1,0 +1,110 @@
+package nexdns
+
+import (
+	"errors"
+	"fmt"
+	"net/netip"
+	"slices"
+	"strings"
+
+	"github.com/DNSControl/dnscontrol/v4/models"
+	"github.com/DNSControl/dnscontrol/v4/pkg/rejectif"
+)
+
+// caaTags are the property tags the API accepts in a CAA record.
+var caaTags = []string{"issue", "issuewild", "iodef"}
+
+// AuditRecords returns a list of errors corresponding to the records
+// that aren't supported by this provider.  If all records are
+// supported, an empty list is returned.
+func AuditRecords(records []*models.RecordConfig) []error {
+	a := rejectif.Auditor{}
+
+	// validation_error - value: Value is required. Last verified 2026-07-31.
+	a.Add("TXT", rejectif.TxtIsEmpty)
+
+	// The API trims a TXT value before storing it, so a leading or trailing
+	// space is dropped without being reported. Left alone that reads as a
+	// permanently pending change - the record never matches what was asked for,
+	// and every run offers the same correction. Refusing it names the reason
+	// once, before anything is written. Interior whitespace is kept and is
+	// therefore not rejected. Last verified 2026-07-31.
+	a.Add("TXT", rejectif.TxtHasTrailingSpace)
+	a.Add("TXT", rejectTxtLeadingSpace)
+
+	// The platform serves public zones only and turns away an address that
+	// cannot be reached from the internet, which is also what stops a zone from
+	// being used to point a public name at a visitor's own network. Catching it
+	// here names the address before a push starts, instead of failing partway
+	// through with some records already written. Last verified 2026-07-31.
+	a.Add("A", rejectUnroutableIP)
+	a.Add("AAAA", rejectUnroutableIP)
+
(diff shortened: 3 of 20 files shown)


TomOnTime (commented):
Hi there!

Thank you for submitting this new provider!  We love to get new submissions!

I'd really like to add NexDNS to the automated testing we do on GHA.  Is there a free tier or some way I can get an API key to do the testing?  It doesn't have to be a real domain (sandbox or unregistered zone is fine).  Since this is an all-volunteer project, we look for any free way to do this testing.

Thanks!
Tom

morozov-alexey:
Hi Tom, happy to. There is now a free account for the project, with no expiry and nothing to pay.

- Test zone: `dnscontrol-test.dev`, unregistered as you suggested, so nothing it does can affect real DNS.
- Token scopes: `zones.read`, `records.read`, `records.write`. No zone creation, so the suite cannot spend the account's one-domain allowance.
- 300 requests/minute, which the suite stays well under.

The `pr_integration_tests.yml` entries (`NEXDNS_DOMAIN` and `NEXDNS_API_TOKEN`) are already in this PR, so only the secret values are missing on your side.

Where would you like the token sent? Happy to use whatever you normally do for this.

TomOnTime:
Fantastic!

I normally use SendSafely.com.  Please send to my email address (tal at whatexit dot org).  I'm also open to any other mechanism (1password, etc).  SendSafely has a free tier.

Tom

morozov-alexey:
Sent via SendSafely to your address.

TomOnTime:
I received the creds. Thanks!  I'll try to get to this in the next few days.

TomOnTime:
Code looks good.  I'll merge this and then move it to the v5 branch.

There's a lot of unit tests for things that don't normally need testing.  I suspect an LLM was used?  I don't mind, that's allowed. However the code does need to be cleaned up a bit.

TomOnTime:
Thank you for contributing this new provider, @morozov-alexey !

A few action items before we can merge this PR:

1. By now you should have received a Github invite to have the "triage" role for this repo.  Please accept the invite so we can assign bugs to you.
2. @fm: Faisal Misle is our “liaison to maintainers”. Please send him an email from your preferred address to dnscontrol so we can stay in touch. Your email address will not be shared, only used for DNSControl communication.  Please email f at faisal dot fm and CC tal at what exit dot org

Thanks again!
Tom

integrationTest/profiles.json
```diff
@@ -300,6 +300,11 @@
     "customer-number": "$NETCUP_CUSTOMER_NUMBER",
     "domain": "$NETCUP_DOMAIN"
   },
+  "NEXDNS": {
```
TomOnTime, line 303: Please put this in alphabetical order. If you run `bin/generate-all.sh` it will do that for you.

providers/nexdns/auditrecords.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	return nil
+}
+
+// rejectTxtLeadingSpace rejects TXT records whose value starts with a space.
+// rejectif has a helper for the trailing case but not for this one, because the
+// integration suite leaves leading whitespace untested - it is nonetheless
+// trimmed the same way here.
+func rejectTxtLeadingSpace(rc *models.RecordConfig) error {
```
TomOnTime, line 97: This seems useful to others.  Pleave move it to `pkg/rejectif/txt.go`

providers/nexdns/auditrecords.go
```diff
@@ -0,0 +1,110 @@
...
+	return nil
+}
+
+// rejectCaaTag rejects CAA records whose property tag the API does not accept.
+func rejectCaaTag(rc *models.RecordConfig) error {
+	if !slices.Contains(caaTags, rc.CaaTag) {
+		return fmt.Errorf("caa tag %q is not supported", rc.CaaTag)
+	}
+	return nil
+}
```
TomOnTime, line 110: This seems useful to others. Please move it to `pkg/rejectif/caa.go`

providers/nexdns/auditrecords_test.go
```diff
@@ -0,0 +1,128 @@
+package nexdns
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/DNSControl/dnscontrol/v4/models"
+)
+
+func TestAuditRecords(t *testing.T) {
```
TomOnTime, line 10: Move these to `pkg/rejectif/*_test.go` as appropriate.

providers/nexdns/convert.go
```diff
@@ -0,0 +1,98 @@
...
+func fromRecordConfig(rc *models.RecordConfig) recordRequest {
+	req := recordRequest{
+		Name: rc.GetLabel(),
+		Type: rc.Type,
+		TTL:  int(rc.TTL),
+	}
+
+	switch rc.Type {
+	case "MX":
+		req.Content = rc.GetTargetField()
```
TomOnTime, line 68: Please move "Content" to the end of each case.  I know this sounds weird, but i try to keep these in the same order as they appear in the struct defined in https://codeberg.org/miekg/dns/src/branch/main/rdata/rdata.go

integrationTest/integration_test.go
```diff
@@ -1992,7 +1993,7 @@ func makeTests() []*TestGroup {
 			),
 			tc("Create some records",
 				a("foo", "1.1.1.1"),
-				a("foo", "10.10.10.10"),
+				a("foo", "9.9.9.9"),
```
TomOnTime, line 1996:
```suggestion
				a("foo", "9.9.9.9"), // Use a non-RFC1918 address since some DSP reject internal IPs
```

providers/nexdns/nexdns_test.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	p, err = newNexdns(map[string]string{"api_token": "nxd_notarealtoken", "api_url": "https://api.example.com/v1"}, json.RawMessage{})
+	if err != nil {
+		t.Fatalf("newNexdns() error = %v", err)
+	}
+	if got := p.(*nexdnsProvider).client.baseURL; got != "https://api.example.com/v1" {
+		t.Errorf("baseURL = %q, want the configured one", got)
+	}
+}
```
TomOnTime: Delete.  Seems superfluous 

```suggestion
```

TomOnTime (changes requested):

morozov-alexey:
Trimmed the tests down to the round trip, the apex NS filter and the audit rules. Left the retry in `api.go` since a push is one request per record, but happy to drop that too.

Invite accepted, and emailed Faisal.

TomOnTime (approved):

PR #4692, opened 2026-08-03
DESEC: Update DNSSEC capabilities

### LINKED ISSUE #4675: DESEC fails integration tests in v5 world
In https://github.com/DNSControl/dnscontrol/issues/4421#issuecomment-5157398637
@androw reports:

> I've run the test suite for desec, results here: https://gist.github.com/androw/9460b645321a13f5323e500f163c63ed Seems to fail because of the NS records (and then because of rate limiting, but not a real problem).

androw:
Definitely close #4675


TomOnTime:
Thanks for the update!

TomOnTime:
P.S.  I'd really love to add DESEC to the automated tests on GHA.  Do they have a free test environment we can access or a free tier?  (anything that's free LOL).

androw:
deSEC is fully free, donation supported. So you can easily register an account and add a domain.

TomOnTime:
> deSEC is fully free, donation supported. So you can easily register an account and add a domain.

Hey that's great!  I'll try that.

PR #4679, opened 2026-08-02
DNSMADEEASY: Unify how labels are made

TomOnTime:
Hi there!

Yes, another request to test a branch. I really appreciate it.

I'm cleaning up a lot of style issues and finding occasional bugs in the process.

Thanks!
Tom

PR #4696, opened 2026-08-04
DNSMADEEASY: Cover label conversion in types_test.go

vojtad:
Test-only change.

After #4679 unified label creation via `dc.LabelFromShort()`, nothing in the test
suite exercised that path — `TestToRecordConfig` asserted only on type, rdata and
`rc.Original`. The zone-apex case, which is what #4679 fixed, had no coverage.

- Label and TTL assertions on every table case, plus two new cases: apex
  (`Name: ""` → `"@"`) and mixed case (`WWW` → `www`).
- `TestApexRoundTrip` — `""` → `"@"` on read, `"@"` → `""` on write.
- `TestSystemNameServerToRecordConfig` — system NS records had no coverage at all.

Reverting `label := dc.LabelFromShort(record.Name)` back to `record.Name` fails
exactly these new assertions, so they guard the fix rather than restate it.

Also switches to `models.MustNewDomainConfig()` and replaces `GetTargetField()`
with `GetRDATA().String()` to keep the file clean under `bin/is_modern.sh`.

Base branch is release_candidate_v5.

### DIFF
--- a/providers/dnsmadeeasy/types_test.go
+++ b/providers/dnsmadeeasy/types_test.go
@@ -7,22 +7,23 @@ import (
 )
 
 func TestToRecordConfig(t *testing.T) {
-	dc, err := models.NewDomainConfig("example.com")
-	if err != nil {
-		t.Fatal(err)
-	}
+	dc := models.MustNewDomainConfig("example.com")
 
 	tests := []struct {
 		name       string
 		record     recordResponseDataEntry
 		wantType   string
+		wantLabel  string
 		wantTarget string
 	}{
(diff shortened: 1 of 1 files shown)


TomOnTime:
Thanks!

vojtad:
Just to confirm - DNSMADEEASY is now passing all tests and there's no more work to be done for now (as far as I know).

TomOnTime:
> Just to confirm - DNSMADEEASY is now passing all tests and there's no more work to be done for now (as far as I know).

`*happy dance*`

PR #4697, opened 2026-08-04
DESEC: Fix DS and DNSKEY tests

TomOnTime:
Hi @androw 

Does this fix the messages in https://github.com/DNSControl/dnscontrol/issues/4675#issuecomment-5169185250 ?

Thanks!
Tom


androw:
I don't think so.
From what I've tried from the deSEC website, we can only:
- create DS record on any non-empty subname
- create DNSKEY record without subname.
Also, I think this correction is creating records not expected by the user (as we are modifying the subname). I don't think we want that. What do you think?

TomOnTime:
Ah!  I took the error message too literally!

I think the solution is to teach dnscontrol to skip the tests where DS is `NOT` at the apex, and DNSKEY *is* at the apex. (Check me to make sure I didn't reverse that)

Look at `pkg/rejectif/ns.go` and clone it to make `ds.go` and `dnskey.go`.  Then you can add to `providers/desec/auditrecords.go`  something like:

```
    a.Add("DS", rejectif.DsNotAtApex)`
    a.Add("DNSKEY", rejectif.DnskeyAtApex)`
```

That will skip those tests and warn users when they appear in dnsconfig.js

Tom

androw:
The first case should already be covered by the CanUseDSForChildren capability. But this solution might be more flexible and allow you to drop this exception (i think it's the only capability like that).

Btw, the DESEC case is already fixed with #4692. I'm going to test this better solution tonight.

TomOnTime:
Sounds good!

FYI: If CanUseDSForChildren fixes it, no need to duplicate that in auditrecords.go

PR #4705, opened 2026-08-05
DESEC, POWERDNS: Add audit rules for DNSKEY records

androw:
DESEC and POWERDNS support DNSKEY records but only at apex.
This PR adds an audit rule and re-enable integration tests.

#4697 no longer required
#4421 v5 integration tests pass for DESEC


TomOnTime (approved):
Perfect1 Thanks!

PR #4732, opened 2026-08-09
WEBSUPPORT: fix MX/SRV empty priority and v5 preparation

mtmn:
Hi @TomOnTime, this is a follow-up to https://github.com/DNSControl/dnscontrol/pull/4584 (apologies for the delay!) that contains the fix for MX/SRV records. It's not ideal as there were some iffy changes upstream but does its job. 

I also reran integration tests and attached them to this PR. I will try to sort out a testing domain in the near future so that I am not blocking your changes.

[20260809_153902.log](https://github.com/user-attachments/files/30873928/20260809_153902.log)


TomOnTime (commented):
Wow!  That's quite a bug you found!

Excellent workaround.  It's good that you isolated it so that when the bug is fixed it can be easily removed.

If you haven't already, please file a bug.

PR #4778, opened 2026-08-18
POWERDNS: Fix SVCB bug

TomOnTime:


jpbede (approved):
Looks good, thanks. Unfortunately I do not use SVCB at all, so I can only review the code as well.

PR #4791, opened 2026-08-20
OVH: Add the ability to disable ListZones() for testing / Fix TXT record handling

TomOnTime:
Fixes https://github.com/DNSControl/dnscontrol/issues/4761

The API key we use to test OVH does not have access to the "list all zones" API call. Therefore we need to be able to work without that call.

When used, ListZones() returns an empty list and the systems that rely (verifying that a zone exists) assume the zone is valid.

This mode is activated by the secret flag `preview --disable-list-zones` (works with `push` too) or when running integration tests, `-disablelistzones`.

Example usage:

```
# Testing:
go test -timeout 1h -failfast -v -args -verbose -profile OVH -disablelistzones
# preview/push:
dnscontrol preview --disable-list-zones
```

This issue was raised in https://github.com/DNSControl/dnscontrol/issues/4761

Now that OVH integration tests work, we discovered that TXT records don't parse correctly. Fixed.

### DIFF
--- a/commands/ppreviewPush.go
+++ b/commands/ppreviewPush.go
@@ -28,6 +28,7 @@ import (
 	"github.com/DNSControl/dnscontrol/v5/pkg/providers"
 	"github.com/DNSControl/dnscontrol/v5/pkg/rfc4183"
 	"github.com/DNSControl/dnscontrol/v5/pkg/zonerecs"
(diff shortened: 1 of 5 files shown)


androw (approved):
Works for me

PR #4794, opened 2026-08-22
FEATURE: updated DMARC_BUILDER for RFC 9989

### LINKED ISSUE #4378: DMARC_BUILDER RFC 9989 updates
With [RFC 9989](https://datatracker.ietf.org/doc/html/rfc9989) being published as a Proposed Standard, the DMARC builder should take into account the new tags.

- `np` policy to apply to non existing subdomains. Its syntax is identical to that of the `p` tag
- `psd` A tag indicating whether the domain is a PSD. Values are y, n, u
-  `t` Should a receiver actually apply the policy stated in the `p` or `sp` tags

Also while some tags are deprecated (namely: `rf` and `ri`), the RFC states they should be ignored, so we should keep them for compatibility reasons but flag them to the people who still use them.

ElliottCepin:
Fixes #4378

Updated DMARC_BUILDER to deprecate `ri`, `rf`, and `pct` tags. Added support for `np`, `t`, and `psd` tags. Updated documentation to represent the changes (removed information about `ri`, `rf`, and `pct` and added information about `np`, `t`, and `psd`). 

`ri`, `rf`, and `pct` logic left in place; they now log warning to the console.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-puppeteer_skip_download=true
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+puppeteer_skip_download=true
--- a/commands/types/dnscontrol.d.ts
+++ b/commands/types/dnscontrol.d.ts
@@ -987,7 +987,6 @@ declare function DKIM_BUILDER(opts: { selector: string; pubkey?: string; label?:
  *   DMARC_BUILDER({
  *     policy: "reject",
  *     subdomainPolicy: "quarantine",
- *     percent: 50,
  *     alignmentSPF: "r",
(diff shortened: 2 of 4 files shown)


TomOnTime (approved):

TomOnTime (approved):

.npmrc
TomOnTime, line 1: What's this?

ElliottCepin: This is an npm environmental variable that I set while troubleshooting that should not have been committed.

TomOnTime: That's fine. Please update the PR without that file.

ElliottCepin: done

PR #4801, opened 2026-08-25
REFACTOR: Simplify boolean returns, hoist regexps, tidy error paths

TomOnTime:
Small, mechanical cleanups from a repo-wide Go style pass. **No behavior change.**

### Changes
- **Boolean returns** — collapse `if cond { return true }; return false` into `return cond` (`models/stutter.go`, `pkg/mustbe/hosts.go`, `providers/gcloud/gcloudProvider.go`).
- **Regexp hoisting** — move constant `regexp.MustCompile(...)` calls out of frequently-called functions into package-level vars so they compile once (`commands/ppreviewPush.go`, `providers/desec/protocol.go`, `providers/softlayer/softlayerProvider.go`).
- **Error paths** — drop a redundant `else` after a terminating branch, lowercase/depunctuate a few error strings, and fix a missing space in an autodns error message (`providers/autodns/api.go`, `providers/route53/route53Provider.go`, `providers/openwrt/api.go`).

### Verification
`gofmt -l` clean · `go vet ./...` clean · `go build ./...` clean · `golangci-lint run` (v2.13.1) → 0 issues.

### DIFF
--- a/models/stutter.go
+++ b/models/stutter.go
@@ -16,6 +16,7 @@ func doesStutter(name, origin string) bool {
 	if name == "@" {
 		return false
 	}
+	// Return true if name is the origin (should be "@") or ends in the origin.
 	return name == origin || strings.HasSuffix(name, "."+origin)
 }
 
--- a/models/stutter.go
+++ b/models/stutter.go
@@ -16,10 +16,7 @@ func doesStutter(name, origin string) bool {
 	if name == "@" {
 		return false
 	}
-	if name == origin || strings.HasSuffix(name, "."+origin) {
-		return true
-	}
-	return false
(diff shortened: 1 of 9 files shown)


models/stutter.go
```diff
@@ -16,10 +16,7 @@ func doesStutter(name, origin string) bool {
 	if name == "@" {
 		return false
 	}
-	if name == origin || strings.HasSuffix(name, "."+origin) {
-		return true
-	}
-	return false
+	return name == origin || strings.HasSuffix(name, "."+origin)
```
TomOnTime, line 20:
```suggestion
	// Return true if name is the origin (should be "@") or ends in the origin.
	return name == origin || strings.HasSuffix(name, "."+origin)
```

PR #4817, opened 2026-08-27
fix(mustbe): Fix 32-bit compatibility

### LINKED ISSUE #4816: Failure to compile on 32-bit ARM GNU/Linux
**Describe the bug**
Unfortunately, version 5.0.0 fails to compile on 32-bit ARM GNU/Linux

```
# github.com/DNSControl/dnscontrol/v5/pkg/mustbe
pkg/mustbe/numbers.go:122:19: math.MaxUint32 (untyped int constant 4294967295) overflows int
```

**To Reproduce**
1. `git clone https://github.com/DNSControl/dnscontrol.git`
2. `cd dnscontrol`
3. `go build .`

**Expected behavior**
Successful compilation on 32-bit ARM GNU/Linux

**DNS Provider**
 - 1.1.1.1

**Additional context**
I strongly suspect that the error is also reproducible on 32-bit x86 GNU/Linux, not only 32-bit ARM GNU/Linux

TomOnTime:
Fixes https://github.com/DNSControl/dnscontrol/issues/4816

* mustbe.Uint32() now compiles on 32-bit systems
* Added doc for how to test for 32-bit compatibility

### DIFF
--- a/documentation/advanced-features/integration-tests.md
+++ b/documentation/advanced-features/integration-tests.md
@@ -1,4 +1,9 @@
-### Integration Tests
+# Integration Tests
+
+- [Integration Tests](#integration-tests)
+  - [Configuration](#configuration)
+  - [Running a test](#running-a-test)
+  - [Bonus: How to test for 32-bit compatibility](#bonus-how-to-test-for-32-bit-compatibility)
 
 This is a simple framework for testing dns providers by making real requests.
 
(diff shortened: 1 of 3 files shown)


robertkirkman (approved):
It fixes the problem for me and when I apply this, `go build .` and `go test -v ./...` now work on 32-bit ARM GNU/Linux!

PR #2708, opened 2023-12-10
TRANSIP: Fix TXT quoting

blackshadev:
Fixes !2702

- Fix TXT quoting, by not quoting at all. TransIP doesn't need it (any more)
- Fix and added some auditrecord entries to let integration tests pass
  - A TransIP TXT record cannot be empty
  - TransIP trims spaces from TXT records.

PR #4824, opened 2026-08-28
TRANSIP: Write TXT records unquoted (v5 regression of #2708)

cafferata:
## What this fixes

Since v5.0.0 the TransIP provider writes TXT record content in RFC presentation format (enclosed in double-quotes). TransIP stores that content verbatim, so the quote characters become literal data in DNS. Every TXT record managed via TransIP (SPF, DKIM, DMARC, verification tokens) is served with a leading/trailing `"`, which breaks anything that expects a bare value. For example, an SPF validator reports `No SPF record found` / `should include ...`. This is a regression of #2708 ("TRANSIP: Fix TXT quoting").

Root cause, in `providers/transip/transipProvider.go` `recordToNative()`: `config.GetRDATA().String()` returns the quoted presentation form for TXT. #2708 had set this field to the unquoted form; the v5 RDATA refactor reintroduced the quoting.

## The fix

Mirror the inwx provider (another provider that stores raw TXT):

- Write the raw, unquoted value with `GetTargetTXTJoined()` for TXT.
- Read it back with the `TxtDontParse` flag so the content is parsed as raw data. The flag only affects TXT; other record types are unchanged.

## Reproduction

```text
$ dig +noall +answer TXT deprobleemoplosser.nl @ns0.transip.net
deprobleemoplosser.nl.  86400  IN  TXT  "\"v=spf1 include:_spf.protonmail.ch include:spf.flowmailer.net -all\""
```

The stored value is `"v=spf1 include:_spf.protonmail.ch include:spf.flowmailer.net -all"`, including the surrounding `"` characters. `dnscontrol preview`/`push` print the normalized model (`v=spf1 ...`), so the diff looks correct; only `dig` reveals the literal quotes, and a re-run reports no drift, so the zone stays silently broken.

## Testing

- Updated `TestNativeToRecordUsesV3RecordConfig` (the TXT case now reflects the raw/unquoted contract) and added `TestRecordToNativeTXTUnquoted` (write-side plus round-trip) and `TestNativeToRecordHealsLegacyQuotedTXT` (a legacy quoted value is read back with quotes intact so a push rewrites it unquoted).
- Regenerated the `recordToNative` golden file with `-update`; the only change is TXT `content` fields losing their enclosing quotes.
- `go test ./providers/transip/...` and `go build ./...` pass.
- Verified against a live TransIP zone: after this change the affected TXT records are served unquoted, and a subsequent `preview` shows no quoting drift.


blackshadev (approved):
Lgtm. Due to limited time I am not able to test it. But the code looks good to me.

providers/transip/transipProvider.go
```diff
@@ -277,16 +278,26 @@ func (n *transipProvider) recordsToNativeObserved(records models.Records) ([]dom
...
+		content = config.GetTargetTXTJoined()
+	}
 	return domain.DNSEntry{
 		Name:    config.Name,
 		Expire:  int(config.TTL),
 		Type:    config.Type,
-		Content: config.GetRDATA().String(),
+		Content: content,
 	}, nil
 }
```
TomOnTime: - Rename "config" to "rc".
- Use a var/switch instead of if/then.
- Less verbose comments.

```suggestion
func recordToNative(rc *models.RecordConfig) (domain.DNSEntry, error) {
	var content string
	switch rc.TypeNum {
	case dnsv2.TypeTXT:
		// TransIP stores the TXT "content" field verbatim.
		content = rc.GetTargetTXTJoined()
	default:
		content = rc.GetRDATA().String()
	}
	return domain.DNSEntry{
		Name:    rc.Name,
		Expire:  int(rc.TTL),
		Type:    rc.Type,
		Content: content,
	}, nil
}
```

providers/transip/transipProvider.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/DNSControl/dnscontrol/v5/models"
```
TomOnTime, line 12:
```suggestion
	dnsv2 "codeberg.org/miekg/dns"
	"github.com/DNSControl/dnscontrol/v5/models"
```

providers/transip/transipProvider.go
```diff
@@ -277,16 +278,26 @@ func (n *transipProvider) recordsToNativeObserved(records models.Records) ([]dom
...
-		Content: config.GetRDATA().String(),
+		Content: content,
 	}, nil
 }
 
 func nativeToRecord(entry domain.DNSEntry, dc *models.DomainConfig) (*models.RecordConfig, error) {
-	return dc.NewRecordConfigParse(dc.LabelFromShort(entry.Name), uint32(entry.Expire), entry.Type, entry.Content)
+	// TransIP returns TXT content unquoted (see recordToNative), so parse it as
+	// raw data. TxtDontParse only affects TXT; other types are unchanged.
+	return dc.NewRecordConfigParse(dc.LabelFromShort(entry.Name), uint32(entry.Expire), entry.Type, entry.Content, nrc.Flags{TxtDontParse: true})
```
TomOnTime: - Style: newline before nrc.Flags{}.  (I need to add this to the style guide)

```suggestion
	// TransIP returns TXT content unquoted (see recordToNative), so parse it as raw data.
	return dc.NewRecordConfigParse(dc.LabelFromShort(entry.Name), uint32(entry.Expire), entry.Type, entry.Content,
		nrc.Flags{TxtDontParse: true})
}
```

TomOnTime (dismissed):

PR #4828, opened 2026-08-28
LINODE: BUGFIX: too picky about hyphens in names of SRV records

### LINKED ISSUE #4812: LINODE: too picky about hyphens in names of SRV records
The LINODE provider does not allow hyphens in SRV service and protocol names.

While trying to add records like

```
_xmpp-server._tcp 28800 SRV 20 0 5269 xmpp-server.example.org.
```

I am receiving the error:
```
provider linode Error: SRV Record must match format "_service._protocol" not _xmpp-server._tcp
```

Note that I can manually add SRV records like these using Akamai's web UI.

The culprit appears to be this regexp:

https://github.com/DNSControl/dnscontrol/blob/6fcc3de94a4b4a7ef109001a2647206e3b95ba04/providers/linode/linodeProvider.go#L47

which should probably be generalized to include hyphens within service and protocol names.
I.e. replace the occurences of `\w+`  with `[-\w]+`.

TomOnTime:
Fixes https://github.com/DNSControl/dnscontrol/issues/4812

* Do the validation in auditrecords.go so that errors are caught earlier.
* Extract service/protocol using parsig, not regex. Use less strict rules to validate labels.  It's up to the API to make the final call. We can't emulate the API perfectly (that's a fools errand) therefore we shouldn't try.


providers/linode/auditrecords.go
```diff
@@ -1,10 +1,19 @@
...
+
 	"github.com/DNSControl/dnscontrol/v5/models"
 	"github.com/DNSControl/dnscontrol/v5/pkg/rejectif"
 )
 
+// srvLabelRegexp matches a valid SRV record label: "_service._protocol",
+// optionally followed by a subdomain (e.g. "_smtp._tcp.sub.domain").  This is
+// used both by AuditRecords to validate labels and by toReq to extract Service
+// and Protocol from the labels.
+var srvLabelRegexp = regexp.MustCompile(`^_(?P<Service>[\w-]+)\._(?P<Protocol>[\w-]+)(?:\.[\w-]+)*$`)
```
dairiki: In order to create a SRV record on a subdomain of the zone apex, the Linode API requires that the subdomain be included in the "protocol" parameter.
As it is now, this regexp captures the protocol to _Protocol_ but does not capture the subdomain.

```suggestion
var srvLabelRegexp = regexp.MustCompile(`^_(?P<Service>[\w-]+)\._(?P<Protocol>[\w-]+(?:\.[\w-]+)*)$`)
```

dairiki: On further thought, it might be better to be stricter about funny characters in the service name.

The Linode API doesn't reject requests with service or protocol names it doesn't like.  It just sanitizes them to its liking, then creates a SRV record on a different name than what was requested.
In practice, this means every `dnscontrol push` will delete the unexpected record that Linode actually created, and try to add the record that Linode didn't like.

As near as I can tell here are the sanitization rules that the Linode API follows:

For service names:
1. strip all leading underscores
2. truncate at the first character that is not a letter, number, or hyphen

For protocol names:
1. strip all leading underscores

Here's a regexp that rejects names that would be mangled by Linode.  (I have tested this.)

```suggestion
var srvLabelRegexp = regexp.MustCompile(`^_(?P<Service>[[:alnum:]-]+)\._(?P<Protocol>[[:alnum:]-][\w-]*(?:\.[\w-]+)*)$`)
```

dairiki (commented):
Thank you for working on this!

I've tested this.  It works for hyphenated service names. 

As it stands, this does not work for SRV records on subdomains.  See in-line comments for suggestions on a fixed and improved regexp.

documentation/provider/linode.md
```diff
@@ -64,7 +68,13 @@ the _default TTL_.
 The provider will automatically round up your TTL to one of these values. For example, 600 seconds would become 3600
 seconds, but 300 seconds would stay 300 seconds.
 
-Linode requires [`SRV`](../language-reference/domain-modifiers/SRV.md) records to have a non-zero priority.
+Linode requires [`SRV`](../language-reference/domain-modifiers/SRV.md) records
+to have a non-zero priority. Linode validates service name and protocol more
```
dairiki: A SRV priority of zero works with the Linode API. (I've tested it.) The priority, for both SRV and MX records, does seem to be limited to the range [0-255], however.

providers/linode/srvlabel.go
```diff
@@ -0,0 +1,51 @@
...
+		Attempts to document Linode's algorithm are here:
+		https://github.com/DNSControl/dnscontrol/issues/4812#issuecomment-5444147757
+
+	*/
+
+	return service[1:], protocol[1:], nil
+}
+
+func validateSrvLabelHelper(label string) (string, string, error) {
+	parts := strings.SplitN(label, ".", 3)
```
dairiki: This latest change has re-broken the ability to put SRV records in subdomains.

The only way to pass a subdomain to the Linode API is to include it in the "protocol" parameter.  (I have no idea whether this is intentional or not, but there appears to be no other way to do it.  As you have noted, it's a pretty crummy API design.)

I.e. if our zone is "example.org", and we want to create a SRV record at "_srv._proto.sub-domain.example.org", we need to pass `service="srv"` and `protocol="proto.sub-domain"` to Linode's API.

Here, the subdomain is in parts[2], which is currently being discarded.

dairiki (commented):
SRV priority of zero is allowed.

The latest changes have re-broken SRV records on subdomains.

I'm working on running the integration tests.  Some failures were reported the first time I ran them.  I ran them again, with the output captured for better examination, and ... no errors.  (My familiarity with go is zero, so it will take awhile to suss this out.)

There appear to be no integration tests that test the creation of SRV records on subdomains.

documentation/provider/linode.md
TomOnTime: Removed.

providers/linode/srvlabel.go
TomOnTime: See if f338e260 fixes that.

PR #4859, opened 2026-09-08
BUGFIX: print-ir and --ir should round-trip

TomOnTime:
# Issue

"dnscontrol print-ir --out filename" and "dnscontrol preview --ir filename" should round-trip.  They don't in v5.0.3

# Resolution

Output the .RDATA field as a string (zonefile format) and parse it back on read.  The .String() function and NewRecordConfigParse() must always round-trip (IIRC), making this a better format.

This is a big change, since "rdata" is now a string instead of a json object.  However if we used a json object, we'd have to deal with the fact that dnsv2 doesn't properly round trip all of its json.

### DIFF
--- a/models/record.go
+++ b/models/record.go
@@ -69,20 +69,71 @@ type RecordConfig struct {
 	UnknownTypeName string `json:"unknown_type_name,omitempty"`
 }
 
-// MarshalJSON marshals RecordConfig.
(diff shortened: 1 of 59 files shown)


TomOnTime:
FYI: The maintainer of dns prefers  RFC 8427 format, which doesn't help you because it encodes everything in hex.

PR #4866, opened 2026-09-09
HOSTINGDE: Fix SOA mailbox with a host being dropped

jonathan8devs:
hosting.de keeps the contact address of a zone in `zoneConfig.emailAddress` rather than in an SOA record, so the provider turns the mailbox of `SOA()` into an address. It only handled a mailbox without a dot, appending `@` and the zone name, and skipped everything else. The spelling that [`SOA.md`](../language-reference/domain-modifiers/SOA.md) uses as its own example, `hostmaster.example.com.`, therefore did nothing at all.

Worse, `SOA()` also accepts a plain email address, and `mustbe.SoaMailbox()` turns `admin@example.com` into the mailbox `admin.example.com`, which has no trailing dot either. The provider appended the zone on top of that and wrote `admin.example.com@example.com`.

This adds `soautil.BindMailToRFC5322()` as the counterpart to the existing `RFC5322MailToBind()` and uses it. A mailbox that names its host becomes the address, one without is qualified with the zone as before.

Verified against a hosting.de account, preview and push, nothing else in the zone touched:

```
hostmaster.example.com.  ->  hostmaster@example.com   next preview clean
admin@example.com        ->  admin@example.com        next preview clean
hostmaster               ->  hostmaster@example.com   unchanged from before
```

The API accepts an address outside the zone, and rejects one ending in a period with `Code:13012`, so neither form causes trouble.

Behaviour change worth noting: the dotted spelling used to be a guaranteed no-op for this provider, and now it takes effect. Anyone using it may see one `Changing SOA Mail` line on the first preview after upgrading. Nothing is written without showing up there first.

`BindMailToRFC5322()` sits next to its counterpart in `pkg/soautil`, but it is only used here. Happy to move it into the provider if you would rather not grow that package.

### DIFF
--- a/documentation/provider/hostingde.md
+++ b/documentation/provider/hostingde.md
@@ -75,9 +75,13 @@ The `HOSTINGDE` provider **ignores the default nameserver set** defined in your 
 Instead, it uses hosting.de's nameservers (`ns1.hosting.de.`, `ns2.hosting.de.`, and `ns3.hosting.de.`) by default, regardless of your account settings.
 Using the `default_ns` metadata, the default nameserver set can be overwritten.
 
-## SOA records
+## Caveats
 
-hosting.de keeps the contact address of a zone in the zone configuration rather than in an SOA record, so the mailbox of [`SOA()`](../language-reference/domain-modifiers/SOA.md) ends up there.
+### SOA
+
+hosting.de derives the primary nameserver of a zone from its nameserver set and offers no way to override it, so the `ns` parameter of [`SOA()`](../language-reference/domain-modifiers/SOA.md) is ignored.
+
+The contact address of a zone lives in the zone configuration rather than in an SOA record, so the mailbox of `SOA()` ends up there.
 A mailbox that names its own host is used as the address, so `hostmaster.example.com.` becomes `hostmaster@example.com`.
 One without a host is qualified with the zone name.
 The address may not end in a period; the API rejects that.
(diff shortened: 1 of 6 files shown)


8devs-service:
Integrated, so #4864 can be closed. The branch now carries its test split, its documentation and the message change.

Two things came out differently.

`Modify SOA ns` is still in the HOSTINGDE group, with `.AllowNoChanges()` rather than commented out. That is the single assertion the case trips over: `helpers_integration_test.go:241` fatals when the first run produces no corrections, and relaxing it lets the case keep running, which still shows the provider does not churn on an ns-only diff. Happy to comment it out instead if you would rather have it that way.

`Modify SOA mbox` is enabled again. It is the regression test for the fix in the first commit: before it, a mailbox ending in a period produced no correction at all, which is the same "Expected changes, but got none" the ns case gives, for a different reason.

The trailing dot on the mailbox went away with the block it lived in, since `soaMailToEmail()` trims the mailbox and never appends one. And `placeholderSOA` returns its error now instead of nil, with the caller passing it up: same intent as your edit, no `Must...` in the push path, without the nil dereference at the call site.

Full suite against the account: 282 passed, 61 skipped, 0 failed, in about 16 minutes. `44:SOA` skips with `excluded by not("HOSTINGDE")`, and the HOSTINGDE group runs all seven cases, the ns one producing no correction as documented:

```
45:SOA:Create_SOA_record   Updating SOARecord ... to (refresh=10000 retry=2400 expire=604800 negativettl=3600 ttl=300)
                           Changing SOA Mail from eee@cloudflare.com to dns@cloudflare.com
45:SOA:Modify_SOA_ns       no correction
45:SOA:Modify_SOA_mbox     Changing SOA Mail from dns@cloudflare.com to eee@cloudflare.com
45:SOA:Modify_SOA_refres   Updating SOARecord ... refresh=10001 ...
45:SOA:Modify_SOA_retry    Updating SOARecord ... retry=2401 ...
45:SOA:Modify_SOA_expire   Updating SOARecord ... expire=604801 ...
45:SOA:Modify_SOA_minttl   Updating SOARecord ... negativettl=3601 ...
```

The group is repeatable: the empty-zone step resets the SOA values to the provider's defaults, so a second run of it behaves the same.

PR #4878, opened 2026-09-13
fix(p/FORTIGATE): ask the TLS and HTTP debug settings as yes/no questions in init

cafferata:
While looking into https://github.com/DNSControl/dnscontrol/issues/4381 (the INWX sandbox question in `dnscontrol init` being confusing), FortiGate turned out to have two prompts of the same kind. `insecure_tls` and `debug_http` are on/off settings, but `init` asks for them as free text and only the `?` help explains that the value has to be `true`. The provider compares against `true` (case insensitive), so answering `yes` or `1` silently leaves TLS verification on.

Both settings are now yes/no questions with no as the default. Answering yes writes `"true"`, so the resulting `creds.json` is the same as before.

Before, after typing `?` at the TLS prompt:

```text
API settings for FortiGate: https://docs.fortinet.com/
? FortiGate host [host] (required) fw.example.com
? VDOM (required) root
? API key [apiKey] (required) [? for help] *****
? Set to "true" to skip TLS certificate verification when connecting to the FortiGate.
? Skip TLS verification (optional) [insecure_tls] (optional)
```

After, after typing `?` at the TLS prompt:

```text
API settings for FortiGate: https://docs.fortinet.com/
? FortiGate host [host] (required) fw.example.com
? VDOM (required) root
? API key [apiKey] (required) [? for help] *****
? Answer yes only when the FortiGate uses a self-signed certificate.
? Skip TLS certificate verification when connecting to the FortiGate? [insecure_tls] (y/N)
```

Answering `y` for TLS and accepting the default for HTTP debugging gives the following, and the `creds.json` entry is identical to answering `true` and leaving HTTP debugging blank before:

```text
API settings for FortiGate: https://docs.fortinet.com/
? FortiGate host [host] (required) fw.example.com
? VDOM (required) root
? API key [apiKey] (required) [? for help] *****
? Skip TLS certificate verification when connecting to the FortiGate? [insecure_tls] Yes
? Log HTTP requests and responses for debugging? [debug_http] No

{
  "example": {
    "TYPE": "FORTIGATE",
    "apiKey": "token",
    "host": "fw.example.com",
    "insecure_tls": "true",
    "vdom": "root"
  }
}
```

The output above comes from running the FortiGate credential prompts (`collectFields()` with the real survey prompts) in a pseudo terminal, on `main` and on this branch. `go test ./providers/fortigate/ ./commands/` passes.

@KlettIT, as the FortiGate maintainer, could you take a look?

Part of https://github.com/DNSControl/dnscontrol/issues/4381

## AI-attributie


TomOnTime (approved):

TomOnTime:
Merging this ahead of the release. It's low risk.

@KlettIT: we haven't heard from you in a while and I don't have an email address for you.  Is this provider still being actively maintained?

KlettIT:
@TomOnTime Yes, we are still checking the provider (since we use it ourselves). However, it is only a side project that I occasionally look at. How can I best send you our email? I would prefer not to write it in the comments here.

KlettIT (approved):

TomOnTime:
> @TomOnTime Yes, we are still checking the provider (since we use it ourselves). However, it is only a side project that I occasionally look at. How can I best send you our email? I would prefer not to write it in the comments here.

Good to know!  This is an all volunteer project so I don't expect super fast replies.

The best way to contact me is  to send email to `tal at whatexit dot org` and I'll get your email address from the From line.

By the way... I used to admin some really big Fortinet devices at my previous job. Good times.

Best,
Tom

PR #4880, opened 2026-09-15
fix(p/AUTODNS): send bare FQDN as MX value, not full RDATA

dkorunic:
## Problem

`recordsToNative()` sets an MX record's `Value` to `rc.GetRDATA().String()` — which includes the preference — while *also* setting the dedicated `Pref` field. AutoDNS receives `"10 mail.example.net."` in a field that must contain only the target FQDN, and rejects the whole zone update:

```
EF020541  The MX resource record value is invalid. The value must be an FQDN.
          objects: [{"type":"rr/value","value":"10 10"}]
E0202     Zone could not be updated on the name server.
```

Because the provider replaces the entire zone on every push, this makes **any** `push` fail for a zone containing an MX record — even when the MX itself is unchanged and the only correction is to some other record. `preview` is unaffected, so it only surfaces at push time.

Reproduced against the live AutoDNS gateway on v5.0.4; the same code is on `main`. In my own use it blocks 87 of 161 zones.

## Cause

v4 built the value from `record.GetTargetField()` (the bare target) and set `Pref` separately. The v5 move to the unified `GetRDATA()` accessor changed that for MX. The correct code was already present in the same switch arm, commented out under `// If that doesn't work, try:` — it turns out it doesn't work.

SRV and CAA are **not** affected: their `GetRDATA().String()` output already matches the format the old hand-rolled `fmt.Sprintf` calls produced, which is why removing those special cases was safe. MX is the one type where the API splits the representation across two fields.

## Fix

Use `f.Mx` for the value and keep the preference in `Pref`.

## Tests

The outbound direction (`recordsToNative`) had no test coverage, which is why this shipped — the existing `TestToRecordConfig` only exercises the inbound direction. Added `TestRecordsToNative`.

On the old code it fails exactly as the gateway reported:

```
--- FAIL: TestRecordsToNative/MX
    autoDnsProvider_test.go:77: Value = "10 mail.example.net.", want "mail.example.net."
```

A and CNAME pass unchanged, isolating the defect to MX. Full `go test ./...`, `go vet`, `go build` and `gofmt` are clean.

## Verification against the live provider

Built from this branch and pushed a real zone whose only pending change was an unrelated TXT/SPF edit:

- before: `FAILURE! ... EF020541`
- after: `SUCCESS! Done. 1 corrections.`
- confirmed at the authoritative nameserver that the MX landed intact as `10 <target>.` and the apex A was untouched.


TomOnTime (approved):

TomOnTime:
CC @arnoschoon for visibility.

Thanks for fixing this!

I don't have an API key to AutoDNS so I can't test AutoDNS myself.  If AutoDNS has a free tier, or if you can create an API key that is restricted to only access a single test domain, it would really help me support AutoDNS better. [Instructions](https://docs.dnscontrol.org/developer-info/byo-secrets).

PR #4882, opened 2026-09-15
fix(p/AUTODNS): transmit an MX preference of 0

dkorunic:
## Problem

`ResourceRecord.Pref` is a plain `int32` tagged `omitempty`, so an MX preference of **0** is dropped from the request body entirely. AutoDNS then applies its own default, and the pushed record silently disagrees with the config.

A zone declaring

```js
MX("@", 0, "mail.example.net.")
```

comes back from the provider as preference **10**, which shows up as a `MODIFY` on the next run that `push` cannot settle:

```
± MODIFY example.com MX (10 mail.example.net. ttl=3600) -> (0 mail.example.net. ttl=3600)
```

Zero is an ordinary MX preference — it is what a single-target "deliver here, no alternatives" setup looks like, and it is what Microsoft 365 and several hosted-mail providers hand out — so the field has to go out on the wire.

Marshalling the record on current `main` shows the field simply missing:

```
{"name":"","ttl":300,"type":"MX","value":"mail.example.net."}
```

## Why this surfaced only now

It was masked. Before #4880, any zone containing an MX record failed to push at all (`EF020541`), so the dropped preference never reached the gateway. Fixing the value encoding let these requests through for the first time and exposed the second defect underneath.

I hit it in production immediately after #4880 merged: a zone of mine with `MX("@", 0, ...)` came back as preference 10.

## Fix

Make `Pref` a `*int32` so an unset preference and a preference of 0 stop sharing a representation. Records that never set it still omit it, so nothing changes for any other type:

| | marshalled |
|---|---|
| before, pref=0 | `{"name":"","type":"MX"}` — dropped |
| after, pref=0 | `{"name":"","pref":0,"type":"MX"}` |
| after, pref=10 | `{"name":"","pref":10,"type":"MX"}` |
| after, unset (A record) | `{"name":"","type":"A"}` — unchanged |

Read paths go through a small `pref()` accessor that treats an absent `"pref"` as 0, preserving today's behaviour for both MX and SRV.

## Tests

`TestRecordsToNativeMarshalsZeroPreference` asserts on the **marshalled JSON** rather than the struct. That distinction matters: the struct field reads back as `0` and looks entirely correct — the value only disappears at encoding time, which is why a struct-level assertion would pass on the broken code.

Verified against the pre-fix sources on `main`:

```
--- FAIL: TestRecordsToNativeMarshalsZeroPreference
    marshalled MX record has no pref field: {"name":"","ttl":300,"type":"MX","value":"mail.example.net."}
```

`TestRecordsToNative` also gains an `MXZeroPreference` case. `go build`, `go vet`, `gofmt` and the full `go test ./...` are clean.

## Note

Follow-up to #4880, which is already merged — `main` currently carries this bug.


providers/autodns/autoDnsProvider.go
```diff
@@ -396,9 +397,9 @@ func toRecordConfig(dc *models.DomainConfig, record *ResourceRecord) (*models.Re
 	ttl := uint32(record.TTL)
 	switch record.Type {
 	case "MX":
-		rc, err = dc.NewRecordConfig(label, ttl, dnsv2.TypeMX, uint16(record.Pref), record.Value)
+		rc, err = dc.NewRecordConfig(label, ttl, dnsv2.TypeMX, uint16(record.pref()), record.Value)
 	case "SRV":
-		rc, err = dc.NewRecordConfigParse(label, ttl, dnsv2.TypeSRV, fmt.Sprintf("%d %s", record.Pref, record.Value))
+		rc, err = dc.NewRecordConfigParse(label, ttl, dnsv2.TypeSRV, fmt.Sprintf("%d %s", record.pref(), record.Value))
```
TomOnTime: We have a special flag for this weird situation. However, please verify this works.

```suggestion
		rc, err = dc.NewRecordConfigParse(label, ttl, dnsv2.TypeSRV, record.pref(), record.Value, nrc.Flag{SrvWeirdSplit: true})
```

providers/autodns/autoDnsProvider.go
```diff
@@ -230,7 +230,8 @@ func recordsToNative(recs models.Records) ([]*models.Nameserver, uint32, []*Reso
 				// and the gateway rejects the entire zone update with
 				// EF020541 "The MX resource record value is invalid.".
 				f := rc.AsMX()
-				resourceRecord.Pref = int32(f.Preference)
+				pref := int32(f.Preference)
+				resourceRecord.Pref = &pref
```
TomOnTime: I'm not sure if this will work but give it a try.

```suggestion
				resourceRecord.Pref = new(int32(f.Preference))
```

dkorunic: Verified — it works, adopted in e3b669c8.

`new()` accepts an expression as of Go 1.26, and this module targets `go 1.27`, so `new(int32(f.Preference))` compiles and yields the right value. Confirmed both by building and by running it (`10` in, `10` out). Nice — it drops the temporary.

providers/autodns/autoDnsProvider.go
dkorunic: Verified, and adopted the flag in e3b669c8 — but it has to go to `NewRecordConfig`, not `NewRecordConfigParse`:

```go
rc, err = dc.NewRecordConfig(label, ttl, dnsv2.TypeSRV, record.pref(), record.Value,
    nrc.Flags{SrvWeirdSplit: true})
```

Three things stopped the suggested line as written:

1. `NewRecordConfigParse` panics on this flag outright — `models/record_factory.go:83`: `panic("NewRecordConfigParse() incompatible with SrvWeirdSplit")`.
2. Its `data` parameter is a `string`, so passing `record.pref()` (an `int32`) doesn't compile.
3. The type is `nrc.Flags`; there's no `nrc.Flag`.

`NewRecordConfig` is where the flag is handled (`record_factory.go:40`), and its handler does exactly the `fmt.Sprintf("%d %s", ...)` this replaces, so the parsed result is identical — the existing `TestToRecordConfig/SRV` case still passes unchanged, which is what I used to confirm the swap is behaviour-preserving. It also now matches how dnsimple, hedns, inwx, namedotcom and vercel call it.

PR #4883, opened 2026-09-15
feat: let providers declare a record identity for duplicate checks

Alice39s:
I ran into this while moving a geo-routed zone to DNSControl.

DNSPod, Gcore, Huawei Cloud and ClouDNS keep one record per line, so several
records share a name and type. That is how these providers express per-region
answers.

Two checks block those configs before the provider ever sees them.

`checkCNAMEs` rejects a second CNAME under the same name, so a geo-routed CNAME
set cannot be declared at all.

`checkDuplicates` compares name, type and RDATA and never looks at provider
metadata, so two records that differ only by line count as duplicates even
though the provider stores them separately.

The `R53_WEIGHT` example in the docs fails the same check, so this is not
specific to any one provider.

The change adds an optional identity function to the provider interface. A
provider that declares one gets its identity text appended to the duplicate
key. A provider that declares nothing keeps today's rules.

Route 53 weighted records and Cloudflare flattened CNAMEs already rely on
hard-coded exceptions to the same checks. A generic hook gives other providers
a supported way in.


TomOnTime:
Hi there!

Thanks for submitting this!  I didn't even realize it was a problem. Thank you for reporting it.

Let me suggest an easier way to fix this.  It might not be better or equivalent. I'm interested in your feedback.

What if we added a metadata flag similar to [DISABLE_REPEATED_DOMAIN_CHECK](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/pkg/js/helpers.js#L2008), perhaps called DISABLE_DUPLICATE_RECORD_CHECK?  Records with that metadata would be exempt from the checkDuplicates and checkCNAMEs checks.

The benefit would be that it would not require provider changes.  The downside is that it could lead to some false negatives which will later be detected/rejected at `push` time.

Tom

Alice39s:
Thanks for the suggestion, and for spelling out the trade-off.

I would like to keep the provider-declared identity. The main argument for it is already in this repo.

The two testgroups already in [`integrationTest/integration_test.go`](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/integrationTest/integration_test.go#L1317) on main describe exactly this shape. [`TENCENTDNS_LINE_WEIGHT`](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/integrationTest/integration_test.go#L1345) declares one name, type and value twice with different `tencentdns_line`. [`R53_WEIGHT_CNAME`](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/integrationTest/integration_test.go#L1317) declares one name twice as a CNAME.

Both pass today because the integration path goes through `zonerecs.CorrectZoneRecords` and never runs `pkg/normalize`. Put the same config through `check` or `preview` and both fail.

The provider side already declares this identity. `recordMetadataComparable` in providers/tencentdns [carries the comment](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/providers/tencentdns/tencentdnsProvider.go#L222) "makes DNSPod's record line and weight part of a record's identity", route53 has [`r53ComparableFunc`](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/providers/route53/route53Provider.go#L664), and `pkg/normalize` [reads provider metadata](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/pkg/normalize/validate.go#L830) when it checks R53 weighted groups. Duplicate detection is the one place that computes its own key.

For DNSPod the key is measurable. The account exposes a read-only conflict check, `DescribeRecordImpactInfo`, and it returns the same [`InvalidParameter.DomainRecordExist`](https://github.com/TencentCloud/tencentcloud-sdk-go/blob/tencentcloud/dnspod/v1.3.131/tencentcloud/dnspod/v20210323/errors.go#L321) that [`CreateRecord`](https://www.tencentcloud.com/document/api/1427/56180) returns. Unknown parameters are rejected, so the input set can be enumerated: it takes `SubDomain`/`RecordLine`/`RecordType`/`Value` and rejects `TTL`, `Weight` and `GroupId`. 

Probing existing records against it returns the key as (name, line, type, value), with the line matched by [`RecordLineId`](https://github.com/TencentCloud/tencentcloud-sdk-go/blob/tencentcloud/dnspod/v1.3.131/tencentcloud/dnspod/v20210323/models.go#L1453). [#4775](https://github.com/DNSControl/dnscontrol/pull/4775) states the same rule in English from the project's own test logs: "duplicate records on the same default line are forbidden".

A flag stops at validation. With the same per-line config, the provider's comparable produces the expected CREATEs; replace it with `nil` and the diff reports zero changes, because the engine sees one record and finds everything in place. Validation can allow or reject that config either way, and the engine still decides on its own what identity means.

`checkCNAMEs` already carries two provider-driven exceptions: AKAMAICDN, and the [Cloudflare proxied CNAME](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/pkg/normalize/validate.go#L688) from [#4189](https://github.com/DNSControl/dnscontrol/pull/4189).

Cost is close either way. The hook is one optional field on [`DspFuncs`](https://github.com/DNSControl/dnscontrol/blob/80299f765ec2a2c5f0a43268620a4c964a6e1847/pkg/providers/providers.go#L62), and providers that do not declare it keep today's behavior. A per-record flag needs a new metadata key, its entry in the generated `others.d.ts`, three conditionals with their messages in `validate.go`, and a documentation page. v4.0.0 moved the other way by collapsing per-record escape hatches into domain-level ones, which is the level such a switch belongs at.

TomOnTime:
Ok!

PR #4891, opened 2026-09-18
feat(p/OPENPROVIDER): add Openprovider as a new provider

nvanlaerebeke:
## Summary

Adds OpenProvider as a DNS service provider.

The provider supports:

- OpenProvider REST API authentication
- DNS zone discovery and creation
- `get-zones`
- Record creation, updates, and deletion
- `A`, `AAAA`, `CAA`, `CNAME`, `MX`, `SPF`, `SRV`, `TLSA`, and `TXT` records
- OpenProvider-specific record auditing and TTL normalization
- `dnscontrol init` credentials metadata

Provider-managed SOA and authoritative apex NS records are treated as read-only.

## Documentation and integration

- Added provider documentation
- Added provider capability matrix entries
- Added README and documentation index entries
- Added CODEOWNERS maintainer
- Added integration-test profile and CI environment variables

## Validation

- [x] `go generate ./...`
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `gofmt` verification
- [x] Live OpenProvider integration test
- [x] Manual TXT-record verification

The live integration test requires an OpenProvider account and disposable test domain.
The openprovider sandbox can be used for that.

I have been using this provider for a couple of weeks and it seems to be working fine.

Please create the GitHub label 'provider-OPENPROVIDER'

## Test Report

The Openprovider API isn't that reliable when doing a lot of requests, when running the entire suite there is a high change it returns with an internal error.
Most of the time it's just a single test or cleanup that fails, re-trying that test case "just works".

This is the test report, as you can see only 1 test failed, re-ran that test and it returned success.
[test-report.log](https://github.com/user-attachments/files/32473890/test-report.log)


providers/openprovider/convert.go
```diff
@@ -0,0 +1,106 @@
...
+	dnsv2 "codeberg.org/miekg/dns"
+	"github.com/DNSControl/dnscontrol/v5/models"
+	"github.com/DNSControl/dnscontrol/v5/pkg/txtutil"
+)
+
+const apexLabel = "@"
+
+func toRecordConfig(record apiRecord, origin string) (*models.RecordConfig, error) {
+	rtype := strings.ToUpper(record.Type)
+	dc := &models.DomainConfig{Name: origin}
```
TomOnTime: The parent function should pass dc as a parameter instead of `origin`.  If you need `origin`, it is available as `origin := dc.Name`

providers/openprovider/convert.go
```diff
@@ -0,0 +1,106 @@
...
+	case "SRV":
+		fields := strings.Fields(value)
+		if len(fields) != 3 {
+			return nil, fmt.Errorf("OPENPROVIDER: SRV record %q has invalid value", record.Name)
+		}
+		var weight, port uint16
+		if _, err := fmt.Sscanf(strings.Join(fields[:2], " "), "%d %d", &weight, &port); err != nil {
+			return nil, fmt.Errorf("OPENPROVIDER: SRV record %q has invalid value: %w", record.Name, err)
+		}
+		rc, err = dc.NewRecordConfig(label, uint32(record.TTL), dnsv2.TypeSRV, uint16(record.Prio), weight, port, absoluteTarget(fields[2], origin))
```
TomOnTime: Rather than using strings.Fields, NewRecordConfig() can do the parsing for you with the SrvWeirdSplit flag.

A good example is providers/namedotcom/records.go:107

```
                rc, err = dc.NewRecordConfig(label, r.TTL, dnsv2.TypeSRV, r.Priority, r.Answer,
                        nrc.Flags{SrvWeirdSplit: true, TargetIsFqdnNoDot: true})
```

See https://docs.dnscontrol.org/developer-info/cookbook#create-a-models.recordconfig

providers/openprovider/convert.go
```diff
@@ -0,0 +1,106 @@
...
+	if label == "" {
+		label = apexLabel
+	}
+
+	value := record.Value
+	var rc *models.RecordConfig
+	var err error
+	switch rtype {
+	case "MX":
+		rc, err = dc.NewRecordConfig(label, uint32(record.TTL), dnsv2.TypeMX, uint16(record.Prio), absoluteTarget(value, origin))
```
TomOnTime: The first parameter to NewRecordConfig must be the result of dc.LabelFromShort(),  dc.LabelFromFQDNNoDot(), or dc.LabelFromFQDNWithDot().  Rather than calling relativeRecordName(), I think LabelFromFQDNWithDot might be all you need.

See https://docs.dnscontrol.org/developer-info/cookbook#create-a-models.recordconfig

providers/openprovider/convert.go
```diff
@@ -0,0 +1,106 @@
...
+			return nil, fmt.Errorf("OPENPROVIDER: SRV record %q has invalid value: %w", record.Name, err)
+		}
+		rc, err = dc.NewRecordConfig(label, uint32(record.TTL), dnsv2.TypeSRV, uint16(record.Prio), weight, port, absoluteTarget(fields[2], origin))
+	case "TXT", "SPF":
+		var decoded string
+		decoded, err = txtutil.ParseQuoted(value)
+		if err == nil {
+			rc, err = dc.NewRecordConfig(label, uint32(record.TTL), dnsv2.TypeTXT, decoded)
+			if rc != nil && rtype == "SPF" {
+				rc.Type = rtype
```
TomOnTime: Do you need to mutate rc?  Please don't if not required.  This may cause future problems if the same zone is used with 2 providers.

providers/openprovider/convert.go
```diff
@@ -0,0 +1,106 @@
...
+		var weight, port uint16
+		if _, err := fmt.Sscanf(strings.Join(fields[:2], " "), "%d %d", &weight, &port); err != nil {
+			return nil, fmt.Errorf("OPENPROVIDER: SRV record %q has invalid value: %w", record.Name, err)
+		}
+		rc, err = dc.NewRecordConfig(label, uint32(record.TTL), dnsv2.TypeSRV, uint16(record.Prio), weight, port, absoluteTarget(fields[2], origin))
+	case "TXT", "SPF":
+		var decoded string
+		decoded, err = txtutil.ParseQuoted(value)
+		if err == nil {
+			rc, err = dc.NewRecordConfig(label, uint32(record.TTL), dnsv2.TypeTXT, decoded)
```
TomOnTime: Rather than using txtutils.ParseQuoted(), use `dc.NewRecordConfigParse(label, uint32(record.TTL), dnsv2.TypeTXT, value)`. It is a more efficient parser.

providers/openprovider/convert.go
```diff
@@ -0,0 +1,106 @@
...
+			if rc != nil && rtype == "SPF" {
+				rc.Type = rtype
+			}
+		}
+	case "CNAME":
+		rc, err = dc.NewRecordConfig(label, uint32(record.TTL), dnsv2.TypeCNAME, absoluteTarget(value, origin))
```
TomOnTime: absoluteTarget isn't needed here if you set TargetIsFqdnNoDot

```suggestion
		rc, err = dc.NewRecordConfig(label, uint32(record.TTL), dnsv2.TypeCNAME, value,
				nrc.Flags{TargetIsFqdnNoDot: True})
```

providers/openprovider/convert.go
TomOnTime: Don't call absoluteTarget.  Add the TargetIsFqdnNoDot flag:

```suggestion
		rc, err = dc.NewRecordConfig(dc.LabelFromFQDNWithDot(record.Name), uint32(record.TTL), dnsv2.TypeMX, uint16(record.Prio), value, nrc.Flags{TargetIsFqdnNoDot: true})
```

providers/openprovider/convert.go
```diff
@@ -0,0 +1,106 @@
...
+	case "CNAME":
+		record.Value = strings.TrimSuffix(rc.AsCNAME().Target, ".")
+	case "CAA":
+		caa := rc.AsCAA()
+		record.Value = fmt.Sprintf("%d %s %q", caa.Flag, caa.Tag, caa.Value)
+	case "TLSA":
+		tlsa := rc.AsTLSA()
+		record.Value = fmt.Sprintf("%d %d %d %s", tlsa.Usage, tlsa.Selector, tlsa.MatchingType, strings.ToLower(tlsa.Certificate))
+	default:
+		record.Value = rc.GetTargetField()
```
TomOnTime, line 92: `GetTargetField()` is deprecated.  
```suggestion
		record.Value = rc.String()
```

providers/openprovider/convert.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return record
+}
+
+func absoluteTarget(target, origin string) string {
```
TomOnTime: absoluteTarget may be unused at this point. Please verify.

providers/openprovider/convert.go
TomOnTime:
```suggestion
				rc.Type = rtype // Mutating rc is ok because we've made a copy in providerDomainConfig
```

providers/openprovider/records.go
```diff
@@ -0,0 +1,248 @@
...
+			return accepted
+		}
+	}
+	return 86400
+}
+
+// providerDomainConfig prevents provider-specific filtering and TTL
+// normalization from changing the desired state used by other providers.
+func providerDomainConfig(dc *models.DomainConfig) *models.DomainConfig {
+	desired := &models.DomainConfig{
```
TomOnTime: Try this instead:

```suggestion
	desired := dc.Copy()
```

providers/openprovider/zones.go
```diff
@@ -0,0 +1,63 @@
...
+	if !isNotFound(err) {
+		return err
+	}
+
+	domain, err := splitZoneName(dc.Name)
+	if err != nil {
+		return err
+	}
+	desired := providerDomainConfig(dc)
+	desired.Records = managedZoneRecords(desired.Records)
```
TomOnTime: The function should create the zone with no records. This might cause confusion downstream.

TomOnTime:
Thank you for contributing this new provider, @nvanlaerebeke !

A few action items before we can merge this PR:

1. I've added a few code suggestions/corrections.
1. By now you should have received a Github invite to have the "triage" role for this repo.  Please accept the invite so we can assign bugs to you.
2. @fm: Faisal Misle is our “liaison to maintainers”. Please send him an email from your preferred address to dnscontrol so we can stay in touch. Your email address will not be shared, only used for DNSControl communication.  Please email f at faisal dot fm and CC tal at what exit dot org
3. Please consider setting up a test account for use by our automated testing. This will mean we can test the provider every release in an automated manner. Info is here: https://docs.dnscontrol.org/developer-info/byo-secrets#donate-secrets-to-the-project 

Thanks again!
Tom

documentation/SUMMARY.md
```diff
@@ -173,8 +173,9 @@
 * [Netnod](provider/netnod.md)
 * [NexDNS](provider/nexdns.md)
 * [NS1](provider/ns1.md)
-* [OpenWrt](provider/openwrt.md)
+* [OpenProvider](provider/openprovider.md)
```
cafferata: The company writes its name as "Openprovider" (lowercase p), for example on https://www.openprovider.com/. This page, the navigation entry and the body text use "OpenProvider".

documentation/provider/openprovider.md
```diff
@@ -0,0 +1,154 @@
...
+    TXT("@", "v=spf1 -all"),
+    CAA("@", "issue", "letsencrypt.org"),
+    SRV("_sip._tcp", 10, 20, 5060, "sip.example.com."),
+);
+```
+{% endcode %}
+
+## Supported record types
+
+The provider supports `A`, `AAAA`, `CAA`, `CNAME`, `MX`, `SPF`, `SRV`, `TLSA`,
```
cafferata: `dnsconfig.js` has no `SPF()` record function (`helpers.js` only builds `TXT` for `SPF_BUILDER()`), so users can't declare an `SPF`-type record. `convert.go` keeps the type when reading (`rc.Type = "SPF"`), while Cloudflare, HEDNS, OVH and Dynu turn an API `SPF` record into `TXT` on read and DNS Made Easy skips it. With the type kept, `get-zones` emits `SPF("@", [v=spf1 -all])`, which doesn't parse (`ReferenceError: 'spf1' is not defined`).

**Question**: have you tested this against an Openprovider zone that contains an `SPF`-type record, with `get-zones` and with `preview`/`push` while the config has the same value as a `TXT` record?

fm: also the SPF type record (RR99) has been deprecated for a while

documentation/SUMMARY.md
nvanlaerebeke: Thanks, I didn't realize, have been mistyping their name for years now.
I've updated the name in the documentation.

documentation/provider/openprovider.md
nvanlaerebeke: I added an extra test for both cases. 

An Openprovider response containing an SPF record is converted to TXT, so get-zones produces valid TXT(...) entries. 

I also tested with the existing SPF record and an a configured TXT record so preview/push reports no change.

I'm still setting up the sandbox account, been a while since I did that and have to figure it out again

cafferata (commented):
Follow-up on my earlier review. The naming and the legacy `SPF` handling are both resolved, thanks for that. The GitBook preview renders the new page correctly: https://docs.dnscontrol.org/~/revisions/FelOOhaEeyEdagCQfOq6/provider/openprovider

All eight files from the checklist in [Writing new DNS providers](https://docs.dnscontrol.org/advanced-features/writing-providers) are present, plus the `.github/labeler.yml` entry. Two things I handled while reviewing: I created the `provider-OPENPROVIDER` label and applied it, which covers the request in your description, and I corrected the spelling in the title of this pull request, since that title becomes the commit message once this is squash-merged.

A few remaining points are in the inline comments. Beyond those I have little to add. My review covers the documentation and the provider checklist; @TomOnTime makes the final call on this pull request.

**Question**: the description ticks "Live OpenProvider integration test", but the output is not on the pull request. Could you post the result of `go test ./integrationTest/ -v -provider OPENPROVIDER` as a comment once the sandbox account is set up? The checklist asks for it, and without it reviewers have no evidence that the scenarios pass against the real API.

README.md
```diff
@@ -58,10 +58,10 @@ DNSControl supports 65 DNS providers and registrars:
...
-| [PowerDNS](https://docs.dnscontrol.org/provider/powerdns) | [Realtime Register](https://docs.dnscontrol.org/provider/realtimeregister)¹ | [RWTH DNS-Admin](https://docs.dnscontrol.org/provider/rwth) | [Sakura Cloud](https://docs.dnscontrol.org/provider/sakuracloud) | [SoftLayer](https://docs.dnscontrol.org/provider/softlayer) |
-| [Tencent Cloud DNS](https://docs.dnscontrol.org/provider/tencentdns)¹ | [TransIP](https://docs.dnscontrol.org/provider/transip) | [UniFi Network](https://docs.dnscontrol.org/provider/unifi) | [Vercel](https://docs.dnscontrol.org/provider/vercel) | [Vultr](https://docs.dnscontrol.org/provider/vultr) |
-| [Netbird](https://docs.dnscontrol.org/provider/netbird) | [WebSupport](https://docs.dnscontrol.org/provider/websupport) |  |  |  |
+| Openprovider | [OpenSRS](https://docs.dnscontrol.org/provider/opensrs)² | [Oracle Cloud](https://docs.dnscontrol.org/provider/oracle) | [OVH](https://docs.dnscontrol.org/provider/ovh)¹ | [Packetframe](https://docs.dnscontrol.org/provider/packetframe) |
```
cafferata: Every other entry in this table links to its documentation page. The single exception is `Windows Server DNS`, which has no page at all. Openprovider does have one, so the plain-text entry here breaks the pattern and leaves a README reader without a route into the new page.

```suggestion
| [Openprovider](https://docs.dnscontrol.org/provider/openprovider) | [OpenSRS](https://docs.dnscontrol.org/provider/opensrs)² | [Oracle Cloud](https://docs.dnscontrol.org/provider/oracle) | [OVH](https://docs.dnscontrol.org/provider/ovh)¹ | [Packetframe](https://docs.dnscontrol.org/provider/packetframe) |
```

documentation/provider/openprovider.md
```diff
@@ -0,0 +1,155 @@
...
+new master zone to contain at least one record, so the domain configuration must
+not be empty on its first push. Zone creation only provisions DNS hosting; it
+does not register, transfer, renew, delegate, or otherwise modify the domain at
+the registrar.
+
+## Limitations
+
+### TTLs
+
+Openprovider's current minimum record TTL is 900 seconds. DNSControl raises a
```
cafferata: This section describes the normalization as a floor: a TTL below 900 is raised to 900. The implementation does more than that. `acceptedTTLs` is a fixed ladder of six values and every requested TTL is rounded up to the next value on it, so a `TTL(1800)` becomes 3600 and a `TTL(7200)` becomes 10800. Those two cases are not predictable from this page, and they also surface in `preview`: the normalization runs before the diff, so a zone sitting at 900 with a configured `TTL(1800)` reports a change to 3600, a value that appears nowhere in the user's `dnsconfig.js`. Someone migrating an existing zone into DNSControl is the most likely to hit this, and such a diff is hard to trace back to the configuration.

providers/openprovider/records.go
```diff
@@ -0,0 +1,248 @@
...
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/DNSControl/dnscontrol/v5/models"
+	"github.com/DNSControl/dnscontrol/v5/pkg/diff2"
+	"github.com/DNSControl/dnscontrol/v5/pkg/printer"
+)
+
+var acceptedTTLs = [...]uint32{900, 3600, 10800, 21600, 43200, 86400}
```
cafferata, line 13: **Question**: does the Openprovider API reject a TTL that is not on this ladder, or does it accept arbitrary values above the 900 second minimum? The difference matters, because rounding a requested 1800 up to 3600 rewrites what the user asked for on every run, and the warning is the only place they learn about it. If the API does enforce these six values, a pointer to where Openprovider documents that would help whoever maintains this next.

providers/openprovider/api.go
```diff
@@ -0,0 +1,284 @@
...
+	defer c.tokenMu.Unlock()
+
+	if c.token != "" {
+		return c.token, nil
+	}
+
+	payload, err := json.Marshal(loginRequest{
+		Username: c.username,
+		Password: c.password,
+		IP:       "0.0.0.0",
```
cafferata: **Question**: what does Openprovider do with this hardcoded `ip`? Their login endpoint uses that field to bind the issued token to an address, and `0.0.0.0` reads as an explicit opt-out of that restriction. If an account relies on IP whitelisting, DNSControl would be quietly weakening it, and the value is not mentioned anywhere on the documentation page.

providers/openprovider/auditrecords.go
```diff
@@ -0,0 +1,68 @@
...
+	"TXT",
+}
+
+var supportedCAATags = []string{"issue", "issuewild", "iodef"}
+
+// AuditRecords returns errors for records that Openprovider cannot represent.
+func AuditRecords(records models.Records) []error {
+	auditor := rejectif.Auditor{}
+	auditor.TypesSupported(supportedRecordTypes)
+	auditor.Add("NS", rejectUnsupportedNS)
```
cafferata: 42 of the existing providers annotate each rejection rule with a `// Last verified YYYY-MM-DD` comment, so a later maintainer can see how old the rule is and when it was last checked against the live API. Those comments are absent here, which means there is no way to tell which of these limits were confirmed against Openprovider and when.

cafferata (approved):

providers/openprovider/records.go
nvanlaerebeke: For DNS records, the supported TTL values are:

900 — 15 minutes
3600 — 1 hour
10800 — 3 hours
21600 — 6 hours
43200 — 12 hours
86400 — 24 hours

Openprovider explicitly says that if you submit another value, 86400 will be saved instead.
There is a place in their documentation they mention 600, but from tests and their UI it's 900.

I've updated the documentation to better explain this with a link to their docs + added a comment in the code as well, so that it's more clear.

documentation/provider/openprovider.md
nvanlaerebeke: I've updated the documentation to be more clear.

The warning I'm printing when normalization happens also mentions `Openprovider only accepts TTLs of 900, 3600, 10800, 21600, 43200, or 86400 seconds.` 

Rounding up is I think the best solution, as any other value provided will automatically make it so the TTL is set to 86400 by openprovider, don't think that's a desired behavior as it would mean someone migrating or creating a record with a TTL of 1800 (30m) would suddenly become a day. While 60m is also not 100% correct, it is at least "more" correct.

providers/openprovider/api.go
nvanlaerebeke: Good catch, I copied that from their examples.

For DNSControl it makes no sense to send this as indeed it disables/overrides the IP white listing as far as I understand.

I've removed the sending of this so the normal whitelisting can be applied

providers/openprovider/auditrecords.go
nvanlaerebeke: I've added the dates now

nvanlaerebeke:
I've attached the test report in the PR.

Note that the Openprovider API isn't very reliable when doing a lot of requests, I've ran the test suite multiple times (make sure to use a higher timeout, example 60m)
Most of the time a single random test fails, I retry that afterwards and it just works.

Next I'll send my contact details and sandbox credentials next, that'll be for tomorrow the earliest though.

providers/openprovider/auditrecords.go
```diff
@@ -0,0 +1,68 @@
...
+	auditor.Add("NS", rejectUnsupportedNS)          // Last verified 2026-09-21
+	auditor.Add("MX", rejectif.MxNull)              // Last verified 2026-09-21
+	auditor.Add("SRV", rejectif.SrvHasNullTarget)   // Last verified 2026-09-21
+	auditor.Add("TXT", rejectif.TxtIsEmpty)         // Last verified 2026-09-21
+	auditor.Add("TXT", rejectif.TxtHasDoubleQuotes) // Last verified 2026-09-21
+	auditor.Add("TXT", rejectif.TxtHasBackslash)    // Last verified 2026-09-21
+	auditor.Add("CAA", rejectUnsupportedCAAFlag)    // Last verified 2026-09-21
+	auditor.Add("CAA", rejectUnsupportedCAATag)     // Last verified 2026-09-21
+	auditor.Add("CAA", rejectUnsupportedCAAFields)  // Last verified 2026-09-21
+	return auditor.Audit(records)
```
TomOnTime: Make this the same as all the other providers.

```suggestion
	a := rejectif.Auditor{}
	a.TypesSupported(supportedRecordTypes)
	a.Add("NS", rejectUnsupportedNS)          // Last verified 2026-09-21
	a.Add("MX", rejectif.MxNull)              // Last verified 2026-09-21
	a.Add("SRV", rejectif.SrvHasNullTarget)   // Last verified 2026-09-21
	a.Add("TXT", rejectif.TxtIsEmpty)         // Last verified 2026-09-21
	a.Add("TXT", rejectif.TxtHasDoubleQuotes) // Last verified 2026-09-21
	a.Add("TXT", rejectif.TxtHasBackslash)    // Last verified 2026-09-21
	a.Add("CAA", rejectUnsupportedCAAFlag)    // Last verified 2026-09-21
	a.Add("CAA", rejectUnsupportedCAATag)     // Last verified 2026-09-21
	a.Add("CAA", rejectUnsupportedCAAFields)  // Last verified 2026-09-21
	return a.Audit(records)
```

nvanlaerebeke:
I think I've applied all the requested changes, also sent the email with my contact information and I've sent the email for the sandbox credentials.

Is there anything else you require?

TomOnTime:
I think we have everything.  I need to add the creds to the system and we should be all done.  I'll get to that in a day or two.

Thanks!

TomOnTime:
The creds work. Thanks for sending those.

Sadly the provider errors (400 Bad Request) as you mentioned, so I won't be adding this to the automated tests.  However they are still useful for my testing loop.

If you could make it pause and re-try on `400 Bad Request`, that would be awesome.  You can crib code from other providers. They tend to do a back-off when they get 429 and other codes.

PR #4893, opened 2026-09-19
docs(p/LINODE): replace the unsupported footnote with a hint

cafferata:
The Linode provider page explains the zone's default TTL in a Markdown footnote (`[^1]`), added in #4835. GitBook does not support footnotes: on https://docs.dnscontrol.org/provider/linode both the marker and the footnote text are dropped, so the explanation of the default TTL is not shown at all. This moves the text into an info hint directly below the TTL list, using the same `{% hint style="info" %}` block that other provider pages use, and joins the soft-wrapped lines because GitBook renders them as line breaks. While here, it adds the missing "to" in "set the minimum TTL to a different value".

## AI-attributie


cafferata:
@dairiki this fixes the default TTL footnote you added in #4835: GitBook drops Markdown footnotes, so the text did not show up on https://docs.dnscontrol.org/provider/linode. Could you take a look?

cafferata:
Here is the GitBook preview URL with the default TTL note as an info hint instead of a footnote:
📚 https://docs.dnscontrol.org/~/revisions/V8tlCiOO770xrTJF3Lj0/provider/linode#caveats

TomOnTime (approved):

dairiki:
> @dairiki this fixes the default TTL footnote you added in #4835: GitBook drops Markdown footnotes, so the text did not show up on https://docs.dnscontrol.org/provider/linode. Could you take a look?

LGTM.  Thank you, @cafferata 

I think I remember that the footnote was rendered in the doc preview, but it was rendered as a tooltip rather than as a real footnote. 
The info callout is a better solution.

PR #4462, opened 2026-07-11
SCALEWAY: Add Scaleway DNS provider

alessiopcc:
Add a DNS service provider for Scaleway Domains & DNS using the official scaleway-sdk-go client. Supports ALIAS, CAA, DNAME, HTTPS, NAPTR, PTR, SRV, SSHFP, SVCB and TLSA, plus get-zones; zones must already exist in the Scaleway console.

Includes provider registration, credentials/integration-test wiring, documentation, and the regenerated feature matrix, CODEOWNERS, labeler and goreleaser metadata.

PR #4896, opened 2026-09-19
docs(p/SCALEWAY): use second-level headings

cafferata:
The Scaleway provider page, added in #4462, uses first-level headings for "Configuration" and "Activation" while the other sections are second-level headings. Every other provider page uses `## Configuration` and `## Activation`, and in GitBook the mismatch nests "Metadata" and "Usage" under "Configuration" in the page outline. This changes both headings to second-level so the structure matches the other provider pages.

## AI-attributie


cafferata:
Here is the GitBook preview URL with "Configuration" and "Activation" as second-level headings:
📚 https://docs.dnscontrol.org/~/revisions/xr4Jh8YxXafuNQulSryB/provider/scaleway

TomOnTime (approved):

PR #1788, opened 2022-10-29
DOCS: Aligned/improved installation section package manager

cafferata:
The installation section within `README.md` and [Getting started](https://stackexchange.github.io/dnscontrol/getting-started#via-packages) has been aligned/improved in terms of 'package manager'.

**Before**
before

**After**
after

PR #4898, opened 2026-09-19
docs: fix typos and a broken anchor in the README

cafferata:
A few small fixes in `README.md`. #4764 introduced "The official Github Action", which now reads "GitHub". #4774 renamed the provider variable in the example to `DSP_ROUTE53` but dropped the trailing semicolon that the line above it has, so this adds it back. #4774 also rewrote the REV() deprecation warning without updating the date in the "Deprecation warnings (updated 2025-11-21)" heading, which now uses the merge date of #4774, 2026-08-17. While here, the "official instructions" link pointed to `#1-install-the-software`, an anchor that GitBook doesn't generate: GitBook uses `#id-1.-install-the-software` for that heading, so the link now jumps to the installation section on https://docs.dnscontrol.org/getting-started/getting-started instead of the top of the page. That anchor dates back to #1788.

## AI-attributie


cafferata:
The README isn't part of GitBook, so here is the rendered README with the fixes:
📚 https://github.com/cafferata/dnscontrol/blob/docs/readme-fixes/README.md#installation

TomOnTime (approved):

Issue #3633, opened 2025-06-20
Provider request: Sav.com


matteotrubini:
I'm opening this issue to track interest in this provider.

Website Ref: https://marketing.sav.com/domains/api
API Docs: https://documenter.getpostman.com/view/9688716/TzzANHFJ

cafferata:
Hi @matteotrubini,

Thanks for opening this issue and for highlighting SAV.com as a potential new provider for DNSControl!

As you may know, DNSControl is a community-driven open source project, and support for new providers is typically added by volunteers or contributors affiliated with the provider itself.

To help increase visibility, I’ve opened GitHub pull request https://github.com/StackExchange/dnscontrol/pull/3635 to add SAV.com to the list of requested providers. The issue has also been tagged accordingly in the hope that someone from the community will take interest.

If you or someone else is interested in developing support for SAV.com, we’d be more than happy to assist with guidance and code reviews along the way.

Thanks again for your contribution and for helping to improve DNSControl!

PR #4900, opened 2026-09-19
chore(deps): update commitlint to v21

cafferata:
`npm install` warns that `git-raw-commits@4.0.0` is deprecated and no longer maintained. It comes in through `@commitlint/read` from `@commitlint/cli` 19, which the `PR: Commitlint` check from https://github.com/DNSControl/dnscontrol/pull/4873 installs on every pull request. Commitlint 21 replaced it with `@conventional-changelog/git-client`, so this bumps `@commitlint/cli` and `@commitlint/config-conventional` to `^21.2.2`.

The `package-lock.json` changes stay within the commitlint dependency tree. `prettier` and `@umbrelladocs/linkspector` are untouched, and `npm install` no longer prints the warning. Commitlint 21 requires Node `>=22.12.0`, and the workflow already uses `node-version: latest`.

It also adds `"name": "dnscontrol"` to `package.json`. Without it, `npm install` writes the name of the checkout directory into `package-lock.json`, so running it from a worktree or a clone with a different directory name changes the lockfile.

I ran the same titles through `commitlint.config.js` with version 19 and 21 and got the same results:

- Pass: `fix(p/ROUTE53): correct TTL rounding`, `feat(p/BIND)!: drop something`, `chore(deps): update dependencies`, `BREAKING CHANGE: drop v3` and `docs: add requested provider Sav.com (#3633)`.
- Fail: `fix(p/ROUTE53): Fix TTL`, `DOCS: Fixed table` and `fix(bad scope): x`.

The `^19.6.1` range came in with https://github.com/DNSControl/dnscontrol/pull/4873; I found no discussion about staying on 19, but let me know if there was a reason.

This went unnoticed because `.github/dependabot.yml` covers GitHub Actions, Go modules and Docker, but not npm. A suggestion, not part of this pull request: add an `npm` entry in the same style as `gomod`. That would also propose updates for `prettier`, which is pinned to `3.9.5`, and `@umbrelladocs/linkspector`, so it may need an `ignore` for `prettier`. Happy to open that separately if you want it.

```yaml
  # Maintain dependencies for npm
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "monthly"
    cooldown:
      default-days: 7
```

## AI-attributie


TomOnTime (approved):

TomOnTime:
@cafferata Yes, please open a PR to add npm to .github/dependabot.yml

PR #4904, opened 2026-09-19
docs: use a repository-relative path in the helpers.js debugging example

cafferata:
Follow-up to #4827. While going through the bug fix pull requests, I noticed that the new "Debug `helpers.js`" section in `documentation/advanced-features/debugging-with-dlv.md` reads `helpers.js` from `/Users/tlimoncelli/gitthings/dnscontrol/pkg/js/helpers.js`, a path that only exists on one machine. Its two code blocks also have no language, unlike the other blocks on the page, and the second one starts with a `$` prompt, so copying it doesn't work as is.

This GitHub pull request changes the path to `pkg/js/helpers.js`, notes that the command runs from the root of the repository, marks both blocks as `shell` and removes the `$` prompt. I ran the updated command from the repository root and it prints `2276849648` for `IP('135.181.247.240')`.

## AI-attributie


cafferata:
Here is the GitBook preview URL with the repository-relative path in the `helpers.js` debugging example:
📚 https://docs.dnscontrol.org/~/revisions/3PzyS5htPbcQpc7HZ88y/developer-info/debugging-with-dlv#debug-helpers.js

TomOnTime (approved):

PR #4905, opened 2026-09-19
docs: redirect the removed HETZNER provider page to HETZNER_V2

cafferata:
Follow-up to #4456. While going through the recent chore pull requests, I noticed that removing the legacy `HETZNER` provider also removed `documentation/provider/hetzner.md` and its redirect in `documentation/.gitbook.yaml`. Both https://docs.dnscontrol.org/provider/hetzner and the older https://docs.dnscontrol.org/service-providers/providers/hetzner now return a 404, so users of the removed provider who follow an old link or search result land on a dead page instead of its successor.

This GitHub pull request adds two redirects to `documentation/.gitbook.yaml` that send both URLs to the `HETZNER_V2` page, https://docs.dnscontrol.org/provider/hetznerv2.

## AI-attributie


das7pad (commented):
Thanks!

PR #4910, opened 2026-09-19
docs: point new users to dnscontrol init first

cafferata:
Tom asked to check the places new users look first and make them consistent about how to get started, now that `dnscontrol init` is available. The documentation homepage and the Migrating page already lead with `init`, but the README did not mention it and the Getting Started guide only offered it halfway through step 3.

The README now has a "Getting Started" section that leads with `dnscontrol init`, followed by the link to the full Getting Started guide and the existing pointer to the `dns-config` starter repo.

The Getting Started guide now opens with two ways to start: the quick path (install DNSControl, run `dnscontrol init`, then continue with steps 7 and 8) and the manual path through all steps with the sample files and the BIND provider, which works without API credentials. The hint in step 3 now also mentions step 4, because `init` writes `creds.json` as well. The links to steps 7 and 8 use the GitBook anchors (`#id-7.-use-your-own-domains`), which I checked against the live page.

The same change for the starter repo is in https://github.com/DNSControl/dns-config/pull/13.

## AI-attributie


cafferata:
Here are the preview URLs for the changed pages:

- 📚 README (not part of GitBook): https://github.com/cafferata/dnscontrol/blob/docs/onboarding-init/README.md#getting-started
- 📚 Getting Started: https://docs.dnscontrol.org/~/revisions/VLR58ZezhIJVC3YBKlOi/getting-started/getting-started

PR #4913, opened 2026-09-19
ci: add npm to dependabot

cafferata:
Follow-up to #4900. While updating commitlint there, I noticed that the deprecated `git-raw-commits` dependency went unnoticed because `.github/dependabot.yml` covers GitHub Actions, Go modules and Docker, but not npm. As agreed in https://github.com/DNSControl/dnscontrol/pull/4900#issuecomment-5742000317, this adds npm.

This GitHub pull request adds an `npm` entry for the root `package.json`, in the same style as the `gomod` and `docker` entries: a monthly schedule with a 7-day cooldown. Dependabot will then also propose updates for `prettier`, which is pinned to `3.9.5`, and `@umbrelladocs/linkspector`. If `prettier` should stay pinned, an `ignore` for it can be added here.

## AI-attributie


.github/dependabot.yml
```diff
@@ -34,3 +34,11 @@ updates:
       interval: "monthly"
     cooldown:
       default-days: 7
+
+  # Maintain dependencies for npm
```
TomOnTime, line 38:
```suggestion
  # Maintain dependencies for npm
  # (Needed for Prettier.  Remove this if we switch to Homebrew prettier or other install method.)
```

cafferata: @TomOnTime small addition to this comment: Prettier isn't the only npm package. `package.json` also has `@commitlint/cli` and `@commitlint/config-conventional`, which the `PR: Commitlint` check (`.github/workflows/pr_commitlint.yml`) installs with `npm install` on every pull request, and `@umbrelladocs/linkspector` for the documentation link checks (`.linkspector.yml`). So if Prettier moves to Homebrew, the npm entry is still needed for those.

PR #4914, opened 2026-09-19
docs: fix the reference links in SPF_BUILDER

cafferata:
Follow-up to #4634. While going through the recently merged pull requests, I noticed that two links on https://docs.dnscontrol.org/language-reference/domain-modifiers/spf_builder don't work. The `txtMaxSize` description shows "[multiple strings][multi-string]" and "[not go higher than 450][record-size]" as plain text, because both reference definitions are on the same line in `SPF_BUILDER.md`, so neither is recognized as a definition. That line dates back to #1809.

This GitHub pull request puts each definition on its own line. Reference links with one definition per line render fine on GitBook, as on the "Writing new DNS providers" page.

## AI-attributie


cafferata:
Here is the GitBook preview URL with the working "multiple strings" and "not go higher than 450" links:
📚 https://docs.dnscontrol.org/~/revisions/4xkuRUgZV8jwwCBsBmXH/language-reference/domain-modifiers/spf_builder

PR #4917, opened 2026-09-19
ci: set a Conventional Commits prefix for Dependabot

cafferata:
This GitHub pull request sets `commit-message.prefix: "build"` with `include: "scope"` for every ecosystem in `.github/dependabot.yml`, so Dependabot titles its pull requests `build(deps): …` and `build(deps-dev): …` instead of `Build(deps): …`. Since https://github.com/DNSControl/dnscontrol/pull/4873 the `PR: Commitlint` check lints the pull request title, and https://github.com/DNSControl/dnscontrol/pull/4916 (`Build(deps-dev): Bump prettier from 3.9.5 to 3.9.6`) is the first Dependabot pull request to fail it, on both `type-enum` and `subject-case`. The `build` type keeps these pull requests in the same "CI/CD:" changelog group as the current `Build(deps)` titles.

The prefix fixes the type, but not the capital in "Bump". Dependabot copies the capitalisation from its last commit on `main` (see `capitalise_first_word_from_last_dependabot_commit_style` in https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator/pr_name_prefixer.rb), which is still `Build(deps): Bump joi from 18.2.3 to 18.2.9 (#4872)`. Squash-merging https://github.com/DNSControl/dnscontrol/pull/4916 as `build(deps-dev): bump prettier from 3.9.5 to 3.9.6` makes that the last Dependabot commit, and from then on Dependabot writes "bump" in lowercase. I checked both titles locally with `npx commitlint`: they pass, while the current title of https://github.com/DNSControl/dnscontrol/pull/4916 fails.

Refs https://github.com/DNSControl/dnscontrol/pull/4916

## AI-attributie


cafferata:
The failing "Check: go (build/fmtjson/mod-tidy/generate/fmt/fix)" isn't caused by this pull request: `main` itself is out of sync since https://github.com/DNSControl/dnscontrol/pull/4914, because `go generate` rewrites `commands/types/dnscontrol.d.ts`. https://github.com/DNSControl/dnscontrol/pull/4920 fixes that. Once it's merged I'll rebase this pull request on `main` so the check runs again.

PR #4919, opened 2026-09-19
fix(p/SCALEWAY): correct TXT round-trip, SVCB params, SSHFP case, NS TTL, and PTR support

alessiopcc:
Five fixes to the SCALEWAY provider, all found by running the integration suite against a live Scaleway zone.

## TXT round-trip

`unquoteTXT` assumed the value Scaleway returns is escaped the same way as the value we send. It is not, and the asymmetry is easy to miss:

* On **write**, Scaleway accepts a BIND-style quoted string and unescapes it.
* On **read**, it wraps the raw stored bytes in quotes and escapes *only* interior double quotes (`"` to `\"`). Backslashes come back verbatim.

So the correct inverse strips the surrounding quotes and unescapes `\"` alone. Unescaping `\` as well double-unescaped the value, halving runs of backslashes: `a\\b` read back as `a\b`, and the record was then offered as a correction on every run.

The unit test now models both sides of the server round-trip rather than just calling `quoteTXT`/`unquoteTXT` back to back, which is what let the bug through: the two functions were each other's inverse, but the server is not.

## TXT values Scaleway silently rewrites

Two new audit rules, because Scaleway modifies these server-side and the record can therefore never match what was asked for:

* leading/trailing **spaces** are trimmed (`rejectif.TxtStartsOrEndsWithSpaces`)
* a leading/trailing **double quote** is dropped (new local rule; interior double quotes are stored as-is and stay allowed)

Without these, such a record reads as a permanently pending change. Rejecting it names the reason once, before anything is written.

## SVCB / HTTPS params

The RDATA's `String()` quotes every SvcParam value (`port="80"`, `alpn="h2,h3"`), which Scaleway's parser rejects. The data is now built with unquoted params (`port=80`) from `models.Svcbv2ValueToString`. This is the same workaround LuaDNS needs, and mirrors the code in `providers/luadns/luadnsProvider.go`.

## SSHFP fingerprint case

DNSControl stores the fingerprint uppercase so comparisons are case-insensitive, but Scaleway only accepts lowercase hex. Lowercased on the way out, as Mythic Beasts already does.

## NS TTL

Scaleway silently clamps NS record TTLs up to 1800. Matching that clamp in `GetZoneRecordsCorrections` stops a lower TTL from being offered as a correction on every run.

## PTR

The API does not support PTR records. `CanUsePTR` is now `Cannot()`, and both generated feature matrices (`documentation/provider/scaleway.md` and the summary table in `documentation/provider/index.md`) were regenerated.

## Testing

`go test ./providers/scaleway/` passes, and the full integration suite passes against a live zone:

```
ok  	github.com/DNSControl/dnscontrol/v5/integrationTest	166.900s
```

The three new audit rules show up there as skips rather than failures, which is the expected outcome:

```
SKIPPED(PROVIDER DOES NOT SUPPORT '[txtstring starts or ends with a doublequote]' ::"28:complex TXT")
SKIPPED(PROVIDER DOES NOT SUPPORT '[txtstring starts or ends with spaces]'        ::"28:complex TXT")
SKIPPED(PROVIDER DOES NOT SUPPORT '[txtstring is empty]'                          ::"28:complex TXT")
SKIPPED(CanUsePTR_not_supported)
```


TomOnTime:
Thanks for fixing these!

PR #4921, opened 2026-09-19
feat(p/SPACESHIP): add Spaceship as a DNS provider and registrar

### LINKED ISSUE #3452: Spaceship support
Dear @StackExchange team,

It is possible to add Spaceship support?
- https://www.spaceship.com/
- https://docs.spaceship.dev/

rootful:
Implemented Spaceship as a registrar and DNS provider. (Actually) closes #3452. Have tested it out with my own domains, working fine so far.

Please create the GitHub label 'provider-SPACESHIP'


providers/spaceship/spaceshipProvider.go
```diff
@@ -0,0 +1,282 @@
...
+	defaultTTL         = 3600
+)
+
+var defaultNS = client.DefaultBasicNameserverHosts()
+
+var features = providers.DocumentationNotes{
+	// The default for unlisted capabilities is 'Cannot'.
+	// See providers/capabilities.go for the entire list of capabilities.
+	providers.CanAutoDNSSEC:          providers.Cannot("Spaceship does not expose DNSSEC via the public API"),
+	providers.CanConcur:              providers.Can(),
```
cafferata, line 31: **Question**: has the concurrency test been run? `documentation/advanced-features/concurrency-verified.md` describes it as building with `go build -race` and running `dnscontrol preview` over at least four domains on the provider. This flag is what puts the ✅ next to "Concurrency Verified" on the provider page, so it tells users that zone gathering is safe to run concurrently.

cafferata (approved):
Thanks for this, @rootful. A note on scope: I do the high-level pass, mostly the documentation, the checklist and the release plumbing. @TomOnTime does the in-depth code review.

I walked through the checklist in `documentation/advanced-features/writing-providers.md` and all eight files from Step 16 are here, including `CODEOWNERS`, the `profiles.json` entry, the env variables in `pr_integration_tests.yml`, the provider count in `README.md` (65 -> 66) and the removal of Spaceship from the "requested providers" list.

The documentation side looks complete: `spaceship.md` follows the layout of the other provider pages, the entry in `SUMMARY.md` is in the right alphabetical spot, the generated feature block is untouched, and the provider registers both `RegisterCredsMetadata` and `RegisterMaintainer`, so it will show up in the `dnscontrol init` picker.

📚 GitBook preview: https://docs.dnscontrol.org/~/revisions/33ocq07NWVH8FcvjX8W8/provider/spaceship

I created the `provider-SPACESHIP` label and applied it to this pull request, so Spaceship changes are filterable like the other providers.

Two things from the checklist are still open:

- Step 16 asks for the output of the integration test as a comment on the pull request (`go test -v -args -verbose -profile SPACESHIP`). Could you post that run?
- **Question**: would you be willing to have SPACESHIP covered by the automated integration tests? The env variables are already in `pr_integration_tests.yml`, so what's left is the secrets themselves, either in your own fork or donated to the project: https://docs.dnscontrol.org/developer-info/byo-secrets That way regressions in the provider show up in every pull request instead of only when someone runs the test by hand.

One more question below.

rootful:
I did run the integration test, and I got hit by a Spaceship ratelimit mid-run. I'll get that fixed real quick and submit the logs.

rootful:
Alright, fixed up a few things now.

First run of the integration test: https://gist.github.com/rootful/db5c03fa4a2f510bab27733c143f473f

Spaceship doesn't like null MX and SRV values so it was failing, that should be fixed in commit 3e9af69. Here are the runs after that commit:

Null MX: https://gist.github.com/rootful/ba05045cfb4a6fc6fb23d84423c3e8ed
Null SRV: https://gist.github.com/rootful/eb690fc3b8ade626348acf92e74c7bb9

So the integration test is working fine it seems like. As for the concurrency thing, you'll have to give me some time to find a few more domains to sacrifice for this test.

providers/spaceship/spaceshipProvider.go
cafferata: Thanks for running the integration tests and posting the logs, and for the rate-limit handling and the null MX/SRV audit rules; the documentation covers both now.

No need to let the concurrency test hold up this pull request. An option is to ship with `providers.CanConcur: providers.Unimplemented()` for now, which renders as ❔ ("hasn't been tested, safety unknown") on the provider page, and flip it to `providers.Can()` in a follow-up pull request once you've run `go build -race` against four or more domains. That keeps this one moving, and the ❔ says exactly what we know today.

Final call is @TomOnTime's, since he does the in-depth code review.

cafferata (approved):
Re-approving after the rate-limit retry and the null MX/SRV audit rules. Same scope as before: documentation and checklist, not the in-depth code review.

providers/spaceship/spaceshipProvider.go
rootful: I've ran `dnscontrol preview` (with a binary built with `go build -race`) on 4 of my domains on Spaceship, it hasn't spat out any race warning. Is this sufficient or do I post logs too?

TomOnTime: @rootful That is sufficient. Thank you for doing that test!

providers/spaceship/convert.go
```diff
@@ -0,0 +1,258 @@
...
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, rec.Value)
+	case "NS":
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, ensureDot(rec.Nameserver))
+	case "PTR":
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, ensureDot(rec.Pointer))
+	case "CAA":
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, derefInt(rec.Flag), rec.Tag, rec.Value)
+	case "SRV":
+		label = joinPrefixedLabel([]string{rec.Service, rec.Protocol}, rec.Name)
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, derefInt(rec.Priority), derefInt(rec.Weight), srvPort(rec.Port), ensureDot(rec.Target))
```
TomOnTime: ensureDot() isn't needed if you add nrc.Flags{TargetIsFqdnNoDot: true}

```suggestion
		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, derefInt(rec.Priority), derefInt(rec.Weight), srvPort(rec.Port), rec.Target, nrc.Flags{TargetIsFqdnNoDot: true})
```

providers/spaceship/convert.go
```diff
@@ -0,0 +1,258 @@
...
+	case "HTTPS", "SVCB":
+		label = joinPrefixedLabel([]string{portString(rec.Port), rec.Scheme}, rec.Name)
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, derefInt(rec.SvcPriority), httpsTarget(rec.TargetName), rec.SvcParams)
```
TomOnTime: First parameter to NewRecordConfig() must be the result of one of the dc.LabelFrom* functions.

```suggestion
		rc, err = dc.NewRecordConfig(dc.LabelFromShort(label), ttl, rec.Type, derefInt(rec.SvcPriority), httpsTarget(rec.TargetName), rec.SvcParams)
```

providers/spaceship/convert.go
```diff
@@ -0,0 +1,258 @@
...
+	case "TLSA":
+		label = joinPrefixedLabel([]string{portString(rec.Port), rec.Protocol}, rec.Name)
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, derefInt(rec.Usage), derefInt(rec.Selector), derefInt(rec.Matching), rec.AssociationData)
```
TomOnTime:
```suggestion
		rc, err = dc.NewRecordConfig(dc.LabelFromShort(label), ttl, rec.Type, derefInt(rec.Usage), derefInt(rec.Selector), derefInt(rec.Matching), rec.AssociationData)
```

providers/spaceship/convert.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/namecheap/go-spaceship-sdk/client"
+)
+
+func toRC(dc *models.DomainConfig, rec client.DNSRecord) (*models.RecordConfig, error) {
+	ttl := clampTTL(uint32(rec.TTL))
+	label := dc.LabelFromShort(rec.Name)
+
+	var rc *models.RecordConfig
+	var err error
+	switch strings.ToUpper(rec.Type) {
```
TomOnTime: Is ToUpper really required?  If so, copy the uppercase version to `rtype` and use it in the cases.

```suggestion
	switch rtype := strings.ToUpper(rec.Type); rtype {
```

providers/spaceship/convert.go
```diff
@@ -0,0 +1,258 @@
...
+	case "A", "AAAA":
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, rec.Address)
+	case "CNAME":
+		rc, err = dc.NewRecordConfig(label, ttl, rec.Type, ensureDot(rec.CName))
```
TomOnTime: Use the constant dnsv2.TypeFOO instead of rec.Type when you are absolutely sure of the type.

Use nrc.Flags{TargetIsFqdnNoDot: true} if the target needs the "." added. No need for ensureDot().

```suggestion
		rc, err = dc.NewRecordConfig(label, ttl, dnsv2.TypeCNAME, rec.CName, nrc.Flags{TargetIsFqdnNoDot: true})
```

providers/spaceship/convert.go
```diff
@@ -0,0 +1,258 @@
...
+	case "ALIAS":
+		rc, err = dc.NewRecordConfig(label, ttl, privatetypes.TypeALIAS, ensureDot(rec.AliasName))
+	case "MX":
```
TomOnTime, line 31: MX, TXT, NS, PTR, CAA, SRV can all use dnsv2.TypeFOO constants instead of rec.Type

providers/spaceship/spaceshipProvider.go
```diff
@@ -0,0 +1,281 @@
...
+
+func nativeFromExisting(rc *models.RecordConfig) (client.DNSRecord, error) {
+	if native, ok := rc.Original.(client.DNSRecord); ok {
+		return native, nil
+	}
+	if native, ok := rc.Original.(*client.DNSRecord); ok && native != nil {
+		return *native, nil
+	}
+	return toNative(rc)
+}
```
TomOnTime: This is overkill. .Original will never be non-nil on an existing record.

I appreciate the defensive programming, but its not needed here.

Change `nativeFromExisting(x)` to `x.Original.(client.DNSRecord)`  and get rid of nativeFromExisting()

TomOnTime (changes requested):
Looks very good!  I just have a few suggestions

rootful:
Thanks for the feedback. I fixed things up as requested.

